PR #1, opened 2020-02-17
Bigger click target = higher conversion rate

dkempner:


flybayer:
😂 You're officially the first ever contributor to Blitz 🎉

PR #3, opened 2020-02-20
Implement Initial Continuous Integration Workflow Using GitHub Actions

medelman17:
This PR implements basic CI using GitHub actions. As it is presently configured, the service will use `yarn`, cache dependencies for faster execution times, and run any/all tests in lerna-enabled packages. This will occur on all `push` events. 

Eventually, once we come with a release workflow, formalize our GitFlow, etc., we'll want to tweak this service to be more nuanced. For now, this should do nicely.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,33 +1,33 @@
-name: Github Stars to Slack
+name: Continuous Integration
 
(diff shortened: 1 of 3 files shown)


flybayer:
Nice!

PR #7, opened 2020-02-23
Add initial CLI project

mabadir:
Initial PR for oclif CLI

PR #8, opened 2020-02-23
feat(cli): Add initial CLI

flybayer:
This replaces #7 by @mabadir. This branch is native to the blitz repo so we can all commit easily.


dkempner:
lgtm

package.json
```diff
@@ -11,24 +11,36 @@
...
-    "@types/jest": "^25.1.3"
+    "typescript": "^3.7.5"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "pretty-quick --staged"
+    }
+  },
+  "dependencies": {
+    "debug": "^4.1.1"
```
medelman17: This likely should not be declared as a dependency in the root `package.json`. Typically, only common `devDependencies` should be pulled up to the root.

packages/cli/README.md
```diff
@@ -0,0 +1,151 @@
+blitz-cli
```
medelman17: Please add a `developer` story to the README.md; it's not apparent how I am to build/test/release, etc. As a new contributor, I should be able to find this info with little friction.

.gitignore
```diff
@@ -4,4 +4,5 @@
 lib
 node_modules
 reports
-*.log
\ No newline at end of file
+*.log
+**/.env*
```
medelman17: Please add test & build artifacts and any other generated files/artifacts to `.gitignore`. Right now, I'm seeing the `.nyc_output` directory being included, which I imagine should be ignored?

packages/cli/package.json
```diff
@@ -1,44 +1,66 @@
...
+  },
+  "author": "Mina Abadir @mabadir",
+  "main": "lib/index.js",
+  "bin": {
+    "blitz": "./bin/run"
+  },
+  "types": "lib/index.d.ts",
+  "files": [
+    "/bin",
+    "/lib",
```
medelman17: Does `lib` need to be included in published package? If not, please remove.

packages/cli/package.json
```diff
@@ -1,44 +1,66 @@
...
   "version": "0.0.1",
   "license": "MIT",
+  "scripts": {
+    "postpack": "rm -f oclif.manifest.json",
+    "posttest": "eslint . --ext .ts --config .eslintrc",
+    "prepack": "rm -rf lib && tsc -b && oclif-dev manifest && oclif-dev readme",
+    "test": "nyc --extension .ts mocha --forbid-only \"test/**/*.test.ts\"",
+    "version": "oclif-dev readme && git add README.md"
+  },
+  "author": "Mina Abadir @mabadir",
```
medelman17: Please revert this change; feel free to add to the contributors.

packages/cli/src/commands/console.ts
```diff
@@ -0,0 +1,25 @@
...
+    force: flags.boolean({char: 'f'}),
+  }
+
+  static args = [{name: 'file'}]
+
+  async run() {
+    const {args, flags} = this.parse(Console)
+
+    const name = flags.name || 'world'
+    this.log(`hello ${name} from /Users/mina/Code/blitz-cli/src/commands/console.ts`)
```
medelman17: Change this string to not include a hardcoded, developer-specific path.

packages/cli/src/commands/db.ts
```diff
@@ -0,0 +1,25 @@
...
+    force: flags.boolean({char: 'f'}),
+  }
+
+  static args = [{name: 'file'}]
+
+  async run() {
+    const {args, flags} = this.parse(Db)
+
+    const name = flags.name || 'world'
+    this.log(`hello ${name} from /Users/mina/Code/blitz-cli/src/commands/db.ts`)
```
medelman17: Same as above.

packages/cli/src/commands/generate.ts
```diff
@@ -0,0 +1,25 @@
...
+    force: flags.boolean({char: 'f'}),
+  }
+
+  static args = [{name: 'file'}]
+
+  async run() {
+    const {args, flags} = this.parse(Generate)
+
+    const name = flags.name || 'world'
+    this.log(`hello ${name} from /Users/mina/Code/blitz-cli/src/commands/generate.ts`)
```
medelman17: Ditto.

packages/cli/src/commands/run.ts
```diff
@@ -0,0 +1,25 @@
...
+    force: flags.boolean({char: 'f'}),
+  }
+
+  static args = [{name: 'file'}]
+
+  async run() {
+    const {args, flags} = this.parse(Run)
+
+    const name = flags.name || 'world'
+    this.log(`hello ${name} from /Users/mina/Code/blitz-cli/src/commands/run.ts`)
```
medelman17: Again.

packages/cli/tsconfig.json
```diff
@@ -1,10 +1,14 @@
 {
```
medelman17: Please revert this change so that the config extends the root config; override as necessary.

medelman17:
Seems to be working within the `monorepo` just fine. Tests failing do not appear to be monorepo-related. Thus, I would say that once the above review is resolved and tests are passing, it's ready to merge.

flybayer:
Current pending items before we can merge this initial work:

- [ ] Switch testing to jest
- [ ] Get tests passing
- [ ] Fix `yarn build` — currently failing

lorenzorapetti:
Some thoughts:

## `blitz new [PATH]`

Right now this command takes `PATH` as an argument but the generator uses the same argument to get the app name. Should we separate path and app name so that the user can write `blitz new ./projects` and then the system asks for the name? We should standardize this so we can use the same standard for all the generators.

## Commands

Except for the `new` command, all the others don't do anything, they're just stubs. I think we should either remove them or implement them before we merge.

flybayer:
Awesome @lorenzorapetti!! For now we can just use the directory name. So if path is `/tmp/blitz` the name is `blitz`. If folks want it, we can add a prompt and flag to specify the name.

Also, as a guiding principle, word the CLI prompts such that Yes is always the default.

mikeattara:
I think we should use `blitz new [NAME]` so we make the path relative to the current directory, just as in CRA

flybayer:
@mikeattara `blitz new [PATH]` does what CRA does but also accepts a directory in  another location.

`blitz new app` -> `./app/` with name `app`
`blitz new projects/app` -> `./projects/app/` with name `app`
`blitz new ../app` -> `../app/` with name `app`

lorenzorapetti:
Ok, now the tests are passing! Too bad that we don't have any yet 😅. I need to study how to mock inquirer.js.

Should we have something in place to define where to put test files (e.g. local `.test.ts` file vs. global `__tests__` or `test` directory)?

mabadir:
We might also want to have global jest config.

prettier.config.js
```diff
@@ -2,5 +2,7 @@ module.exports = {
   bracketSpacing: false,
   jsxBracketSameLine: true,
   singleQuote: true,
-  trailingComma: "all"
-};
+  trailingComma: 'all',
+  semi: false,
+  printWidth: 110,
+}
```
jasonblalock: Just curious, since we're changing the repo formatting as part of this CLI PR, is this the standard that we want for all blitzjs repos (website repo included)?

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "es2015",
```
jasonblalock: This is probably my being new to Typescript, but do we not need to target es5? Or are we assuming users of our library will transpile our es6/es2015 as part of their build?

And same question for the es2017 in [packages/cli/tsconfig.json](https://github.com/blitz-js/blitz/pull/8/files#diff-fea82da13cd5f2a75bd510aaab6f18ad)

prettier.config.js
flybayer: I added this. I'm willing to flex on the printWidth, but not the `semi` 😄

tsconfig.json
flybayer: Yes, we need to leave this as `es5` like @medelman17 set up. If the CLI has different needs, then overside the top level tsconfig in the CLI tsconfig.

flybayer:
I really don't like `__tests__` :D So use a combination of `tests` directory and `a.test.ts` depending on what makes the most sense to y'all.

prettier.config.js
jasonblalock: Oh man, that's it I'm out :door: J/k, I can ignore the lack of semi :laughing: And I don't have an opinion on the print width, I've always gone with the default based on their recommendation https://prettier.io/docs/en/options.html. Mostly was just curious so we can keep the website repo in sync.

jasonblalock:
I've been a fan of the local tests in the past. Feels like it removes the friction of clicking through directories to get to the one that matches your src. It's just the file right below the one you're working on. Just my two cents.

medelman17:
Tests should all go in one place. And it should be the same place in each package in the monorepo. Consistency across the monorepo is one of its KEY benefits. To do otherwise would be to foist ourselves upon our own petards. Cheers!

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

kamerontanseli:
Tried running the `cli new` within an empty directory to check if everything generates okay.

```
kamerontrayio:example kamerontanseli$ ../packages/cli/bin/run new .
? How do you want your project to be called? example
   create package.json
   create README.md
   create pages/index.js
   create .gitignore
yarn add v1.22.0
error package.json: Name contains illegal characters
info Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command.
App created!
```

NPM complains about the `package.json` name. It should be just `example`.

```json
{
  "name": "/Users/kamerontanseli/Desktop/blitz/example",
  "version": "0.0.1",
  "private": true,
  "scripts": {
    "dev": "next dev",
    "build": "next build",
    "start": "next start"
  }
}
```

lorenzorapetti:
Thank you for reporting! I've just pushed a fix

tsconfig.json
lorenzorapetti: I reverted the change. I'm just curious, why do we need to compile down to es5? The latest node LTS supports es2019

flybayer: @lorenzorapetti `es5` is for the other packages, not the cli. The cli can target something newer.

flybayer:
Fyi: I'm working right now on merging this into canary

PR #10, opened 2020-02-26
Use sqlite as default database for demo project

ewgenius:
- switched to `sqlite` provider
- applied default formatting rules to schema file ([prisma docs](https://github.com/prisma/prisma2/blob/master/docs/prisma-schema-file.md#auto-formatting))
- fixed few type issues on blitzjs/core and example controllers

image

image

### DIFF
--- a/examples/first-demo/controllers/posts.ts
+++ b/examples/first-demo/controllers/posts.ts
@@ -1,4 +1,5 @@
 import {Controller} from '@blitzjs/core'
+import {PostCreateInput} from '@prisma/client'
 
 export const PostsController = Controller(({db}) => ({
   name: 'PostsController',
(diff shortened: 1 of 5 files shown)


flybayer:
I'm closing this, because i have a bunch of changing incoming shortly. We can revisit sqlite after this.

examples/first-demo/controllers/posts.ts
```diff
@@ -23,7 +24,7 @@ export const PostsController = Controller(({db}) => ({
 
   async create(params, newData) {
     const data = await db.post.create({
-      data: newData,
+      data: newData as PostCreateInput,
```
flybayer: Instead of this change, we need to set up the types so that `newData` here automatically includes the attributes defined in `permit`

PR #16, opened 2020-02-27
chore(core): Add core tests

### LINKED ISSUE #14: Add tests to blitz core
Blitz core is currently untested. We need to add comprehensive tests. This would be a good task for anyone who wants it!

Comment below if you're getting ready to work on it

MrLeebo:
As I commented earlier, I couldn't use `next-test-utils` even though that seemed like a good way to approach this, so I did something basic with ordinary mock methods.

I've never worked with TypeScript before, so I was mostly stumbling through the syntax (turns out VS Code is pretty helpful with that) but I ended up leaving a lot of declarations as `: any` that can probably be better defined.

- Closes #14 or at least moves the needle in the right direction
- Tests cover Form and Controller
- You can check code coverage with "yarn test --coverage"

Current code coverage
Screen Shot 2020-02-27 at 12 17 56 AM


flybayer:
Thank you @MrLeebo! I'll review this later

packages/core/src/controller.ts
```diff
@@ -38,7 +38,6 @@ export const harnessController = (Controller: ControllerInstance) => async (
   res: NextApiResponse,
 ) => {
   const startTime = new Date().getTime()
-  console.log('')
```
flybayer: Leave this one in — it adds a new line between logs so they are easier to read

flybayer:
@MrLeebo this is great, thank you!! I already added back that console.log I commented on.

packages/core/src/controller.ts
emeraldsanto: My bad, I thought it was useless. Why not simply use a newline character?

flybayer: Ha, yes you are right — I just hadn't thought about that :)

PR #25, opened 2020-03-01
refactor(cli): CLI refactoring

### LINKED ISSUE #20: Change CLI build to use pkg
Currently, the cli uses the default oclif packager. We want to use [pkg](https://github.com/zeit/pkg) to build it into a single binary without any install dependencies.

@lorenzorapetti is working on this

lorenzorapetti:
This is a _very_ early look at the refactor I'm making to remove the yeoman dependency so that the build with `pkg` works. I'm putting this out so we can gather our thoughts around this.

## New dependencies

We are going to mainly depend on [`enquirer`](https://github.com/enquirer/enquirer) for prompting and [`mem-fs-editor`](https://github.com/SBoudrias/mem-fs-editor) for templating.

## Commands

There's now a base `Command` class from which every command inherits. This class will contain an instance of enquirer to enable prompting directly in the command class. We can automate the prompting by adding a `prompt` static property:

```javascript
export default class New extends Command {
  static description = 'Create a new Blitz project'

  static args = [
    {
      name: 'path',
      required: false,
      description: 'path to the new project, defaults to the current directory',
    },
  ]

  static flags = {
    help: flags.help({char: 'h'}),
    ts: flags.boolean({
      char: 't',
      description: 'generate a TypeScript project',
      default: true,
      allowNo: true,
    }),
    yarn: flags.boolean({description: 'use Yarn as the package manager', default: true}),
  }

  // Here is the new property
  static prompt = [
    {
      name: 'path',
      args: ['path'],
      type: 'input',
      message: 'App name?'
    },
    {
      name: 'ts',
      flags: ['ts'],
      type: 'confirm',
      message: 'Do you want to use TypeScript?'
    }
  ]
}
```

The `args` and `flags` properties inside `prompt` will filter out the questions that have already been answered by an arg or a flag (or more than one, hence the array). For example, the `ts` prompt will not be shown if the user launched the command with the `--ts` flag.

## Generators

There's now a base `Generator` class from which every generator inherits. It will have:

* An instance of `mem-fs-editor` to interact with the filesystem
* Automatic conflict detection
* A "dry run" mode
* Automatic installation of packages with either `yarn` or `npm`
* Automatic creation of the destination folder and call to `mem-fs-editor`'s `commit` to write the edits on disk.

This is the actual structure of the class for now: https://github.com/blitz-js/blitz/blob/b91389a70258c585df5c166b3a96179202ffa4fb/packages/cli/src/generator.ts

There's an `enquirer` instance to ask the user what to do in case of conflicts, but it can technically be used everywhere.

## Todo

For generators:

- [x] Handle conflicts
- [x] Handle possible errors
- [x] Handle installation with yarn or npm
- [x] Implement dry run feature

For commands:

- [x] Create the base command class
- [ ] Integrate `enquirer` with the `prompt` static property

## Conclusion

I made this structure the way that it is not only to support building with `pkg`, but also to ease the work that needs to be done to enable the development of the cli plugins.

That's all I could think of for now. I would really appreciate any kind of feedback you have.

Closes #20


flybayer:
Looking good to me! Thank you!

lorenzorapetti:
I've just added a conflict checker for the generator! Right now you can:

* Skip the file
* Overwrite
* Show the changes

You can see a preview here:

[asciicast](https://asciinema.org/a/0JHuNAlEe3Rde2XYIW26GW8tJ)

flybayer:
@lorenzorapetti Hey Lorenzo, do you have an update on this? It's totally fine it's been sitting since we've been so focused on architecture, but I'm just curious if you'll be able to finish this?

lorenzorapetti:
Yeah I'm currently super busy with a project. I'm planning on finishing this in the weekend so it can be finally merged

flybayer:
@lorenzorapetti ok, no problem. I appreciate the update and all your work on this so far!!

lorenzorapetti:
> @lorenzorapetti ok, no problem. I appreciate the update and all your work on this so far!!

Thank you! I think this is good enough to merge. Eventually, we'll want to make other changes like generalize the environment on which the command is run (right now it's only the terminal) to enable things like a VSCode extension but, for now this will do.

P.S.: in the end i didn't go for the static `prompt` property for the command class because i think it doesn't give enough flexibility for all the things `enquirer` can do

Issue #22, opened 2020-03-01
Add cli command: blitz test


flybayer:
- [ ] `blitz test` should run the `test` script in package.json. 
- [ ] `blitz t` should run `blitz test`
- [ ] Add tests

camilo86:
If still open, I can submit a PR for this later today

flybayer:
Yep, still open. Go for it! @camilo86

camilo86:
Just sent a PR for this issue https://github.com/blitz-js/blitz/pull/30 Let me know what you think

PR #30, opened 2020-03-03
feat(cli): Add `blitz test` command

camilo86:
PR for issue #22 

This adds a `test` CLI command to run the project test command via yarn
You can run the command via `blitz test` or `blitz t`

This is a very simplistic approach; maybe later, we would want to check if the CLI is being executed within a Blitz project and then execute the desired test commands. For now, it just delegates that work to yarn

*Note:* Since the command is called `test` (filename: `test.ts`), I had to add `src/commands/test.ts` to the ignore jest patterns list so that it would not thing that the test command is unit test itself 😅

### DIFF
--- a/packages/cli/jest.config.js
+++ b/packages/cli/jest.config.js
@@ -6,6 +6,7 @@ module.exports = {
   // collectCoverage: !!`Boolean(process.env.CI)`,
   collectCoverageFrom: ['src/**/*.ts'],
   coveragePathIgnorePatterns: ['/templates/'],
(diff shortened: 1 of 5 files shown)


flybayer:
Sweet, thanks! One thing: A blitz app can be set up with npm or yarn, so we should use npm or yarn depending on which the project uses. This is very common, so I'm sure there's an npm package that makes it easy.

camilo86:
Awesome. I added [has-yarn](https://www.npmjs.com/package/has-yarn) to check for yarn/npm. Also updated tests accordingly

PR #35, opened 2020-03-04
fix(core): Add support for multiple attribute types for id

gielcobben:
Support all default values for IDs:

- `String`:
  - `cuid`: Prisma will generate a globally unqiue identifier based on the [`cuid`](https://github.com/ericelliott/cuid) spec and set it as the record's ID value
  - `uuid`: Prisma will generate a globally unqiue identifier based on the [UUID](https://en.wikipedia.org/wiki/Universally_unique_identifier) spec and set it as the record's ID value
- `Int`: 
  - `autoincrement`: Prisma will create a sequence of integers in the underlying database and assign the incremented values to the ID values of the created records based on the sequence. This corresponds to e.g. using [`SERIAL`](https://www.postgresql.org/docs/9.1/datatype-numeric.html#DATATYPE-SERIAL) in PostgreSQL.

Here are examples for using default values for the model:

```prisma
model User {
  id    String  @id @default(cuid())
  name  String
}
```

or

```prisma
model User {
  id    String  @id @default(uuid())
  name  String
}
```

or

```prisma
model User {
  id    Int     @id @default(autoincrement())
  name  String
}
```

### DIFF
--- a/examples/first-demo/controllers/comments.ts
+++ b/examples/first-demo/controllers/comments.ts
@@ -15,7 +15,7 @@ export const CommentsController = Controller(({db}) => ({
       data: {
         ...newData,
         post: {
-          connect: {id: parseInt(params.query.postId)},
+          connect: {id: params.query.postId},
         },
       },
     })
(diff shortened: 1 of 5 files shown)


flybayer:
Fantastic @gielcobben! 

1. Could you add some comments above the `stringId` line in `controller.ts` that describes the three possible id types? To ensure everyone touching that knows what's going on.
2. Could you add tests for the three cases? Input is a mocked req object, output is that `Controller.[action]` is called with the correct `params.id`.

Issue #41, opened 2020-03-04
Create separate @blitzjs/webpack package for node context


ryardley:
Currently from the same package we are exporting code that runs:
* Within a node context. (webpack config)
* Within an isomorphic JSX / React context.

Because of the nature of JS when modules are loaded they are also run. Unfortunately that means stuff breaks when it is in the wrong context. 

For example, in order to allow our users to import using absolute paths such as. `app/posts/controllers`, if I try and add a node using plugin such as `tsconfig-paths-webpack-plugin` to `withBlitz` function export of `@blitzjs/core` , we get an error because that webpack code is included with the other exports from that package that are run in the wrong "non-node" context. 

This means we need to deep link to the node stuff by itself perhaps or create a new package like `@blitzjs/webpack`. 

We could alternatively try to create a custom rollup config to not concatenate files which would allow for `@blitzjs/core/webpack` but I think that would be much harder than just making a new package mainly because tsdx doesn't support it.

If no one objects I might make a new package?

PR #42, opened 2020-03-05
bug(webpack): Create webpack package to avoid context bugs

ryardley:
Closes https://github.com/blitz-js/blitz/issues/41

Extracted out a `@blitzjs/webpack` package.
Added test for webpack config changes.


packages/core/src/components/Form.tsx
```diff
@@ -9,7 +9,7 @@ export default function({children, action, method, ...props}: {[index: string]:
     // Warm the lamba
     // TODO: Only do this once per route
     fetch(action, {method: 'HEAD'})
-  }, [])
+  }, [action])
```
jimthedev: Seems unrelated to the rest of the PR but good catch :)

yarn.lock
```diff
@@ -10307,7 +10307,7 @@ neo-async@^2.5.0, neo-async@^2.6.0, neo-async@^2.6.1:
   resolved "https://registry.yarnpkg.com/neo-async/-/neo-async-2.6.1.tgz#ac27ada66167fa8849a6addd837f6b189ad2081c"
   integrity sha512-iyam8fBuCUpWeKPGpaNMetEocMt364qkCsfL9JuhjXX6dRnguRVOfk2GZaDpPjcOKiiXCPINZC1GczQ7iTq3Zw==
 
-next-compose-plugins@2.2.0:
+next-compose-plugins@^2.2.0:
```
jimthedev: @flybayer Were these previously pinned to a version instead of a range for a reason? If not then probably ok.

packages/core/src/components/Form.tsx
ryardley: yup linting error was preventing me from committing.

yarn.lock
ryardley: Nice catch will pin them

packages/webpack/test/withBlitz.test.ts
```diff
@@ -0,0 +1,19 @@
+import {withBlitz} from '../src'
+
+describe('withBlitz', () => {
+  it('alters the webpack config as expeted', () => {
```
chickencoder: Just a spelling correction. Should be 'expected'.

PR #36, opened 2020-03-04
feat(cli): Add `blitz start` command #21

tsawan:
This is PR for issue #21. (adds `start` command with alias `s`)

PR #45, opened 2020-03-05
feat(cli): Compile to Next

ryardley:
## Compiling to next

This is the beginning of our start (development) script for blitz. 

1. Start sets up a file watcher that builds all next files into a `.blitz/caches/dev` or `.blitz/caches/build` folder depending on process and then runs `next dev` or `next build` over it
1. Server management commands are then proxied to run within `.blitz/` 
1. In Blitz apps out of the box you can refer to your dependencies with absolute paths based on your tsconfig baseUrl. 

### Ancillary things

1. TypeScript compile target needed to be changed to es6 to support @oclif testing 

## This PR

Apologies for the size of this PR. Happy to break it down a little better if too much just let me know and I will reset/rebase it. I did my best to point out the main locations for all the features here:

- [x] Removes hard prisma setup peer dependency so we can test using a simple next app [here](https://github.com/blitz-js/blitz/pull/45/files#diff-a93f507f51aa5d12ae321a37c9c1afa7R11)
- [x] Creates a [vanilla next](https://github.com/blitz-js/blitz/pull/45/files#diff-8680dc92258310730428d1ad3d80a906) app for testing under `/examples/vanilla-next`.
- [x] Starts watching and building to `.blitz` folder using this file synchroniser [here](https://github.com/blitz-js/blitz/pull/45/files#diff-8bc7305f734d5c115b342ebb93737b01)
- [x] Provides a simple noop file transformation to be extended at a later point (https://github.com/blitz-js/blitz/pull/45/files#diff-8bc7305f734d5c115b342ebb93737b01R73)
- [x] Enables CTN in Blitz sample application [here](https://github.com/blitz-js/blitz/pull/45/files#diff-8f5920f757ee7907c8ce08771ae89853R6)
- [x] Merge in the start syntax and testing strategy for long running process PR from https://github.com/blitz-js/blitz/pull/36 once ready
- [x] Production script [here](https://github.com/blitz-js/blitz/pull/45/files#diff-7977878778d75ba992fb6584dba6b3ad)
- [x] Build script [here](https://github.com/blitz-js/blitz/pull/45/files#diff-00d74ea29c26b79f6bd7e78a3913f4c3R7) (blitz app build only no prisma)
- [x] Testing strategy/testing/mocking [here](https://github.com/blitz-js/blitz/pull/45/files#diff-070d7ad9fdb87d909493100dc89ef228)
- [x] TypeScript compile target had to be changed to es6 to support testing oclif. [here](https://github.com/blitz-js/blitz/pull/45/files#diff-e5e546dd2eb0351f813d63d1b39dbc48R3)

Acceptance criteria (to be tested)
- [x] Hot loading works editing the source files
- [x] Vanilla Next app runs without error
- [x] Deleted files tested that they are being removed
- [x] Build and start can still work 
- [x] `now` can still work

### Other Separate PRs
- Refactor scripts out to a separate package with a node functions as entry point.
- Output errors change file paths according to manifest (see [below](https://github.com/blitz-js/blitz/pull/45#issuecomment-596129945))
- `/pages` folder transform rules & file content transformation system
- Transform manifest/map to check integrity and increase performance
- Validation rules eg no dupes etc.
- Efficiency/performance - avoid copying files that are the same / ensure proper filestream use
- Test against a next codebase with a few hundred pages.


flybayer:
@ryardley we're so lucky to have you!! This is awesome.

And good news! It's deploying to zeit with zero config! By adding the following package.json script:
```
"build": "cd .blitz && next build && cd .. && cp -r .blitz/.next ."
```

So for the real deal, we can use the default `.next` build directory or set now to use a `.blitz-dist` directory or something. 

https://vanilla-next.flybayer.now.sh/

I had to do hack around stuff since we don't have any packages deployed yet, but it works.

Btw, I do have a request in to npm for the `@blitzjs` scope. There's been no reply from the owner, so once 4 weeks is up, it's all mine! So I think about 1 week left.

ryardley:
Thanks man. Super stoked to be on the team!

> ```
> "build": "cd .blitz && next build && cd .. && cp -r .blitz/.next ."
> ```

Yeah I hadn't checked but was hoping something like that might be the case. The only thing I was thinking was that we may want to use node within the cli for that to stay compatible with our windows friends.

flybayer:
> The only thing I was thinking was that we may want to use node within the cli for that to stay compatible with our windows friends.

Can you clarify? (is there a typo?)

ryardley:
My understanding is that windows users don't necessarily have `cp` available. Every project I have been on with .NET devs usually want build scripts written in node as it is cross platform.

flybayer:
Ah yes, yes, for sure.

ryardley:
PSA: Just force pushed this branch.

ryardley:
One issue I have discovered with this approach is we will somehow need to map compiler errors in next outside of the compiled folder. The problem is when you click through they send you to the compiled file and not the original. This might suck a bit. 🤔

ryardley:
Ok I think this is ready for review.

flybayer:
Nice, @ryardley this is a beauty!

Last thing before merging, can you go ahead and add `blitz build` which will do the build without starting the server? This is needed for Now deployment.

@tsawan looks like this covers everything from #36. I probably shouldn't have created that issue for adding blitz start apart from this PR 🙁 I think we should just roll with this. Is there anything missing in here from your PR that should be copied over?

ryardley:
I noticed I forgot command descriptions. I will update them and create the build script.

ryardley:
> Last thing before merging, can you go ahead and add blitz build which will do the build without starting the server? This is needed for Now deployment.

Am I right that it should also run `prisma2 generate`?

packages/cli/test/commands/start.test.ts
```diff
@@ -0,0 +1,29 @@
...
+    devFolder: '.blitz/caches/dev',
+    rootFolder: resolve(__dirname, '../../'),
+  }
+
+  it('runs the dev script', async () => {
+    await StartCmd.run([])
+    expect(dev).toBeCalledWith(options)
+  })
+
+  it('runs the dev script', async () => {
```
tsawan: should change test description to 'run the production server'

tsawan:
@flybayer I am not too sure. @ryardley can confirm if #36 is covered here.

ryardley:
> Am I right that it should also run prisma2 generate?

@flybayer I am thinking maybe I will skip doing that for now as we probably should talk about the migration process from next -> blitz.

ryardley:
> @ryardley can confirm if #36 is covered here.

Yeah it does. Sorry @tsawan that is probably my fault I should have coordinated a bit more. Was trying to bring in what you had done as I was waiting on it.

ryardley:
I just realised I missed the test on prod. Will throw that together and push tonight

ryardley:
ok @flybayer I did a couple of fix ups. I think this ought to work as a start.

tsawan:
> > @ryardley can confirm if #36 is covered here.
> 
> Yeah it does. Sorry @tsawan I apologise that is probably my fault too I should have coordinated a bit more. Was trying to synchronise and wait to bring in what you had done but had scratch code there to test and run the command.

@ryardley no worries, I am very happy learning here and able to contribute.

Issue #24, opened 2020-03-01
Add cli command: blitz db migrate


flybayer:
- [ ] `blitz db migrate` should effectively run `prisma2 migrate save --experimental && prisma2 migrate up --experimental`
- [ ] Add tests

tsawan:
@flybayer can you comment if blitz db will have more options (commands) in the future? This may require us to structure this command in a specific way ( I need to read more from oclif docs ).

flybayer:
Sure, some possible examples:

```
blitz db seed
blitz db migrate
blitz db setup
```

tsawan:
@flybayer do you want 'd' alias for 'db'? and how about replacing setup with init since 's' alias is already taken by 'seed'?

flybayer:
Don't alias `db` since it's already so short. Don't add `seed` and `setup` yet, let's wait till we know how to implement those. But `blitz db init` is fine. Might be better than `setup` anyways.

PR #51, opened 2020-03-10
feat(cli): Add `blitz db` command #24

tsawan:
This is PR for issue #24. It supports
- blitz db migrate (or blitz db m)
- blitz db init (or blitz db i)


flybayer:
Looks great! At some point we should create a promise wrapper around `spawn` to make the code cleaner

PR #67, opened 2020-03-12
Architecture experimentation with Store example

flybayer:
### This PR is just experimentation and doesn't necessarily reflect the latest arch design

Note: this store app does not currently run. 

I'm using this to experiment with Blitz app architecture and file structure.


examples/store/app/product/mutations/createProduct.ts
```diff
@@ -0,0 +1,20 @@
+import {createMutation} from '@blitzjs/core'
+import {Product, ProductModel} from 'app/product/ProductModel'
+import db, {ProductCreateInput} from 'prisma/db'
+
+export const createProduct = createMutation<ProductCreateInput, ProductModel>({
+  validateInput: Product.validate,
+  handler: async input => {
```
ryardley: So just had an idea would this sort of thing be better with an express middleware type API? In express you can add as many arguments to use as you want and flow the results through. I think the advantage is we can then add and compose arbitrary filters.

```ts
createMutation(Auth.isAuthorized, Product.validation, () => {
  // do stuff
})
```

ryardley: API might be with helpers to fashion the return output maybe?

```ts
import {next, error} from 'blitzjs/helpers';

function isAuthorised<T,U>(input:T) {
  if(!checkIsAuthorized(input)) {
    return error(msgs.NOT_AUTHORIZED, input);
  };
  return next(input);
} 
```

flybayer: Yeah, that middleware style API could be a good option!

PR #68, opened 2020-03-14
feat(cli): load dependencies in blitz console

quirk0o:


flybayer:
Once you fix the failing test, then we can merge

quirk0o:
Should be fixed now ;) Sorry it took so long!

PR #74, opened 2020-04-02
[RFC] Blitz File Structure & Routing

flybayer:
The purpose of this RFC (Request for Comments) is to gather as much feedback as possible.

We welcome all feedback, whether good or bad! This is your chance to ensure Blitz meets the needs of your company or project.

### 👉 [View the Rendered RFC](https://github.com/blitz-js/blitz/blob/rfc-structure/rfc-docs/02-file-structure-routing.md) 👈

### DIFF
--- a/rfc-docs/02-file-structure-routing.md
+++ b/rfc-docs/02-file-structure-routing.md
@@ -1,0 +1,190 @@
+# [RFC] Blitz File Structure & Routing
+
+**The purpose of this RFC is to gather as much feedback as possible.**
+
+**We welcome all feedback, whether good or bad! This is your chance to ensure Blitz meets the needs of your company or project.**
+
+<hr/>
+
+## File Structure
+
+Blitz is defining a standard file structure to answer the age old question of "How should I organize my react app files?"
+
+#### Guiding Principles
+
+1. Files that change together should live together.
+2. Minimal requirements, maximum flexibility
+
+```
+├── app/
+│   ├── components/
+│   │   ├── Button.js
+│   │   ├── Image.js
+│   │   ├── Input.js
+│   │   ├── Link.js
+│   │   └── Text.js
+│   ├── layouts/
+│   │   ├── Authenticated.js
+│   │   └── Public.js
+│   ├── routes/  (same functionality as Next.js pages/)
+│   │   ├── dashboard.js
+│   │   ├── log-in.js
+│   │   ├── settings.js
+│   │   ├── sign-up.js
+│   │   └── api/
+│   │       └── stripe-webhook.js
+│   ├── marketing/
+│   │   ├── components/
+│   │   │   ├── FeatureSection.js
+│   │   │   ├── Footer.js
+│   │   │   ├── Header.js
+│   │   │   ├── Screenshot.js
+│   │   │   └── Testimonial.js
+│   │   └── routes/  (same functionality as Next.js pages/)
+│   │       ├── about.js
+│   │       ├── features.js
+│   │       ├── index.js
+│   │       └── pricing.js
+│   ├── projects/
+│   │   ├── components/
+│   │   │   ├── Project.js
+│   │   │   ├── ProjectForm.js
+│   │   │   └── Projects.js
+│   │   ├── mutations/
+│   │   │   ├── createProject.js
+│   │   │   ├── createProject.test.js
+│   │   │   ├── deleteProject.js
+│   │   │   ├── deleteProject.test.js
+│   │   │   ├── updateProject.js
+│   │   │   └── updateProject.test.js
+│   │   ├── queries/
+│   │   │   ├── getProject.js
+│   │   │   └── getProjects.js
+│   │   └── routes/  (same functionality as Next.js pages/)
+│   │       └── projects/
+│   │           ├── [id]/
+│   │           │   └── edit.js
+│   │           ├── [id].js
+│   │           ├── index.js
+│   │           └── new.js
+│   ├── tasks/
+│   │   ├── components/
+│   │   │   ├── Task.js
+│   │   │   ├── TaskForm.js
+│   │   │   └── Tasks.js
+│   │   ├── mutations/
+│   │   │   ├── createTask.js
+│   │   │   ├── deleteTask.js
+│   │   │   └── updateTask.js
+│   │   ├── queries/
+│   │   │   ├── getTask.js
+│   │   │   └── getTasks.js
+│   │   └── routes/  (same functionality as Next.js pages/)
+│   │       └── projects/
+│   │           └── [projectId]/
+│   │               └── tasks/
+│   │                   ├── [id]/
+│   │                   │   └── edit.js
+│   │                   ├── [id].js
+│   │                   ├── index.js
+│   │                   └── new.js
+│   └── tests/
+│       ├── userAuthentication.js
+│       └── userOnboarding.js
+├── db/
+│   ├── index.js  (exports your db client)
+│   ├── migrations/
+│   └── schema.prisma
+├── integrations/
+│   └── twilio.js
+├── jobs/
+│   └── generateReport.js
+├── public/
+│   └── favicon.ico
+├── routes/  (same functionality as Next.js pages/)
+├── utils/
+└── blitz.config.js  (same format as next.config.js)
+```
+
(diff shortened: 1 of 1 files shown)


PullJosh:
Overall, this looks awesome!

> While you can place any route files [in the top-level routes directory], we recommend placing route files inside `app`. But if you want, you can instead place all your route files in this top level folders instead of being spread out in various directories inside `app`.

I'm struggling to understand the purpose of the top-level routes folder. If I'm not supposed to use it, why is it there?

flybayer:
@PullJosh good question! 

The top level `routes` folder exists for two reasons:
1. Flexibility — for small, simple apps it can be easier to put all your routes here.
2. Ease of converting an existing Next.js app to a Blitz app.

But this is up for debate! We wouldn't have to support this.

PullJosh:
> 2. Ease of converting an existing Next.js app to a Blitz app.

That second point is pretty compelling. In general, this feels like the kind of decision that I don't want to have to make. But if it's presented as a Next.js transition tool, most decisions are still easy as a Blitz user.

jamonholmgren:
I like pretty much everything you’ve detailed here. I especially like the freedom to compose how I want and then Blitz composes it all together at build time.

I would put in a vote for naming it `pages` rather than `routes` — even before I used Next, I liked that nomenclature.

I also like the top level routes(pages) folder, because I could see a use case for some websites being small enough to not warrant splitting into concerns.

flybayer:
@jamonholmgren awesome!

We're proposing `routes` instead of `pages` because:

1. `pages` makes total sense until you add `pages/api/...`. Then you have non-page API routes inside a pages folders which is not ideal
2. Next.js is planning to rename `pages` to `routes` (and support both for backwards compatibility I'm sure)

Anyone else have thoughts on this part?

PullJosh:
> Next.js is planning to rename pages to routes (and support both for backwards compatibility I'm sure)

Interesting! I didn't know that. Where was that future change announced?

flybayer:
@PullJosh it hasn't been announced and there's no planned timeline for it. The way I understood is that it's something they plan to do at some point.

flybayer:
What do you all think about this for the pages/routes folder?

- Keep the folder named `pages` for all your React pages
- Move the `/api/` folder out of pages so it becomes a sibling of `pages`

This preserve the pages name but removes the discrepancy for the non-page files inside api/

```
app/
   products/
      components/
      pages/
      api/
      queries/
      mutations/
```

Now, most apps won’t need the `api/` folder because most code will be in queries and mutations. You only need the api folder if you need a raw Node.js request handler for something custom.

MrLeebo:
It feels off-putting to see queries and mutations next to api. Initial reaction upon seeing it is "those folders belong under api," at least from a graphql interpretation, which isn't the right interpretation but the naming convention makes it an easy mistake to make.

wKovacs64:
I think I prefer the idea of separating them (`pages` and `api` as siblings). Not a big fan of calling it `pages` when they are combined, though (for the [reasons originally mentioned](https://github.com/blitz-js/blitz/pull/74#issuecomment-609171090)), so would vote for `routes` in that case (the original proposal).

ryardley:
I am not sure I have an answer necessarily but I am wondering is it common enough to make it so routes map along context unless you choose to escape it somehow? 

The problem with this system is that there are often repeated folders within the structure.

Eg.

```
app
├── products
│   └── routes
│       ├── admin
│       |   ├── index.tsx
│       |   └── products
│       |       ├── [id].tsx
│       |       ├── index.tsx
│       |       └── new.tsx
│       └── products
│           ├── [handle].tsx
│           └── index.tsx
├── users
│   └── routes
│       └── users
│           ├── [id]/
│           │   └── edit.tsx
│           ├── [id].tsx
│           ├── index.tsx
│           └── new.tsx
└── routes
    ├── about.tsx
    └── index.tsx
``` 

So you get `app/users/routes/users/[id].tsx`

---

What if we leverage convention and used a special folder (maybe `_` ?) to delineate paths originating from root?

```
app
├── products
│   └── routes
│       ├── [handle].tsx
│       ├── _
│       │   └── admin
│       │       └── products
│       │           ├── [id].tsx
│       │           ├── index.tsx
│       │           └── new.tsx
│       └── index.tsx
└── users
    └── routes
        ├── [id]
        │   └── edit.tsx
        ├── [id].tsx
        ├── index.tsx
        └── new.tsx
```

flybayer:
I really like that proposal @ryardley. I think the common case will indeed be for the url path to match the context path.

We could put the special root symbol directly in the first directory/file name instead of the special root symbol being it it's own directory.

```
app
├── products/
│   └── routes/
│       ├──_admin/
│       │       └── products
│       │           ├── [id].tsx
│       │           ├── index.tsx
│       │           └── new.tsx
│       ├── [handle].tsx
│       └── index.tsx
```

Not sure `_` is the best though since this open signifies something internal/not public.

How about `^`?
```
app
├── products/
│   └── routes/
│       ├──^admin/
│       │       └── products
│       │           ├── [id].tsx
│       │           ├── index.tsx
│       │           └── new.tsx
│       ├── [handle].tsx
│       └── index.tsx
```

Or `;`?
```
app
├── products/
│   └── routes/
│       ├──admin/
│       │       └── products
│       │           ├── [id].tsx
│       │           ├── index.tsx
│       │           └── new.tsx
│       ├── [handle].tsx
│       └── index.tsx
```

ryardley:
> We could put the special root symbol directly in the first directory/file name instead of the special root symbol being it it's own directory.

Hmm interesting idea.

I guess it is a personal taste thing and I can see how that could flatten the folder structure but in my head that is an extra mapping and could get noisey with lots of root folders.

```
app
├── widgets/
│   └── routes/
│       ├──^admin/
│       ├──^checkout/
│       ├──^affiliate/
│       ├── [handle].tsx
│       └── index.tsx
```

For me once I am in the `_` folder I am in 'root' land which works basically like next `/pages` with no special rules so it seems simpler.

flybayer:
I just updated this file structure RFC with the following changes:

1. Rename `routes` folder to `pages`
2. Move `routes/api` folder to `api` (`pages` and `api` are now siblings instead of nested)

I think we should hold off on @ryardley' suggestion for changing the url root for nested pages folders. Because, while it reduces nesting for some cases, it increases nesting for others. For example, the file structure example in the RFC has `app/tasks/pages/products/[productId]/tasks/index.ts` which would have to change to `app/tasks/pages/_/products/[productId]/tasks/index.ts`. 

But perhaps there's another way to reduce nesting we haven't thought about that's even more elegant.

sijad:
is it possible to define `_app` and `_document` per `pages/` directories?

is it possible to have `components` folder inside a page and ignored as a route?

```
├── pages
    └── mypage
        ├── components
        │   └── MyPageItem.tsx
        └── index.tsx
```

flybayer:
> is it possible to define `_app` and `_document` per `pages/` directories?

Yes, the following works just like Next.js:
```
app/pages/_app.js
app/pages/_document.js
```

> is it possible to have `components` folder inside a page and ignored as a route?
> 
> ```
> ├── pages
>     └── mypage
>         ├── components
>         │   └── MyPageItem.tsx
>         └── index.tsx
> ```

Currently no, but we could add support to prefix folders with `_`. So you could put `_components/` inside pages without it being counted as a page.

ryardley:
> is it possible to define _app and _document per pages/ directories?

They should be able to be defined in any `pages` folder however I would imagine that best practice would be to keep them in your global `/pages` folder.

dwightwatson:
I find `integrations` to be a little awkward.

I guess `integrations` it's sort of like the `initializers` directory of a Rails app where you're meant to configure third-party clients before the app has booted? Will everything in that directory be auto-loaded by default? The name feels a little too open-ended and I wonder if it even needs to be there by default.

flybayer:
@dwightwatson yes, sort of like Rails, but not autoloaded.

We may drop it as a default before 1.0 if we don't use it, but I expect our installers (RFC #242) will put any type of third-party integration in there.

jamonholmgren:
I'm a big fan of the `/pages` and `/api` siblings. Feels great to me.

flybayer:
This is now documented at [blitzjs.com](https://blitzjs.com)

PR #95, opened 2020-04-11
feat: Implement the new Architecture!

flybayer:
Woohoo!! Finally bringing the vision to life!

- [x] RPC working for queries and mutations with hard coded rpc files
- [x] Change RPC stuff to be generated in the build
- [x] Ensure static build is working?


package.json
```diff
@@ -25,7 +25,7 @@
     "cross-env": "^7.0.0",
     "debug": "^4.1.1",
     "husky": "^4.2.3",
-    "jest": "25.3.0",
+    "jest": "24.9.0",
```
flybayer, line 28: @ryardley is the issue with jest 25?

ryardley: I was finding having multiple versions of jest seemed to cause some tests that should have failed to pass. We should try and upgrade it all to 25 to avoid the warning with `ts-jest` but I vaguely remember having issues with that.

ryardley:
Hey @flybayer how do you want to do this static data build? Normally you would go to a remote database somewhere right? Should we be having an opinion here?

package.json
ryardley: Yes I just checked and tsdx can't handle the newer jest as it throws these errors: 

```
TypeError: Cannot read property 'run' of undefined
    at /Users/rudiyardley/projects/blitz/blitz/node_modules/tsdx/dist/index.js:418:20
error Command failed with exit code 1.
```

ryardley: I just downgraded ts-jest and everything appears to be working with no warnings.

flybayer:
@ryardley I’m ready to merge this. Are you?

ryardley:
Let's do it!

PR #106, opened 2020-04-13
feat(server): Warning about invalid routes/api structure

### LINKED ISSUE #55: Make blitz server stop and log a nice error if it finds duplicate react pages
Ensure that Blitz throws and error when there are duplicate pages within the app folder.

merelinguist:
Closes #55 

This is my first attempt at providing a warning if duplicate / overlapping routes are created. N.B. this is my first time looking at the server code so it might be totally off the mark!

**Before:** routes are created and duplicates are silently swallowed.

**After:** user gets a pretty warning about duplicate routes.

carbon


ryardley:
Thanks heaps for the PR @merelinguist!

packages/server/src/synchronizer/index.ts
```diff
@@ -58,6 +59,28 @@ export async function synchronizeFiles({
...
+
+    duplicateRoutesPaths.forEach((route) => {
+      route.forEach((path) =>
+        console.log(`- ${path.split('routes/')[0]}routes/${chalk.bold.redBright(path.split('routes/')[1])}`),
+      )
+    })
+
+    process.exit(0)
+  }
+
```
ryardley, line 83: This drafts this out and is really helpful. 

Couple of things I am thinking:
1. What do you think about putting this all into its own function?
1. As this is an invariant should this all happen before everything else?
1. Can/should this be generalised? precondition throws error which is caught and then displayed nicely. The answer could be no/later.

yarn.lock
```diff
@@ -2232,13 +2232,13 @@
     once "^1.4.0"
 
 "@octokit/request@^5.2.0":
-  version "5.3.4"
```
ryardley, line 2235: Maybe all this stuff can stay out of this PR.

packages/cli/README.md
```diff
@@ -36,7 +36,7 @@ $ npm install -g @blitzjs/cli
 $ blitz COMMAND
 running command...
 $ blitz (-v|--version|version)
-@blitzjs/cli/0.0.1-canary.1 darwin-x64 node-v12.16.1
+@blitzjs/cli/0.0.1-canary.1 darwin-x64 node-v13.9.0
```
ryardley, line 39: This stuff is noise too.

packages/cli/src/commands/db.ts
```diff
@@ -16,10 +16,14 @@ export default class Db extends Command {
...
-      const cp = spawn('prisma', ['migrate', 'save', '--experimental'], {stdio: 'inherit'})
+      const cp = spawn('prisma', ['migrate', 'save', '--experimental'], {
+        stdio: 'inherit',
+      })
       cp.on('exit', (code: number) => {
         if (code == 0) {
-          const cp = spawn('prisma', ['migrate', 'up', '--experimental'], {stdio: 'inherit'})
+          const cp = spawn('prisma', ['migrate', 'up', '--experimental'], {
+            stdio: 'inherit',
+          })
```
ryardley, line 26: Also here is noise

flybayer:
The CLI output is gorgeous!! Feel free to open other PRs for improving all the CLI output! :)

packages/server/src/synchronizer/index.ts
```diff
@@ -58,6 +59,28 @@ export async function synchronizeFiles({
 
   await clean(destPath)
 
+  const allRoutes = entries.filter((route) => route.includes('routes'))
```
ryardley, line 62: Routes folder can be `routes` or `pages`

packages/cli/README.md
flybayer: @merelinguist is your editor set up to use the `prettier.config.js` in the project root? That should keep this from happening.

Also, the husky pre-commit hook should be running to format it according to `prettier.config.js`

yarn.lock
```diff
@@ -2232,13 +2232,13 @@
...
-  resolved "https://registry.yarnpkg.com/@octokit/request/-/request-5.3.4.tgz#fbc950bf785d59da3b0399fc6d042c8cf52e2905"
-  integrity sha512-qyj8G8BxQyXjt9Xu6NvfvOr1E0l35lsXtwm3SopsYg/JWXjlsnwqLc8rsD2OLguEL/JjLfBvrXr4az7z8Lch2A==
+  version "5.4.0"
+  resolved "https://registry.yarnpkg.com/@octokit/request/-/request-5.4.0.tgz#52382c830f0cf3295b5a03e308872ac0f5ccba9b"
+  integrity sha512-uAJO6GI8z8VHBqtY7VTL9iFy1Y+UTp5ShpI97tY5z0qBfYKE9rZCRsCm23VmF00x+IoNJ7a0nuVITs/+wS9/mg==
```
flybayer, line 2237: I think pinning all the dependencies to a specific version, instead of a range, would keep this from happening to any of us (not something you caused @merelinguist)

packages/server/src/synchronizer/index.ts
merelinguist: Done 👍

packages/cli/README.md
merelinguist: Fixed, thanks!

packages/cli/src/commands/db.ts
merelinguist: Fixed.

packages/server/src/synchronizer/index.ts
merelinguist: Is it _either_ routes or pages, or could someone have both `routes` and `pages` in the same app?

yarn.lock
ryardley: I would have thought the yarn.lock file should prevent carrotted events from upgrading automatically. Got a feeling that some kind of accidental dep upgrade possibly happened. Could the cli build have caused this?

packages/server/src/synchronizer/index.ts
flybayer: I think we should abandon supporting both. It's added complexity in our code that we don't have to have since we are a greenfield project :)

Note, we are changing back to `pages` from `routes` 😏 #107

merelinguist: Okay cool! For now should I just leave as is?

ryardley: Hmm good question. At this point I am making the assumption it could be both in the same app. Perhaps you have a legacy next app and you want your blitz modules to use `routes`. If we disallow one we should have a warning saying the user has accidentally used the other.  I kind of think it is better to just allow both as it helps when people are refactoring although I can see an argument for making it one or the other.

yarn.lock
merelinguist: I think the problem was that I deleted yarn.lock while trying to figure out the prisma enoent problem. I’ve just reverted the file for safety.

packages/server/src/synchronizer/index.ts
ryardley: Ok that's great @flybayer I will adjust the pages rules.

yarn.lock
merelinguist: And yes, I think normally the lock file would prevented careted packages being upgraded randomly.

packages/server/src/synchronizer/index.ts
merelinguist: @flybayer Up to you! 😄

flybayer: Yeah, let's only support `pages`. So @merelinguist you can change this to be pages.

The change in #107 affects two things here:
- Now also need to detect duplicates for `api` directories scattered throughout the app
- Need to warn the developer if they try to use `pages/api`

merelinguist:
@flybayer Have now finished:

- [ ]  Duplicate api routes throughout the app are detected
- [ ] Nested `pages/api` routes are detected

packages/server/src/synchronizer/check-duplicate-pages.ts
```diff
@@ -0,0 +1,23 @@
+import chalk from 'chalk'
+
+export const checkDuplicatePages = (entries: string[]) => {
+  const allPages = entries.filter((page) => page.includes('pages'))
```
ryardley, line 4: One last thing hope you don't mind. This one looks a lot like the other one. Any chance we can create a shared function?

merelinguist: Good idea, done!

merelinguist:
@ryardley All done!

packages/server/src/synchronizer/check-duplicate-routes.ts
```diff
@@ -0,0 +1,41 @@
+import chalk from 'chalk'
+
+const getDuplicatePaths = (entries: string[], type: string) => {
```
ryardley, line 3: This is perfect!

packages/server/src/synchronizer/check-duplicate-routes.ts
```diff
@@ -0,0 +1,41 @@
...
+  if (duplicatePagesPaths.length > 0) {
+    console.log(chalk.yellow('\n Warning: You have created conflicting pages: \n'))
+
+    duplicatePagesPaths.forEach((page) => {
+      page.forEach((path) => console.log(buildErrorMessage(path, 'pages')))
+    })
+
+    process.exit(0)
+  }
+}
```
ryardley, line 41: Would it be neater to split these into two methods? Seems they are more versatile that way.

ryardley:
Just to let you know @merelinguist I have been doing some refactoring of the synchroniser so we can put things that are related together in a rule so once this is merged I can add these invariants to their appropriate rules.

packages/server/src/synchronizer/check-duplicate-pages.ts
```diff
@@ -0,0 +1,23 @@
+import chalk from 'chalk'
```
ryardley, line 1: Should this be deleted?

packages/server/src/synchronizer/check-duplicate-api-routes.ts
```diff
@@ -0,0 +1,25 @@
+import chalk from 'chalk'
+
+export const checkDuplicateApiRoutes = (entries: string[]) => {
```
ryardley, line 3: Should this be deleted?

packages/server/src/synchronizer/check-duplicate-pages.ts
merelinguist: Whoops!

ryardley:
Hi @flybayer just in case you missed it we still need a review to merge. This isn't the final resting place for this code but I would like to incorporate this into https://github.com/blitz-js/blitz/pull/110 which has a better way of handling our build rules.

flybayer:
@all-contributors add @merelinguist for code

PR #109, opened 2020-04-13
fix(cli): spawn Windows .cmd entrypoints on win32 platform

wKovacs64:
Closes #108

Issue #108, opened 2020-04-13
[cli] db and test commands don't work on Windows


wKovacs64:
yarn blitz db migrate

```
Error: spawn prisma ENOENT
    at Process.ChildProcess._handle.onexit (internal/child_process.js:267:19)
    at onErrorNT (internal/child_process.js:469:16)
    at processTicksAndRejections (internal/process/task_queues.js:84:21)
```

yarn blitz test

```
Error: spawn npm ENOENT
    at Process.ChildProcess._handle.onexit (internal/child_process.js:267:19)
    at onErrorNT (internal/child_process.js:469:16)
    at processTicksAndRejections (internal/process/task_queues.js:84:21)
```

(Side note: `test` tried to run `npm` instead of `yarn`, but that's another issue.)

PR incoming.

PR #122, opened 2020-04-14
fix(cli): leverage cross-spawn for cross-platform compatibility

wKovacs64:
I ran into problems with `yarn blitz start` and tracked it down to the same root cause as #108. Rather than duplicate logic similar to #109, I figured we should just let [`cross-spawn`](https://www.npmjs.com/package/cross-spawn) resolve the appropriate executable for us. This effectively reverts #109 and replaces it with a more comprehensive solution (IMHO).

Caveat: going forward, if we need to spawn processes, we need to remember to import from `cross-spawn` rather than `child_process`.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -30,6 +30,7 @@
     "@oclif/plugin-not-found": "^1.2.3",
     "chalk": "^4.0.0",
     "chokidar": "^3.3.1",
+    "cross-spawn": "^7.0.2",
     "diff": "^4.0.2",
     "enquirer": "^2.3.4",
     "fs-extra": "^9.0.0",
@@ -43,7 +44,7 @@
   "devDependencies": {
     "@oclif/dev-cli": "^1.22.2",
     "@oclif/test": "^1.2.5",
-    "@types/cross-spawn": "6.0.1",
+    "@types/cross-spawn": "^6.0.1",
     "@types/diff": "^4.0.2",
     "@types/fs-extra": "^8.1.0",
     "@types/fs-readdir-recursive": "1.0.0",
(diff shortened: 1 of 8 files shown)


ryardley:
This might also be needed in `server/next-utils`? Also is it still possible to run tests on windows? There is quite a bit of spawn mocking going on there.

wKovacs64:
> This might also be needed in `server/next-utils`?

Hm, I do see the `child_process` spawn calls in there now (I was focused on `@blitzjs/cli` earlier), although I was able to get things working without changing it in that file. But we probably should anyway, for consistency if nothing else. New commit added.

> Also is it still possible to run tests on windows? There is quite a bit of spawn mocking going on there.

I've never gotten all the tests to pass on Windows - working on that now. Almost done, except for [`packages/server/test/rules.test.ts`](https://github.com/blitz-js/blitz/blob/4f21628365bb487a3f9498fe26e7e46c62f992ed/packages/server/test/rules.test.ts), which I haven't been able to resolve yet. I mentioned it in Slack (not sure if it needs its own issue?), but the gist is the [`remove`](https://github.com/blitz-js/blitz/blob/4f21628365bb487a3f9498fe26e7e46c62f992ed/packages/server/test/rules.test.ts#L35) call can't be done while the watcher is running or it will throw an `EPERM: operation not permitted` error. Stopping the watcher before that call eliminates that error, but the tests fail as the expected directory tree differs from the actual one. Would love another set of eyes on it!

ryardley:
The new architecture for the synchroniser night fix that as I’m not going to be running the watcher in tests

flybayer:
@ryardley These changes look good to me, so it's up to you to merge when you're ready

PR #124, opened 2020-04-15
Generate prisma client into project instead of node_modules

flybayer:
It turns out that generating prisma client into `node_modules/@prisma/client` is flaky in that it often get's blown away when installing or updating dependencies.

So Blitz apps will generate it into the project, but in a hidden folder: `db/.generated-prisma-client`

Currently not working because of a prisma bug

- [ ] Get prisma actually working
- [ ] Update new app template

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,5 @@ dist
 **/.env.development.local
 **/.env.test.local
 **/.env.production.local
+
+.generated-prisma-client
--- a/examples/store/.gitignore
+++ b/examples/store/.gitignore
@@ -8,11 +8,11 @@
 # testing
 /coverage
 
-# next.js
-/.next/
-/out/
-/.blitz/
+# blitz.js
+.next
+.blitz
 *.sqlite
+.generated-prisma-client
 
 # production
 /build
--- a/examples/store/app/admin/pages/admin/products/[id].tsx
+++ b/examples/store/app/admin/pages/admin/products/[id].tsx
@@ -5,7 +5,7 @@ import ProductForm from 'app/products/components/ProductForm'
 
 function Product() {
   const router = useRouter()
-  const id = parseInt(router.query.id as string)
+  const id = parseInt(router?.query.id as string)
   const [product] = useQuery(getProduct, {where: {id}})
 
   return <ProductForm product={product} onSuccess={() => router.push('/admin/products')} />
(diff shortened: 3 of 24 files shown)


flybayer:
So the main issue here is `__dirname` not working right with Next.js.

- Maybe the different is the client is getting transpiled now since not in node_modules?
- Can we fix this with webpack?
- Could swap out `__dirname` in our server transform with a hard coded value, but only for development?
- Need to see if it's only broken in dev, but not production

flybayer:
Update: I got it working!

flybayer:
Actually not...

flybayer:
Well, I'm giving up on this for now.

- Adding the client to webpack externals makes it work in development, but that fails in production since it's not bundled
- Perhaps a webpack think like this would work? https://gist.github.com/timiscoding/2416182c5d7b3ce68966883906cb1f57
	- I tried this but was still getting errors

sijad:
can I take a look? if yes, let me know how can I reproduce it please

flybayer:
@sijad you sure can!!

Steps to reproduce are here: https://github.com/prisma/prisma/issues/2195

flybayer:
So this branch as of right now does work properly on the local machine with `blitz start` and `blitz start --production`.

But it is still failing a production deployment to Zeit with:
```
Invalid `prisma.product.findMany()` invocation:
Query engine binary for current platform "rhel-openssl-1.0.x" could not be found.
This probably happens, because you built Prisma Client on a different platform.
(Prisma Client looked in "/zeit/59caaf16/.blitz/caches/build/db/.generated-prisma-client/runtime/query-engine-rhel-openssl-1.0.x")
Files in /zeit/59caaf16/.blitz/caches/build/db/.generated-prisma-client/runtime:

```

Not sure how to fix this. Might be as easy as copying moving some files around.

sijad:
this can be related to https://github.com/zeit/now/issues/2824

flybayer:
No it's only related to our build step and how we are moving files around. That error shows that there are no files where it's looking. If it was a native binary issue, the wrong binary would show there

sijad:
is it possible to download possible binaries using [@prisma/fetch-engine](https://github.com/prisma/prisma/tree/master/src/packages/fetch-engine) and include all of them in production?

flybayer:
@sijad so the issue is that there are _no_ binaries. Not that it has the _wrong_ binaries.

The webpack plugin added in this PR substitutes __dirname for what it is at build time. And at build time the files are in `.blitz`. But the actual build output for Zeit is in `.next`. So we might need to change the webpack plugin to use the `.next` path instead.

But in any case, this is low priority right now, so I'm not too worried about it.

PR #128, opened 2020-04-15
Add the `blitz` package and proper global/local CLI handling

### LINKED ISSUE #126: [Help Wanted!] Make CLI work from the `blitz` package and with npx
We need the following:

- [x] Be able to install the blitz cli globally with yarn/npm from the `blitz` package instead of `@blitzjs/cli`
- [x] `npx blitz --help` to work
- [x] The globally installed binary delegates to the locally installed version in the app. So `blitz start` in an app will use the blitz version in package.json.

This is needed for our imminent alpha release, so I'd love if someone can jump on this!

flybayer:
This PR does the following:

- [x] Adds the `blitz` parent package
- [x] `blitz` has a CLI that can be used when install globally
- [x] The `blitz` CLI runs the locally installed CLI version if it exists
- [x] Deletes the outdated `vanilla-next` example app
- [x] Makes some internal DX improvements
- [x] Fixes a server issue where it wasn't using the local version of next (this is fixed by `resolveCwd()` instead of `resolve()` which resolves from the package, not the project.
- [x] Fixes a server issue where it wasn't properly finding next-patched bin
- [x] Fixes prisma spawn from global CLI

Closes #126


sijad:
1. there's an open issue about first issue https://github.com/jaredpalmer/tsdx/issues/275

2. second issue is related to https://github.com/blitz-js/blitz/blob/80a5a1f63db2f5e79362cb294a2fa2eb27595c76/packages/server/src/config.ts#L46
  in dev mode it refers to `blitz/examples/store/node_modules/.bin/next-patched` but this file does not exists...

ryardley:
> 2. second issue is related to https://github.com/blitz-js/blitz/blob/80a5a1f63db2f5e79362cb294a2fa2eb27595c76/packages/server/src/config.ts#L46
> 
>   in dev mode it refers to `blitz/examples/store/node_modules/.bin/next-patched` but this file does not exists...

Probably need to apply `requireBinAsync` there too

medelman17:
IMHO, there's almost no value add in using `tsdx` for the `blitz` 'all-in-one' package given that, apart from the 'cli-wrapper' that I included above (to enable global and local cli-usage), the lib's sole other job is to export other blitz packages. So, I stripped it out and went with a straight-forward `rollup` setup that makes building both the cli-wrapper and library a breeze. 

Regarding the second issue above, I added a 'does this patched-next file exist' check to the `config.ts` file, which solves the 'this won't run in dev mode' issue. This had the unfortunate consequence of causing a test to fail, which I, then, skipped. I don't quite follow the intent of the code and or why we would be testing this functionality. So, please excuse any overreach, here.

flybayer:
WOOHOO! This PR is finally ready to merge!

Please review @ryardley 

The `blitz` package set up was mostly from @medelman17, so a huge thanks to him.

package.json
```diff
@@ -16,35 +16,50 @@
...
     "prettier": "^2.0.4",
     "pretty-quick": "2.0.1",
+    "prompt": "^1.0.0",
+    "rollup": "^2.6.1",
+    "rollup-plugin-commonjs": "^10.1.0",
+    "rollup-plugin-json": "^4.0.0",
+    "rollup-plugin-node-polyfills": "^0.2.1",
+    "rollup-plugin-node-resolve": "^5.2.0",
+    "rollup-plugin-peer-deps-external": "^2.2.2",
+    "semver": "^7.3.2",
```
ryardley, line 52: Out of interest/curiosity why are we installing everything in the root? Was workspaces screwing up stuff?

packages/cli/test/commands/db.test.ts
```diff
@@ -13,14 +13,15 @@ jest.doMock('cross-spawn', () => ({spawn}))
 import DbCmd from '../../src/commands/db'
 
 const schemaArg = `--schema=${path.join(process.cwd(), 'db', 'schema.prisma')}`
+const prismaBin = path.join(process.cwd(), 'node_modules/.bin/prisma')
```
ryardley, line 16: Should we be using `resolveCwd` here at all?

package.json
flybayer: I'm not sure. @medelman17 did this. Why in the root vs in the package Micheal?

packages/cli/test/commands/db.test.ts
ryardley: ah just noticed this was in a test - all good.

packages/blitz/src/bin/cli.ts
```diff
@@ -0,0 +1,61 @@
...
+  usageType = 'global'
+  pkgPath = globalCLIPkgPath
+} else {
+  usageType = 'global-linked'
+  pkgPath = globalLinkedCLIPkgPath
+}
+
+// TODO: remove
+console.log('DEBUG:', usageType)
+console.log('Path:', pkgPath)
```
ryardley, line 43: When do we remove this?

flybayer: After I publish everything NPM and see if it's working right

packages/cli/test/commands/db.test.ts
flybayer: Good call. I just pushed a commit that changes this to use resloveBinAsync from the server package.

package.json
flybayer: He says all dev dependencies should be installed at root so that everything absolutely stays in sync and uses the same version. Makes sense to me.

PR #130, opened 2020-04-16
feat(server): `blitz.config.js` config support without `withBlitz`

ryardley:
1. Will warn when two configurations are present.
1. Allows either `blitz.config.js` or `next.config.js` to be used in place of next config. 

Only supports javascript in configurations
We should ask users permission to rename the file in a separate PR


flybayer:
We should remove support for blitz.config.js.

ryardley:
No worries

flybayer:
Ahh, did you miss my edit? My first comment was "remove blitz.config.js" but I immediately edited the comment to be "remove next.config.js"

I want to support "blitz.config.js" but not "next.config.js"

Sorry about that!!

ryardley:
That’s what this does

ryardley:
I had a stray blitz config in the example app

PR #139, opened 2020-04-17
feat(cli): add help for db commands

### LINKED ISSUE #135: Add help information to `blitz db` command
Currently `blitz db` and `blitz db -h` both print an error about an unsupported command.

Both of these should print a list of supported commands.

The alpha user guide lists the supported `blitz db` commands: https://github.com/blitz-js/blitz/blob/flybayer-patch-1/USER_GUIDE.md#blitz-cli-commands

sijad:
fixes #135

### DIFF
--- a/packages/cli/src/commands/db.ts
+++ b/packages/cli/src/commands/db.ts
@@ -1,9 +1,22 @@
 import {spawn} from 'cross-spawn'
+import {flags} from '@oclif/command'
 import {Command} from '@oclif/command'
+import chalk from 'chalk'
 import * as path from 'path'
 
 export default class Db extends Command {
-  static description = 'Run database commands'
+  static description = `Run database commands
+
+${chalk.bold('migrate')}   Run any needed migrations via Prisma 2 and generate Prisma Client.
+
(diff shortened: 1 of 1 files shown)


sijad:
I think it's better to use [topics](https://oclif.io/docs/topics) but commands would be like `db:migrate`, `db:introspect`

flybayer:
@sijad yeah, I really want the "db migrate" format instead of "db:migrate".

At some point we need to build a nice abstraction that makes it easy to support these without ugly code :)

flybayer:
Can you add more detailed help for each db command? I.e. basically what I have here: https://github.com/blitz-js/blitz/blob/flybayer-patch-1/USER_GUIDE.md#blitz-cli-commands

tsawan:
would be nice to support "db migrate" and "db m".

sijad:
currently it looks like this:

2020-04-17-16:49:23-screenshot

please let me know what you think

flybayer:
Yeah, that works. Thank you!!

@all-contributors add @sijad for code

PR #140, opened 2020-04-18
Skeleton for Webpack Config

dkempner:


flybayer:
@all-contributors add @dkempner for code

flybayer:
@dkempner so your approach didn't work. But the good news is I did get it to work with the following:

```js
    if (options.isServer) {
      config.externals = config.externals || []
      config.externals.push({
        '.generated-prisma-client': path.resolve(__dirname, '.generated-prisma-client'),
      })
    }
```

I appreciate your help here!

dkempner:
Awesome @flybayer good work!

PR #143, opened 2020-04-18
Fix server for deployment to Zeit

flybayer:
Zeit automatically adds the serverless target to next.config.js, so it's always present during Zeit build time. Blitz build was yelling because that file shouldn't be there.

And we do actually need that config that Zeit sets, so this will preserve any config Zeit sets.

This is done except the test is failing and I can't figure out why. @ryardley could you take a look at it?

### DIFF
--- a/packages/server/src/synchronizer/rules/config.ts
+++ b/packages/server/src/synchronizer/rules/config.ts
@@ -1,4 +1,4 @@
-import {readFile, writeFile, rename} from 'fs-extra'
+import {readFile, writeFile, copyFile} from 'fs-extra'
 import {resolve} from 'path'
 import File from 'vinyl'
 import {Rule} from '../types'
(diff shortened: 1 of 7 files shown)


ryardley:
Ok. Where does the NOW_BUILD env var come from? Is that a zeit thing?

flybayer:
Yes, Zeit always provides it during build time

packages/server/src/synchronizer/rules/config.ts
```diff
@@ -25,7 +25,20 @@ export async function copyConfig(entries: string[], srcPath: string, destPath: s
 
   await writeFile(resolve(destPath, 'blitz.config.js'), Buffer.from(fileContents))
 
-  const nextConfigShellTpl = `
+  // Zeit now adds configuration needed for Now, like serverless target,
+  // so we need to keep and use that
+  if (process.env.NOW_BUILDER) {
+    rename(resolve(srcPath, 'next.config.js'), resolve(destPath, 'next-zeit.config.js'))
```
ryardley, line 31: Problem is you are missing an await here I think

ryardley: Also we can't rename here. We can't alter the source folder as this is run after the entries have been taken.

ryardley: Will push a fix

flybayer:
@ryardley great, thank you!!

Issue #56, opened 2020-03-10
Improve Blitz Server Efficiency/performance


ryardley:
Avoid copying files that are the same. 
Avoid upfront copying when starting the server. 
Test with a large next codebase with a few hundred pages.

PR #146, opened 2020-04-19
Refactor synchronizer

### LINKED ISSUE #107: Change blitz server to support `pages` and `api` as siblings instead of nested
We just updated the file structure RFC to switch from `routes` and `routes/api` to `pages` and `api`: https://github.com/blitz-js/blitz/pull/74

- The server already supports the `pages` directory
- [ ] `api` directory at top level or nested anywhere inside `app/` should be moved to `pages/api` inside the blitz build directory.
- [x] Presence of `pages/api/` should error.

ryardley:
### Pull Request Type

- [ ] Feature
- [ ] Bug fix
- [ ] Build related changes
- [ ] Tests related changes

### Checklist

Synchronizer was not structured in an extensible way that allows us to move forward quickly.

- [x] Agnostic input file watcher / glob = no watcher in unit tests
- [x] Define stream error rendering
- [x] Related logic needs to live together
- [x] Everything is a stream
- [x] More efficiency potential
- [x] Parallel processing
- [x] Cleaner Architecture
- [x] Prepare for dirty restart functionality
- [x] Migrate existing functionality over to streams 
  - [x] Incorporate zeit now build logic
  - [x] Incorporate duplicate page detection logic
  - [x] Add `app/foo/api/bar` -> `pages/api/bar` to paths rule
- [x] Simplify tests
- [x] Docs
- [x] Move deps to root
- [x] Support `VERCEL_BUILDER`
- [x] "stuck" on the last file progress indicator
- [x] ~~Cross package loader animation~~ (probably best done in the cli package)

- [ ] Yes
- [x] No

Closes #107 

# Synchronizer Refactor

# [Rendered docs](https://github.com/blitz-js/blitz/blob/refactor-synchronizer/packages/server/src/synchronizer/README.md)

Design goals

- [x] Rules: Related logic needs to live together
- [x] Everything is a rule
- [x] Efficiency
- [x] Parallel processing via streams
- [x] Cleaner Architecture to allow for future Dirty restart functionality
- [x] Agnostic input file watcher / glob
- [x] Define error decoration pattern

Tasks

Fuzzy / Concerns 

* How do we handle calls to `process.exit()`? Currently I handle exit in a try catch for upfront stuff that prevents execution but that is wrong as it doesn't take into account what happens after the promise has resolved and the watcher 

Things to look at still possibly later PR

* Use in memory fs for tests


flybayer:
It'd be really great if you can also add documentation to all of this! I have no idea what's going on, when looking in from the outside 😅  Ideally some high level docs and then also some file level docs for the workhorse files.

ryardley:
Yep was planning to here is a starter: https://github.com/blitz-js/blitz/blob/3b1dd9071cc10e33abf0c5da1e1f130fa54e8d81/packages/server/src/synchronizer/refactor.md

ryardley:
Will also make a diagram

ryardley:
We're really close on this jst a few more things to do and a bunch of docmentation.

flybayer:
You are our resident wizard!! 🧙 😄

ryardley:
Ok I think the code is ready. Please test for bugs/regressions. I will try and work through some docs this afternoon. Please add any questions to the PR and I will try and include in docs if relevant.

packages/server/package.json
```diff
@@ -49,6 +54,24 @@
...
+    "@types/pumpify": "^1.4.1",
+    "@types/readable-stream": "^2.3.5",
+    "@types/through2": "^2.0.34",
+    "@types/vinyl-fs": "^2.4.11",
+    "directory-tree": "^2.2.4",
+    "husky": "^4.2.3",
+    "tsdx": "^0.13.1",
+    "tslib": "^1.11.1",
+    "typescript": "^3.8.3"
+  },
```
flybayer, line 75: Can you move all these up to the root? (I think most are already there)

packages/server/src/synchronizer/pipeline/rules/config/index.ts
```diff
@@ -0,0 +1,82 @@
...
+      }),
+    )
+  }
+
+  // No need to filter yet
+  const stream = through({objectMode: true}, (file: File, _, next) => {
+    if (!isNextConfigPath(file.path)) return next(null, file)
+    // Zeit now adds configuration needed for Now, like serverless target,
+    // so we need to keep and use that
+    if (process.env.NOW_BUILDER) {
```
flybayer, line 57: We need to support `VERCEL_BUILDER` here too. Or instead, just check for `next.config.js` and use that regardless of the env variable. They won't be able to run blitz in dev mode with that file, so I think this would be fine.

flybayer:
The new blitz progress thing is sweet! One thing I'm seeing tho is that it often appears to get "stuck" on the last file. I'm assuming it's processing something else, so it'd be nice if it can show that with a spinner instead of the file path

flybayer:
Everything working so far! Need to test a zeit deployment tho

flybayer:
There's a small pause in `blitz start` before the server progress appears because it's generating the prisma client there (`commands/start.ts`). Is there some way we can use that same loader for the prisma client generation so there's not this awkward pause?

ryardley:
Ok added docs and some JavaDoc comments for editors such as VSCode. Perhaps it might be good to have @merelinguist have a look at this to see what they think about the view rendering stuff?

packages/server/src/synchronizer/pipeline/rules/config/index.ts
ryardley: Hmm probably easier at this point to check for either vars

packages/server/src/synchronizer/errors.ts
```diff
@@ -0,0 +1,72 @@
...
+  process.stdout.write('\n')
+
+  err.routes.forEach((route) => {
+    console.log(`- ${removeCwd(route)}`)
+  })
+
+  process.stdout.write('\n')
+
+  error(err.secondary)
+}
```
ryardley, line 72: @merelinguist Might be good if you can grok this and see if it makes sense.

packages/server/src/synchronizer/reporter.ts
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    // Capture last event incase we need to tidy up the console
+    lastEvent = event
+
+    next()
+  })
+
+  return {stream}
+}
```
ryardley, line 51: @merelinguist Might be good if you can check this out too

ryardley: If you squint it is sort of working like Redux...

packages/server/src/synchronizer/errors.ts
ryardley: If anything doesn't make sense I can explain and add it to the docs.

ryardley:
> The new blitz progress thing is sweet! One thing I'm seeing tho is that it often appears to get "stuck" on the last file. I'm assuming it's processing something else, so it'd be nice if it can show that with a spinner instead of the file path

What this is that blitz has a ready event that is fired based on input stream inactivity. Maybe the easiest thing right now is to lower the threshold.

ryardley:
> There's a small pause in blitz start before the server progress appears because it's generating the prisma client there (commands/start.ts). Is there some way we can use that same loader for the prisma client generation so there's not this awkward pause?

This could be done completely in the CLI package perhaps we can do that in a separate PR?

flybayer:
Does this also close #56?

ryardley:
> Does this also close #56?

I would probably say it lays out the groundwork but does not finish it.

packages/server/src/synchronizer/errors.ts
merelinguist: Okay, this looks good! Took a second to see how everything fits together but I really like the way this is laid out 👍

packages/server/src/synchronizer/reporter.ts
merelinguist: Hm do the events just come from us, or from Next/elsewhere?

merelinguist:
@flybayer What’s left to be done in #56? I could do one of those today.

merelinguist:
Also, @flybayer @ryardley, I just tested this with an app of 10000 pages:

```js
// createDummyPages.js

const crypto = require('crypto')
const fs = require('fs')

const times = 10000
for (let i = 0; i < times; i++) {
  const randomString = crypto.randomBytes(20).toString('hex')
  fs.writeFileSync(
    `./examples/store/app/products/pages/${randomString}.tsx`,
    `export default () => <h1>${randomString}</h1>`,
  )
}
```

And noticed no real difference in speed between that and the regular app of just a few pages! ✨

packages/server/src/synchronizer/reporter.ts
ryardley: They come from us in the rules. You can search for the event type to find the source. I might have a play after this is merged with trying to get some oversight into next

ryardley:
> @flybayer What’s left to be done in #56? I could do one of those today.

It's a big job I think. We should merge this first and tackle it separately in my opinion.

In my head:
- [.] Build metrics to report how long a sync took - test for this in a unit test
- [ ] Test against a huge open source next project with the sync report so we get a good idea of the problem
- [ ] Use streaming vinyl file contents https://github.com/gulpjs/vinyl#filecontents
- [ ] Solve dirty start problem by setting up a better output cache and doing some sort of file analysis there is a sketch of what it might look like [here](https://github.com/blitz-js/blitz/blob/refactor-synchronizer/packages/server/src/synchronizer/README.md#future-thinking)
- [ ] Remove the clean function from each run

flybayer:
> Also, @flybayer @ryardley, I just tested this with an app of 10000 pages:

@merelinguist WHOA!! Sweeet

ryardley:
> Also, @flybayer @ryardley, I just tested this with an app of 10000 pages:

Man that's awesome. Maybe we don't need to optimise?

merelinguist:
> > Also, @flybayer @ryardley, I just tested this with an app of 10000 pages:
> 
> @merelinguist WHOA!! Sweeet

It can happily go much bigger I think, I ran the script a couple of times!

ryardley:
> It can happily go much bigger I think, I ran the script a couple of times!

Maybe we should add a test like that?

merelinguist:
Just tested the build too, Blitz barely slows it down.

merelinguist:
@flybayer I can pr a test for large projects once this is merged?

flybayer:
> I can pr a test for large projects once this is merged?

Yes please! :)

flybayer:
@ryardley merge at your leisure

ryardley:
Yeah it's not super crazy slow at 10K pages. I thought it would take longer. With enough analysis we might be able to make it so if you dirty start it is close to instantaneous and we will be just be waiting for NextJS

dmzza:
In case anyone is looking, the docs referenced in this PR have since moved
# [Rendered Docs](https://github.com/blitz-js/blitz/blob/canary/packages/file-pipeline/README.md)
😄 Happy Reading

Issue #40, opened 2020-03-04
add blitz generate command


marcoseoane:

tsawan:
@flybayer any details here then I might be able to pick it up?

flybayer:
@tsawan it's already mostly done in #46

flybayer:
Ok @marcoseoane! Here's the details for the generate command. For now it's only one generator command.

Are you able to work on this soon? If not, we can label "help wanted" until someone else (or you) is able to do it.

---

Goal: a single command to generate **working** crud code for a model. Then you can customize from there. At some point we'll add ability to define the templates inside a blitz app, so that an app can have custom templates.

```
blitz generate crud [context/]MODEL(s)
```

Model input can be singular or plural. The result is the same.

```
blitz generate crud project
blitz generate crud projects
```

The above commands will both generate the following:

```
app/projects/queries/getProject.ts
app/projects/queries/getProjects.ts
app/projects/mutations/createProject.ts
app/projects/mutations/updateProject.ts
app/projects/mutations/deleteProject.ts
app/projects/components/ProjectForm.ts
app/projects/pages/projects/index.ts
app/projects/pages/projects/new.ts
app/projects/pages/projects/[id].ts
app/projects/pages/projects/[id]/edit.ts
```


Once we figure out the testing system, we'll also generate working tests.

The content of all those files can be roughly what's in the store example. I'm not sure what to do for the form component yet. How we should handle a form library. I think to start, we shouldn't use a form library, and just have some vanilla JS that does the job.

marcoseoane:
Hey! Yeah sure I'll get to work on having this completed for this week thurs/fri?, by all means put up a help wanted in case theres an extra eager contributor laying in wait for a ticket :)

flybayer:
Ok sounds great!!

flybayer:
So I'd also like to add an optional `--parent MODEL` argument:

```
blitz generate crud task --parent project
```

Would generate the following:

```
app/tasks/queries/getTask.ts
app/tasks/queries/getTasks.ts
app/tasks/mutations/createTask.ts
app/tasks/mutations/updateTask.ts
app/tasks/mutations/deleteTask.ts
app/tasks/components/TaskForm.ts
app/tasks/pages/projects/[projectId]/tasks/index.ts
app/tasks/pages/projects/[projectId]/tasks/new.ts
app/tasks/pages/projects/[projectId]/tasks/[id].ts
[...]

PR #150, opened 2020-04-20
Many updates to blitz new template, including prettier, eslint, husky, and git init

flybayer:


osirvent:
It WORKED!!

image

Just define `any model in your schema.prisma` and change `.env` to connect to a working postgresql DB. Then `npx blitz start`

Now everything seems ready to recreate store example through CLI.

I have some experience on prisma and nextjs so I think I can help on #40.

I must say I am a newbie on dev projects so be patient.

flybayer:
@osirvent that's so awesome!

I think Adam is already working on #40, but there's a bunch of other issues you can work on! Let me know if you need any help or guidance. You can also DM me in slack.

PR #153, opened 2020-04-20
Print help message for unsupported commands

### LINKED ISSUE #149: Unsupported blitz command prints error instead of help message
The below should print the blitz help information, saying that this command is not supported.

```
~/c/b/e/store> yarn blitz hacker
yarn run v1.19.1
$ /Users/b/c/blitz/node_modules/.bin/blitz hacker
DEBUG: monorepo
Path: /Users/b/c/blitz/node_modules/@blitzjs/cli
(node:87628) TypeError Plugin: @blitzjs/cli: Cannot read property 'dim' of undefined
module: @oclif/config@1.15.1
task: runHook command_not_found
plugin: @blitzjs/cli
root: /Users/b/c/blitz/packages/cli
See more details with DEBUG=*
 ›   Error: command hacker not found
error Command failed with exit code 2.
```

merelinguist:
Closes #149


aem:
I wonder if we could compute some kind of distance algorithm (e.g. https://github.com/trekhleb/javascript-algorithms/tree/master/src/algorithms/string/levenshtein-distance) and then suggest a command to run instead, maybe even with a way for the user to run it immediately? An interaction like this might be cool:

```bash
$ blitz nwe

Command `nwe` not found.

# only if we compute a close match, cache that
# command until the next CLI invocation
Did you mean to run the `new` command? If so run `blitz -y`
```

The next invocation would always clear the cache, regardless of what's called, but if the command is `blitz -y` we'll execute the cached command, passing in the same arguments as before if any were supplied.

flybayer:
That's a great idea @aem! We could probably streamline it even further by prompting user with `y/n`. Can you open a new issue for this and tag it ready to work on?

flybayer:
@merelinguist can you also make this print the actual help information, like the below?

```
Blitz CLI

VERSION
  @blitzjs/cli/0.5.0-canary.5 darwin-x64 node-v12.16.2

USAGE
  $ blitz [COMMAND]

COMMANDS
  build    Create a production build
  console  Run project REPL
  db       Run database commands
  help     display help for blitz
  new      Create a new Blitz project
  start    Start a development server
  test     Run project tests
```

merelinguist:
@flybayer It turns out we already have all of this functionality, it just wasn’t working, for some reason! Upgrading deps seemed to fix it.

flybayer:
Oh wow lol.. thanks!

aem:
@flybayer #173

PR #159, opened 2020-04-21
Only run Lambda warm-up HEAD requests when target is serverless

### LINKED ISSUE #104: Don't run the lambda warming code if not serverless
Currently the query/mutation lambdas are always being sent a HEAD request on page load to warm the lambda.

- [ ] We need to automatically detect if running as a traditional server, in which case don't make the HEAD call
- [ ] Open question: should we make the HEAD calls in development? If I'm deploying serverless, then I'd like to see the HEAD calls in development. But if not, I don't want to see them. How should we handle this?

Zeko369:
Dont run warump HEAD request when not running in serverless environment (`start --production`)

Closes #104


flybayer:
Looks good to me. @ryardley please review this

packages/cli/src/commands/build.ts
```diff
@@ -1,13 +1,15 @@
...
 
 export default class Build extends Command {
   static description = 'Create a production build'
   static aliases = ['b']
 
   async run() {
-    const config = {
+    const config: ServerConfig = {
       rootFolder: process.cwd(),
+      serverless: true,
```
ryardley, line 12: I am just wondering is it the right approach to assume it is serverless just because it is a build? What if we deploy elsewhere other than zeit? Perhaps this should be an optional or passed in via blitz.config.js or a cli flag?

Zeko369: Probably best to keep it in the blitz.config.js

flybayer: Ah, yes I forgot about `target: 'serverless'` in the config! Let's use that.

Need to check if `target.includes('serverless')` because it can also be something like `"serverless-trace-experimental"`

merelinguist:
@Zeko369 How’s this going?

Zeko369:
This is done, can someone please test this?

packages/server/src/synchronizer/pipeline/rules/rpc/index.ts
```diff
@@ -1,15 +1,29 @@
...
+  let serverless = false
+  for (const configFile of configFiles) {
+    if (pathExistsSync(resolve(src, configFile))) {
+      const config = require(resolve(src, configFile))
+      if (Object.keys(config).includes('target') && config.target.includes('serverless')) {
+        serverless = true
+      }
+    }
+  }
+
```
ryardley, line 26: I think I would be more comfortable if we did elsewhere possibly as a global singleton config loader. I am pretty sure there is a package out there to handle this.

ryardley: Happy to merge with a big TODO and then open an issue for it?

ryardley: Ok I was thinking of https://github.com/davidtheclark/cosmiconfig - this has ramifications to the config rule so best addressed in a separate PR

ryardley: Here is an issue https://github.com/blitz-js/blitz/issues/332

flybayer:
Hey @Zeko369, are you able to wrap this one up?

Zeko369:
@flybayer completely forgot about this PR, will finish it by the of tomorrow

Zeko369:
Depends on #653

flybayer:
@Zeko369 looks good! just need to fix merge conflict

flybayer:
Sorry @Zeko369!! I caused more merge conflicts by merging my middleware PR.

flybayer:
Hey @Zeko369, think you could fix the merge conflicts soon?

Zeko369:
Sorry @flybayer been super busy at the university, will finish it on friday

merelinguist:
@Zeko369 Couple more merge conflicts here.

flybayer:
@Zeko369 👀

PR #170, opened 2020-04-21
Use latest versions for app template

### LINKED ISSUE #166: Improve dependency versioning of the blitz new template
Currently most of the dependency versions in the `blitz new` template are defined as `latest`. This is not ideal.

It's much better for versions to be explicitly defined to prevent unintended dependency version changes. Unknown dependency version changes are a common source of bugs.

Currently the template is using `latest` to ensure the new app always has the latest version.

**Possible implementation(s)**

- Change the template source to explicit versions and then add tooling to make sure they are always at the latest
- Change the template source to explicit versions and then automatically run yarn upgrade during blitz new generation. Problem with this approach, is that we would install all the deps, then turn around and install the updated deps again.
- Keep as `latest` and then during blitz new generation automatically update that to the explicit version

merelinguist:
### Pull Request Type

- [x] Feature
- [ ] Bug fix
- [ ] Build related changes
- [ ] Tests related changes

### Checklist

Closes #166 

The latest versions are overwritten to explicit ones, e.g.:

```json
 "dependencies": {
    "@prisma/cli": "2.0.0-beta.2",
    "@prisma/client": "2.0.0-beta.2",
    "blitz": "0.5.1-alpha",
    "react": "16.13.1",
    "react-dom": "16.13.1",
    "typescript": "3.8.3"
  },
```

- [ ] Yes
- [x] No


packages/cli/package.json
```diff
@@ -50,7 +51,8 @@
     "@blitzjs/server": "0.5.0-canary.1",
     "@oclif/dev-cli": "^1.22.2",
     "@oclif/test": "^1.2.5",
-    "@prisma/cli": "2.0.0-beta.2"
+    "@prisma/cli": "2.0.0-beta.2",
+    "@types/node-fetch": "^2.5.6"
```
flybayer, line 55: Can you move this types dependency to the root? All dev dependencies should be installed at root from now on (less weird monorepo issues)

packages/cli/src/generators/app.ts
```diff
@@ -37,6 +40,19 @@ class AppGenerator extends Generator<AppGeneratorOptions> {
...
+    const pkgDependencies = Object.keys(pkg.dependencies)
+    const latestVersions = await Promise.all(
+      pkgDependencies.map(async (dependency) => {
+        const res = await fetch(`https://registry.npmjs.org/-/package/${dependency}/dist-tags`)
+        const json = await res.json()
+        return json.latest as string
+      }),
+    )
+    pkg.dependencies = pkgDependencies.reduce((o, k, i) => ({...o, [k]: latestVersions[i]}), {})
+    writeJSONSync(join(this.destinationPath(), 'package.json'), pkg)
```
flybayer, line 54: What are the ways this could fail and how can we handle it well?

We should probably add 1 or 2 retries here. And maybe then fallback to whatever the template originally had for the failing dependency?

flybayer:
Another thing: how do we handle breaking changes from dependencies? Like maybe we should only automatically upgrade the version within the major version. So major version changes would have to be made to the template here in the repo. That way we can actually test it.

packages/cli/src/generators/app.ts
merelinguist: I now have it retry a few times, and use the old versions if any fetching fails.

merelinguist:
@flybayer Good idea! Working on this now.

merelinguist:
@flybayer Fixed the updating strategy 👍

flybayer:
@merelinguist this is looking great!! I pushed a couple fixes.

One last thing, it seems your retry stuff got accidentally removed? Can you add that back in, and make sure when something fails to not crash but fallback to the template version?

merelinguist:
Oh yeah I just simplified that situation slightly. If something goes wrong with the fetching we just skip it and use the versions originally defined. Notice the `try catch` 🙂

merelinguist:
cc/ @flybayer

flybayer:
Ok, I see that now! Would probably be better to handle the failure per package, rather than fail the whole thing. Is this simple to change?

merelinguist:
Oh sure, will do.

merelinguist:
@flybayer Done!

PR #174, opened 2020-04-21
Fix `yarn test` by also running the build

flybayer:
### Pull Request Type

- [ ] Feature
- [ ] Bug fix
- [ ] Build related changes
- [x] Tests related changes

`yarn test` fails for a clean repo because `yarn build` must first be run

This changes `yarn test` to also run `yarn build`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -35,10 +35,6 @@ jobs:
         run: yarn install --frozen-lockfile --silent
         env:
           CI: true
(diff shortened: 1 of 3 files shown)


aem:
Might be worth setting up a watch build command to run in parallel with `test:watch` so that types are regenerated if you make changes to the code?

flybayer:
@aem yeah, you're welcome to add that if you need.

PR #186, opened 2020-04-22
fix(server): make synchronizer pages rule cross-platform

wKovacs64:
### Pull Request Type

- [ ] Feature
- [x] Bug fix
- [x] Build related changes
- [x] Tests related changes

### Checklist

The synchronizer rule for handling pages was not working on Windows due to the difference in the path segment separator (`/` vs `\\`).

- Updated the regular expressions used for detecting `pages` and `api` paths
- Updated the tests to normalize the inputs and expected results

- [ ] Yes
- [x] No

### Other information

@ryardley I had to change the regular expression from the `new RegExp` syntax to slash syntax in order to properly escape the Windows path segment separator and combine it with the non-Windows character in a set (`\\/` → `[\\\/]`). Hopefully I understood the original intention properly, or I might've just broke something. But all the tests pass now. :)

### DIFF
--- a/packages/server/src/synchronizer/pipeline/rules/pages/api.test.ts
+++ b/packages/server/src/synchronizer/pipeline/rules/pages/api.test.ts
@@ -1,31 +1,32 @@
+import {normalize} from 'path'
 import {apiPathTransformer} from '.'
 
 describe('createPagesPathTransformer', () => {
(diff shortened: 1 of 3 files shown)


flybayer:
Thank you @wKovacs64!!

PR #46, opened 2020-03-05
feat(cli): blitz generate command

marcoseoane:
## blitz generate

`blitz generate context/entity` produces the following folder structure plus the respective files

```
context/
-- entity/
---- pages/
    ContextEntityController.js
    ContextEntityModel.js
```

Nested file structures are also supported, with the prefix on the entity file names always being the preceding directory as seen in the next example:

`blitz generate context/nestedContext/entity`

Will generate the structure below

```
context/
-- nestedContext/
---- entity/
------ pages/
      NestedContextEntityController.js
      NestedContextEntityModel.js
```
And finally just providing an entity name will give you

```
entity/
-- pages/
  EntityController.js
  EntityModel.js
  ```
  (although I have spotted an error with the current code that puts the entity name in filenames twice, so will have to push a fix for this)
##
One design decision I have put into the code which I want everyone to give their thoughts on is the formatting of the top level context directory name. 
Currently the code is setting it to be lowercase, so if we take the `context/nestedContext` example and change `context` to be `Context`... the directory structure will come out looking like `context/nestedContext`. 

I took this direction with the idea that it would be nice to keep the display/formatting of all the root level directories in the blitz app cohesive, anything nested deeper will be up to the developer.

Next step will be setting up the template for the `Model` files, and any other default files we would like to include as the result of `generate`

PR #187, opened 2020-04-22
feat(cli): generate command

### LINKED ISSUE #40: add blitz generate command

### LINKED ISSUE #157: Command `blitz generate` does not exist on canary.
Following the instructions in [this document](https://github.com/blitz-js/blitz/blob/canary/USER_GUIDE.md), the user is met with an error when attempting to run the 'blitz generate` command. 

It does not exist as a CLI command.

I tried installing `@blitzjs/cli@latest` and  `@blitzjs/cli@canary` but it's not present on either tag.

aem:
### Pull Request Type

- [x] Feature

### Checklist

- n/a Any added terminal logging uses `packages/server/src/log.ts` (or N/A)
  - I thought it might be worth having a separate discussion about creating a separate logging package rather than having logging tied to the server package. For now I just reimplemented the minimal logging required within `command.ts` so all commands have access

Closes #40, closes #157 

Introduces `blitz generate`. Here's the docs for `blitz generate --help`, which should be fairly self-explanatory as far as the implications of this PR:

```
Generate new files for your Blitz project

USAGE
  $ blitz generate TYPE NAME

ARGUMENTS
  TYPE  (all|crud|mutation|page|query|resource) Type of files to generate
  NAME  Namespace for generated files (e.g. [name]Query.ts).

OPTIONS
  -d, --dry-run        Show what files will be created without writing them to disk
  -h, --help           show CLI help

  -p, --parent=parent  The parent context for nested generation. For example, generating `products` within a `store` would supply `-c 
                       store`. For nested contexts you may supply the full path.

ALIASES
  $ blitz g

EXAMPLES
  # We can just generate a single file type, for example queries. In this case,
  # the generator will ask any questions necessary to generate the proper files.
  blitz generate query task
    
  # Or, we can generate a full set of models, mutations, pages, etc.
  blitz generate all task
    
  # We can specify where the files belong with the parent flag, this allows you
  # to generate nested routes automatically. The example below would generate a route
  # for https://myapp.com/taskManager/task.
  blitz generate page task -c=taskManager
```

The one thing left up for debate is the terminology of `resource`. In my mind I've been thinking about `resource` to mean a full set of files (model + mutation + page + query), but I can see how that can be confusing. For example, in the Java/REST world "resource" usually means a single REST endpoint or a small collection of related endpoints.

Having said that, I do think if our goal is to truly create a full app ecosystem, I think we can reasonably take some creative license here to define our own DSL. As long as our docs are clear and have good guides for these words (e.g. a glossary that defines what we mean when we say "resource") I think it's totally reasonable to redefine these words to have special meaning in the context of Blitz. 

- [ ] Yes
- [x] No

### Other information

The `generate` command functions as a wizard and attempts to prevent you from hitting pitfalls as much as possible. You'll notice >100 lines of validation code in the command, the idea is to provide guardrails as much as possible to prevent users from doing anything bad, and either correcting for them or giving them a prompt to correct the error themselves rather than either (a) making assumptions that could be destructive, or (b) just failing hard.


packages/cli/src/command.ts
```diff
@@ -1,8 +1,34 @@
...
+  indent?: boolean
+}
 
 abstract class Command extends OclifCommand {
   protected enquirer = new Enquirer()
+  protected themeColor = '6700AB'
+  protected indent = false
+
+  logTheme(str: string, options?: LogOptions) {
+    this.log(chalk.bgWhite.hex(this.themeColor).bold(str), options)
```
flybayer, line 16: Can we reuse the `log` util for these? I've opened #194 for adding a utils package, but importing from `server` is fine for now.

packages/cli/src/commands/generate.ts
```diff
@@ -0,0 +1,150 @@
...
+import {flags} from '@oclif/command'
+import * as fs from 'fs'
+import * as path from 'path'
+import chalk from 'chalk'
+import enquirer from 'enquirer'
+const debug = require('debug')('blitz:generate')
+
+enum ResourceType {
+  Resource = 'resource',
+  Model = 'model',
```
flybayer, line 11: We don't currently have model files, so just leave this out for now

flybayer:
Sweet!

Over in the main issue I proposed `blitz generate crud [context/]MODEL --parent PARENT_MODEL`, which I like pretty well for defining context path and parent path. Sorry if you missed those comments since they aren't in the first post of the thread. See here: https://github.com/blitz-js/blitz/issues/40#issuecomment-613473059 and https://github.com/blitz-js/blitz/issues/40#issuecomment-614538737

All the input checks are great!!

flybayer:
Here's what I think we should use for naming. Rails and redwood use `scaffold` for generating everything. And rails uses `resource` for just model + controllers. So being consistent here is nice for folks if we don't have a good reason to deviate.

| Fri. Alpha | Command          | Result |
|------------|------------------|-----------------------|
| Critical  |`blitz generate crud` | All crud queries/mutations |
| X   |`blitz generate resource` | All crud queries/mutations + model schema |
| Preferably   |`blitz generate scaffold` | All crud queries/mutations + pages + model schema |
| X  |`blitz generate page` | A page |
| X  |`blitz generate component` | A component |
| X  |`blitz generate query` | A query |
| X  |`blitz generate mutation` | A mutation |

packages/cli/src/commands/generate.ts
```diff
@@ -0,0 +1,150 @@
...
+      // otherwise, validate the provided path, prompting the user if it's absent or invalid
+      if (!flags.path) {
+        this.log(
+          chalk.yellow(
+            `No path flag (--path, -p) was detected, but multiple roots under /app were present (${appPaths.join()})`,
+          ),
+        )
+        outputTarget = await this.promptForTargetDirectory(appPaths)
+      } else if (!appPaths.includes(flags.path)) {
+        this.log(chalk.yellow(`Given path '${flags.path}' doesn't exist in /app. Please select a valid path`))
```
merelinguist, line 141: All this should ideally use our `log.ts` stuff.

flybayer:
oops!

flybayer:
@aem there's some template processing code in https://github.com/blitz-js/blitz/pull/46 that you might can re-use.

packages/cli/src/commands/generate.ts
aem: I was figuring that model would add the model to the prisma file, generate a query + mutation, and then run `prisma migrate`. Just a shorthand for adding data to the `db` without generating a UI for it

packages/cli/src/commands/generate.ts
aem: yeah this was addressed in the PR description and in @flybayer's comment

packages/cli/src/command.ts
flybayer: Thanks! Better to make a step towards the right solution than no step at all :)

aem:
@flybayer couple of thoughts re: your naming/architecture comments:

> Rails and redwood use scaffold for generating everything

Agree that deviation is not ideal, although I do think the term `scaffold` is a bit wordy for what we're trying to convey. What if we just went with `blitz generate all` for that case? Feels like it might be a little more clear, accessible to anyone who doesn't natively speak English, and will save a few keystrokes :)

As for `crud`, it seems like that's just what I was using `model` for. Again, given that `model` feels like shared terminology I opted for that. I'm not super intent on that naming, `crud` just felt a little verbose/technical compared to `model` which felt a little more agnostic.

> So I think using Context/User fits nicely for us. We should support both \ and /

Totally agree here. I hadn't gotten up to supporting nested paths yet, was building this pretty incrementally. Should be pretty simple to make this happen. I'll build in some fallback support for what to do if a particular context path doesn't exist (e.g. find best match, provide option to create the context resources similar to `mkdir -p`).

flybayer:
> Agree that deviation is not ideal, although I do think the term `scaffold` is a bit wordy for what we're trying to convey. What if we just went with `blitz generate all` for that case? Feels like it might be a little more clear, accessible to anyone who doesn't natively speak English, and will save a few keystrokes :)

Ok, great suggestion. Go with `all` for generating all :) 
 
> As for `crud`, it seems like that's just what I was using `model` for. Again, given that `model` feels like shared terminology I opted for that. I'm not super intent on that naming, `crud` just felt a little verbose/technical compared to `model` which felt a little more agnostic.

We definitely shouldn't use model for also generating queries/mutations. Model should be for an actual model, i.e. database table and model validation/authn.

I'm open to something better than crud for queries/mutations, but I think crud is very accurate here. Most people know what crud is. If not, they will learn quickly because it's a simple concept.

> Totally agree here. I hadn't gotten up to supporting nested paths yet, was building this pretty incrementally. Should be pretty simple to make this happen. I'll build in some fallback support for what to do if a particular context path doesn't exist (e.g. find best match, provide option to create the context resources similar to `mkdir -p`).

Sounds good! Just make sure you finish the core stuff first before polishing. Alpha release will be about 21 hours from now.

flybayer:
It's 9p for me now, so during my day tomorrow I can finish up some stuff if you need. Just let me know where you're at when you are done for your day :)

flybayer:
Just curious why the template id directory is `__id__` instead of `[id]`?

```
packages/cli/templates/page/__id__/edit.tsx
```

aem:
@flybayer yeoman has a bug where it doesn't handle `[]` in filenames well, so in order to get the generator to recognize the files we have to use a different pattern and then replace it at runtime. I spent about half an hour and couldn't find a way around it so I decided this was the fastest way to an alpha

aem:
Outstanding known issues that we'll want to fix (I can open issues for these):

1. When generating Prisma models the ConflictChecker always prompts to see if the user wants to commit the changes. We should build a way to disable the conflict checker in cases like this when we're manually writing to files but know that it's safe
2. The generated `pages` don't have any useful code in them, just a placeholder header telling the user what the page should be used for. We should generate some useful actions/components based on the prisma models (if they exist)
3. Never implemented `component` generator. I wasn't quite sure what we'd put in there yet, didn't seem worth the thought experiment at 12:30am ET 🙂
4. Context paths in the positional argument wasn't built out. I need to rewrite how we calculate the final object name + context path in order to make it a little more generic first, also seemed out of scope for the alpha so I opted to force users to provide context via the `--parent` flag for now. I'll start on the more flexible solution this weekend

All things considered, the generator is pretty robust right now. It's not incredibly smart as far as how it generates things, but definitely smart enough to give people something to play with this weekend. I'd be comfortable merging as is.

flybayer:
Reorder the options like below (so help command prints in this order)
```
  All = 'all',
  Resource = 'resource',
  Crud = 'crud',
  Query = 'query',
  Mutation = 'mutation',
  Page = 'page',
```

Running outside a project logs error but then continues instead of quitting:
```
~/c/blitz> blitz g query task
✕ No blitz.config.js found. `generate` must be run from the root of the project.
? No parent flag (--parent, -p) was found. Would you like to create a new context folder under /app for 'tasks'? …
❯ Yes
  No
```

Change `blitz g` to print out the available generator commands and descriptions instead of this:
```
~/c/b/e/store> blitz g
 ›   Error: Missing 2 required args:
 ›   type  Type of files to generate
 ›   name  Namespace for generated files (e.g. [name]Query.ts).
 ›   See more help with --help
```

flybayer:
**High priority**
Logged path of created files should have full path from root instead of this:
```
~/c/b/e/store> blitz g query task
✔ No parent flag (--parent, -p) was found. Would you like to create a new context folder under /app for 'tasks'? · Yes

CREATE    getTask.ts
CREATE    getTasks.ts
```

flybayer:
@aem I disabled `--context` and the model generator for now. Let's open issues for them and make them nice without the pressure for this alpha :)

PR #195, opened 2020-04-23
Change blitz new install to cli version

### LINKED ISSUE #191: Change `blitz new` to set the blitz version to `latest` or `canary` based on the version of the CLI that's running

- `blitz new` from a global install of `blitz@latest` will result in `"blitz": "latest"` in package.json
- `blitz new` from a global install of `blitz@canary` will result in `"blitz": "canary"` in package.json

This will ensure the generated app always matches the CLI version tag being used.

merelinguist:
### Pull Request Type

- [x] Feature
- [ ] Bug fix
- [ ] Build related changes
- [ ] Tests related changes

### Checklist

Closes #191 

- [ ] Yes
- [x] No

### DIFF
--- a/packages/cli/src/utils/replace-blitz-dependency.ts
+++ b/packages/cli/src/utils/replace-blitz-dependency.ts
@@ -13,8 +13,7 @@ export const replaceBlitzDependency = async (pkg: any, desinationPath: string, c
   // `blitz new` from a global install of `blitz@latest` will result in `"blitz": "latest"` in package.json
   // `blitz new` from a global install of `blitz@canary` will result in `"blitz": "canary"` in package.json
   // Otherwise, fall back on canary
-  const newVersion =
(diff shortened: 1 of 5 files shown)


packages/cli/src/utils/replace-blitz-dependency.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  if (!distTags) {
+    return
+  }
+
+  // `blitz new` from a global install of `blitz@latest` will result in `"blitz": "latest"` in package.json
+  // `blitz new` from a global install of `blitz@canary` will result in `"blitz": "canary"` in package.json
+  // Otherwise, fall back on canary
+  const newVersion =
+    cliVersion === distTags.latest ? 'latest' : cliVersion === distTags.canary ? 'canary' : 'canary'
```
flybayer, line 17: I think this logic needs to be more loose and not check for an exact version match. So basically this:

```js
newVersion = cliVersion.includes('canary') ? 'canary' : 'latest'
```

Issue #189, opened 2020-04-23
Exit gracefully and show helpful message when running the blitz cli outside a blitz app


flybayer:
**What is the problem?**

App specific commands like `blitz console` currently fail with odd errors when ran outside of a blitz app directory.

The only command that should work globally is `blitz new`.

All others need to exit gracefully and print a helpful error message.

PR #203, opened 2020-04-23
Print helpful error if try to use CLI commands outside of blitz app

Zeko369:
### Pull Request Type

- [x] Feature
- [x] Bug fix
- [ ] Build related changes
- [ ] Tests related changes

Closes: #189

Check if inside of blitz app

- [x] Yes
- [ ] No

### Other information

I made it so it checks you're in the root of the project, because although some commands can work throughout the project structure, some fail so we should encourage people to just run it from the root path

image


Zeko369:
Closes #189

packages/cli/src/checkBeforeRunning.ts
```diff
@@ -0,0 +1,16 @@
...
+const whitelistGlobal = ['new']
+
+const hook: Hook<'init'> = async function (options) {
+  if (options.id && whitelistGlobal.includes(options.id)) return
+
+  try {
+    await readFile('blitz.config.js')
+  } catch (err) {
+    this.error(`It appears that you're not running in a blitz project`)
+  }
```
flybayer, line 13: Ok few things:

1. `blitz.config.js` is optional, so this won't work. We do need robust logic for detecting if we are in a blitz app. Maybe if `blitz` is a dependency of the package.json? 
2. Put the above detection logic in it's own file so we can reuse elsewhere
3. If not at root of a blitz project, check if we are in a sub directory of a blitz project and have a specific error message for this.
4. Else have an error about not being in a blitz project.

flybayer: **Error message for in a sub-folder:** "You are currently in a sub-folder of your Blitz app, but this command must be used from the root of your project."

**Error message for not in a blitz app:** "You are not inside a Blitz project, so this command won't work. You can create a new app with `blitz new myapp` or see help with `blitz help`

flybayer:
Sweeet! Two things:

1. Can you use the `pkg-dir` package that's already installed instead of `find-package-json`?
2. Could you add a detailed comment in the utils file describing the blitz app detection and all the possible outcomes?

Thank you so much!

flybayer:
This is beautiful — great work!!!

aem:
@Zeko369 i think we have commands that check for this manually, can you follow up and convert those usages to use the new utils?

Zeko369:
@aem you pushed the new utils?

aem:
Ah sorry my comment was a bit unclear. Since this hook/util you added handles it for all commands, can you remove any checks for `blitz.config` in existing commands?

PR #216, opened 2020-04-24
Remove synchronizer logging during runtime

### LINKED ISSUE #213: Blitz synchronizer should report different during runtime than during start
**What is the problem?**

We have a nice spinner and success message for at the start of blitz start, but this same message is being shown if the server is already running and you change a file.

During runtime I think we should only show error messages. Don't show spinner or success.

**Supporting Documentation**
```
> Using "webpackDevMiddleware" config function defined in default.
[ info ]  bundled successfully, waiting for typecheck results...
[ wait ]  compiling ...
[ info ]  bundled successfully, waiting for typecheck results...
[ ready ] compiled successfully - ready on http://localhost:3000
Processing file...db/schema.prisma
✔ Prepped for launch
Processing file...db/schema.prisma
✔ Prepped for launch
Processing file...app/admin/pages/admin/index.tsx
✔ Prepped for launch
```

merelinguist:
### Pull Request Type

- [ ] Feature
- [x] Bug fix
- [ ] Build related changes
- [ ] Tests related changes

### Checklist

Closes #213 

During runtime only show error messages. Don't show spinner or success.

- [ ] Yes
- [x] No

### Other information

- [x] Remove `Processing file...db/schema.prisma`
- [ ] Remove ✔ Prepped for launch


packages/server/src/synchronizer/index.ts
```diff
@@ -53,6 +53,7 @@ export async function synchronizeFiles({
     }
 
     const readyHandler = () => {
+      reporter.stream.write({type: IDLE, payload: null})
       reporter.stream.write({type: READY, payload: null})
```
ryardley, line 57: Only thing I would be thinking maybe we might do well to send these events from within the actual stream? What do you think?

ryardley: This handler is mainly to help with promise conversion

ryardley: Also if we do that we don't actually need to pass a callback into the stream at all we can simply attach an `on('data', ...)` listener to the reporter stream.

ryardley: I think the above is fine for now. 

In the future not in this PR we could make an 'EventStream' decorator that accepts a stream and does the switching for us.

```ts
reporter.on(READY, (payload) => {
  // ...
})
```

Brainstorm 1: another option would be that we could use the fact that the stream is an eventemitter and simply define and emit events on the stream but I think that might have implications for our type system...

merelinguist: @ryardley Would you mind finishing this? Struggling to understand the whole data flow of the server and want to focus on finishing the tutorial.

ryardley: I can certainly do it but I would love to understand what you need help with and see if there is some way I can create enough docs or resources to make it easier to grok.

merelinguist: Finished the tutorial :). So, just not sure how to make READY fire once? Have played around with the order of firing events but I’m sure there’s something I’m missing,

ryardley:
Hey @merelinguist this seemed stale so I fixed it up on another branch will make a PR.

ryardley:
https://github.com/blitz-js/blitz/pull/369

PR #218, opened 2020-04-24
Create utils package

### LINKED ISSUE #194: Create the `@blitzjs/utils` package for shared functionality like logging
We are starting to have utilities like logging that is used by multiple packages. Let's move them into their own package.

**Possible implementation(s)**
Essentially you can copy the `core` package into `utils` for the package.

Then move shared utils into like logging stuff.

merelinguist:
### Pull Request Type

- [ ] Feature
- [ ] Bug fix
- [ ] Build related changes
- [ ] Tests related changes

### Checklist

Closes #194 

- [ ] Yes
- [x] No


ryardley:
Stuff I can think of that is pretty general that could be useful to share:

`packages/server/src/synchronizer/streams.ts`
`packages/server/src/resolve-bin-async.ts`

merelinguist:
@ryardley Finished those three.

packages/utils/src/streams.ts
```diff
@@ -1,25 +1,17 @@
...
+
 const pipeline = (pumpify.ctor({
   autoDestroy: false,
   destroy: false,
   objectMode: true,
   highWaterMark: 160,
 }) as any) as PumpifyFn
 
-export {pipeline}
+export const streams = {pipe, through, parallel, from, pipeline}
```
ryardley, line 17: Each one of these is a stream helper. Any chance we can't just export each of these individually? 

```
import {through} from '@blitzjs/utils'
```

Maybe namespacing them might make sense if we are concerned with execution context.

```
import {through} from '@blitzjs/utils/streams'
```

ryardley:
As this grows is there a good way to increase discoverability of the packages in here?

packages/utils/src/streams.ts
merelinguist: Happy to change this but I think it makes it simpler for each module to have a single export–similar to how the logger now just exports `log` instead of doing `import { error } from ...`. Thoughts?

flybayer:
I think we need to split utils up into `@blitzjs/utils-client` and `@blitzjs/utils-server` to ensure server code never gets in the client bundle.

ryardley:
@flybayer what do we do about shared isomorphic utils?

ryardley:
I think it might be a bit too heavy handed doing a separate package for client and server esp as we may not have a huge number of these truly shared dependencies but perhaps having a deep import rollup targets set for `@blitzjs/utils/node`, `@blitzjs/utils/browser` and `@blitzjs/utils` to ensure we cannot mix execution context might be a neat way to do it?

packages/utils/src/streams.ts
ryardley: Hmm spider dependencies like this need to be handled with care as they become hard to refactor the more they are used. Also maybe this is bikeshedding so tell me to shut up if so but whatever we decide here will probably stay.

There are tradeoffs on both sides:
* `streams.through.obj(...)` Provides more context as to what it is is. (through is just a bad name - although it is a standard and well known module in Node)
* Grouping by namespace (ie. streams) uses less global utils namespace by compacting to a standard lib
* It makes sense to consider our view rendering layer as a single dependency because it is somewhat application specific and will likely contain app specific things such as ascii logos or special layouts and standard colours etc. 
 
However
* None of the stuff within `streams` depends on each other they are all top level helpers that happen to deal with node streams and could be used in any node project.
* By using a named `streams` export the `streams` variable declaration becomes unusable where it would be handy to call let's say a variable containing an array of streams `streams`. Perhaps `streamUtils` might be more accurate but I really think that deep imports are the best way to manage this `@blitzjs/utils/streams` as it also allows destructured imports.
* It is common to pipeline nested pipelines together:
```
pipeline(inp, pipeline(a, b, c), pipe(d, e))
```
With namespacing and without aliasing this gets messy:
```
streams.pipeline(inp, streams.pipeline(a, b, c), streams.pipe(d, e))
```
* If we get the namespacing wrong it is hard to fix up later so using single level namespacing or context level namespacing is one less decision we need to make. 

My preference is for deep imports.

merelinguist:
Looks like tsdx doesn’t support deep imports, but generally like the idea. Will investigate

ryardley:
I think we should hold off on this until we have resolved a few things here https://github.com/blitz-js/blitz/issues/306

Issue #207, opened 2020-04-23
Long delay between `blitz new` and user feedback


wKovacs64:
**What is the problem?**

After running `blitz new`, there is a very long delay (30-60 seconds on machines I have available) before any feedback is displayed to the user.

**Steps to Reproduce:**

1. `blitz new`
1. ⌚ 
1. Observe feedback.

**Versions:**

- Blitz: 0.6.0-canary.0
- OS: Windows 10 and macOS

```
debug: local
debug: pkgPath: C:\dev\contrib\blitz\node_modules\@blitzjs\cli

env: win32-x64 node-v12.16.1
blitz: 0.6.0-canary.0 (global)
blitz: 0.6.0-canary.0 (local)
```

**Supporting Documentation**

N/A

flybayer:
I think the solution is to ensure every command lazy loads it’s dependencies

flybayer:
I added a message that logs immediately about it being alpha and could be slow. At least this will help a bit for right now.

wKovacs64:
Cool, good enough for now. 👍

merelinguist:
Experiencing this issue too.

flybayer:
We definitely need to lazy load everything in the CLI. I bet this will make a big difference

ntgussoni:
Hello guys, I've been trying to test this and im not seeing such delays.

Would any of you mind to add this at the very top of new.ts and run the new command?

```javascript
const {performance, PerformanceObserver} = require('perf_hooks')
const mod = require('module')
// Monkey patch the require function
mod.Module.prototype.require = performance.timerify(mod.Module.prototype.require)
require = performance.timerify(require)
// Activate the observer
let sum: number = 0
const obs = new PerformanceObserver((list: {getEntries: () => any}) => {
  const entries = list.getEntries()
  entries.forEach((entry: any[]) => {
    sum = parseFloat(sum) + parseFloat(entry.duration)
    console.log(`require('${entry[0]}')`, parseFloat(entry.duration) / 1000)
  })

  console.log('TOTAL', sum)
  obs.disconnect()
})
obs.observe({entryTypes: ['function'], buffered: true})
```

This is my output


```
require('path') 0.000323605
require('path') 0.00000411
require('@oclif/command') 0.000043021999999999996
require('@oclif/command') 0.000030842
require('../command') 0.000027757000000000002
require('../command') 0.000020818
require('../generators/app') 0.082966314
require('../generators/app') 0.08296059700000001
[...]

Issue #127, opened 2020-04-15
Fix CLI test failure — always fails on first test run


flybayer:
The cli tests are always failing on the first test run.

**To reproduce:**

1. Checkout canary
2. `git clean -xfd`
3. `yarn`
4. `yarn test` <-- succeeded
5. `git clean -xfd packages`
6. `yarn`
7. `yarn test` <-- failed

```
~/c/blitz> y build
yarn run v1.19.1
$ lerna run build --no-private --stream
tt^C⏎                                                                                                             ~/c/blitz> y test
yarn run v1.19.1
$ yarn run build && lerna run test --parallel
$ lerna run build --no-private --stream
lerna notice cli v3.20.2
lerna info Executing command in 4 packages: "yarn run build"
@blitzjs/server: $ tsdx build
@blitzjs/core: $ tsdx build
✓ Creating entry file 4.7 secs
✓ Creating entry file 4.7 secs
⠹ Building modules
✓ Building modules 12.6 secs
@blitzjs/cli: $ rimraf lib && tsc && yarn build:templates
@blitzjs/cli: $ rimraf templates/**/node_modules && cpy --dot --parents '!/node_modules/' templates lib
blitz: $ rollup -c rollup.config.js
blitz: ./src/index.ts → dist/index.js...
blitz: created dist/index.js in 2.9s
blitz: ./src/bin/cli.ts → ./dist/cli.js...
blitz: created ./dist/cli.js in 476ms
lerna success run Ran npm script 'build' in 4 packages in 32.0s:
lerna success - blitz
lerna success - @blitzjs/cli
lerna success - @blitzjs/core
lerna success - @blitzjs/server
lerna notice cli v3.20.2
lerna info Executing command in 3 packages: "yarn run test"
@blitzjs/cli: $ jest
@blitzjs/server: $ tsdx test
@blitzjs/core: $ tsdx test
@blitzjs/cli: PASS test/commands/new.test.ts
@blitzjs/cli: PASS test/commands/start.test.ts (7.153s)
@blitzjs/core: PASS test/todo.test.ts
@blitzjs/core:   ✓ todo (2ms)
@blitzjs/cli: PASS test/commands/build.test.ts (7.565s)
@blitzjs/core: Test Suites: 1 passed, 1 total
@blitzjs/core: Tests:       1 passed, 1 total
@blitzjs/core: Snapshots:   0 total
@blitzjs/core: Time:        4.846s
@blitzjs/core: Ran all test suites.
[...]

PR #245, opened 2020-04-25
Fixes tests timing out on first try

ntgussoni:
### Pull Request Type

- [ ] Feature
- [ ] Bug fix
- [ ] Build related changes
- [x] Tests related changes

### Checklist

The CLI takes too long to start on the first try, causing JEST to timeout with the default testTimeout of 5000

I arbitrarily doubled the timeout, this doesn't mean that is a solution

- [ ] Yes
- [x] No

### Other information

This might be related to #207

### DIFF
--- a/packages/cli/jest.config.js
+++ b/packages/cli/jest.config.js
@@ -8,6 +8,7 @@ module.exports = {
   coveragePathIgnorePatterns: ['/templates/'],
   modulePathIgnorePatterns: ['tmp', 'lib'],
   testPathIgnorePatterns: ['src/commands/test.ts'],
+  testTimeout: 10000,
   // TODO enable threshold
   // coverageThreshold: {
   //   global: {


ntgussoni:
More than a fix this is just sweeping under the rug, but at least we can see if it doesn't make it fail on most environments. 
I believe it's closely related to #207, so once it's solved we might not need to have the timeout increased.

flybayer:
Thanks! How is this related to #207?

flybayer:
So #207 is fixed but tests still fail on first run. I think I see the issue, which I've commented about here: https://github.com/blitz-js/blitz/issues/127#issuecomment-619475161

So I'm going to close this PR and wait for a real fix (I already have the workaround in place to run the tests twice in a row)

@all-contributors add @ntgussoni for code

Issue #227, opened 2020-04-25
Bug: `blitz h` doesn't display the help information


flybayer:
**What is the problem?**

`blitz h` should display the help information instead of this:

```
~/c/blitz> blitz h
You are using alpha software - if you have any problems, please open an issue here:
  https://github.com/blitz-js/blitz/issues/new/choose
Also, this CLI may be slow as it's not yet optimized for speed

 ›   Warning: h is not a blitz command.
Did you mean b? [y/n]:
 ›   Error: Run blitz help for a list of available commands.
```

PR #252, opened 2020-04-25
Make `blitz h` print the help menu

jclancy93:
### Pull Request Type

- [ ] Feature
- [x] Bug fix
- [ ] Build related changes
- [ ] Tests related changes

### Checklist

**Was unsure on how to get test coverage output so left this section blank**

Closes: #227 

Not sure if this is the right way to do this. The reason `blitz h` does not work is because the `help` command comes from https://github.com/oclif/plugin-help. 

The approach I took was to override this in blitz locally. There are some alternatives, including:

* Use a fork of `oclif/plugin-help`
* Use https://github.com/ds300/patch-package (unsure if this would work)

- [ ] Yes
- [x] No

### Other information

image

### DIFF
--- a/packages/cli/src/commands/help.ts
+++ b/packages/cli/src/commands/help.ts
@@ -1,0 +1,23 @@
+import {Command, flags} from '@oclif/command'
+
+import Help from '@oclif/plugin-help'
+
+export default class HelpCommand extends Command {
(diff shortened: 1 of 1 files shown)


flybayer:
This is perfect, thank you!!

@all-contributors add @jclancy93 for code

PR #270, opened 2020-04-26
Add start script to new app template package.json

kandros:
As a new user, I was surprised not to find a start script.
I believe this should be opinionated by the framework

Thanks for the project, looks amazing ❤️


flybayer:
Thank you @kandros!

@all-contributor add @kandros for code

PR #272, opened 2020-04-26
Change blitz version to be pinned in new app template (Closes #262)

### LINKED ISSUE #262: Change blitz version to be pinned in new app template
### What do you want and why?

Currently `blitz new myapp` results in the blitz version being set to `latest`. We need to change that so the latest explicit version is set, just like we are doing with the other dependencies.

- If cliVersion is canary, then blitz version should be set to most recent version of `canary` release tag
- Else, blitz version should be set to most recent version of `latest` release tag

Code: https://github.com/blitz-js/blitz/blob/canary/packages/cli/src/generators/app.ts#L54..L62

Skn0tt:
### Type: bug fix, refactor, tests

Closes: #262 

- Setup Integration test for `blitz new`
- Refactored dependency version replacement
- Use canary tags for canary CLI, latest tag for regular CLI (as described in #262)

For further explanation on the given changes, please refer to the commit messages of this PR.

### Checklist

- [x] Tests added for changes

### Breaking change: no


Skn0tt:
@all-contributors please add @skn0tt for code, test

flybayer:
This looks great, thank you!

One issues: running the `blitz new` test also runs yarn install — can we mock the actual dependency installation?

I'll mark this as draft — change it back when ready for another review

packages/cli/src/generators/app.ts
```diff
@@ -45,21 +45,14 @@ class AppGenerator extends Generator<AppGeneratorOptions> {
...
 
     console.log('') // New line needed
     const spinner = log.spinner(log.withBranded('Retrieving the freshest of dependencies')).start()
 
-    const dependenciesArray = await Promise.all([
-      replaceDependencies(pkg, pkgDependencies, 'dependencies'),
-      replaceDependencies(pkg, pkgDevDependencies, 'devDependencies'),
-    ])
+    pkg.dependencies = await fetchLatestVersionsFor(pkg.dependencies);
+    pkg.devDependencies = await fetchLatestVersionsFor(pkg.devDependencies);
```
flybayer, line 53: Also, is there a reason you removed the Promise.all?

Skn0tt: I thought it'd be easier to read - but a day later, I don't think that's necessary true. I'll change it to use Promise.all again ^^

packages/cli/src/commands/new.ts
```diff
@@ -37,6 +38,11 @@ export default class New extends Command {
       default: hasbin.sync('yarn'),
       allowNo: true,
     }),
+    "skip-install": flags.boolean({
```
aem, line 41: can we find a way to make this option undocumented? i'm not sure we want to expose this to users if it's just for testing. maybe do this as an environment variable instead? `BLITZ_SKIP_INSTALL=true blitz new my-proj`?

packages/cli/src/utils/npm-fetch.ts
```diff
@@ -10,12 +10,17 @@ export const fetchAllVersions = async (dependency: string) => {
   }
 }
 
-export const fetchLatestDistVersion = async (dependency: string) => {
+export const fetchDistTags = async (dependency: string) => {
   const res = await fetch(`https://registry.npmjs.org/-/package/${dependency}/dist-tags`)
   if (res.ok) {
     const json = await res.json()
-    return json.latest as string
+    return json;
```
aem, line 17: nit: remove semicolon

packages/cli/src/utils/get-blitz-dependency-version.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export const getBlitzDependencyVersion = async (cliVersion: string) => {
+  const { latest, canary } = await fetchDistTags("blitz");
+  
+  if (cliVersion.includes("canary")) {
+    return canary;
+  }
+
+  return latest;
+}
```
aem, line 11: nit: please add newline at EOF

packages/cli/test/commands/new.test.ts
```diff
@@ -1,3 +1,49 @@
...
+        { encoding: "utf8", flag: "r" }
+      );
+      const packageJson = JSON.parse(packageJsonFile);
+      const { dependencies: { blitz: blitzVersion } } = packageJson;
+
+      expect(blitzVersion).toEqual(await getLatestBlitzVersion());
+    });
+
+  })
+})
```
aem, line 49: nit: newline at EOF please

packages/cli/src/commands/new.ts
Skn0tt: We could this undocumented by setting `hidden: false` for the flag, which would simply hide it from `blitz new --help`.
I actually wondered wether this is helpful for testing purposes only - isn't there some use case where a user would want to skip installations? I cannot come up with one on the spot, but still ... 😅

Skn0tt:
> if you could also make sure you're running prettier against your commits, it should solve the problems with the newlines and semicolons
Somehow the husky installation didn't go through, so the pre-commit prettier didn't work. Now it does 🙈

packages/cli/src/commands/new.ts
Skn0tt: @aem so do we want to simply hide it or remove the option altogether and use an environment variable instead?

aem: Oops I thought I had commented back, sorry about that!! We can leave it like this but let’s still hide the arg from the docs. Thanks!

Skn0tt:
I cannot reproduce the test failure locally. @aem could you have a look at this?

flybayer:
Looks like this is because CI doesn't have git installed

```
@blitzjs/cli:     console.error src/generators/app.ts:99
@blitzjs/cli:       Failed to run git commit
```

Want to take a stab at adding this to the github actions workflow in `.github/workflows/main.yml` ?

Skn0tt:
Judging from [GitHub's Docs on this](https://github.com/actions/virtual-environments/blob/master/images/linux/Ubuntu1804-README.md), Git is installed on Ubuntu-latest. What I think causes the tests to fail is some interdependency between them: While only test for `blitz new` should output to a temp folder, other tests fail showing a temp folder as the reason.

Skn0tt:
Alright, so I can reproduce to locally with `--runInBand`. Let's have a look 🤨

Skn0tt:
Figured it out, apparently creating a new project for the `blitz new` inside of `beforeAll` messed up the remaining tests. Closing on favor of #284

Issue #235, opened 2020-04-25
Consider changing from ejs templates to plain string replace


flybayer:
**What do you want and why do you want it?.**

Working on the ejs templates for `blitz generate` (which has syntax like `<div<%= name %></div>`) is currently a nightmare because syntax TS can't parse it, prettier can't parse it, eslint can't parse it etc.

This means you have to actually run the generator each time you make a change to see it's correct or not. And you have to manually format your code.

I think we should consider just doing plain string replaces instead.

Look how difficult this is to read and write:
```
import create<%= ModelName %> from 'app/<%= modelNames %>/mutations/create<%= ModelName %>'
```

This is SO much easier to read and write:
```
import createModelName from 'app/modelNames/mutations/createModelName`
```

**Other motivations**:
-  We will be adding template ejection for users to customize. So this affects all blitz users, not just blitz contributors.
- We can run tests directly against the templates
- This is also nicer for the template file names: `createModelName.ts` -> `createProduct.ts`

/cc @aem

aem:
I didn't realize this was a discussion rather than a ready-to-implement issue. I personally see very little downside here but lots of upside. Owning our own templating system increases flexibility, it means we own the full pipeline rather than being tied to an external system that can break out from under us, and the guaranteed cross-system compatibility is nice. The implementation is also a nice forcing function for centralizing the generator logic in the superclass since all of the generator code is functionally the same.

aem:
This seems uncontroversial, relabeling as ready for dev since the PR seems to be acceptable

PR #281, opened 2020-04-26
feat(cli): use custom template format instead of EJS

aem:
### Type: Feature

Closes: https://github.com/blitz-js/blitz/issues/235

This removes our usages of the `.ejs` templating language and replaces it with a custom templating system. Since we don't actually use any templating functionality other than pure variable replacement, we can move from `mem-fs-copy`'s `copyTpl` function to the raw `copy` function which lets us supply a custom processing function. In our case, we just do global string replaces on any code that matches the `__variable__` pattern with a variable that exists in the template. 

As an extension, we had to centralize the generation logic to ensure that each generator doesn't have to re-implement this logic themselves. This _drastically_ simplifies each `Generator` subclass which now just needs to supply a template name, a subdirectory within the project where its files should be placed, and a function that will return the values to pass to the template. I was initially worried that this API would be a bit too abstract, but I've come to like the fact that creating a new generator is now more about writing up the template and less about writing generation code. There's of course an escape hatch which is just to override the `async write` function. 

### Checklist

- [ ] Tests added for changes


flybayer:
This is looking much better!!

- `.gitignore` still isn't working right (I verified I pulled your latest code)
```
CREATE    db/schema.prisma
CREATE    gitignore
CREATE    integrations/.keep
```

- Something in here is causing a significant hang between "Dependencies successfully installed." and "Your new Blitz app is ready! Next steps:" during `blitz new`

aem:
@flybayer I'm guessing the delay is because the gitignore isn't getting copied, the CLI is probably trying to do some processing to the whole tree, some of which is now things like `node_modules` if the gitignore is malformed

flybayer:
Ok, delay is fixed and `.gitignore` is generated, but the log for gitignore is wrong:

```
CREATE    db/schema.prisma
CREATE    gitignore
CREATE    integrations/.keep
```

Missing the dot

aem:
@flybayer 🤦 i fixed that but had the commit stashed while resolving some conflicts. pushing it up now

merelinguist:
Okay I might be wrong but I think @Zeko369’s recent pull broke the ability to create new projects! 

```sh
$ /Users/dylan/dev/blitz/node_modules/.bin/blitz new aem-test

You are not inside a Blitz project, so this command won't work.
You can create a new app with blitz new myapp or see help with 
 ›   Error: Not in correct folder
error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

aem:
@merelinguist you're definitely correct. I'm going to open up a revert PR for that

merelinguist:
Thanks!

merelinguist:
Otherwise this all looks great, by the way 👍

merelinguist:
@flybayer Looks grand to me, will let you merge.

Issue #282, opened 2020-04-27
Add links to instructions for installing postgres database to the new app template page


flybayer:
### What do you want and why?

Currently the new app template page instructs users to run `blitz db migrate` but doesn't say anything about installing postgres. We need to add this.

- Link to outside sources on installing postgres on macOS, windows, and linux
- The file to update is `packages/cli/templates/app/app/pages/index.tsx`

Let me know if you have any questions!

Issue #253, opened 2020-04-26
Bundle docker-compose for local database


iRoachie:
**What do you want and why do you want it?.**
Awesome project! On starting, the demo asking you to run a db migrate. Of course this doesn't work cause there's no database up to this point. The instructions on the README then suggest you installing postgres locally. 

Instead, I think we can provide a better experience by bundling an already setup docker-compose file and maybe even start the container on `blitz start`

**Possible implementation(s)**
Add a docker-compose file to the db folder. Maybe something like the old prisma2 one https://github.com/prisma/prisma2-cli/blob/master/.buildkite/docker-compose.yml

Zeko369:
@iRoachie wanna work on this?, if not I'll add "ready to work on label" so other people can work on this

iRoachie:
What's the best way to do it? I could definitely add a docker-compose file, unsure about the running it part.

Does that mean that docker now is a dependency of the project we're asking users to have?

Zeko369:
If they have postgres running on their system they can just use that via the ENV variable. Otherwise they could use docker-compose.

It's an optional dependency if you don't have postgres/don't want to install it.

@flybayer what do you think about this?

You'll need to add a docker file and a docker-compose.yml file for this to work.

ryardley:
Would it be easier to bundle sqlite3 by default instead? We can use npm to install that and then we don't necessarily need to muck around with docker. We may want to make docker/postgres combo an installer later?

flybayer:
Here's my thoughts:

The goal here is to have a database running as easy as possible.
- On macOS, this is already very easy `brew install postgres`. So Docker is probably worse here
- Windows it seems is much harder, so Docker may be the easiest for this

I think we should stick with Postgres because it's likely what people will use in production. If we use sqlite, then the project grinds to a halt when they try to deploy. I'd rather frontload this difficulty than defer it. And once PG is set up on their machine, they are set for "life". 

I think we should solve this with tooling & docs. An installer is a great idea.

ryardley:
> If we use sqlite, then the project grinds to a halt when they try to deploy.

I think that happens no matter what you do. Deploying a stateful system is always a little difficult. The danger in changing dialect might be important if you are deep in the thick of app development and have used non compatible queries. I doubt that would be the case when you first have to do this which would happen as you add state to the application. 

In my opinion the best case scenario is that blitz just works with no db.

When you first run `blitz db migrate` the cli should prompt you to install and select a db.

[...]

PR #283, opened 2020-04-27
Change default DB to SQLite instead of postgres

flybayer:
### Type: change 

Closes: #253 
Closes: #282 

This changes the default database for newly created apps to use sqlite instead of postgres. See #253 for more context

### Checklist

- [ ] Tests added for changes


merelinguist:
Much better!

PR #284, opened 2020-04-27
Major cleanup to `blitz new` dependency update + graceful failure (Closes #202, #284)

### LINKED ISSUE #202: Blitz new dependency updates can fail, but it should not
**What is the problem?**

If there's a network error or something that interrupts fetching the latest dependencies, the entire `blitz new` process fails. It should not.

1. It should retry each request 1 time
2. If still fails, then fallback to the template version and keep going.
3. Also we should write tests for this whole update functionality

**Versions:**

```
blitz: 0.6.0-canary.0 (global)
```

**Supporting Documentation**
```
/t/mm> blitz new app

Hang tight while we set up your new Blitz app!

CREATE    .babelrc.js
CREATE    .env
CREATE    .eslintrc.js
CREATE    .npmrc
CREATE    .prettierignore
CREATE    README.md
CREATE    app/components/.keep
CREATE    app/components/ErrorBoundary.tsx
CREATE    app/layouts/.keep
CREATE    app/pages/_app.tsx
CREATE    app/pages/_document.tsx
CREATE    app/pages/index.tsx
CREATE    blitz.config.js
CREATE    db/index.ts
CREATE    db/migrations/.keep
CREATE    db/schema.prisma
CREATE    integrations/.keep
CREATE    jobs/.keep
CREATE    package.json
CREATE    public/favicon.ico
CREATE    public/logo.png
CREATE    tsconfig.json
CREATE    utils/.keep
CREATE    .gitignore

◢ Retrieving the freshest of dependenciesFetchError: invalid json response body at https://registry.npmjs.org/next reason: Unexpected end of JSON input
    at ~/.config/yarn/global/node_modules/node-fetch/lib/index.js:272:32
    at runMicrotasks (<anonymous>)
```

### LINKED ISSUE #262: Change blitz version to be pinned in new app template

Skn0tt:
### Type: Feature, tests

Closes: #202, closes #262 

As to #202:

- Each request is retried 1 time
- If still fails, template version is used as fallback
- On network errors, the user is prompted to run `yarn` afterwards
- added tests emulating network errors
### Checklist

- [x] Tests added for changes


Skn0tt:
This PR is based on #272, which should be merged first.

Skn0tt:
I'm facing some problems with testing: Somehow, the tests terminate before `blitz new` run through. Could somebody (preferably experienced with Jest) have a look at this?

packages/cli/src/generators/app.ts
```diff
@@ -45,46 +46,62 @@ class AppGenerator extends Generator<AppGeneratorOptions> {
...
+        throw new Error('Failed running prettier')
+      }
+    } else {
+      console.log(
+        chalk
+          .hex(themeColor)
+          .bold(
+            `\nThere seems to be some problem with your network. We'll skip installing your dependencies for now, make sure to compensate by running ${
+              this.options.yarn ? "'yarn'" : "'npm install'"
+            } once your connection is working again.`,
```
aem, line 102:
```suggestion
            `\nWe had some trouble connecting to the network, so we'll skip installing your dependencies right now. Make sure to run ${
              this.options.yarn ? "'yarn'" : "'npm install'"
            } once you're connected again.`,
```

Just some small copy edits on this message

packages/cli/src/utils/get-latest-version.ts
```diff
@@ -1,24 +1,36 @@
 import {fetchAllVersions, fetchLatestDistVersion} from './npm-fetch'
+import {log} from '@blitzjs/server'
 
-export const getLatestVersion = async (dependency: string, templateVersion: string) => {
+/**
+ * @example const [ version, isFallback ] = await getLatestVersion("blitz", "2.x");
```
aem, line 5: I like the new explicit API! Can we update these utilities to return an object that conforms to something like an `interface NpmResult {dependency: string; falback: boolean, version: string}` so the return value is a little more explicit? Right now you have to read into the function to really understand what the return value signifies

Skn0tt: I agree with you that an object is clearly a better choice for readability, though I don't really understand why `dependency` should be part of it. Callers already know the dependency since they pass it into it in the first place, why should it be returned?

Skn0tt: Please have a look at 90722873e8439cec4b97583cb69a48f4614acc50, I think that captures it nicely.

flybayer:
@Skn0tt I just merged in latest canary and test is failing. Can you take a look?

Also, it looks like #272 should be canceled because this PR also fixes #262, right?

flybayer:
Fyi I'm working right on this to bring it over the finish line

flybayer:
Great work on this @Skn0tt!!

Issue #295, opened 2020-04-27
Uniform code style in monorepo 


aem:
### What do you want and why?

We should standardize the code style in the monorepo. As the project grows and we add more functionality, consistency will be crucial in allowing the core team to review PRs efficiently, as well as allowing contributors to move between parts of the codebase without too much friction.

As discussed in Slack, we'd like to start enforcing two rules:

1. Named exports for all non-TSX files. 
2. `lower-kebab-case` file names
3. For single export files, the filename must match the name of the export

For example:

```typescript
// in cli/src/generators/page-generator.ts
export class PageGenerator extends Generator<PageGeneratorOptions> {
  // ...
}

// in utils/src/log.ts
// filename does not need to match since there are multiple exports,
// but should still be lower-kebab case.
export function info(...) {}
export function warn(...) {}
export function error(...) {}

// in cli/templates/page/index.tsx
// function is a default export since it's a React component, the one
// exception to the named exports rule.
export default function __ModelName__Home(props) {}
```

### Possible implementation(s)

We should make these changes incrementally, they're changes that will almost certainly cause merge conflicts, so the smaller the batches are (maybe one directory at a time) the less disruptive the changes will be.

It may also be worth looking into eslint rules, but only if they're autofixable (mostly for the named exports). eslint rules that aren't autofixable for code style issues (not for rules that are there to prevent bugs) can cause developer overhead that isn't really necessary.

### Additional context

Slack thread here: https://blitzjs.slack.com/archives/CUM2C68T1/p1587925285162700

ryardley:
I think this sort of thing should be done completely by linting rules or not at all. There is also Kent Dodd’s prettier for file structures idea which might be an interesting thing to throw in the mix.

ryardley:
Maybe this too? https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-default-export.md

Zeko369:
We should defiantly also add this https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/first.md to keep imports organised, currently we just add them left and right without organising them.

We should keep them
1. Packages
2. Internal packages (core/cli/server)
3. Local files

jportela:
[...]

PR #298, opened 2020-04-27
Enables linting rules for kebabCase and no-default-export

jportela:
### Type: linting

Start of #295 

* Adds a `lint` script with warnings for default exports and non kebab-case filenames
* Starts applying these rules to the `core` package

PS: This `lint` script makes use of the default tsdx linter

### Next steps

* Apply these rules to the other packages
* Change warnings to errors
* Include this script into the pre-push hook

### Checklist

N/A

### Other information

N/A


packages/core/src/use-query.ts
```diff
@@ -47,10 +47,14 @@ export function useQuery<T extends QueryFn, O extends Options<T>>(
   }
 
   if (options?.paginated) {
+    // TODO: this could be an issue, unless we assume the `paginated` option will always be constant when it's called
```
jportela, line 50: I refrained from touching this. I can disable this rule, but IMO this is rule we should have to avoid issues later. I think the best solution would be to have `useQuery` and `usePaginatedQuery` as they are different enough ways to query data

.eslintrc
```diff
@@ -0,0 +1,10 @@
+{
+  "rules": {
+    "import/no-default-export": ["warn"],
+    "unicorn/filename-case": ["warn", {
+      "case": "kebabCase"
+    }],
+    "import/first": 0
+  },
+  "plugins": ["unicorn"]
```
ryardley, line 9: I didn't realise you need to install a plugin here. Would it be simpler to use. https://www.npmjs.com/package/eslint-plugin-filenames

packages/core/src/use-query.ts
ryardley: This will likely not change but it certainly can and should be addressed as it is a bug. Hmm to refactor this to be hooks safe without changing API we could run both and switch out a noop as queryFunction an also set `manual` to `true` on the one we don't want.

.eslintrc
jportela: we would need to install the `eslint-plugin-filenames` plugin instead. The advantage of `eslint-plugin-unicorn` for this case is that it actually accepts the option as `kebabCase` for the filename. The `filenames` plugin accepts a regex, which would look like this:
`"filenames/match-regex": ["warn", "^[a-z-]+$", true]`

packages/core/src/use-query.ts
flybayer: We can make breaking changes if needed, but I really want the unified API if at all possible.

We can ignore this on this PR but open an issue to discuss a real fix?

package.json
```diff
@@ -21,6 +21,7 @@
     "link-cli": "yarn workspace blitz link",
     "unlink-cli": "yarn workspace blitz unlink",
     "build": "lerna run build --no-private --stream",
+    "lint": "lerna run lint --stream --parallel",
     "test": "yarn run build && yarn run testonly",
```
flybayer, line 25:
```suggestion
    "test": "yarn run lint && yarn run build && yarn run testonly",
```

packages/core/src/use-query.ts
aem: unless we can guarantee that `paginated` is a constant (maybe we expect the query to have a `paginated` static field rather than a prop to the hook?) react will break if the flag ever changes. When the hooks called during render changes between renders React throws an error which will cause everything within the nearest error boundary to unmount from the page

flybayer: Easy approach would be to not support this changing — throw an error if it does.

We can also fork the react-query code if needed

flybayer:
@all-contributors add @jportela for code

packages/core/src/use-query.ts
ryardley: https://github.com/blitz-js/blitz/issues/334

flybayer:
@jportela can you fix the yarn.lock conflict? Then it's good to merge

jportela:
Sure thing, I’ll merge this after I fix the conflict and I’ll continue to work on fixing these linting rules on other PRs

jportela:
@flybayer the approval got stale from fixing the conflict, the only change I did was done automatically by `yarn install`, to fix the conflict, so this should be good to merge

PR #300, opened 2020-04-27
test(server): normalize CWD in relative paths synchronizer rule test

wKovacs64:
### Type: tests 

Tests were failing on Windows as the string replacement to strip the CWD from the final absolute path wasn't working due to it trying to strip `/projects/blitz/blitz` from `\projects\blitz\blitz\app\components\three`.

Normalized the CWD passed to `relativeToAbsolute` in test.

### Checklist

- [x] Tests added for changes

### Breaking change: no 

N/A

### Other information

N/A


ryardley:
Ahh crap sorry! Maybe we should run our unit tests on both windows and Linux in CI? Will create an issue

wKovacs64:
No sweat. And yeah, we should definitely test on both platforms in CI (I think Brandon mentioned that in another thread, but I could be imagining things). Speaking of CI, not sure why the tests haven't run for this PR...

flybayer:
CI didn't run because it was out of date with canary

Issue #296, opened 2020-04-27
Pipe yarn/npm output to log file rather than stdout


aem:
### What do you want and why?

The `blitz new` process looks scary because of all of the `npm`/`yarn` install output. We should hide this from the user and instead show them a nice loading spinner and maybe even some helpful docs to read while the process is running.

### Possible implementation(s)

`cross-env` should give us a way to pipe the output to, well, not the console. Worst case we can update the terminal command to `> ./[name]/blitz-log.log` or something.

eliasjohansson:
I can look into this!

flybayer:
We should show more progress than just a spinner. Ideally we could just hide all the warnings and leave the rest. 

Npm has a log level flag we can use but yarn doesn’t.

ryardley:
We can run `yarn --json` which serializes output (we should do this too eventually)

eliasjohansson:
So set loglevel for npm and manually format yarn output with `--json`? Looks like npm is going to implement a json flag aswell: https://docs.npmjs.com/misc/config#json. So maybe we can handle both the same way further down the road.

PR #302, opened 2020-04-27
Hide all the scary warning messages during `blitz new`

eliasjohansson:
### Type: DX improvement

Closes: #296 

`blitz new`
- Hide yarn/npm install output
- Pipe yarn/npm install output to blitz-log.log
- Show spinner while installing dependencies

### Breaking change: no


ryardley:
Are we happy to write a log file every time regardless of success failure? Just wondering if for example it would make sense to have a flag configure the install log and then buffer all output before sending to blitz-error.log only if the process fails?

flybayer:
Pending changes according to: https://github.com/blitz-js/blitz/issues/296#issuecomment-620528980

eliasjohansson:
Alright, second version of this. Set npm logs to silent. Used the `--json` flag for yarn and then took the `step` messages and made them look like the rest of the CLI. I might be missing something, but is this closer to what we want? @aem @flybayer

flybayer:
Yarn output is awesome!!!

image

---

Npm not so much (see below). There's two other bugs here I just opened issues for:
- Prettier error when using npm: https://github.com/blitz-js/blitz/issues/331
- Git repo not being initialized before dependency install so husky fails: https://github.com/blitz-js/blitz/issues/330

Ugh, we need those e2e tests so bad

**Let's try also using a spinner for npm and see if we get any pushback (probably won't)** We can add more fine grained status updates for npm when they add the json flag

--- 
image

eliasjohansson:
So add back a spinner for NPM. Do we want to output the npm logs to a file instead? And im guessing we want to output any errors in console still?

flybayer:
The UX of this is good now, thank you!

One question: if npm or yarn throws an error, will that error be printed to the console (it should be)

eliasjohansson:
@flybayer Now all errors should be logged. Had alot of issues with actually testing this. Prisma throws the following error when installing with npm:
`Error: Either provide --schema or make sure that you are in a folder with a schema.prisma file.`
However it does not result in a failed install.

flybayer:
Fyi I'm working right now to fully land this

flybayer:
Sweeeet — great work @eliasjohansson!

Issue #241, opened 2020-04-25
Implement blitz db reset command


flybayer:
**What do you want and why do you want it?.**

Sometimes during dev things get out of wack and you need a sledgehammer.

`blitz db reset` should totally delete the database and then run migration to create the db fresh.

- Prompt user for confirmation
- Should work with SQLite, Postgres, and Mysql
- Should force close any existing connections to the DB. (A postgres `DROP TABLE` command will fail if anything is connected to it)

LoriKarikari:
I'm up for this one 😊

ryardley:
Would it be worth considering a blitz db seed command as well?

flybayer:
@ryardley yes definitely!

vavdoshka:
in case if someone will read this - this is how I managed to do it as of today:

`blitz prisma migrate reset && blitz db seed` with postgres DB.

PR #304, opened 2020-04-27
Add db reset

LoriKarikari:
### Type: feature

Closes: [#241](https://github.com/blitz-js/blitz/issues/241)

 Added the `blitz db reset` command which deletes the database and runs a fresh migration afterwards. It works with PostgreSQL, MySQL and  SQLite.

### Checklist

- [ ] Tests added for changes

### Other information
Since Prisma 2 doesn't have a `reset` command anymore I had to write some custom code to interact with the database. Currently this is done by getting the database url from the .env file and connecting with the database clients (except for SQLite). This currently works if the user sticks to the default options like using the public schema in Postgres and storing the SQLite database in the db folder. This can be optimized to use other schemas and database locations. Also maybe add a --force flag to skip the prompt.


LoriKarikari:
> 😬 It seems like your prettier messed up reading configs, try rerunning it

Sorry, my bad. Opening VSCode from WSL messed with my plugins. Should be fixed now.

packages/cli/package.json
```diff
@@ -1,80 +1,85 @@
...
+	},
+	"keywords": [
+		"blitz",
+		"cli"
+	],
+	"repository": {
+		"type": "git",
+		"url": "https://github.com/blitz-js/blitz"
+	},
+	"gitHead": "9c54cfaef19b63c30a7e41447541c7cc30b5c6d4"
```
flybayer, line 84: Sorry for the delay on review here!

This pkg.json is converted to tabs. Looks like this is still left over from the prettier mess up :)

flybayer:
Overall this looks great!!

Did you know that prisma has `db.raw('select * ...')` for SQL access? 😬 I think you can remove the pg and mysql clients and use that.

flybayer:
(converting to draft until ready for review again)

LoriKarikari:
@flybayer Must have missed that, still learning a lot about Prisma. Will edit this and remove the clients.

flybayer:
Sorry about that — I should have added this into the original issue description

wKovacs64:
Closing in favor of #339

Issue #160, opened 2020-04-21
Support vanilla Javascript for the blitz new template


flybayer:
You can use vanilla JS right now in a Blitz app by creating new files in JS or converting the existing files to JS. 

However, currently blitz new generates all Typescript files. We need to add a prompt during blitz new for selecting Typescript or plain Javascript.

The main unknown on implementing this is how to minimize code duplication for template files. Can we write all templates in TS and then have a tool that strips all the TS stuff? 

Note: we also need this same JS support for all the `blitz generate` commands.

PR #305, opened 2020-04-28
Add plain JavaScript support with `blitz new myapp --js`

aem:
### Type: Feature 🆕
Closes: #160 

We now fully support JavaScript file generation out of the box! If the user supplies the `--js` flag to `blitz new` we'll convert all file contents to JS using Babel (and move all file extensions to be `.js`) before committing our changes, and we'll ignore generation of the `tsconfig.json` for these projects as well. In future instances when calling `blitz generate`, we'll look for the `tsconfig.json` and only generate TS files for projects that already have a `tsconfig.json`.

Making this automatic based on the config means that if you'd like to migrate to TS at a later date it's as simple as adding a `tsconfig.json` and we'll start generating your files in TS instead of JS. Currently there's no way to go in the other direction (generate JS files in a TS project), but if we find a significant need we can figure that out.

~This depends on the changes in #281 so currently this branch is based off of that one, I'll rebase against `canary` once that gets merged.~

### Checklist

- [ ] Tests added for changes

### DIFF
--- a/packages/cli/src/commands/new.ts
+++ b/packages/cli/src/commands/new.ts
@@ -26,12 +26,12 @@ export default class New extends Command {
 
   static flags = {
     help: flags.help({char: 'h'}),
-    // ts: flags.boolean({
-    //   char: 't',
-    //   description: 'generate a TypeScript project',
-    //   default: true,
-    //   allowNo: true,
-    // }),
+    ts: flags.boolean({
+      char: 't',
+      description: 'generate a TypeScript project',
+      default: true,
+      allowNo: true,
+    }),
     yarn: flags.boolean({
       description: 'use Yarn as the package manager',
       default: hasbin.sync('yarn'),
@@ -53,6 +53,7 @@ export default class New extends Command {
       destinationRoot,
       appName,
       dryRun: flags['dry-run'],
+      useTs: flags.ts,
       yarn: flags.yarn,
       version: this.config.version,
     })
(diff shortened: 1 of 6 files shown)


aem:
> blitz generate still generates TS files inside a JS project

Oof. Looks like I accidentally dropped a commit while rebasing against `canary` 😨 Let me try to revive that

packages/cli/src/commands/new.ts
```diff
@@ -26,10 +26,9 @@ export default class New extends Command {
 
   static flags = {
     help: flags.help({char: 'h'}),
-    ts: flags.boolean({
-      description: 'generate a TypeScript project',
-      default: true,
-      allowNo: true,
+    js: flags.boolean({
+      description: 'Generates a JS project. Omitting this flag will TypeScript as the default language.',
```
flybayer, line 30:
```suggestion
      description: 'Generates a JS project. TypeScript is the default unless you add this flag.',
```

I went to fix a typo and decided to tweak it a bit

flybayer:
Still generating TS

1. `yarn link-cli` in root
2. Outside monorepo somewhere, run `blitz new -h` to verify the flag is there
3. `blitz new myapp --js`
4. JS files are generated properly here
5. `cd myapp`
6. `blitz g all user`
7. Fail: ts files are generated

aem:
if you run `blitz new -h` in between 5 and 6 there what do you get @flybayer? I think once you install deps in the new project it's going to use the locally installed versinon of `blitz` rather than the linked/global one

flybayer:
Oh yes 🤦

PR #316, opened 2020-04-28
Revert "Print helpful error if try to use CLI commands outside of blitz app"

aem:
Reverts blitz-js/blitz#203 - it broke the ability to create new apps since that command inherently has to be run outside of a blitz app


merelinguist:
@Zeko369 FYI. Free free to create another pr which fixes these problems

PR #319, opened 2020-04-28
Add helpful error messages to CLI when trying to use commands outside an app

Zeko369:
### Type: feature

Another shot at #203 because of Oclif bug

### Breaking change: No, but the same was sad last time 😄


packages/cli/src/checkBeforeRunning.ts
```diff
@@ -0,0 +1,39 @@
+import {Hook} from '@oclif/config'
+import isBlitzRoot, {IsBlitzRootError} from './utils/is-blitz-root'
+import chalk from 'chalk'
+
+const whitelistGlobal = ['new']
+
+const hook: Hook<'init'> = async function (options) {
+  // Bug with oclif
```
aem, line 8: did you open an issue with their project? would be good to get this fixed at some point

packages/cli/src/checkBeforeRunning.ts
```diff
@@ -0,0 +1,39 @@
...
+  const id = (options as any).Command.id
+  if (id && whitelistGlobal.includes(id)) return
+
+  const {err, message, depth} = await isBlitzRoot()
+
+  if (err) {
+    switch (message) {
+      case IsBlitzRootError.notBlitz:
+        this.log(
```
aem, line 17: can we use the `@blitzjs/server/utils/log` utility rather than the Command logger?

packages/cli/src/utils/is-blitz-root.ts
```diff
@@ -0,0 +1,70 @@
+import {readJSON} from 'fs-extra'
+import pkgDir from 'pkg-dir'
+import {resolve} from 'path'
+
+export enum IsBlitzRootError {
+  notBlitz,
```
aem, line 6: can we use `PascalCase` for enum values? we've used that convention in a few other places in the CLI

packages/cli/src/utils/is-blitz-root.ts
```diff
@@ -0,0 +1,70 @@
...
+ * @name isBlitzRoot
+ * @returns IsBlitzRootError enum
+ * notBlitz -> when can't find package.json in current folder and first found in parent
+ *             doesn't have blitz in dependencies
+ * notRoot -> if in a nested folder of blitz project (found blitz as depend in a parent package.json)
+ * badPackageJson -> an error occurred while reading local package.json
+ */
+
+const isBlitzRoot = async (): Promise<{err: boolean; message?: IsBlitzRootError; depth?: number}> => {
+  try {
```
aem, line 35: i wonder if this check is a bit complex and instead if we can just check for the existence of a `blitz.config.js`?

Zeko369: As @flybayer said, we're not requiring `blitz.config.js` nor `next.config.js`, so we should also be able to detect vanilla apps without those files

aem: Oops missed that. Disregard then

packages/cli/src/checkBeforeRunning.ts
Zeko369: https://github.com/oclif/config/issues/108

flybayer:
Thank you @Zeko369!!

PR #320, opened 2020-04-28
Update new app instructions in template for SQLite

kandros:

Since the default, Prisma DB is now SQLite instructions needed to be changed as there is no more a requirement to have PostgreSQL running.

.env is also no longer required due to no PostgreSQL
some instructions have been moved from app index page to readme because the app would not run before running these command


packages/cli/templates/app/.env
```diff
@@ -1,3 +0,0 @@
-# THIS FILE SHOULD NOT BE CHECKED INTO YOUR VERSION CONTROL SYSTEM
```
aem, line 1: I think we still want the `.env` file in the template version that we put into the generated app. We're not using Postgres by default anymore, but it's likely that any nontrivial production app will want to use postgres or something similar, so we can put a comment here telling the user how to use this file if that's the case. Something like:

```
# You can use this file to specify secrets like your database connection information
# if you move to a more sophisticated database than sqlite. For example, for postgres:
DATABASE_URL=postgresql://__username__@localhost:5432/__name__
```

packages/cli/templates/app/README.md
```diff
@@ -2,20 +2,22 @@
...
 
-3. DB migrate
+2. DB migrate
 
 ```
 blitz db migrate
 ```
 
-4. Start the dev server
+2. Start the dev server
```
aem, line 20: Two `2` steps here :)

packages/cli/templates/app/app/pages/index.tsx
```diff
@@ -1,12 +1,6 @@
 import {Head, Link} from 'blitz'
 
-const modelSnippet = `model Project {
```
aem, line 3: We should still include the `model` step in these instructions to match the README

flybayer:
Thank you!

Issue #179, opened 2020-04-22
Setup memory file system for tests


ryardley:
**What do you want and why do you want it?.**
We need to setup an in-memory file system for our file writing mocks here https://github.com/blitz-js/blitz/tree/canary/packages/server/test

Problem that is happening occasionally is that if a test fails to clean up the file system the files left over could be accidentally checked into git also an inmem fs would have a clean up/ reset method we could call on beforeEach.

Also this should theoretically speed up our tests as they would require less disk access.

**Possible implementation(s)**
Possibly metro-memory-fs mem-fs memfs or jest-plugin-fs not sure which one is best

**Background**
I had a super quick go but was getting errors on first pass so I put it down and wrote this up to see if anyone wants to have a go. I think there is going to be a bit of investigation to work out what's going on.

dajinchu:
down to try working on this

PR #323, opened 2020-04-28
In-memory FS for tests

dajinchu:
### Type: tests

Closes: #179 

Using `mock-fs` to mock the file system. Entire fixtures folder is deleted and mock files are specified on a per-test basis. `mock-fs` mocks out all uses of fs, so `resolveBinAsync` no longer works, since there is no node_modules in the mock file system. Added a mock for it as well.

### Checklist

- [ ] Tests added for changes


dajinchu:
These tests actually exposed a real bug as shown by test case `dev > when with next.config > should fail when passed a next.config.js` where we don't log to console when there's a next.config.js, we just throw and exit. The test is skipped for now as its failure causes other tests not to run.

ryardley:
> These tests actually exposed a real bug as shown by test case dev > when with next.config > should fail when passed a next.config.js where we don't log to console when there's a next.config.js, we just throw and exit. The test is skipped for now as its failure causes other tests not to run.

Yep this needs refactoring. I will have a go at fixing this up today.

flybayer:
@all-contributors add @dajinchu for code

wKovacs64:
Yeah so this change doesn't work in Windows. 😔 I'll open a new issue with details.

PR #215, opened 2020-04-24
Cypress E2E for examples

Zeko369:
### Pull Request Type

Setup and add Cypress tests to example/store

- [x] Feature

### Checklist

- [x] Tests added for changes

### What's the reason for the change? :question:

Closes: #97

### What are the changes and their implications? :gear:

- [x] Setup Cypress
- [x] Update eslint to not yell at us for using global stuff
- [x] Write cypress tests
	- [x] /
    - [x] /products
    - [x] /products/ssr
    - [x] /products/[handle]
    - [x] /admin
	- [x] /admin/products
	- [x] /admin/products/new
	- [x] /admin/products/[handle]

### Does this introduce a breaking change? :warning:
NO

### Other information

Some progress 🧪

image

PR #335, opened 2020-04-29
Add cypress E2E test for store example + CI (+ switch store to SQLite)

flybayer:
I'm added a few things to https://github.com/blitz-js/blitz/pull/215. Doing it here since I don't have write access to that branch

cc @Zeko369


flybayer:
YAY E2E testing for example with CI

PR #339, opened 2020-04-29
Add `blitz db reset` command

LoriKarikari:
### Type: feature
Closes: [#241 ](https://github.com/blitz-js/blitz/issues/241)

Added a `blitz db reset` command. Works with PostgreSQL, MySQL and SQLite
I changed the branch name because the previous `add-db-migrate` was wrong and confusing so it opened a new pull request. Previously I was getting the database urls from the .env file and connecting with the database through client packages. Now I'm using the Prisma client to determine which type of database is connected and use `db.raw()` to run the queries against the database.

### Checklist

- [ ] Tests added for changes

### Other information

Sorry for creating another pull request for the same issue. Could've renamed the remote branch too.


packages/cli/src/commands/db.ts
```diff
@@ -1,13 +1,17 @@
...
+import {resolveBinAsync} from '@blitzjs/server'
+import {Command, flags} from '@oclif/command'
 import chalk from 'chalk'
+import {spawn} from 'cross-spawn'
+import {prompt} from 'enquirer'
+import * as fs from 'fs'
 import * as path from 'path'
-import {resolveBinAsync} from '@blitzjs/server'
+import {promisify} from 'util'
+import {loadBlitz} from '../utils/load-blitz'
```
flybayer, line 9:
```suggestion
```

packages/cli/src/commands/db.ts
```diff
@@ -1,13 +1,17 @@
...
 
 const schemaPath = path.join(process.cwd(), 'db', 'schema.prisma')
 const schemaArg = `--schema=${schemaPath}`
 const getPrismaBin = () => resolveBinAsync('@prisma/cli', 'prisma')
+const {db} = loadBlitz()
```
flybayer, line 14:
```suggestion
const db = require(path.join(process.cwd(), 'db').default
```
I didn't actually test this, but I think this is correct

packages/cli/types/index.d.ts
```diff
@@ -1 +1,2 @@
 declare module 'hasbin'
+declare module 'mysql2/promise'
```
flybayer, line 2:
```suggestion
```

flybayer:
Nice, looking good!

One other thing, it looks like you need to run `yarn` again to get a fresh lock file (it still has the dependencies you added and then removed)

packages/cli/src/commands/db.ts
LoriKarikari: I was doing it like this first but it says `Error Plugin: @blitzjs/cli: @prisma/client did not initialize yet. Please run "prisma generate" and try to import it again.` Without default I was able to get the database connections but raw wasn't working.

packages/cli/src/commands/db.ts
```diff
@@ -57,6 +61,66 @@ export const runMigrate = async () => {
...
+    )
+    // currently assuming the public schema is being used
+    // delete schema and recreate with the appropriate privileges
+    await db.raw('DROP SCHEMA public cascade;')
+    await db.raw('CREATE SCHEMA public;')
+    await db.raw('GRANT ALL ON schema public TO postgres;')
+    await db.raw('GRANT ALL ON schema public TO public;')
+    // run migration
+    await runMigrate()
+    console.log('Your database has been reset.')
```
aem, line 79: This all looks great! Love the clarity with the platform-specific implementations. As a note to myself, we'll need to eventually support customizing this hook in database plugins in the future.

Would you be able to use the built-in logging utilities we have in `@blitzjs/server/utils/log.ts` rather than `console.log`/`console.error`? If everything is passing through those log utilities we can control the look and feel of the CLI a bit better

packages/cli/src/commands/db.ts
LoriKarikari: Okay it works. Had to run migrate again forgot about that with all the swapping between databases. But I can't get the test to pass with this import it says `'Cannot find module '/home/lori/dev/oss/blitz/blitz/packages/cli/db' from 'db.ts'' from 'db.ts'`. It searches for this file at the cli package root. Do I have to add something to the tests?

packages/cli/src/commands/db.ts
LoriKarikari: Wasn't sure what to do with logging so thanks for the heads up. I'm currently just logging the errors when the reset fails but should I replace them by more generic messages like "Database reset failed" or something? The database errors are pretty clear about what's going wrong but maybe it's too much clutter in the CLI?

aem: @LoriKarikari maybe both? Let's try something like this:
```
log.error('[[insert operation name here]] failed with an error from the database:')
log.error(error)
```
we may need to update the type of `log.error` to handle `Error` objects, I think right now it only accepts `string`, but the console knows how to stringify `Error`s so we should be able to pass those as well

packages/cli/src/commands/db.ts
flybayer: Ah yeah, you need to do the same thing that `loadBlitz` is doing.
```ts
import pkgDir from 'pkg-dir'
const projectRoot = pkgDir.sync() || process.cwd()
```

You can move that into a utils file or something if you want.

LoriKarikari:
I was still getting the `cannot find module ...` error even after using the package root. The test was failing for the `blitz start` command because it runs `runPrismaGeneration` which is located at the same file as the `reset` command. I was importing the `db` globally but it doesn't exist until the generation is complete. So now I dynamically load the `db` at the `reset` command. Added some nicer logging, and removed some unnecessary stuff. I think it should be okay now?

flybayer:
Thank you Lori, great work!

hauptrolle:
First of all thank you for your great work here!!

However I still get the `cannot find module ... db` error when running `blitz db reset`. Am I doing something wrong? 🙈

LoriKarikari:
> First of all thank you for your great work here!!
> 
> However I still get the `cannot find module ... db` error when running `blitz db reset`. Am I doing something wrong? 🙈

Thanks for reporting this issue @hauptrolle. You are indeed right, hower I am not able to reproduce this error when using this command with the code from the repository so I'm not sure what's going on. Also the error is pointing at a folder that exists `'/home/lori/test/db'` as before it was looking at the CLI root itself. @flybayer any ideas what could be wrong?

LoriKarikari:
@hauptrolle can you open an issue for this problem?

hauptrolle:
@LoriKarikari Sure. Just created one: https://github.com/blitz-js/blitz/issues/401

PR #342, opened 2020-04-29
Improve error message for api routes in wrong folder

kandros:


packages/server/src/synchronizer/pipeline/rules/pages/index.ts
```diff
@@ -52,7 +52,7 @@ const createRulePages = ({config, errors, getInputCache}: RuleArgs) => {
           ? 'Warning: You have tried to put an api route inside a pages directory:'
           : 'Warning: You have tried to put api routes inside a pages directory:'
 
-      const secondary = 'All api routes should be in their own directory.'
+      const secondary = 'All api routes should be in their own directory (/app/api)'
```
ryardley, line 55: This sounds like it must be at the root `app` level where as it would be fine to be in a context. `app/products/api`

kandros: What about that?

Api routes should be in their own directory (outside of pages folder):
- Context-specific routes can be nested in a context directory (app/products/api) 
- Global Api routes such as webhooks can be at root level (app/api)

ryardley:
@merelinguist are you happy with this output? 
image

packages/server/src/synchronizer/pipeline/rules/pages/index.ts
```diff
@@ -52,7 +52,10 @@ const createRulePages = ({config, errors, getInputCache}: RuleArgs) => {
           ? 'Warning: You have tried to put an api route inside a pages directory:'
           : 'Warning: You have tried to put api routes inside a pages directory:'
 
-      const secondary = 'All api routes should be in their own directory.'
+      const secondary = `Api routes should be in their own directory (outside of pages folders):
+- Context-specific routes can be nested in a context directory (app/products/api)
+- Global Api routes such as webhooks can be at root level (app/api)
+`
```
flybayer, line 58: This is pretty good — I'll suggest a tweak:

```suggestion
      const secondary = `API routes should be in their own 'api/' folder as a sibling of 'pages/':
- Examples: app/api/, app/products/api/
`
```

---

If we want to **really** nail it, we could do this:

```
You need to move `app/products/pages/api/` to `app/products/api/`

Would you like us to do that for you? [Y/n]
```

ryardley: Yes! We should definitely provide these sorts of patterns.

kandros: agree, it would be a great UX

flybayer:
Awesome!

PR #344, opened 2020-04-29
docs(tutorial): fix a couple typos and change some copy

wKovacs64:

Fixed a couple typos, formatted a couple things for consistency, generated the directory listing with `tree`, etc.

### Checklist

- [ ] Tests added for changes

### Other information

@merelinguist Thoughts on these changes? I can remove anything you don't care for, just say the word. Also, I noticed a lot of the punctuation (particularly apostrophes and double quotation marks) are currently represented using codes that are part of the ASCII extended character set rather than the basic set (e.g. `<E2><80><99>` instead of `'`). Do we care to change those? It's probably a locale and/or OS thing - not sure how the extended characters look to others.

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -27,52 +27,67 @@ This should create a `mysite` directory in your current directory.
 Let’s look at what `blitz new` created:
 
 ```
-mysite/
-  app/
-    components/
-      ErrorBoundary.tsx
-    layouts/
-    pages/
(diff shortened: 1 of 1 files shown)


wKovacs64:
> Thank you!
> 
> @all-contributors add @wKovacs64 for docs

Does this bot fail sometimes? I certainty don't mind, just curious.

Issue #306, opened 2020-04-28
Architecture direction (discussion)


ryardley:
### What do you want and why?

This is a big one but could work as a roadmap for what we want in terms of the internal codebase of blitz moving forward.

We need to guide our internal architecture to avoid unnecessary dependencies and spaghetti code.

We want to somehow keep things orthogonal which means following "things that change together stay together" but also allow for horizontal layers to be grouped where appropriate.

### Possible implementation(s)

I have a suggestion (feel free to workshop this with me - also if you are a code owner and want to update the description of any package please comment):

# ~~app~~ framework

~~User app~~ Code that actually runs in user apps.  
* rpc, 
* cache management
* rpc and reactQuery templates used by generator and server-manager's synchronizer rules should live here and be imported

Code should be built for both node and browser and deeply linked. 

# blitz 
__Blitz wrapper package.__

Determines launch environment exports public api for users. This stays the same

# cli 
__Argument parsing__

Responsible for parsing command-line arguments and delegating data to the correct downstream packages and displaying help text. Data qualification with enquirer can be managed here as that is part of parsing arguments (or not?) 

Will probably accept configuration from the prisma package and generate an oclif plugin internally here to configure new commands.

# repl

__REPL__

Launch the blitz REPL 

# generator

__Dynamically generate source files in a users project__

Generate files within client app directory. Pull in templates and generators from the prisma package and elsewhere eventually and use that to generate assets appropriately. 

# prisma
__Database plugin__

Fledgeling db package and first plugin. Anything related to prisma should live here and get exported into wherever it is needed. API can be refactored internally and get ready for release

We want to avoid any system dependencies here that we don't control (eg. oclif) as whatever dependencies we have here that aren't prisma related will need to eventually become part of the plugin framework. Naturally anything prisma/db related is fine

# synchronizer 

__Accept rules and pipe and transform files based on those rules__

Accept config and setup a vinyl file transform pipeline with dynamic stages (could possibly eventually be used by generator) pipe them to a file writer interface. 

# server-manager

__Manage next-server process__

Accept an input stream listen for events and manage the next server start, stop and output appropriately

# ~~utils~~

~~__Shared general development libraries__~~

~~General small focussed development utilities similar to lodash and not blitz specific. Nothing should have any dependencies on anything else in the codebase aside from potentially other utilities.~~

EDIT: Avoiding a utils library will mean less refactoring in the future as we rarely will need to extract out libs from utils which may happen if people start adding things to utils.

# display

[...]

PR #346, opened 2020-04-29
refactor(cli): split out generator package

aem:
### Type: Refactor

Related: https://github.com/blitz-js/blitz/issues/306

The first bit in the new package architecture - regardless of where certain utils end up, we'll definitely want the generator code to be split out of the CLI. This creates a new `@blitzjs/generator` package which houses all of the files needed for codegen. I'm not sure we'll want to use them outside of the CLI, but it'll still be useful to at least have them logically separated from the CLI architecture so we can evolve the architectures separately.

@flybayer I made myself the codeowner for the generator package since I was effectively the codeowner for both anyway

### Checklist

- [ ] Tests added for changes


packages/generator/src/generator.ts
```diff
@@ -37,6 +35,8 @@ abstract class Generator<T extends GeneratorOptions = GeneratorOptions> extends
   private performedActions: string[] = []
   private useTs: boolean
 
+  abstract sourceRoot: string
```
aem, line 38: we had to make this change since templates are now relative to the generator rather than the `generate` CLI command.

aem:
alright. finally got template compilation across packages working properly. this is now ready to go @flybayer @merelinguist

Issue #345, opened 2020-04-29
[server] Blitz server tests are failing on Windows


wKovacs64:
### What is the problem?

Blitz tests are failing on Windows after merging #323.

### Steps to Reproduce: 

1. Clone repo
1. `yarn`
1. `yarn test`

### Versions:

Windows 10

### Supporting Documentation

<details>
<pre>
@blitzjs/server: FAIL test/api-routes.test.ts
@blitzjs/server:   ● Dev command › should copy the correct files to the dev folder
@blitzjs/server:     expect(received).toEqual(expected) // deep equality
@blitzjs/server:     - Expected
@blitzjs/server:     + Received
@blitzjs/server:       Object {
@blitzjs/server:     -   "children": Array [
@blitzjs/server:     -     Object {
@blitzjs/server:     -       "name": "blitz.config.js",
@blitzjs/server:     -     },
@blitzjs/server:     -     Object {
@blitzjs/server:     -       "name": "next.config.js",
@blitzjs/server:     -     },
@blitzjs/server:     -     Object {
@blitzjs/server:     -       "children": Array [
@blitzjs/server:     -         Object {
@blitzjs/server:     -           "children": Array [
@blitzjs/server:     -             Object {
@blitzjs/server:     -               "name": "bar.ts",
@blitzjs/server:     -             },
@blitzjs/server:     -             Object {
@blitzjs/server:     -               "name": "foo.ts",
@blitzjs/server:     -             },
@blitzjs/server:     -           ],
@blitzjs/server:     -           "name": "api",
@blitzjs/server:     -         },
@blitzjs/server:     -       ],
@blitzjs/server:     -       "name": "pages",
@blitzjs/server:     -     },
@blitzjs/server:     -   ],
@blitzjs/server:         "name": ".blitz-rules",
@blitzjs/server:       }
@blitzjs/server:       55 |   it('should copy the correct files to the dev folder', async () => {
@blitzjs/server:       56 |     const tree = directoryTree(devFolder)
@blitzjs/server:     > 57 |     expect(tree).toEqual({
@blitzjs/server:          |                  ^
@blitzjs/server:       58 |       name: '.blitz-rules',
@blitzjs/server:       59 |       children: [
@blitzjs/server:       60 |         {name: 'blitz.config.js'},
@blitzjs/server:       at test/api-routes.test.ts:57:18
@blitzjs/server:       at step (../../node_modules/tslib/tslib.js:139:27)
[...]

PR #347, opened 2020-04-29
Fix server tests not running on windows

dajinchu:
### Type: bug fix

Closes: #345 

Use resolve and relative paths with mockfs so we don't have issues between `/dev` and `C:\dev`. 
😡 Windows :rage:

### Checklist

- [ ] Tests added for changes

### Breaking change: no


ryardley:
I don't have a win vm on my machine @wKovacs64 can you help review this?

dajinchu:
yeah, we could try something with https://www.npmjs.com/package/unionfs if needed. It plays nicely with `mock-fs`. 

Not sure where we could document this, but the workaround for console.log is to reference `console` anywhere before the mock happens, and then it will work even after mock.

wKovacs64:
I tried a unionfs/memfs combo, but ran into trouble (don't recall what exactly, would have to revisit), but that type of overlay approach would probably be better than a pave-over. Maybe unionfs/mock-fs is the magic pairing. 🤷‍♂️

ryardley:
Sounds like this solves our immediate problem but we need to revisit the approach

wKovacs64:
Can we get this merged for now @flybayer? We'll circle back in #352.

Issue #348, opened 2020-04-29
Remove purgecss from tailwind example


merelinguist:
### What do you want and why?

Tailwind now supports purgecss by default! We should use this rather than including it in the example’s `postcss.config.js`

### Possible implementation(s)

1. Upgrade tailwind to the latest version

2. Uninstall `postcss`

3. Remove `postcss` related code from `examples/tailwind/postcss.config.js`

4. Edit the `tailwind.config.js` as described here https://github.com/tailwindcss/tailwindcss/releases/tag/v1.4.0

Jfelix61:
Would love to work on this!

PR #362, opened 2020-04-30
Native tailwindcss purgecss integration

Jfelix61:
### Type: enhancement

Closes: #348 

Integrate tailwind's default purgecss instead of having it as a separate package.


examples/tailwind/package.json
```diff
@@ -50,11 +49,10 @@
     "eslint-plugin-react-hooks": "3.0.0",
     "husky": "4.2.5",
     "lint-staged": "10.1.7",
-    "next": "9.3.5",
```
merelinguist, line 53: I believe we still need next installed as per the note above

merelinguist:
Thank you for the great work here! 🥳

flybayer:
🔥

osirvent:
Since Nextjs is in `./app`, `tailwind.config.js` shouldn't reflect that?

```
module.exports = {
  purge: [
    "./app/pages/**/*.{js,jsx,ts,tsx}",
    "./app/components/**/*.{js,jsx,ts,tsx}"
  ],
  theme: {
```

flybayer:
@osirvent yes, you're exactly right — looks like the example is probably broken right now. Want to open a PR or issue for this?

osirvent:
Perfect, I'll do.

osirvent:
#381

Issue #360, opened 2020-04-30
Blitz server manager doesn't ignore files defined in .gitignore


flybayer:
### What is the problem?

The Blitz internal server manager is processing files that are ignored by git.

This isn't a critical issue, but it could cause perf problems if there are ignored directories with thousands of files.

### Steps to Reproduce: 

1. Add a file to .gitignore
2. blitz start
3. Change that file
4. See the following:
```
Processing file....ignoredFile.txt
✔ Prepped for launch
```

### Versions:

Canary

cc @ryardley

PR #365, opened 2020-04-30
Fix git ignored files being processed when they shouldn't

Skn0tt:
### Type: bug fix

Closes: #360

The chokidar file watching config is extended by ignored / included files based on all present `.gitignore`s.
These are read in recursively.

### Checklist

- [ ] Tests added for changes


Skn0tt:
Where should I put tests for this? While developing, I tested using the store example, but I don't think that's where they belong.

ryardley:
Hey cheers for the PR! In my opinion tests should go as close to the code as possible.

packages/server/src/parse-chokidar-rules-from-gitignore.ts
```diff
@@ -0,0 +1,55 @@
+import parseGitignore from 'parse-gitignore'
+import fs from 'fs'
+import {resolve} from 'path'
+import _ from 'lodash'
+
+function getAllGitIgnores(rootFolder: string) {
+  function getGitIgnoresRecursively(folder: string, prefix: string): {gitIgnore: string; prefix: string}[] {
```
ryardley, line 7: Any reason we can't just fast-glob for gitignore files?

packages/server/src/parse-chokidar-rules-from-gitignore.ts
```diff
@@ -0,0 +1,55 @@
+import parseGitignore from 'parse-gitignore'
+import fs from 'fs'
+import {resolve} from 'path'
+import _ from 'lodash'
```
ryardley, line 4: Would we get a tny install bundle optimization by using a destructured import? 

```ts
import {partition} from 'lodash'
```

packages/server/test/parse-gitignore.test.ts
```diff
@@ -0,0 +1,25 @@
+import parseGitignore from 'parse-gitignore'
```
ryardley, line 1: Why do we need to test an external library?

packages/server/src/parse-chokidar-rules-from-gitignore.ts
```diff
@@ -0,0 +1,55 @@
...
+  return getGitIgnoresRecursively(rootFolder, '')
+}
+
+export function parseChokidarRulesFromGitignore(rootFolder: string) {
+  const result: {ignoredPaths: string[]; includePaths: string[]} = {
+    includePaths: [],
+    ignoredPaths: [],
+  }
+
+  for (const {gitIgnore, prefix} of getAllGitIgnores(rootFolder)) {
```
ryardley, line 33: Perhaps the following block could be its own unit tested function? Just that it is is unclear on first pass the intent around the prefix stuff.

packages/server/src/parse-chokidar-rules-from-gitignore.ts
Skn0tt: Didn't know about fast-glob, will use 👍🏻

packages/server/src/parse-chokidar-rules-from-gitignore.ts
Skn0tt: I don't know, do we have tree shaking working? I generally prefer having the `_` there, since it makes it obvious we're using lodash (familiar to a lot of people).

ryardley: Also I can't see we're using lodash anywhere else as it is probably brought in by a sub dependency. If we are using explicitly I would probably suggest we explicitly add it.

packages/server/test/parse-gitignore.test.ts
Skn0tt: - to validate and fixate our assumptions about its behaviour. Gives you peace of mind when upgrading its version down the road
- to have an easy-to-reproduce playground for testing out libraries

ryardley: Hmm this is kind of already tested here: https://github.com/jonschlinkert/parse-gitignore/blob/master/test/test.js Perhaps it makes more sense to unit test this library in the context of the way we are using it?

Skn0tt: We can do that. There's a test in https://github.com/blitz-js/blitz/pull/365/commits/2648d55b53be210b8e6656ef3c1430894263cc29, I'll remove this one.

Also, I googled what these kinds of tests are called again: Contract tests.

packages/server/src/parse-chokidar-rules-from-gitignore.ts
```diff
@@ -1,6 +1,6 @@
 import parseGitignore from 'parse-gitignore'
 import fs from 'fs'
-import _ from 'lodash'
+import {partition} from 'lodash'
```
kandros, line 3: this will only reduce bundle size when using [babel-plugin-lodash](https://github.com/lodash/babel-plugin-lodash) or `lodash/es`

you could use `import partition from 'lodash/partition` which does not require tree-shake and take advantage of `lodash` "per file" repo strucure

Skn0tt: Thanks for the input, I'll change it to `lodash/partition`. I'll open an issue on tree shaking anyway, since I don't feel like this is the kind of issues contributors should have to worry about.

Skn0tt: Welp, now that I think about it - do we have to care about bundle size? I don't think Lodash will be bundled into our code, anyway, since this won't be run client-side, does it?

ryardley: You are right we shouldn't be too concerned about this instance and I am not as it is there as a transient dependency anyway and we don't control it in a bunch of stuff. I am more worried about precedent honestly especially if it is used in the frontend as this is the first time our code uses it. 

I don't think it is a big deal but honestly the best practice here is probably to use a deep import: 

```
import partition from 'lodash/partition';
```

That way if that was copied everywhere in the codebase we might be able to save size in the packages we could.

packages/server/src/config.ts
```diff
@@ -2,6 +2,8 @@ import {resolve} from 'path'
 import {ciLog} from './ciLog'
 import {resolveBinAsync} from './resolve-bin-async'
 import {synchronizeFiles} from './synchronizer'
+import _ from 'lodash'
```
ryardley, line 5: This is unused. One other reason why we shouldn't use the `_` is that typescript considers `_ `as a possibly unused value and won't complain when we don't use it.

packages/server/src/parse-chokidar-rules-from-gitignore.ts
ryardley: I don't think we need an issue on tree shaking btw.

packages/server/src/config.ts
ryardley: Happy to merge if we can clean this up. I will leave it up to you for whether or not you want to import deeply.

Skn0tt:
I've changed it to a deep import,. not placing a precedent makes absolute sense. Could you approve again? The commit staled your previous review.

ryardley:
Have to resolve conflicts. I am moving today. Will merge once I am settled.

PR #366, opened 2020-04-30
Fix additional lint rules

jportela:
### Type: linting

Closes: #295 
Follow-up to: #298 

Uniform linting rules across all subpackages. 0 warnings, 0 errors.

### Checklist

- [x] Tests added for changes


packages/cli/src/commands/db.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable eqeqeq */
```
jportela, line 1: it seems `==` is being used intentionally here over `===`, so I'm disabling this rule for this file only. In general, I think this is a rule we want to promote, so I didn't disable it globally

ryardley: Hmm. Exit code is a number so to me that looks like it should be referential equality.

jportela: `code` can actually be `null`:
https://nodejs.org/api/child_process.html#child_process_event_exit

> If the process exited, code is the final exit code of the process, otherwise null

Looking at this,`===` would definitely be what we want here, since we are targeting the success case. I'll remove the disable and fix the equality comparison

jportela: Typescript is also inferring it properly as `number | null` so I'm removing the type declaration on the argument of the exit event

ryardley: I love it when linting solved real world bugs! :)

packages/cli/src/commands/build.ts
```diff
@@ -1,8 +1,7 @@
 import {Command} from '@oclif/command'
 import {build} from '@blitzjs/server'
 
-// eslint-disable-next-line import/no-default-export
```
jportela, line 4: just checking, because of the eslint-disable comments, do we intend to export these as `default`?

aem: nope, that's just legacy. moving forward we'd like everything to be named exports

flybayer:
Uh oh, lot's of conflicts 😬

jportela:
no worries, that's to be expected, should be straightforward to solve. If we plan to merge soon, I can do that now

flybayer:
@jportela yeah I can merge soon

jportela:
@flybayer how do you feel of upgrading these two linting rules to `error` instead of `warning`, to actually fail the build?

flybayer:
@jportela which rules?

jportela:
the `No Default Export` and `Kebab Case for filenames`

flybayer:
I.e. the current code is all correct and that we can now change those to be an error to ensure new PRs conform? If this is what you are asking, then yes let's error them.

flybayer:
@jportela Good to merge?

jportela:
@flybayer yes, just need to wait for the build checks to finish :)

jportela:
@flybayer I actually found a typo on my .eslintrc changes, I'm pushing the fix for it now

jportela:
@flybayer done, feel free to merge whenever 👍

flybayer:
Sweeet, thank you!

Issue #188, opened 2020-04-23
Set up Windows CI test runner


flybayer:
**What do you want and why do you want it?.**
Most of us are on macOS or Linux and easily break things from running on Windows.

We need a CI windows test runnner to ensure we always run good on Windows too.

**Possible implementation(s)**

We are currently using Github actions. I'm not sure if this supports windows? We can add another CI service for windows if needed.

wKovacs64:
We need to get this going ASAP, I can't keep up with all the Windows issues, haha! 😅 

Most projects I've seen targeting Windows in CI use AppVeyor, but it appears [GitHub Actions supports Windows Server 2019](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#github-hosted-runners). I'll try to mess around with this tomorrow.

flybayer:
Awesome, let me know if you need any help from me! Slack DM or a direct tag here in GH are the best ways to get my attention

PR #368, opened 2020-04-30
ci: run tests on both Windows and Ubuntu

wKovacs64:
### Type: tests 

Closes: #188

Modified the main CI workflow to run against an OS matrix including `ubuntu-latest` and `windows-latest` (Windows Server 2019 at the time of this PR).

### Checklist

- [x] Tests added for changes

### Breaking change: no 

N/A

### Other information

N/A

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,7 +13,10 @@ on:
 jobs:
   build_and_test:
     name: Build & Test
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v2
       - name: Use Node.js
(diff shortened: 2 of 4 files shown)


flybayer:
@wKovacs64 awesome! Did you look at what's currently failing on the windows CI?

wKovacs64:
Briefly, but I can't reproduce it on Windows 10 locally. I should have more time tomorrow to investigate further.

flybayer:
Ok cool, mark as ready for review when done

wKovacs64:
Do we want all Windows issues resolved before merging this so we start with a fully passing CI?

flybayer:
Good question — what do you think we should do?

wKovacs64:
I guess there's no point in merging before we clear up the known issues, or every PR going forward will fail CI. 😄

wKovacs64:
Fixed the lint issue by enforcing LF line endings at the repository level (Git for Windows defaults to auto convert to CRLF, which was causing eslint to stroke out). This should really be in the repo regardless of CI, so let me know if you'd prefer a separate PR for just that and I'll extract it out.

Anyway, now we're only failing on known issues so... that's... good... I guess... heh.

wKovacs64:
This is finally ready! 

I think someone with repository settings access may need to adjust the branch protection rules, as the "Build & Test" job is now two separate checks. See https://github.community/t5/GitHub-Actions/Expected-Waiting-for-status-to-be-reported/m-p/52179#M8428.

flybayer:
Yay!!!

Issue #331, opened 2020-04-29
Bug: `blitz new myapp --no-yarn` fails before completion because of prettier


flybayer:
### What is the problem?

Creating a new blitz app with npm fails without completing

### Steps to Reproduce: 

1. `blitz new myapp --no-yarn`
2. It will fail with error `Error: Failed running prettier`

### Versions:

Latest canary

### Supporting Documentation

```
CREATE    tsconfig.json
CREATE    utils/.keep
CREATE    .gitignore

✔ Retrieving the freshest of dependencies

Installing those dependencies...
Scary warning messages during this part are unfortunately normal.

npm WARN deprecated mkdirp@0.5.3: Legacy versions of mkdirp are no longer supported. Please update to mkdirp 1.x. (Note that the API surface has changed to use Promises in 1.x.)
npm WARN deprecated path-is-absolute@2.0.0: This package is no longer relevant as Node.js 0.12 is unmaintained.
npm WARN deprecated @types/serialize-error@4.0.1: This is a stub types definition. serialize-error provides its own type definitions, so you do not need this installed.

> @prisma/cli@2.0.0-beta.3 preinstall /private/tmp/tt/node_modules/@prisma/cli
> node preinstall/index.js


> fsevents@1.2.12 install /private/tmp/tt/node_modules/webpack/node_modules/fsevents
> node-gyp rebuild

  SOLINK_MODULE(target) Release/.node
  CXX(target) Release/obj.target/fse/fsevents.o
  SOLINK_MODULE(target) Release/fse.node

> @prisma/cli@2.0.0-beta.3 install /private/tmp/tt/node_modules/@prisma/cli
> node download-build/index.js


> husky@4.2.5 install /private/tmp/tt/node_modules/husky
> node husky install

husky > Setting up git hooks
fatal: not a git repository (or any of the parent directories): .git
husky > Failed to install

> @scarf/scarf@1.0.4 postinstall /private/tmp/tt/node_modules/@scarf/scarf
> node ./report.js


    The dependency 'react-query' is tracking installation
    statistics using Scarf (https://scarf.sh), which helps open-source developers
    fund and maintain their projects. Scarf securely logs basic installation
    details when this package is installed. The Scarf npm library is open source
    and permissively licensed at https://github.com/scarf-sh/scarf-js. For more
    details about your project's dependencies, try running 'npm ls'. To opt out of
    analytics, set the environment variable 'SCARF_ANALYTICS=false'.


[...]

PR #370, opened 2020-04-30
Fix prettier failure on `blitz new` with npm

Skn0tt:
### Type: bug fix

Closes: #331

The code was trying to run `npm run prettier`, but that's not that easy:
You'd have to create a special `prettier` script in your package.json to make that work.
The workaround for NPM users is to use NPX, which ships since NPM v5.2.

### Checklist

- [ ] Tests added for changes


flybayer:
Awesome, thank you!!

Issue #372, opened 2020-04-30
VSCode ESLint integration not working


wKovacs64:
### What is the problem?

The introduction of `.eslintrc` in #298 broke VSCode/ESLint integration (reproduced on multiple devices).

### Steps to Reproduce: 

Assuming you had ESLint working in VSCode previously...

1. Clone Blitz repo (current `canary` branch as of this issue being logged)
1. Install dependencies (`yarn`)
1. Open in VSCode
1. Open a file (e.g. `packages/core/src/index.ts`) and observe lint errors (e.g. `Parsing error: The keyword 'export' is reserved`)
1. Delete `.eslintrc` and the errors go away and integration is restored

### Versions:

N/A

### Supporting Documentation

The tsdx CLI supports [merging `.eslintrc.js`](https://github.com/jaredpalmer/tsdx#eslint) (don't think it honors `.eslintrc` though), but that isn't enough for VSCode, which doesn't know to perform said merge (so it only looks at the config it finds, which in this case is missing things like the TS parser, etc.). The [solution from tsdx](https://github.com/jaredpalmer/tsdx/issues/354) is to "eject" the ESLint config if you need to extend it while retaining editor integration.

PR #373, opened 2020-04-30
style: fix editor integration with ESLint

wKovacs64:
### Type: tooling fix 

Closes: #372

Ran `tsdx lint --write-file` to "eject" the tsdx ESLint config and merged in our recent adoption of the `unicorn` plugin and corresponding rules.

### Checklist

- [ ] Tests added for changes

### Breaking change: no 

N/A

### Other information

The tsdx CLI supports [merging `.eslintrc.js`](https://github.com/jaredpalmer/tsdx#eslint) (don't think it honors `.eslintrc` though), but that isn't enough for VSCode, which doesn't know to perform said merge (so it only looks at the config it finds, which in this case is missing things like the TS parser, etc.). The [solution from tsdx](https://github.com/jaredpalmer/tsdx/issues/354) is to "eject" the ESLint config if you need to extend it while retaining editor integration.


flybayer:
Awesome, thank you so much!

PR #381, opened 2020-05-01
Fix postcss purge routes on Tailwind example

osirvent:

Fix `examples/tailwind`. Nextjs lives on `/app`.
```
module.exports = {
  purge: [
    "./app/pages/**/*.{js,jsx,ts,tsx}",
    "./app/components/**/*.{js,jsx,ts,tsx}"
  ],
  theme: {
```

### Checklist

- [ ] Tests added for changes

### DIFF
--- a/examples/tailwind/tailwind.config.js
+++ b/examples/tailwind/tailwind.config.js
@@ -1,7 +1,6 @@
 module.exports = {
   purge: [
-    "./app/pages/**/*.{js,jsx,ts,tsx}",
-    "./app/components/**/*.{js,jsx,ts,tsx}"
+    "./**/{pages,components}/**/*.{js,jsx,ts,tsx}",
   ],
   theme: {
     extend: {},
--- a/examples/tailwind/tailwind.config.js
+++ b/examples/tailwind/tailwind.config.js
@@ -1,5 +1,8 @@
 module.exports = {
-  purge: ["./pages/**/*.{js,jsx,ts,tsx}", "./components/**/*.{js,jsx,ts,tsx}"],
+  purge: [
+    "./app/pages/**/*.{js,jsx,ts,tsx}",
+    "./app/components/**/*.{js,jsx,ts,tsx}"
+  ],
   theme: {
     extend: {},
   },


Skn0tt:
Thanks for taking the time to provide a PR 🎉

examples/tailwind/tailwind.config.js
```diff
@@ -1,5 +1,8 @@
 module.exports = {
-  purge: ["./pages/**/*.{js,jsx,ts,tsx}", "./components/**/*.{js,jsx,ts,tsx}"],
+  purge: [
+    "./app/pages/**/*.{js,jsx,ts,tsx}",
+    "./app/components/**/*.{js,jsx,ts,tsx}"
+  ],
```
flybayer, line 5: Thanks! One tweak:

```suggestion
  purge: [
    "./**/{pages,components}/**/*.{js,jsx,ts,tsx}",
  ],
```

I'm not sure if this actually works, but it should be something close to this. Match file inside any page or component directory anywhere inside the project, including at root.

osirvent:
Nice! Im gonna check if it is everything working as expected. We must create some tailwind styling and see if purgecss works as it should be

osirvent:
I've been testing `.css` files size with diferent `purge` routes on `/examples/tailwind`. After running `blitz build` with different routes:

**1. NO Purge: 146 kB**
`purge: []`
imagen

**2. No `/app` route: 1.48 kB**
```
purge: [
  "./pages/**/*.{js,jsx,ts,tsx}",
  "./components/**/*.{js,jsx,ts,tsx}"
]
```
imagen

**3. `/app` route: 1.57 kB**
```
purge: [
  "./app/pages/**/*.{js,jsx,ts,tsx}",
  "./app/components/**/*.{js,jsx,ts,tsx}"
]
```
imagen

**4. `./**/{pages,components}` route: same as 3.** with this simple example
```
 purge: [
    "./**/{pages,components}/**/*.{js,jsx,ts,tsx}",
  ]
```

I don't know why but it seems that case number 2 works. But number 4 is gonna work better on no trivial examples.

Jfelix61:
I don't think modifying the routes like this will work since blitz compiles into a next.js file structure.

image


Instead, I believe it should be like this to match: 

```
purge: [
  "./pages/**/*.{js,jsx,ts,tsx}",
  "./app/**/*.{js,jsx,ts,tsx}"
]
```

osirvent:
Hmmm, so there's no need to add `/app`. That's why number 2 option works!

@flybayer should close the PR then?

flybayer:
Yep, looks great! Thank ya!

Issue #334, opened 2020-04-29
Fix `useQuery` rules of hooks violation


ryardley:
### What do you want and why?

We unfortunately accidentally have had a rules of hooks violation:

https://github.com/blitz-js/blitz/blob/canary/packages/core/src/useQuery.ts#L49

__To avoid a breaking change:__

Investigate switching on hooks configuration to swap out noop objects so we don't have to run hooks in an if

### Possible implementation(s)

```ts
const paginatedData =  usePaginated(paginated, queryFunction , queryObject)
const nonPaginatedData =  useNonPaginated(paginated, queryFunction , queryObject)

return paginated ? paginatedData : nonPaginatedData
```

We may want to ensure that the manual option is set to `true` for whatever react-query hook is running with noop data.

eliasjohansson:
Hmm, did we just implement hooks validation? Did not think of this. It's unfortunate react-query doesnt have a `skip` parameter like apollo useQuery..

ryardley:
I found a `manual` property that might work for our needs but needs to be tested

flybayer:
The other solution is simply not support the paginated value changing — throw an error if `paginated` ever changes

Skn0tt:
I'm not that familiar with useQuery's characteristics, but wouldn't smth like this work?

```ts
useQuery([key, skip], ([key, skip]) => {
  if (skip) {
    return Promise.resolve(null);
  }
  return fetcher(key);
})
```

eliasjohansson:
Yea something like that. Im off work tomorrow so can have a go at fixing it then. If  nobody else jumps on it while im sleeping 😅

MrLeebo:
What's wrong with going forward with two separate `useQuery()` and `usePaginatedQuery()` hooks? A boolean parameter that toggles the behavior of a function might as well be two functions.

eliasjohansson:
Hm, yeah im starting to feel like it makes more sense with seperate hooks as well. Especially when we will make `useInfiniteQuery` seperate. Also people who have used react-query before will be more familiar.

flybayer:
As discussed in the other PR, we will change to have separate hooks for each, just like react-query

flybayer:
Closing this because I opened #517 for separating the hooks.

PR #391, opened 2020-05-01
fix useQuery react hooks violation

eliasjohansson:
### Type: refactor | fix

Closes: #334

Moved usePaginatedQuery out of its if statement to comply with rules of hooks.
Both `react-query` hooks will now be called, but the one not active will simply return null. Only the return value of blitz `useQuery` is now instead conditional based on `options.paginated`.

### Checklist

- [ ] Tests added for changes

### Other information

I had to add 'Paginated' to the query key since both hooks cant use the same key. Does anyone know if this might cause an issue somewhere else?

### DIFF
--- a/packages/core/src/use-query.ts
+++ b/packages/core/src/use-query.ts
@@ -38,23 +38,24 @@ export function useQuery<T extends QueryFn, O extends Options<T>>(
   params?: InferUnaryParam<T>,
   options?: O,
 ): [PromiseReturnType<T>, RestReturnType<T, O>] {
-  const queryKey: [string, any] = [(queryFn as any).cacheKey, params]
-  const queryFunction = (_: string, params: {}) => queryFn(params)
   const queryOptions = {
     suspense: true,
     retry: process.env.NODE_ENV === 'production' ? 3 : false,
     ...options,
   }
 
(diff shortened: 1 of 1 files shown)


eliasjohansson:
We could also use `react-query` [manual](https://github.com/tannerlinsley/react-query#manual-querying) option. But that makes it not possible toggle pagination runtime. Atleast not with how its done right now. (not sure why/when you would wanna toggle it though)

MrLeebo:
Thank you for tackling this issue!

However, I think this should just be two hooks. The "paginated" param fundamentally changes the behavior of the hook and I don't understand why a component would ever need to toggle it in the same component. We shouldn't present users with an option if we don't want them to take it because inevitably, someone will. 

I'm worried that this solution, while clever, introduces new potential issues. Anytime a method has a boolean parameter that drastically changes its behavior like this, that method is usually written better as two methods. Also, attaching 'Paginated' to the cache key is surprising and makes it possible to collide with another cache key by that name.

eliasjohansson:
Should we close this and make seperate query hooks instead?

flybayer:
@eliasjohansson yeah, let's switch to using separate hooks, just like react-query has

Issue #353, opened 2020-04-30
Building store example fails on Windows


wKovacs64:
### What is the problem?

Maybe the relative synchronizer rule is normalizing import paths when it shouldn't? Not sure yet.

### Steps to Reproduce: 

1. Clone Blitz repo on Windows
1. Either run the tests, or cd into `examples/store` and try to build the store example
1. Observe `Bad character escape sequence` error

### Versions:

```
debug: monorepo
debug: pkgPath: C:\dev\contrib\blitz\node_modules\@blitzjs\cli

Windows 10 | win32-x64 | Node: v12.16.1

blitz: 0.7.0 (global)
blitz: 0.7.1-canary.1 (local)
```

### Supporting Documentation:

```
Creating an optimized production build

Failed to compile.

./app/products/components/ProductForm.tsx
SyntaxError: C:\dev\contrib\blitz\examples\store\.blitz\caches\build\app\products\components\ProductForm.tsx: Bad character escape sequence (4:51)

  2 | import {Product, ProductCreateInput, ProductUpdateInput} from 'db'
  3 | import createProduct from 'app\products\mutations\createProduct'
> 4 | import updateProduct from 'app\products\mutations\updateProduct'
    |                                                    ^
  5 |
  6 | type ProductInput = ProductCreateInput | ProductUpdateInput
  7 |


> Build error occurred
```

Blitz code (`examples\store\app\products\components\ProductForm.tsx`):

```ts
import createProduct from '../mutations/createProduct'
import updateProduct from '../mutations/updateProduct'
```

Compiled code (`examples\store\.blitz\caches\build\app\products\components\ProductForm.tsx`):

```ts
import createProduct from 'app\products\mutations\createProduct'
import updateProduct from 'app\products\mutations\updateProduct'
```

The resulting backslashes are incorrect, they should remain forward slashes (in code).

### Note:

I don't seem to run into this in a newly generated project, but we're still getting the issue reported in #299 so we might just not be making it that far yet? 🤔

Jfelix61:
#### Problem
After some investigation, the [`relativeToAbsolute`](https://github.com/blitz-js/blitz/blob/99b4cd628c25a6d647a33e0bcaf249e2a11f6da6/packages/server/src/synchronizer/pipeline/rules/relative/index.ts#L42) function is the culprit. `path.sep`, `process.cwd()` and the file's paths are being returned with back slashes.

[...]

PR #393, opened 2020-05-01
fix(server): force forward slashes in absolute imports

wKovacs64:
### Type: bug fix 

Closes: #353

Sindre Sorhus to the rescue again. Slashified the result of `path.join` (it was trying to be helpful, but it wasn't in this particular case) to enforce forward slashes in absolute paths. A hack? Maybe. I don't know anymore.

### Checklist

- [x] Tests added for changes

### Breaking change: no 

At least, I sure hope not.

### Other information

CrossPlatformFileAccess50

### DIFF
--- a/packages/server/src/synchronizer/pipeline/rules/relative/relative-paths.test.ts
+++ b/packages/server/src/synchronizer/pipeline/rules/relative/relative-paths.test.ts
@@ -1,21 +1,21 @@
 import {relativeToAbsolute, replaceRelativeImports} from './index'
 
+// I don't like this one bit, but Node's path module behaves so differently
+// based on the runtime platform that it's just easier to test paths as they
+// would be on the platform currently executing the test suite. Tests should
+// be run on both platforms, so both test paths will be executed - just not
+// during the same run.
+const platformSensitiveAbsolutePath = (p: string) =>
+  process.platform === 'win32' ? 'C:' + p.split('/').join('\\') : p
+
 describe('relativeToAbsolute', () => {
-  // I don't like this one bit, but Node's path module behaves so differently
-  // based on the runtime platform that it's just easier to test paths as they
-  // would be on the platform currently executing the test suite. Tests should
-  // be run on both platforms, so both test paths will be executed - just not
-  // during the same run.
-  const isWindows = process.platform === 'win32'
   const tests = [
     {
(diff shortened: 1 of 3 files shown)


wKovacs64:
Still working on this, avert your eyes.

simonpeterdebbarma:
@wKovacs64 Sure, I'm just organizing all Issues and PRs. Let me know if I assigned the wrong tags.

wKovacs64:
@simonpeterdebbarma Oh you're good, that wasn't directed at you - more of a general warning for anyone who stumbled along (although I guess that should be obvious from being in draft). 🙂

wKovacs64:
OK, I think this is ready.

The testing situation sucks because the implementation relies on Node's `path` module which returns different values based on the operating system on which the tests are running at the time. To deal with this, I added a platform check in the tests and pass the platform-specific paths based on that result. This is kind of bad because now there are 2 code paths (Windows and not-Windows) in the test suite for this module, where one executes on Windows and the other on macOS and Linux. That said, they will both be run once we have Windows as part of our OS matrix in CI (PR ready to go), so it should be clear when a change breaks one platform but not the other. An alternative approach might be to spy on `path` functions and mock their implementations per test but that felt worse. Open to other suggestions!

packages/server/src/synchronizer/pipeline/rules/relative/relative-paths.test.ts
```diff
@@ -1,38 +1,51 @@
...
+  const isWindows = process.platform === 'win32'
   const tests = [
     {
       name: 'Provides an absolute path within app',
       input: {
-        relativeImport: normalize('../components/three'),
-        filename: normalize('/projects/blitz/blitz/app/users/pages.ts'),
+        relativeImport: '../components/three',
+        filename: isWindows
+          ? 'C:\\projects\\blitz\\blitz\\app\\users\\pages.ts'
```
ryardley, line 16: Maybe could make a function to do this so we don't get so much testing noise? 

```ts
const platformSensitiveAbsolutePath = p => process.platform === 'win32' ?  'C:' + p.split('/').join('\\') : p
```

wKovacs64: Done!

Issue #395, opened 2020-05-02
[Linux] Retrieving the freshest of dependenciesTypeError


itranga:
Retrieving the freshest of dependenciesTypeError: Object.fromEntries is not a function
    at /usr/lib/node_modules/blitz/node_modules/@blitzjs/cli/lib/src/utils/fetch-latest-version-for.js:22:23
    at Generator.next (<anonymous>)
    at fulfilled (/usr/lib/node_modules/blitz/node_modules/tslib/tslib.js:110:62)

Ubuntu

simonpeterdebbarma:
@itranga Could you kindly provide more information on the version of node and yarn you are using?

flybayer:
@itranga this should be fixed in `0.8.1` — can you test and verify?

itranga:
"react-native-web": "^0.11.6"
node - v10.20.1
npm - 6.14.4 
yarn - 1.21.1

flybayer:
@itranga did you try Blitz `0.8.1`?

itranga:
blitz -v
You are using alpha software - if you have any problems, please open an issue here:
  https://github.com/blitz-js/blitz/issues/new/choose

Linux 4.15 | linux-x64 | Node: v10.20.1

blitz: 0.8.1 (global)Retrieving the freshest of dependenciesTypeError: Object.fromEntries is not a function
    at /usr/local/share/.config/yarn/global/node_modules/@blitzjs/cli/lib/src/utils/fetch-latest-version-for.js:22:23
    at Generator.next (<anonymous>)
    at fulfilled (/usr/local/share/.config/yarn/global/node_modules/tslib/tslib.js:110:62)

flybayer:
I wonder if this is a Node version issue.

Could you test with Node v12?

cktang88:
FWIW, I'm able to run it fine with the newest version.
```
Linux 5.3 | linux-x64 | Node: v14.4.0

blitz: 0.16.4 (global)
blitz: 0.16.4 (local)
```

flybayer:
Thanks @cktang88! I'll go ahead and close it then

Issue #394, opened 2020-05-01
[Linux] Eslint Plugin React Hooks error while resolving packages. 


Jfelix61:
### What is the problem?

Resolving packages step from `blitz new app` fails because it can't find version `39` (in fact, does not exist) from `eslint-plugin-react-hooks`. 

### Steps to Reproduce: 

1. Run `blitz new app`

### Versions:

```
debug: local
debug: pkgPath: /home/jfelix/Documents/programming/open-source/blitz/node_modules/@blitzjs/cli 

Linux 5.3 | linux-x64 | Node: v12.16.1

blitz: 0.8.0 (global)
blitz: 0.8.0 (local)
```

### Supporting Documentation

Please include applicable logs and screenshots that show your problem.
![image](https://user-images.githubusercontent.com/21092519/80848712-28b45c80-8be2-11ea-8154-08eb65c52757.png)

Issue #392, opened 2020-05-01
[Linux] getting "Object.fromEntries is not a function" when retrieving dependencies


pcrunn:
### What is the problem?
I can't create a new project because the blitz cli can't retrieve the dependencies

### Steps to Reproduce: 

1. Clear npm module cache (if it exists and blitz's project dependencies are in there)
2. Create a new project using
`blitz new myapp`

### Versions:

```
debug: global
debug: pkgPath: /usr/local/lib/node_modules/blitz/node_modules/@blitzjs/cli/lib/src/index.js

Linux 4.19 | linux-x64 | Node: v10.20.1

blitz: 0.8.0 (global)
```

### Supporting Documentation
Here is the exact error i'm getting

```
◤ Retrieving the freshest of dependenciesTypeError: Object.fromEntries is not a function
    at /usr/local/lib/node_modules/blitz/node_modules/@blitzjs/cli/lib/src/utils/fetch-latest-version-for.js:22:23
    at Generator.next (<anonymous>)
    at fulfilled (/usr/local/lib/node_modules/blitz/node_modules/tslib/tslib.js:110:62)
```
(this is after the project's file structure is created)

LoriKarikari:
Thanks for reporting this issue @pcrunn 👍 we will look into this asap.

LoriKarikari:
Btw, I just checked this out and I'm having the same problem. The error message is slightly different though.
```
Linux 4.19 | linux-x64 | Node: v13.13.0

blitz: 0.8.0 (global)

```
```
Couldn't find any versions for "eslint-plugin-react-hooks" that matches "39"
Error: Failed running prettier
    at AppGenerator.<anonymous> (~/.npm-global/lib/node_modules/blitz/node_modules/@blitzjs/cli/lib/src/generators/app.js:127:27)
    at Generator.next (<anonymous>)
    at fulfilled (~/.npm-global/lib/node_modules/blitz/node_modules/tslib/tslib.js:110:62)
```

flybayer:
@pcrunn  this should be fixed in `0.8.1` — can you test and verify?

pcrunn:
> @pcrunn this should be fixed in `0.8.1` — can you test and verify?

hmmm, not sure about that. i'm still getting the same error

simonpeterdebbarma:
@pcrunn It might have to do with your Node version. Can you try switching to the latest version from v10.20.1?

is2ei:
It seems the error occurs here:
[...]

PR #396, opened 2020-05-02
Fix `blitz new` on Linux (fromEntries/eslint-plugin error)

Jfelix61:
### Type: bug fix

Closes: #392 
Closes: #394 
Closes: #395 

`latestVersion` now returns version instead of array's index number. This will return the correct version when `latestDistVersion` does not contain desired version located in template.

### Checklist

- [x] Tests added for changes


flybayer:
@Jfelix61 thank you!! Does this also fix #392?

Jfelix61:
Let me check real quick that issue.

Jfelix61:
Yes, just ran the build successfully!

Issue #405, opened 2020-05-03
[Linux] fetchRemote.sh script fails when interpreted with non-bash shell


tmns:
### What is the problem?

When running `./scripts/fetchRemote.sh` as per the [contributing docs](https://github.com/blitz-js/blitz/blob/canary/CONTRIBUTING.md#sync-your-fork) with `sh` (and `dash` and possibly other shells?), the script errors out with a `[[: not found` error:

```bash
$ ./scripts/fetchRemote.sh 
./scripts/fetchRemote.sh: 5: ./scripts/fetchRemote.sh: [[: not found
fatal: 'upstream' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

This is due to the first line of the script being `!#/bin/sh`, which [does not understand `[[` in `if` statements](https://stackoverflow.com/a/3401197). In turn, this can be fixed by either changing the line to `!#/bin/bash` or by making the script `sh` compatible.

### Steps to Reproduce:

1. Attempt to run `fetchRemote.sh` with `sh` or another non-compatible shell

### Versions:

```
Linux 4.13 | linux-x64 | Node: v12.3.1

blitz: 0.8.1 (global)
blitz: 0.8.1 (local)

```

### Supporting Documentation
* https://github.com/blitz-js/blitz/blob/canary/CONTRIBUTING.md#sync-your-fork
* https://stackoverflow.com/questions/3401183/bash-syntax-error-not-found

*Note: I edited this to more accurately reflect the issue of the problem, which occurs regardless if `upstream` is set, as long as a non compatible shell is used.*

Zeko369:
I forgot that's a bashism, could we maybe refactor that into something else?

Zeko369:
@tmns Could you run `ls -l /bin/sh` and copy the output here, I want to know which shell is default on your system

tmns:
Hi @Zeko369! Yes, this can also be refactored to just using single brackets, like so:

```
if [ $(git remote -v | grep 'upstream' | wc -l) -eq 0 ]; then
  echo "Upstream remote not found, adding one"
  git remote add upstream https://github.com/blitz-js/blitz.git
fi
```

And as for the default shell, it looks like `dash`:
```
ls -l /bin/sh
lrwxrwxrwx 1 root root 4 Ιούν 19  2018 /bin/sh -> dash
```

Zeko369:
DAMN YOU DASH.

[...]

PR #406, opened 2020-05-03
Fix fetchRemote dev script to be sh compatible

tmns:
### Type bug fix

Closes: #405 

* Does not use the `bash` `[[` syntax. Also uses `=` instead of `eq`.

### Checklist

- [ ] Tests added for changes

### Other information

*Note: The original pull request simply changed the script to use `/bin/bash` but has since been refactored to continue using `sh` with compatible syntax.*

### DIFF
--- a/scripts/fetchRemote.sh
+++ b/scripts/fetchRemote.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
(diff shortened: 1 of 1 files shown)


flybayer:
Thank you @tmns!!

@all-contributors add @tmns for code

PR #408, opened 2020-05-03
update db.ts for #383

nabi009:
I have changed in javascript, plz check typescript if I am not wrong so 

https://github.com/blitz-js/blitz/issues/383

exports.default = Db;
Db.description =`Run database

${chalk_1.default.bold('👉 migrate')}   Run any needed migrations via Prisma 2 and generate Prisma Client.

${chalk_1.default.bold('👉 introspect')}   Will introspect the database defined in db/schema.prisma and automatically \ngenerate a complete schema.prisma file for you. Lastly, it\'ll generate Prisma Client.

${chalk_1.default.bold('👉 studio')}   Open the Prisma Studio UI at http://localhost:5555 so you can easily see and\n change data in your database.
`;

Db.args = [

    {
        name: 'Commands',
        description: Db.description,
        required: true,
    },
   
];

Issue #383, opened 2020-05-01
`blitz db`  should print help message instead of error


nabi009:
command: blitz db

 ›   Error: Missing 1 required arg:
 ›   command  Run specific db command
 ›   See more help with --help

the description should be clear to user, that missed command or shows db related command.
such as:
**blitz db migrate
blitz db introspect
blitz db studio
etc ....**

flybayer:
Thank you @nabi009, you're absolutely right!

peaonunes:
Happy to work on that if possible 😄

flybayer:
@peaonunes yeah, that'd be awesome! Let us know if you need help

nabi009:
![Screen Shot 2020-05-03 at 05 49 50](https://user-images.githubusercontent.com/3170831/80896327-5f819400-8d02-11ea-84d3-36dc5155fa61.png)
I have done it.

peaonunes:
Looks nice! Feel free to open up the PR 😄 
**suggestion:** Perhaps we should use `log.warning` as it is a warning instead of a normal message? Idk. Seems to be missing the `reset` command, you can interpolate the `descriptions` static property to be in sync with the command defaults.

nabi009:
@peaonunes thanks so much from your suggestion, your right but I did not change anything too much.
the original code is in blitz/packages/cli/src/commands/db.js
`exports.default = Db;
Db.description =`Run database commands

${chalk_1.default.bold(' migrate')}   Run any needed migrations via Prisma 2 and generate Prisma Client.

${chalk_1.default.bold(' introspect')}   Will introspect the database defined in db/schema.prisma and automatically \ngenerate a complete schema.prisma file for you. Lastly, it\'ll generate Prisma Client.

${chalk_1.default.bold(' studio')}   Open the Prisma Studio UI at http://localhost:5555 so you can easily see and\n change data in your database.
`;

Db.args = [

    {
        name: 'Commands',
        description: 'Run specific db command',
        required: true,
    },
   
];

Db.flags = {
    help: command_1.flags.help({ char: 'h' }),
};
`
and I changed it to this
`exports.default = Db;
Db.description =`Run database

${chalk_1.default.bold('👉 migrate')}   Run any needed migrations via Prisma 2 and generate Prisma Client.

${chalk_1.default.bold('👉 introspect')}   Will introspect the database defined in db/schema.prisma and automatically \ngenerate a complete schema.prisma file for you. Lastly, it\'ll generate Prisma Client.

[...]

PR #409, opened 2020-05-03
Update log.ts

nabi009:
### Type: ?? 

Closes: ??

??

### Checklist

- [x] Tests added for changes


flybayer:
Hey @nabi009, thank you for your contribution!

It looks like you `log.pointing` isn't being used anywhere. Did you intend to log something with it?

nabi009:
@flybayer, I removed it for now.

flybayer:
@nabi009 CI is failing for this. Can you look into it?

nabi009:
`const withPointing = (str: string) => {
   return `👉  ${chalk.yellow(str)}`
 }`

it is just returning icon with string. what will be the error, anyhow I will check it. thanks

merelinguist:
Is `withPointing` used anywhere?

nabi009:
> Is `withPointing` used anywhere?

Yes in cli/command/db.ts

ryardley:
> Yes in cli/command/db.ts #383

Does it make sense to perhaps simply include it with that work?

nabi009:
> > Yes in cli/command/db.ts #383
> 
> 
> 
> Does it make sense to perhaps simply include it with that work?

It pointing commands to users that you can use these specific comments.

ryardley:
> Does it make sense to perhaps simply include it with that work?

What I mean is that I think we should merge this change into https://github.com/blitz-js/blitz/pull/408 as they depend on each other.

flybayer:
Closing — see https://github.com/blitz-js/blitz/pull/408#issuecomment-632130242

Issue #301, opened 2020-04-27
Support custom port + friendly help if port already in use


alvaromashiro:
### What is the problem?
Starting blitz app with command blitz start, and if default port which is  3000 and that port is used by example by an app in react dont change the port 

### Steps to Reproduce: 

1. 2. start react app 
1. 3. start blitz app

### Versions:

```
[Add the output of `blitz --version --verbose` here]
➜  myapp git:(master) blitz --version --verbose
You are using alpha software - if you have any problems, please open an issue here:
  https://github.com/blitz-js/blitz/issues/new/choose

debug: local
debug: pkgPath: C:\Users\alvar\react\myapp\node_modules\@blitzjs\cli

Windows 10 | win32-x64 | Node: v13.13.0

blitz: 0.7.0 (global)
blitz: 0.7.0 (local)
```

### Supporting Documentation

Please include applicable logs and screenshots that show your problem.

``` 

➜  myapp git:(master) blitz start
You are using alpha software - if you have any problems, please open an issue here:
  https://github.com/blitz-js/blitz/issues/new/choose

√ Prepped for launch
[ wait ]  starting the development server ...
[ info ]  waiting on http://localhost:3000 ...
√ Prepped for launch
Warning: You have enabled experimental feature(s).
Experimental features are not covered by semver, and may cause unexpected or broken application behavior. Use them at your own risk.

> Using "webpackDevMiddleware" config function defined in default.
[ info ]  bundled successfully, waiting for typecheck results...
[ wait ]  compiling ...
Attention: Next.js now collects completely anonymous telemetry regarding usage.
This information is used to shape Next.js' roadmap and prioritize features.
You can learn more, including how to opt-out if you'd not like to participate in this anonymous program, by visiting the following URL:
https://nextjs.org/telemetry

[ info ]  bundled successfully, waiting for typecheck results...
[ ready ] compiled successfully - ready on http://localhost:3000

```

Zeko369:
Yup, noticed this today. We should add a CLI prompt to run it on another PORT or exit

ivandevp:
I'd like to work on this issue 😃

aem:
I think it's probably not worth prompting the user, we should just keep incrementing by 1 until we find an open port

ivandevp:
[...]

PR #413, opened 2020-05-04
Adds --port and --hostname to blitz start and dev commands

ivandevp:
### Type: feature 

Closes: #301 

- Add `--port` and `--hostname` flags to start and dev commands
- Move `-p` alias to `--port` instead of `--production`
- Add `detect-port` library to get available port on hostname and pass it to nextjs

### Checklist

- [X] Tests added for changes

### Breaking change: yes 

Now projects running the build with `-p` for _production_ won't longer work since that alias will refer to `--port`.


ivandevp:
Is the error in Github Actions related with `react-query` normal? Is it happening because of any dependency update?

flybayer:
Looks good to me @ivandevp! But yeah, looks like the react-query update has a breaking type change. Can you back out that update? Then we should be good to go as long as @ryardley approves the server related changes

flybayer:
Just change react-query to be pinned at `1.2.9`

ivandevp:
@flybayer @ryardley I've updated the PR with a fixed version of `react-query`. Now it doesn't break the build.

ivandevp:
May I ask to be added to all-contributors? 😊

aem:
absolutely!

@all-contributors add @ivandevp for code

Issue #377, opened 2020-05-01
`blitz` takes too long to install


flybayer:
### What is the problem?

`yarn global add blitz` or `npm i -g blitz` takes a long time, presumably because we have so many dependencies.

We need to see how we can optimize this. Do we have some really heavy dependencies that swap out with something else? Or maybe we are only using one small part of a large dependency that we can inline into our code?

### Versions:

```
macOS Catalina | darwin-x64 | Node: v12.16.2

blitz: 0.7.0 (global)
```

Skn0tt:
I'll have a look at this.

Skn0tt:
If install time is a key metric we're focusing on, we should think about having some kind of hard limits on file size that we enforce as part of our CI.

Skn0tt:
Judging from [this](https://arve0.github.io/npm-download-size/#blitz), Next.js makes up 90% of our dependency size (14mb total, 12mb of that is Next.js).

Skn0tt:
## Some totally non-scientific measurements:

In a clean environment (`docker run node:12.16`), `yarn global add blitz` finishes in 42s and fetches 950 different packages. For me, it finishes in 40secs.

On my host system, where there's already lots of dependencies installed globally, it fetches ~2400 different packages and finishes in 60secs. A lot of this time is spent "building fresh packages", which seems to be a common problem with globally installed packages.

`yarn list --flat --prod | wc -l` counts ~2900 individual dependencies to be installed. Maybe the high number of individual packages contributes to slow install time.

---

We'd need to find some consensus on what "a long time" means to us. For me personally, 30secs is more than alright, the 60secs get on my nerves - but mostly because "building fresh packages" takes soooo long without any substantial progress.
How long does installing take for you @flybayer ?

LoriKarikari:
Here the installation takes 32s. Seems not too bad?

wKovacs64:
blitz@0.8.1 took 13s on my Windows host machine (most/all dependencies probably cached) and 42s in a VM with no cached dependencies. Off the cuff, I think I agree with Simon's thresholds: <30s is great, ~45s is fine, >=60s is starting to get irritating.

Skn0tt:
We could add a CI check to enforce those thresholds. @flybayer what do you think?

flybayer:
Ok cool, if Next is the large majority then that's really good. I wanted to make sure we weren't adding a ton ourselves.

[...]

PR #420, opened 2020-05-04
Add CI step to monitor install time

Skn0tt:
### Type:tests

Closes: #377

- Added an install time check to the CI

### Checklist

- [ ] Tests added for changes

### Other information

A cute otter


.github/checkInstallTime.js
```diff
@@ -0,0 +1,18 @@
...
+
+const fs = require('fs')
+const yarnOut = fs.readFileSync(0, {encoding: 'utf8'})
+
+const [installTimeString] = /(?<=^Done in )\d+\.\d+(?=s\.$)/m.exec(yarnOut)
+const installTime = Number(installTimeString)
+
+console.log(`Install time: ${installTime}s`)
+
+if (installTime < 30) {
```
Skn0tt, line 11: We'll have to account for CI installations being faster than local ones. Do y'all think we should leave it at < 30 green / < 50 orange | >= 50 red or lower our thresholds?

ryardley: I would be thinking we probably just want to know how large the delta is.

.github/workflows/main.yml
```diff
@@ -39,3 +39,9 @@ jobs:
         run: yarn test
         env:
           CI: true
+      - name: Test Install time
+        run: |
+          cd ../ && mkdir test && cd test
+          yarn cache clean
+          yarn add file:../blitz/packages/blitz --non-interactive | tee out.txt
+          cat out.txt | ../blitz/.github/checkInstallTime.js
```
ryardley, line 47: Random question I was wondering though why we need the intermediary file here?

.github/workflows/main.yml
```diff
@@ -39,3 +39,9 @@ jobs:
         run: yarn test
         env:
           CI: true
+      - name: Test Install time
+        run: |
+          cd ../ && mkdir test && cd test
+          yarn cache clean
```
ryardley, line 45: We probably want to use a different yarn here as clearing the cache will slow our build.

.github/workflows/main.yml
Skn0tt: You mean `out.txt`? I don't pipe the yarn output directly so it's not swallowed from the logs.

.github/workflows/main.yml
Skn0tt: Makes sense. I hoped to just use some `--no-cache` or smth, but Yarn apparently doesn't have that flag ... What do you recommend?

.github/checkInstallTime.js
Skn0tt: That doesn't seem trivial to do. Can you point me towards an implementation route?

.github/workflows/main.yml
ryardley: Can we get away with running this as a separate job?

.github/checkInstallTime.js
ryardley: Yeah that would be an ideal solution. This will be ok for now but will not age well as it will change over time and need revisiting unless we can somehow store previous vals in state somewhere. Perhaps we can us a gh actions addon (perhaps something similar to node_modules caching) to give us a file system to store previous values.

.github/workflows/main.yml
flybayer: I think that would be ideal to run it as a separate job

.github/checkInstallTime.js
Skn0tt: Tbh, that actually seems pretty crazy - but I don't think we're the only ones having the need for such a tool, so maybe there's already something in existance.

.github/workflows/main.yml
Skn0tt: Done in c55ba95

flybayer:
Thank you @Skn0tt!

PR #428, opened 2020-05-04
Missing part of path

harris1717:
part of the path to the questions' index.tsx was missing. (VS code hides a layer when there are no other folders or files)

### Type: Documentation

Closes: None

Filepath in the tutorial was missing a folder


harris1717:
a bug in progress will address this

PR #429, opened 2020-05-04
Fix file path in TUTORIAL

harris1717:
the filepath under /pages is the actual URL path. The path to the questions index page will be /questions/pages/questions/index.tsx (localhost:3000/questions)


flybayer:
Awesome, thank you @harris1717 !

@all-contributors add @harris1717 for docs

Issue #242, opened 2020-04-25
[RFC] Blitz Installers


flybayer:
### Problem

There's hundreds of tools and libraries that different people want to integrate with their app.

Next.js solves this problem by having all these integrations as [example apps](https://github.com/zeit/next.js/tree/canary/examples). This works fairly well, but it's still cumbersome to go find the proper example, copy the code, paste the code, try it, go back and copy something you forgot to copy, and finally get it working.

### Solution

Add the `blitz install` command that can run any available installer.

Examples:

`blitz install tailwind`
`blitz install styled-components`
`blitz install emotion`
`blitz install theme-ui`
`blitz install sentry`
`blitz install storybook`
`blitz install auth0`
`blitz install fauna`

The two most common things performed by an installer are:
1. Install npm dependencies
2. Add or update files

### Implementation

At the most basic level, an installer is a plain Node script. But we'll provide utilities that make it easy to write an installer.

- Installers will be hosted in an official Blitz repo
- When an installer is invoked, the blitz cli will download it from github and run it
- We could also add ability to use private installers from a private repo or private npm package

Installer helpers:

1. `installDependencies` and `installDevDependencies` — will automatically use npm or yarn based on the project
2. File create and update helpers

Perhaps we could have a `@blitzjs/installer` package to import the helpers from. It would always be available in a blitz app for the installer to use.

### Feedback

I'd love to hear all your comments on this. Love it? Hate it? Ideas on making this better?

nainardev:
Why not `npm install`?
How is `blitz install` different from above?

remi2j:
Sounds great!

I wonder how apps that already have custom code will be handled. Sounds like Blitz will need some kind of conflict resolution system.

> Why not `npm install`?
> How is `blitz install` different from it?

npm can only install packages in `node_modules` and save the dependencies in `package.json`.

Having this in the CLI also allows us to also generate the boilerplate in our code.

amine-benselim:
some of the items in the list seem to be of the same category (`styled-components`, `emotion`, or for example `auth0` and `firebase auth` which is not mentioned). so adding two of them does not make sense. 
How about creating a sane categorization for these installers:
`blitz install --css-in-js emotion --auth auth0 --cms prismic`

NateRadebaugh:
Sounds like Gatsby recipes. Does it make sense to standardize on a similar format for this? Eg MDX with custom components for install vs update

cristianbgp:
[...]

PR #434, opened 2020-05-05
Add base infrastructure for `blitz install`

aem:
### Type: Feature

Related: https://github.com/blitz-js/blitz/issues/242

This PR adds the infrastructure and initial implementation for Blitz installers. The installer framework allows us to write a declarative set of steps to install third-party dependencies into a Blitz application.

This version is missing a few key features that we'll need for a public release so it's still hidden from the CLI. The goal is to get the MVP in here, and we can iterate and continue to implement features after this.

 🚨 **Before reviewing, your question may be answered by an unimplemented feature!** 🚨

#### Known unimplemented or not-working features

Ordered from most important to least

1. **Remote installer execution.** Currently I only support local file paths. We should support "native" installers built by us at `github.com/blitz-js/installers` by just a name (e.g. `blitz install tailwindcss`), or by an arbitrary GitHub URL as well (e.g. `https://github.com/aem/my-custom-blitz-installer`). Both of these should download the repo's source as a tarball, extract the proper directory into a temp directory, and then import it as the installer. Next.js examples do this, so we should be able to take inspiration from their implementation.
2. **Proper npm version resolution for `AddDependencyExecutor`.** Currently I just blindly pass the arguments into `yarn`/`npm` `add`. We should use similar logic to what we use in the app generator to resolve the latest valid version and pin it
3. **Transforms that target multiple files.** I put a placeholder field `selectTargetFiles` to allow an installer author to dynamically select multiple files to transform rather than a single target. I don't really know what we want this UX to look like so I left it unimplemented for now.
4. **Maybe a dry-run mode for the transform executor?** This one isn't a must-have for me but might be nice to add.
5. **Write up some consumable docs on `recast`.** We might not need them if our library of default transforms is extensive enough, but we won't really know that until we release this more widely.

### Checklist

- [ ] Tests added for changes


aem:
To test the installer, you can use this custom installer i've already built:

1. `yarn && yarn build`
2. `cd examples`
3. `blitz new installer-test`
4. `cd installer-test`
5. `git clone git@github.com:aem/sample-blitz-tailwind-installer.git`
6. `blitz install ./sample-blitz-tailwind-installer/index.ts`

merelinguist:
Tried this locally and it’s amazing!

wKovacs64:
sick.

aem:
> We could just add the tailwind installer to the installer package for now, just to get it to work with blitz install tailwind.

Definitely. Remote installer support is the first thing on my list to tackle once this gets in

> When do you expect it could be merged?

It's ready now! Once I add the dep to the `blitz` package we should be all set. The command is hidden in the CLI right now so we can merge and publish future releases without exposing this to users until we're ready.

Issue #322, opened 2020-04-28
Enable React Fast Refresh


merelinguist:
### What do you want and why?

We should enable Fast Refresh for apps by default.

### Implementation

1. Add the fastRefresh flag inside the `withBlitz` config wrapper
```
  experimental: {
    reactRefresh: true
  }
```
2. Upgrade Next.js to latest canary (>= `9.3.7-canary.5`)

See here for an explanation: https://github.com/zeit/next.js/pull/12008

Timer:
This feature is not ready to be enabled by default. Please do not enable it for all users until (at a minimum) https://github.com/zeit/next.js/pull/12222 has been landed.

merelinguist:
@timer Oh, okay! I saw a tweet from Dan asking people to try it out so I assumed it was at least alpha-ready. Thanks so much for the heads up 👍

Timer:
https://twitter.com/dan_abramov/status/1255233660317503490

flybayer:
Thanks @timer!

Reopening this because we still want to do it at some point

Timer:
Good to enable now if the underlying Next version is `^9.3.7-canary.5` or newer.

flybayer:
Awesome, thank you @Timer!!

simonpeterdebbarma:
Thank you very much, @dwightwatson !

PR #435, opened 2020-05-05
Enable React Fast Refresh

dwightwatson:
### Type: Feature

Closes: #322

This enables Fast Refresh for Blitz apps by default.

It bumps the minimum supported version of Next.js to `9.3.7` (as-yet unreleased) as it will be the first version to properly support fast refresh (as noted in the attached issue).

This can't be merged until `9.3.7` has properly landed.

### Checklist

- [ ] Tests added for changes

### Breaking change: No

As this is just opting-in to a new development configuration for React I wouldn't expect it to break anything.

### Other information

This is my first Blitz PR so I hope I've gone about it the right way!


dwightwatson:
Failure expected as it depends on the unreleased Next.js - but I assume we can rebuild once it's been tagged.

package.json
```diff
@@ -73,7 +73,7 @@
     "jest": "24.9.0",
     "lerna": "^3.20.2",
     "lint-staged": "^10.0.8",
-    "next": "^9.3.0",
+    "next": "^9.3.7",
```
merelinguist, line 76: Thanks for the PR! Any reason not to just set this to the canary version? `9.3.7-canary.10`

dwightwatson: Can do that - just thought it would be weird to commit a canary version. Would it be better to just do that and get on with it?

flybayer: I'm fine to do this. Also, we need to pin the next.js version. It shouldn't be a range.

dwightwatson:
And we're back to green 🎉

flybayer:
Great, thank you!!

@all-contributors add @dwightwatson for code

PR #441, opened 2020-05-05
Fix Node 10 error for Object.fromEntries

is2ei:
### Type: bug fix

Closes: #392

Replaced `Object.fromEntries` with `reduce` method with callback function.

### Checklist

- [ ] Tests added for changes

### Breaking change: ?? 
no

### DIFF
--- a/packages/generator/src/utils/fetch-latest-version-for.ts
+++ b/packages/generator/src/utils/fetch-latest-version-for.ts
@@ -26,6 +26,8 @@ export const fetchLatestVersionsFor = async <T extends Record<string, string>>(
 
   return {
     isFallback: fallbackUsed,
-    value: Object.fromEntries(updated),
+    value: updated.reduce((result, [key, value]) => {
+      return Object.assign({}, result, {[key]: value})
+    }, {} as T),
   }
 }


packages/generator/src/utils/fetch-latest-version-for.ts
```diff
@@ -26,6 +26,8 @@ export const fetchLatestVersionsFor = async <T extends Record<string, string>>(
 
   return {
     isFallback: fallbackUsed,
-    value: Object.fromEntries(updated),
+    value: updated.reduce((object, [key, value]) => {
```
aem, line 29: can we use a more descriptive variable for the accumulator, something like `acc` or `result`? `object` is just a bit of a loaded term.

packages/generator/src/utils/fetch-latest-version-for.ts
```diff
@@ -26,6 +26,8 @@ export const fetchLatestVersionsFor = async <T extends Record<string, string>>(
...
+      return Object.assign(object, {[key]: value})
```
aem, line 30: can we add `{}` as the first argument to `Object.assign` so we're updating immutably rather than mutating the object every iteration?

is2ei:
@aem 

Thanks for your comments! Updated:
- `object` -> `result`
- Added add `{}` as the first argument to `Object.assign`.

flybayer:
Thank you @is2ei!

@all-contributors add @is2ei for code

Skn0tt:
Is there a way to make sure nobody uses v12-Features? Maybe some setting in `tsconfig.json`?

flybayer:
@Skn0tt 👉 #447

PR #450, opened 2020-05-05
Update react-query to 1.3.3

ivandevp:
### Type: bugfix 

Closes: ??

In previous PR #413, `react-query` dep was set to a fixed version because of inconsistency in their types definitions. Since, there's been a bugfix release on their side, we can now go back to accept minor releases when installing deps.

### Checklist

- [ ] Tests added for changes

### Other information

Builds continue succeeding.


ivandevp:
Oh.. I see. So, I've updated it to a fixed version @flybayer.

flybayer:
Great, thank you!

Issue #440, opened 2020-05-05
[Feature] Run prettier on output of `blitz generate`


flybayer:
### What do you want and why?

All files generated by `blitz generate` should be automatically formatted by Prettier

### Possible implementation(s)

Probably this should be added to `write()` inside `packages/generator/src/generator.ts`

lednhatkhanh:
I would like to take this issue, is it ok?

simonpeterdebbarma:
@lednhatkhanh Yes, assigning this Issue to you. Feel free to create a Pull Request.

PR #453, opened 2020-05-06
Run prettier on output of `blitz generate` commands

lednhatkhanh:
### Type: feature 

Closes: #440 

- Run `prettier` on generated files

### Checklist

- [ ] Tests added for changes


packages/generator/src/generator.ts
```diff
@@ -99,7 +105,16 @@ export abstract class Generator<T extends GeneratorOptions = GeneratorOptions> e
         const templateValues = await this.getTemplateValues()
 
         this.fs.copy(this.sourcePath(filePath), this.destinationPath(pathSuffix), {
-          process: (input) => this.process(input, pathSuffix, templateValues),
+          process: (input) => {
+            let formattedData = this.process(input, pathSuffix, templateValues)
+
+            if (prettierExtensions.test(pathSuffix) && typeof formattedData === 'string') {
+              log.progress(`Formatting file ${pathSuffix}`)
```
aem, line 112: Can we remove this log line for now? I think prettier can be a silent step since most users would expect that we're generating formatted files

packages/generator/src/generator.ts
```diff
@@ -99,7 +105,16 @@ export abstract class Generator<T extends GeneratorOptions = GeneratorOptions> e
         const templateValues = await this.getTemplateValues()
 
         this.fs.copy(this.sourcePath(filePath), this.destinationPath(pathSuffix), {
-          process: (input) => this.process(input, pathSuffix, templateValues),
+          process: (input) => {
+            let formattedData = this.process(input, pathSuffix, templateValues)
+
+            if (prettierExtensions.test(pathSuffix) && typeof formattedData === 'string') {
```
aem, line 111: would you mind moving this logic into `this.process`? That'll avoid splitting our file manipulation into two different spots

lednhatkhanh: Since the `process` function need to be synchronous so I have to get the `prettier options` in the `run` function, is that Ok, do you have any suggestions, thank you.

aem: I don't believe `process` is overridden anywhere, you should be able to make `process` a `private` method and make it `async`, the only reason I made `process` synchronous was because there was nothing async inside, so feel free to change that!

lednhatkhanh: Sorry for my bad explanation, `mem-fs-editor` requires the `process` function to be synchronous so I cannot change it to be async, even if I try to do that then it will not work so I left the logic to get the prettier config in the `write` function.

aem: ohhhh i see. that's fine then!

packages/generator/src/generator.ts
```diff
@@ -91,6 +102,10 @@ export abstract class Generator<T extends GeneratorOptions = GeneratorOptions> e
       const additionalFilesToIgnore = this.filesToIgnore()
       return ![...alwaysIgnoreFiles, ...additionalFilesToIgnore].includes(name)
     })
+    const prettierOptions = (await prettier.resolveConfig(this.sourcePath())) || {
+      semi: false,
+      printWidth: 110,
+    }
```
flybayer, line 108: I think we should not set default config values here. Instead, if config isn't found, then just fallback to the prettier default by not explicitly setting options.

Make sense?

lednhatkhanh: Yes that makes sense, let me change that

lednhatkhanh:
> @aem this will cause generators to crash if Prettier isn't installed. Prettier isn't a requirement for Blitz, so it's possible some users will remove Prettier.
> 
> I think we should change this so that it won't crash if prettier isn't installed. What do you think?

I can try to use dynamic import to see if I works or not.

aem:
> I think we should change this so that it won't crash if prettier isn't installed. What do you think?

That's fine with me, I think however unlikely it is that people remove prettier, we should still support it

flybayer:
@lednhatkhanh great, change to a lazy import with a silent try/catch around it, and then we can merge!

lednhatkhanh:
> @lednhatkhanh great, change to a lazy import with a silent try/catch around it, and then we can merge!

Done, I've tested it locally on my machine and it's working fine without prettier installed

flybayer:
@lednhatkhanh awesome, thank you!!

@all-contributors add @lednhatkhanh for code

Issue #447, opened 2020-05-05
Enable `eslint-plugin-node/no-unsupported-features`


aem:
### What do you want and why?

If we enable [`eslint-plugin-node`](https://github.com/mysticatea/eslint-plugin-node/blob/c833e11ebd50515142cfd4988674a6615833ed03/docs/rules/no-unsupported-features/node-builtins.md) we can have ESLint detect when we're using unsupported Node APIs based on our `engine` definition in `package.json`.

Related: #392 

### Possible implementation(s)

Install the plugin and add to our eslint config

PR #454, opened 2020-05-06
Fix some node v10 compatibility issues

### LINKED ISSUE #458: Find an alternative to REPL#setupHistory
### What do you want and why?

https://github.com/blitz-js/blitz/blob/e12d17b6c14e29d8fd179e44e9b6d9ceceee8240/packages/cli/src/commands/console.ts#L67

[`REPL#setupHistory`](https://nodejs.org/dist/latest-v12.x/docs/api/repl.html#repl_replserver_setuphistory_historypath_callback) is not available in Node v10, see #454 for reference.

We'll need to find a way around using that :D

Maybe there's some way to re-implement or polyfill setupHistory on our side?

### Additional context

A cute duck

Skn0tt:
### Type: tooling

Closes: #447
Closes #458 

#447 recommended to add an eslint-Plugin to prevent usage of unsupported features.
This PR In contrast aims to achieve the same by downgrading available APIs in TypeScript.
This has the added benefit that users won't see unsupported features in their editors etc.

- Downgraded supported ECMAScript version to ES2018
- Dowgraded @types/node to v10

### Checklist

- [ ] Tests added for changes

### Other information

A cute fox


Skn0tt:
The CI failure is due to Node v10 not supporting REPL#setupHistory (see [REPL docs for v10](https://nodejs.org/dist/latest-v10.x/docs/api/repl.html)). @flybayer you're the last editor of that line, could you have a look at this?

https://github.com/blitz-js/blitz/blob/2b1d4897f8779c7f5570debb45269fa1006a61f8/packages/cli/src/commands/console.ts#L67

flybayer:
Well shucks. Probably need to open an issue to make that Node 10 compatible. Use `setupHistory` if it's available, otherwise do what works for v10.

Skn0tt:
Will open an issue, but for this PR we'll just have it skip `setupHistory` on v10.

Skn0tt:
I had a look into `no-unsupported-features` and to me it seems like you've mixed up the purpose of it. Allow me to elaborate 😄
There are basically two distinct feature levels that we need to ensure compatibility with:

**Supported Node.js stdlib Features**
(The version of available Node.js APIs, e.g. [availability of `REPL#setupHistory`](https://github.com/blitz-js/blitz/pull/454#issuecomment-624528581) or [`fs.rmdirSync` recursive mode](https://github.com/blitz-js/blitz/pull/454/commits/d697102ea5506ca6c7b49ef813d3f21629a368d3))
Ensuring compatibility on this level can be achieved by [downgrading `@types/node`](https://github.com/blitz-js/blitz/pull/454/commits/b16911a3ff7c246d6b0b1431ca11fc6f2a2ab915) (as this PR does), but not by using `no-unsupported-features`:
Judging from its docs, `no-unsupported-features`  does not check correct usage of Node APIs but rather the used ECMAScript features (see below)

**Supported ECMAScript Features**
(The version of ES that's supported, e.g. [availability of stuff like `Object.fromEntries`](https://github.com/blitz-js/blitz/pull/441))
This PR ensures this in https://github.com/blitz-js/blitz/pull/454/commits/b093bee76627ccd63c00d8e4627622e784eb18aa.
You propose to ensure this by using [`no-unsupported-features`](https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-unsupported-features/es-syntax.md), which is functionally equivalent but provides a nicer DX in your opinion.

---

Some more thoughts:

> @types/node being pinned to an early version can pretty easily be read as a mistake

It's pinned down to the exact version which should make it obvious that there's some thought behind that.
I don't think any dev should just drive-by upgrade something as important as the Node typings.
Also we do reviews to prevent this kind of stuff from happening.

> However, this is likely confusing for an experience node dev who knows that the API definitely exists.

Well to my understanding, if it's not in the types, it doesn't exist!  Node type definitions are as good as it gets - if something isn't included there, it's not a supported feature on our target. Remember: Our target is v10, not Nightly. Not every feature that "definitely exists" in Node also exists on our target. That assumption is basically what lead to https://github.com/blitz-js/blitz/issues/392 (I was the one who caused that btw 😅)

aem:
@Skn0tt there's a few different bits to the `no-unsupported-features` plugin...you liked to the ES features plugin, but there's also a Node.js API plugin: https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-unsupported-features/node-builtins.md

Skn0tt:
My bad 😅 It wan't linked in [the deprecated `no-unsupported-features` README](https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-unsupported-features.md), so I assumed there wasn't one.

If that's the case, I'm with you - the eslint errors seem like a nicer DX, also they enable conditional usage of new APIs on v12.

Skn0tt:
So I've setup `eslint-plugin-node` (see 
432206f), but it doesn't really catch all errors (see [`file-with-unsupported-features.ts`](https://github.com/blitz-js/blitz/blob/432206f8b137b2e8004ead3b6b522ef6c9187116/packages/blitz/src/file-with-unsupported-features.ts)) - so we go with `@types/node@v10` instead? 😅

flybayer:
I linked #458 to this since this fixes that

aem:
Yeah if it's not working for that edge case then we can do the @types solution for now, but we should open an issue on the ESLint plugin that it doesn't catch that usecase and open an issue here to make sure we follow up so we can have the nicer developer experience eventually. Would you be able to open those?

Skn0tt:
Done @ryardley 😄

packages/cli/package.json
```diff
@@ -10,7 +10,7 @@
     "dev": "rimraf lib && tsc --watch --preserveWatchOutput",
     "build": "rimraf lib && tsc",
     "lint": "tsdx lint",
-    "test": "jest || jest",
+    "test": "tsdx test",
```
flybayer, line 13: Why did you make this change? (tsdx isn't used for the CLI package)

packages/cli/package.json
```diff
@@ -56,7 +56,9 @@
     "@prisma/cli": "2.0.0-beta.3",
     "@rollup/pluginutils": "^3.0.8",
     "@types/pluralize": "^0.0.29",
-    "nock": "13.0.0-beta.3"
+    "@types/rimraf": "^3.0.0",
+    "nock": "13.0.0-beta.3",
+    "rimraf": "^3.0.2"
```
flybayer, line 61: Can you move these to the root package.json? (rimraf is already installed there, btw). Then general rule is to install all dev dependencies in the project root.

packages/cli/src/commands/console.ts
```diff
@@ -65,11 +66,44 @@ export class Console extends Command {
...
+    }
+
+    function onExit() {
+      const addedHistory = repl.lines.join(os.EOL)
+      fs.appendFileSync(path, addedHistory)
+    }
+
+    init()
+    repl.on('exit', onExit)
+  }
```
flybayer, line 89: This looks nice! Are there any known limitations or gotcha's with this vs the built in `setupHistory()` in Node 12+?

tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "target": "es5",
     "module": "esnext",
-    "lib": ["dom", "dom.iterable", "esnext"],
+    "lib": ["dom", "dom.iterable", "ES2018"],
```
flybayer, line 5: Why did we have to make this change?

packages/cli/package.json
Skn0tt: Well, I had a look at all the other packages and saw that all of them use `tsdx test` for testing, only `cli` does not. I changed it to have consistent a output across all packages.

Another reason I was confused is that tests were run twice before - if this was to counter flaky tests, we should do something about those other than rerunning them. Adding more of them down the road will only amplify flakiness ^^

Even if there are tests we cant get stable, we should limit retrying to those.
Maybe [`jest-retry`](https://github.com/bluzi/jest-retries) or [`jest-circus`](https://github.com/facebook/jest/tree/master/packages/jest-circus) could come in handy?

packages/cli/src/commands/console.ts
Skn0tt: Thx 😄 Compared to v12's `setupHistory`, history is saved in reverse order. This will break when a user switches between v10 and v12. If we were to fix this, we'd need to prepend the file instead of append, making this rather costly for large histories.

Other than that, it basically felt like native history - although there must be some differences, as Node v10's internal implementation was quite a big larger.

tsconfig.json
Skn0tt: Node v10 implements ES2018 (see [node.green](https://node.green)), so thats our lib level.
This does not affect language features like optional chaining etc, but "stdlib" availability of features like `Object.fromEntries`, which cannot be transpiled by `tsc`.

flybayer: @Skn0tt ok, I think we only need to make this change in the cli package since it's the only one with a plain `tsc` build. All the other packages are transpiled and bundled with rollup.

packages/cli/package.json
flybayer: Ok we can try rolling with this

tsconfig.json
Skn0tt: Done in 323e3b3

flybayer:
Yay!!

Issue #455, opened 2020-05-06
Editorconfig support


simonedelmann:
### What do you want and why?

It would be great to include `.editorconfig` support out of the box. (More details about .editorconfig files [here](https://editorconfig.org).)

### Possible implementation(s)

.editorconfig
```
root = true

[*]
charset = utf-8
end_of_line = lf
insert_final_newline = true
indent_style = space
indent_size = 2
trim_trailing_whitespace = true

[*.md]
trim_trailing_whitespace = false
```

### Additional context

- It seems like there is support for `.editorconfig` files within prettier. (https://prettier.io/docs/en/api.html) Should be possible to link them!

eliasjohansson:
Thank you for the request! If it isn't already supported I personally agree that it should be.

simonedelmann:
I did a PR for this! It turned out, linking prettier to editorconfig was super easy.

simonedelmann:
Closed, see #460 for discussion.

PR #460, opened 2020-05-06
Add .editorconfig support

simonedelmann:
### Type: Feature

Closes: #455 

- Added an `.editorconfig` file to the app template used by `blitz new` 
- Configured prettier to rely on that editorconfig file

### Breaking change: no

If existing apps want to adopt this, they just need to create the `.editorconfig` file and the following configuration in packages.json:
```
  "prettier": {
    "editorconfig": true,
    "semi": false
  },
```

### Other information

The tailwind example is updated with an editorconfig file.


lednhatkhanh:
IMHO they serve the same purpose: code style, by adding `editorconfig` and make prettier config rely on it, I now have to change the editorconfig for tab size and other stuff while if I only use prettier than I only need to update prettier config, also prettier config is easier to remember than editorconfig's. It's just my personal preference.

flybayer:
Hmm, yeah I was hoping we could link it the other way around, so prettier is the source of truth. 

I think more people will use prettier than editorconfig, and so most people would prefer configuring prettier than editorconfig.

flybayer:
So if we can't link this the other way around, then I think this would be a good candidate for an installer! (#242)

simonedelmann:
I do get the point. To be honest: we don't **really** need .editorconfig support. 
When we edit an existing file the files settings for intendation and tabs will be discovered by the editor. The only situation where an existing .editorconfig makes a difference is for new files, because they are created with the editors default values (which don't necessary match our settings). 

We could:
1. leave everything like it is. If somebody uses .editorconfig it is easy to add that single file.
2. add an .editorconfig file not linked to prettier. Good for people using an editorconfig extension, bad for people who wants to change prettier settings AND using an editorconfig extension.
3. add an .editorconfig file linked to prettier. Good for people using an editorconfig extension, but has some drawbacks (editorconfig is source of truth)

Personally I like option 3 but I can totally understand if you don't agree. ;)

simonedelmann:
> So if we can't link this the other way around, then I think this would be a good candidate for an installer! (#242)

Is this really worth it for an installer? Installer sounds like 'framework' or 'library' to me, not like a config file for editor styling. At least: If I would like to use .editorconfig, I would not have the idea to have a look at the list of installers.

flybayer:
@simonedelmann yeah maybe not, but it totally could be. We'll have thousands of installers because anyone can create them for anything they want.

Because of the complications, I think we should leave this like it is (without .editorconfig). I'm really sorry about the time you spent on this PR without it being merged!

simonedelmann:
I see, thanks for your attention!

Issue #471, opened 2020-05-07
Don't bundle next install it within the app generator


ryardley:
### What do you want and why?

A common usecase for blitz is to enhance an existing Next app and add blitz manually. Out of this I found a couple of things that are preventing this workflow. 

What I want to do:

1. Take an existing next app
2. `yarn add blitz --dev`
3. Have the blitz commands all work.

I noticed a couple of things break if I tried to do this this is one of them

---

Currently we are bundling next as a dependency of the @blitzjs/server package which might be a mistake as it limits the use of blitz when enhancing an existing next app

### Possible implementation(s)

Move next to a peer dependency and add next to the app generator

### Additional context

This would be a breaking change as it will mean existing apps would need to install next separately.

flybayer:
Initially, I really think we should keep bundling Next, but let's discuss it.

Can you explain more in-depth what you mean by "have blitz commands work" in a Next.js app? 

Also, why are you trying to enhance a Next app instead of converting it wholesale to Blitz?

aem:
Bundling Next could have some unintended side effects for this case, so I'm with @ryardley here. If Blitz has a different version of Next pinned than the host app (speaking strictly in the additive case Rudi suggested here) then we could inadvertently either try to use APIs that didn't exist in the host version (not that bad, just weird) or upgrade them to a version where an API they were using has been deprecated + removed (very bad, will be a breaking change to install Blitz). Either way, the user isn't actually using the version of `next` that they have in their `package.json` which would be confusing.

I think the best bet is a peer dependency + a preinit hook in the `blitz` command that checks the installed version of Next to make sure it satisfies our peer dep. Normally I'd say `npm`/`yarn`'s warnings are enough, but since we will absolutely break if the peer dep isn't met I think an extra safeguard is probably a good idea

flybayer:
If we want to support Blitz as an addon to Next, then yes I agree it needs to be peer dependency.

But I'm wondering why would we support Blitz as an addon to Next?

aem:
Isn't that effectively the upgrade path to converting an existing `next` app to a `blitz` app?

flybayer:
No. As of today, the upgrade path is:

1. Remove Next dependency
2. Add Blitz dependency
3. Move `/pages/api/` to `/api/`
4. Rename `next.config.js` to `blitz.config.js`

wKovacs64:
Current upgrade path sounds almost like an installer. `blitz install blitz` 😉

[...]

PR #472, opened 2020-05-07
Don't bundle next install it within the app generator

ryardley:
### Type: refactor

Closes: #471 

* Move next from a `dependency` to a `peerDependency` in `@blitzjs/server`
* Add next to the app generator template

### Checklist

- [x] Tests added for changes (not testable)
- [x] Any added terminal logging uses `packages/server/src/log.ts` (n/a)

### Breaking change: ‼️ yes

Apps that do not have next installed already will need to install next 

### Extra info

When an incompatible version of next is installed blitz attempts to install the correct one: 

[asciicast](https://asciinema.org/a/93ebYAbfdfpolhDr9Ey9H7WNa)


flybayer:
Don't merge this — pending discussion in #471

ryardley:
FYI: Force pushed this branch.

ryardley:
I am going to close this as i looks like we are holding off on this for now

PR #475, opened 2020-05-07
[RFC] Blitz Auth Session Management

flybayer:
Woohoo! Finally we have the RFC for session management.

The purpose of this RFC (Request for Comments) is to gather any feedback from the community.

We welcome all feedback, whether good or bad! This is your chance to ensure Blitz meets the needs of your company or project.

Also we'd like your feedback on the Blitz developer interface. This can, and probably will, be refined and improved.

### 👉 [View the Rendered RFC](https://github.com/blitz-js/blitz/blob/rfc-session/rfc-docs/03-session-management.md) 👈

---

### Todo

- [ ] How to expose the session inside `getServerSideProps`?

### DIFF
--- a/rfc-docs/03-session-management.md
+++ b/rfc-docs/03-session-management.md
@@ -215,10 +215,13 @@ export default async function exampleQuery(args: SomArgs, ctx: Context) {
       // Can use publicData to get the role or other public info of this session.
       let publicData = await Session.getPublicData(session)
 
+      // Can change publicData for this session
+      await Session.setPublicData(session, {...publicData /* ... */})
+
       // get private session data for this session
       let privateData = await Session.getPrivateData(session)
 
-      // Can change privateData for this session only
+      // Can change privateData for this session
       await Session.setPrivateData(session, {...privateData /* ... */})
 
       // Log user out of this specific device
@@ -328,7 +331,19 @@ This is significantly more secure than the essential method, but adds extra comp
 
 ### Blitz Developer Interface
 
-Same as the essential method, but with an extra http endpoint for refreshing tokens
+Same as the essential method, but with a few additions:
+
+- An extra http endpoint for refreshing tokens
+- A callback function for when token theft is detected (ideally where custom error handlers are defined). We will provide a default implementation in which the affected session will be revoked, logging out the victim and the attacker.
+  ```
+  Session.onTokenTheft((sessionHandle: string, userId: string, req: BlitzApiRequest, res: BlitzApiResponse) => {
+    // can revoke this session using sessionHandle
+    // OR
+    // can revoked all sessions belonging to userId
+    // OR
+    // send an email alert to the user etc..
+  })
+  ```
 
 ### Implementation Details
 
--- a/rfc-docs/03-session-management.md
+++ b/rfc-docs/03-session-management.md
@@ -1,0 +1,412 @@
+# [RFC] Blitz Auth Session Management
+
+Authentication has two key components:
+
+1. Identity verification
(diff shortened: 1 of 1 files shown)


dajinchu:
This looks amazing and I can't wait for it to be available. Will the session be available in `getServerSideProps`?

goldo:
Hi! Thanks for the RFC. I was wondering, would passwordless be an option ? 
To me, passwordless is a great solution:
- you don't have to store passwords
- you will avoid some attackers who are looking for DBs with passwords
- you will rely on the security of your email provider, which is way more secure than what you will build, and.. it's free

I mean no database in the world is safe, if you want to to rely on something robust, without dragging attention, I think it's a good solution

raulingg:
Pretty neat solution! 👌

flybayer:
> This looks amazing and I can't wait for it to be available. Will the session be available in `getServerSideProps`?

Great question @dajinchu! Hadn't thought about that. We can definitely expose the session there. Just need to think about the best way to do that.

flybayer:
> Hi! Thanks for the RFC. I was wondering, would passwordless be an option ?
> To me, passwordless is a great solution:

@goldo yes, for sure! This session management will work with any identity verification method/service, including passwordless.

merelinguist:
This all looks really great! I love the session hook. What’s the timeline for the authorisation RFC?

flybayer:
@merelinguist here's the latest word from @rishabhpoddar on implementation timeline, which starts now:

> Will take a couple of weeks. First to architect the solution keeping in mind future changes we have planned, and then about a week to build and test.

This will be for the essential security level without anonymous sessions. The rest will follow.

dwightwatson:
Something I'd be curious to know more about (that I'm not sure is covered here) is whether Blitz auth solution will handle things on the database end as well - as in, provide migrations and boilerplate to register users, login users etc. - or will it just provide the session token abstraction and leave the implementation to the developer?

flybayer:
@dwightwatson user registration/login/password reset/etc will be provided as a Blitz recipe/installer (#242). It'll scaffold out all the code into your project. So you get the benefit of not writing it from scratch, **and** you get the benefit of having total ownership over it. So you aren't locked into a certain library or anything. 

As a reference for Rails people, this will be the "devise" for Blitz apps.

dwightwatson:
Fantastic news, appreciate the clarification on that point.

Romstar:
@flybayer @rishabhpoddar 

I am planning to utilize a middleware for the session given that Redis is always significantly more performant than PostgreSQL.  Can we make adding a middleware for the session management intuitive?

flybayer:
@Romstar yes for sure! I think we'll have a config similar to this so you have full control over where and how sessions are stored:

```js
const config = {
  getSession: sessionHandle => /* access persistent storage and return */,
  getAllSessionsForUser: userId => /* access persistent storage and return */,
  createSession: session => /* access persistent storage and return */,
  updateSession: session => /* access persistent storage and return */,
  deleteSession: session => /* access persistent storage and return */,
}
```

What do you think about that?

matepapp:
Hey! 👋 

Have you thought about combining forces and integrating [`next-auth`](https://github.com/iaincollins/next-auth) into Blitz? It could be an opt-in solution, like Prisma, when we're talking about database management. I know it's pretty new, but I think it worths checking out, and it would fit into the project pretty well.

flybayer:
@matepapp thanks for the tip! I did not know about that. Cool project!

From what I can tell our custom session management library will be even more full featured and have a lot better integration with Blitz than next-auth could.

Issue #446, opened 2020-05-05
Plain JS blitz app fails to compile


flybayer:
### What is the problem?

Plain JS blitz app is unusable — fails to compile

### Steps to Reproduce: 

1. `blitz new myapp --js`
2. `cd myapp`
3. Follow the steps in the default index.js page for migrating your DB **and** and running `blitz generate all projects`
4. `blitz start`
5. Try to access localhost:3000/projects, but it fails with:

```
./app/projects/queries/getProjects.js
SyntaxError: /private/tmp/8c/.blitz/caches/dev/app/projects/queries/getProjects.js: Unexpected token, expected ";" (4:80)

  2 | import {isomorphicRpc} from '@blitzjs/core'
  3 | import resolver from 'app/_rpc/projects/queries/getProjects'
> 4 | export default isomorphicRpc(resolver, 'app/_rpc/projects/queries/getProjects') as typeof resolver
    |                                                                                 ^
```

### Versions:

```
0.9.0
```

### How to fix

This template is defined with the `isomorphicRpcTemplate` variable in `packages/server/src/synchronizer/pipeline/rules/rpc/index.ts`.

I see two ways to approach this:

1. Detect if it is JS/TS project, then modify the template accordingly
2. Or, leave template as is but add a pipeline rule that runs `tsc` to compile to JS, if it is a JS only project.

Thoughts on how to fix this @ryardley?

aem:
is it possible that the synchronizer needs to learn about our `jsconfig.js`? we added it to the template but is the synchronizer not copying it into the `.blitz` directory?

flybayer:
@aem, no that is working fine. The issue is our rpc template is hardcoded as Typescript.

aem:
ohhh, i see. missed that before.

would distributing the server files as JS be reasonable here? they only actually need to be in TS for development, but when publishing we could ship plain JS + type definitions, right?

tmns:
I also noticed the following error today with a freshly generated JS project when attempting to access the `db` object from `blitz console`:
```javascript
$ blitz console
[..snip..]
◢ Loading your codeCouldn't find tsconfig.json. tsconfig-paths will be skipped
✔ Loading your code
⚡️ > db.question.findMany().then(console.log)
Thrown:
ReferenceError: db is not defined
```
Note the first error about `tsconfig.json` - maybe these issues are related?

flybayer:
[...]

PR #476, opened 2020-05-08
Enable JavaScript projects

Jfelix61:
### Type: bug fix

Closes: #446 

Add a rule to convert Typescript to Javascript whenever a project lacks a tsconfig.json. This will remove errors related to parsing of TS on JS Blitz projects.

### Checklist

- [x] Tests added for changes


packages/server/src/synchronizer/pipeline/rules/typescript/index.ts
```diff
@@ -0,0 +1,56 @@
...
+ * Returns a Rule that compiles TS files to JS.
+ */
+export const createRuleTypescript: Rule = ({config, reporter}) => {
+  const tsConfigPath = path.resolve(config.src, 'tsconfig.json')
+  const isTsProject = pathExistsSync(tsConfigPath)
+
+  const stream = through({objectMode: true}, function (file: File, _, next) {
+    if (isTsProject) next(null, file)
+
+    const tsconfig = readConfigFile(tsConfigPath)
```
ryardley, line 21: Perhaps we don't need to do the read here? FS operations are slow.

packages/server/src/synchronizer/pipeline/rules/typescript/index.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    next(null, newFile)
+  })
+
+  return {stream}
+}
+
+const readConfigFile = (configFileName: string) => {
+  // Read config file
+  const configFileText = fs.readFileSync(configFileName).toString()
```
ryardley, line 41: I would avoid a file read here as it will block the event loop.

packages/server/src/synchronizer/pipeline/rules/typescript/index.ts
```diff
@@ -0,0 +1,56 @@
...
+  const result = ts.parseConfigFileTextToJson(configFileName, configFileText)
+  const configObject = result.config
+  if (!configObject) {
+    process.exit(1)
+  }
+
+  // Extract config information
+  const configParseResult = ts.parseJsonConfigFileContent(configObject, ts.sys, path.dirname(configFileName))
+  if (configParseResult.errors.length > 0) {
+    process.exit(1)
```
ryardley, line 53: We might want to clean up here. Failing to read the config file probably should not have the right to control the node process.

packages/server/src/synchronizer/pipeline/rules/typescript/index.ts
ryardley: Also if this is a TS project this will not exist.

ryardley:
Sorry I just realised you probably weren't ready for review.

ryardley:
So i think I found a bug here. 

When the rpc rule creates a file it does so by cloning an existing resolver file and changing the contents which means we can end up with typescript within a js file. 

https://github.com/blitz-js/blitz/blob/canary/packages/server/src/synchronizer/pipeline/rules/rpc/index.ts#L45

We probably want to select a ts or javascript template based on the file type it is presented with in the rpc rule.

Jfelix61:
> We can end up with typescript within a js file.

Currently the rule compiles just `.js` files with Typescript within them, since the user should not have ts files inside their project without a `tsconfig`. I wrote a test on thi, can you check please? Otherwise, I can add `.ts` files also.

ryardley:
@Jfelix61 You are correct! Apologies I forgot to build.

yarn.lock
```diff
@@ -3152,16 +3152,6 @@
     regexpp "^3.0.0"
     tsutils "^3.17.1"
 
-"@typescript-eslint/eslint-plugin@2.31.0":
```
ryardley, line 3155: Just wondering why we have any changes here as I cannot see corresponding package.json changes anywhere? Was yarn.lock not updated by a prior PR?

Jfelix61: :thinking: I don't know. I'll merge changes from canary.

PR #479, opened 2020-05-08
Pin react experimental version & fix graceful fallback for prisma version

flybayer:

1. Pin the react experimental version to one that works (latest is broken)
2. Change prisma version in package.json template to be a valid version `2.0.0-beta.4` instead of an invalid version `2.x` so that `npm i` will work if the dependency updater fails for some reason during `blitz new`

### Checklist

- [ ] Tests added for changes

### DIFF
--- a/packages/generator/src/utils/fetch-latest-version-for.ts
+++ b/packages/generator/src/utils/fetch-latest-version-for.ts
@@ -10,7 +10,8 @@ export const fetchLatestVersionsFor = async <T extends Record<string, string>>(
 
   const updated = await Promise.all(
     entries.map(async ([dep, version]) => {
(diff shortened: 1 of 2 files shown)


osirvent:
Latest prisma version is 2.0.0-beta.4. On `package.json` appears `2.0.0-beta.3`

flybayer:
@osirvent yeah this will get out of date and can be updated periodically in our code, but it's automatically updated during `blitz new`.

osirvent:
@flybayer I understand for what your are saying that now in a `blitz new` app should appear `2.0.0-beta.4` but it is not. Shouldn't be using something like `2.0.0-beta.X`

Issue #404, opened 2020-05-03
Bug: Dependent Queries do not work with useQuery() 


MrLeebo:
### What do you want and why?

I want to be able to implement a [Dependent Query](https://github.com/tannerlinsley/react-query#dependent-queries) using blitz useQuery(), but the query key is internally managed

I want this because sometimes my params are state-dependent

```ts
const [profile, setProfile] = useState(null)

// I don't want this to run until a profile is selected
useQuery(getHistory, { userId: profile.id })
```

### Possible implementation(s)

Use a query key function and allow `params` to be a function, perhaps?

```ts
const [profile, setProfile] = useState(null)

// if profile is null, the query key function will throw, canceling the query
useQuery(getHistory, () => ({ userId: profile.id }))
```

### Additional context

https://github.com/tannerlinsley/react-query#dependent-queries

Skn0tt:
> If a query isn’t ready to be requested yet, just pass a falsy value as the query key or as an item in the query key:
> ~ React Query Docs

Judging from this and [our sources](https://github.com/blitz-js/blitz/blob/canary/packages/core/src/use-query.ts#L36), dependent queries should be possible by passing a falsy value for params.
I don’t feel like this is convenient though, since people may want to pass null etc. for params for legitimate reasons.

Having said that, I quite like your suggestion with the throwing param suppliers!

bobzoller:
Dependent Queries are called out in blitz-js/blitz#89 (note the lack of a _not implemented_ label?), but I came here looking because they don't seem to work --- which is to say if I pass a function that throws or a falsy value for params, the RPC still fires.

blitz@0.6.3

Skn0tt:
Good catch! That's definitely a bug then.
So if someone wants to provide a PR for this, this is where you'd have to look:
https://github.com/blitz-js/blitz/blob/2c9e8e131889e6e323dd94f408b31295dd65fcca/packages/core/src/use-query.ts#L41

flybayer:
## Update

Yes, this is indeed a bug. It should work exactly like @MrLeebo expected with:

```
const [profile, setProfile] = useState(null)

// if profile is null, the query key function will throw, canceling the query
useQuery(getHistory, () => ({ userId: profile.id }))
```

PR #487, opened 2020-05-10
Add support for Dependent Queries to useQuery and usePaginatedQuery

MrLeebo:
### Type: ?? 

Closes: #404 

Support dependent params in useQuery

### Checklist

- [x] Tests added for changes
- [x] Any added terminal logging uses `packages/server/src/log.ts` (n/a)


packages/core/src/use-query.ts
```diff
@@ -35,10 +35,15 @@ type RestReturnType<T extends QueryFn, O extends Options<T>> = O['paginated'] ex
...
   queryFn: T,
-  params?: InferUnaryParam<T>,
+  params?: InferUnaryParam<T> | (() => InferUnaryParam<T>),
   options?: O,
 ): [PromiseReturnType<T>, RestReturnType<T, O>] {
-  const queryKey: [string, any] = [(queryFn as any).cacheKey, params]
+  const queryKey: () => [string, any] = () => {
+    return [
+      (queryFn as any).cacheKey,
+      (typeof params === 'function' ? (params as () => InferUnaryParam<T>)() : params) as InferUnaryParam<T>,
```
flybayer, line 44: This is all you need to make TS happy for this line:

```suggestion
      typeof params === 'function' ? (params as Function)() : params
```

flybayer:
@MrLeebo I just saw you requested a review from me on this, but it's still in draft mode.

Is this finished or still wip?

MrLeebo:
@flybayer I merged latest from master and applied the same changes to usePaginatedQuery also. I noticed that the `initialData` prop also seems to infer its type from the query function (e.g. expects the initial data to be an async function rather than just the function result) but I don't think that was introduced by this change.

MrLeebo:
I think the github 500 errors are preventing the new commit from showing up.

flybayer:
@MrLeebo ok cool. I'm still not seeing your commit. Maybe try pushing again?

MrLeebo:
Looks like they are there now.

flybayer:
Thank you @MrLeebo!!!

Issue #463, opened 2020-05-07
Support installation of remote Blitz installers


aem:
### What do you want and why?

Currently we don't have a way to run installers that aren't local files. We should have a seamless way to execute installers that are hosted on Github without having to install them as npm dependencies; an installer should be as easy as `blitz install tailwind`. We should also support an installer URL that points to a github repo which holds an installer, e.g. `blitz install https://github.com/aem/fake-blitz-installer`. That implementation should be fairly similar to the native installers.

### Possible implementation(s)

The process should be fairly straightforward:

1. Download the repo's latest default branch as a tarball into a temp directory and extract
2. Install any `npm` dependencies in the `package.json`
3. Get path to main installer file
4. Execute using the same codepath as the local installer

### Additional context

If a user supplies a bad installer name, we should list all of the installers in an `enquirer` prompt and allow the user to choose an existing one.

PR #491, opened 2020-05-11
Support remote packages in the installer

aem:
### Type: feature

Closes: https://github.com/blitz-js/blitz/issues/463

This adds support in the `install` CLI command for GitHub-hosted installers. We'll now support the following installer arguments:

1. Local file path (`blitz install ../my-installer/index.ts`)
2. Name of Blitz native installer (`blitz install tailwind`)
3. Full GitHub repo name (`blitz install aem/sample-blitz-tailwind-installer`)
4. Full URL of GitHub repo (`blitz install https://github.com/aem/sample-blitz-tailwind-installer`)

For all three of the remote options, the install process is relatively straightforward:

1. Ensure project exists (check for a `package.json` in the repo root)
2. Fetch default branch for repo (e.g. GH defaults to `master` but we use `canary`)
3. Download tarball of latest code on default branch into temp directory, extract tarball
4. Install installer's dependencies
4. Execute installer
5. Clean up temp directory

The current system has three limitations:

1. There's currently no native Blitz installers to install :) 
2. Since there's no native installers to execute, I haven't built the prompt that lists all installers if the one you listed doesn't exist :) 
3. We don't support nested installer paths in third-party installers, we only support installers hosted at the repo root. This wouldn't be _that_ hard to add, but I think the likelihood that we have a third-party monorepo of installers is unlikely

All-in-all, it's relatively crude at the moment, but it's absolutely robust enough for an alpha and will be good enough to start building our own installers and getting this into peoples' hands. 

### Checklist

- [ ] Tests added for changes


aem:
Quick note about testing these changes - I had to shift some `installer` dependencies from `devDependencies` to `dependencies` so the actual running of the installer won't work right now without modifying the `install` command since the published version of the `installer` package doesn't have the right dependency setup. If you want this to work locally, first modify `packages/cli/src/commands/install.ts` line 111 to look like this:

```js
const installerDir = path.resolve(`blitz-installer-${repoFullName.replace('/', '-')}`)
```

That will put the temp directory under the current directory, so when you run the command in the monorepo it'll use the workspace version of the packages rather than the published versions. Then you can run `blitz install aem/sample-blitz-tailwind-installer` inside the monorepo, say on `examples/store`

flybayer:
Since the installer itself has `@blitzjs/installer` as a dependency, can we remove `@blitzjs/installer` as a dependency of `blitz` ?

aem:
@flybayer yeah we can probably remove it. We need it as a `devDependency` since we use it for typechecking and in tests, but you're correct that it's not used in the main CLI. I'll update this to reflect that

Issue #232, opened 2020-04-25
Add context path to blitz generate


flybayer:
**What do you want and why do you want it?.**

Currently blitz generate command generates all files at the top level. So `blitz generate all tasks` puts the tasks files at `app/tasks/*`.

We need to add a context/namespace support that works like this:

```
blitz generate all admin/tasks
```
generates all the normal tasks files but at `app/admin/tasks/*`

General format is:

```
blitz generate TYPE [CONTEXT_PATH/]MODEL
```

- The model name is always the last segment of the input path.
- The `/app/` path is always implicit. 
- The context path can be many layers deep like `/clients/admin/internal/`

aem:
For queries and mutations this is pretty self-explanatory. For pages, will the context path also create a path prefix or just a top-level page with a logical organization in the file system? As a concrete example, using the `admin/tasks` example above, which of the following would we generate:

1. `app/admin/tasks/pages/admin/index.tsx`
2. `app/admin/tasks/pages/admin/tasks/index.tsx`
3. `app/admin/tasks/pages/tasks/index.tsx`
4. `app/admin/pages/admin/tasks/index.tsx`
5. `app/admin/pages/tasks/index.tsx`

I could make arguments for any of them based on how our file structure works

flybayer:
The context path will not go inside `pages/`, so the result would be (3)

`--parent` will be used for page nesting

We might need to add another flag later for arbitrary page nesting

PR #493, opened 2020-05-11
🚀 Add context path option to `blitz generate`

aem:
### Type: feature

Closes: https://github.com/blitz-js/blitz/issues/232

Adds a new CLI flag to support contextual file generation. Both Windows and UNIX paths are accepted, and you can pass the context in the model name rather than using the CLI flag if you'd like. That means all of the following are valid commands on any operating system:

* `blitz generate queries admin/tasks/todos`
* `blitz generate queries admin\tasks\todos`
* `blitz generate queries todos --context=admin/tasks`
* `blitz generate queries todos --context=admin\tasks`

All 4 of those output the exact same files - `app/admin/tasks/todos/queries/getTodo(s).ts`

### Checklist

- [x] Tests added for changes

### DIFF
--- a/packages/cli/src/commands/generate.ts
+++ b/packages/cli/src/commands/generate.ts
@@ -93,11 +93,11 @@ export class Generate extends Command {
 
   static flags = {
     help: flags.help({char: 'h'}),
-    // context: flags.string({
-    //   char: 'c',
-    //   description:
-    //     'The parent folder for nested generation. For example, generating `products` within a `store` would supply `-c store`. For nested contexts you may supply the full path.',
-    // }),
+    context: flags.string({
+      char: 'c',
+      description:
+        "Provide a context folder within which we'll place the generated files for better code organization. You can also supply this in the name of the model to be generated (e.g. `blitz generate query admin/projects`). Combining the `--context` flags and supplying context via the model name in the same command is not supported.",
+    }),
     'dry-run': flags.boolean({
       char: 'd',
       description: 'Show what files will be created without writing them to disk',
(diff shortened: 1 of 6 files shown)


flybayer:
Looks great! But fails for `admin\tasks`. I committed a failing tests to this branch.

```
~/c/b/e/store> blitz g all admin\pizza -d

Would create    app/adminpizzas/pages/adminpizzas/[id]/edit.tsx
Would create    app/adminpizzas/pages/adminpizzas/[id].tsx
Would create    app/adminpizzas/pages/adminpizzas/index.tsx
Would create    app/adminpizzas/pages/adminpizzas/new.tsx
Would create    app/adminpizzas/queries/getAdminpizza.ts
Would create    app/adminpizzas/queries/getAdminpizzas.ts
Would create    app/adminpizzas/mutations/createAdminpizza.ts
Would create    app/adminpizzas/mutations/deleteAdminpizza.ts
Would create    app/adminpizzas/mutations/updateAdminpizza.ts
```

aem:
@flybayer it seems like the test you committed will never pass on unix-based systems. any POSIX-compliant shell will escape the \t before it even gets into the node context. the last test in the suite that has the escaped `\` characters is correctly testing the windows behavior since the string we're operating on has the slashes in it, but the test with a single slash is testing how the script handles a context path of `admin    asks`. I think it makes sense to revert that test and merge as is

aem:
here's a contrived example in node of how JS handles this - 
```
> node
Welcome to Node.js v12.16.3.
Type ".help" for more information.
> let a = '\t'
undefined
> for (let i = 0; i < a.length; i++) {
... console.log(a[i]);
... }

undefined
> 
```

flybayer:
Ok that makes sense! I was just going off what you said in the first post:

>  That means all of the following are valid commands on any operating system:

aem:
@flybayer yeah my wording was incorrect there, backslashes will only work on windows. having said that, i don’t expect that to be an issue, i assume anyone on mac/linux is using forward slashes

Issue #233, opened 2020-04-25
Add --parent to blitz generate pages


flybayer:
**What do you want and why do you want it?.**

The Blitz routing convention often results in the URL having a parent model and a child model. For example, if your DB has Project and Task and Project has many Tasks, your page structure should be:

So this:

```
blitz generate pages task --parent project
```

should generate this:

```
app/tasks/pages/projects/[projectId]/tasks/index.ts
app/tasks/pages/projects/[projectId]/tasks/new.ts
app/tasks/pages/projects/[projectId]/tasks/[id].ts
app/tasks/pages/projects/[projectId]/tasks/[id]/edit.ts
```

Also, `--parent` argument must also work with the other commands that invoke the pages generator.

So this:
```
blitz generate all task --parent project
```
should generate this:
```
app/tasks/queries/getTask.ts
app/tasks/queries/getTasks.ts
app/tasks/mutations/createTask.ts
app/tasks/mutations/updateTask.ts
app/tasks/mutations/deleteTask.ts
app/tasks/components/TaskForm.ts
app/tasks/pages/projects/[projectId]/tasks/index.ts
app/tasks/pages/projects/[projectId]/tasks/new.ts
app/tasks/pages/projects/[projectId]/tasks/[id].ts
app/tasks/pages/projects/[projectId]/tasks/[id]/edit.ts
```

PR #494, opened 2020-05-11
Add `--parent` to `blitz generate` + Lots of code template updates + change route id to modelId

aem:
### Type: feature
Closes: https://github.com/blitz-js/blitz/issues/233

Adds a CLI flag for the user to specify a `--parent` while generating pages. These pages will be nested under `{{parentModelName}}/{{parentModelId}}`, enabling the user to generate pages for dependent data models:

```
$ blitz g pages tasks --parent=projects

CREATE    app/tasks/pages/projects/[projectId]/tasks/[id]/edit.tsx
CREATE    app/tasks/pages/projects/[projectId]/tasks/[id].tsx
CREATE    app/tasks/pages/projects/[projectId]/tasks/index.tsx
CREATE    app/tasks/pages/projects/[projectId]/tasks/new.tsx
```

### Checklist

- [ ] Tests added for changes


flybayer:
@aem Looking good!

The generated "children" pages don't work because they aren't taking into account that they are children. They need to do this for the query argument to only get the child models based on the parent model id. And the links need to also use the parentId. So we need some more sophisticated templating here.

You can add that here, or we can hide the --parent flag and merge this. Then open another issue for doing the above.

aem:
@flybayer sorry, those are all outstanding issues I'm aware of, I meant to open this as a draft since it builds on the context generation in #493, I just wanted to get the path generation stuff working first and then circle back to this after the other PR gets merged.

flybayer:
This would be really great to have by end of the weekend if possible. So that it can be released before my live stream with Kitze on Wednesday.

aem:
got slowed down a little bit by having to update the generator to be able to do conditional templates (we need to generate different code based on whether or not there's a parent, and having different templates for those isn't ideal since you'll have to update two templates to change the pages templates, similar problem to having both JS and TS versions of templates). made some good progress to day and should be able to wrap things up tomorrow

aem:
Got this all working locally! Things that work now:

* All templates support conditional generation. I'm about to push a readme update that explains how this works
* The Page templates have conditional code paths for parent/non-parent generation
* The queries + mutations generated in a page with a parent model automatically filter by the parent model ID, and creates will associate the new objects with the parent model by default

Tested also generating the same model without a parent and confirming that we get all of the objects returned _without_ the parent filtering.

To test this:

```
// in schema.prisma
model Todo {
  id          Int      @default(autoincrement()) @id
  name        String
  description String   @default("")
  completed   Boolean  @default(false)
  TodoList     TodoList? @relation(fields: [todoListId], references: [id])
  todoListId   Int?
}

model TodoList {
  id    Int    @default(autoincrement()) @id
  name  String
  todos Todo[]
}

$ blitz g all todolist
$ blitz g pages todos --parent=todolist
```

You'll want to seed some data via prisma studio so that you can observe the filtering in action.

aem:
Also, you may have noticed that we have a pretty wild leg up on every other web framework now......our "todo list" example (a rite of passage, no?) is modifying the schema to add the models, running two terminal commands, and adding a form input for the todo title, and that's all you need for a fully persistent todo app in Blitz. I think we win for fewest user-involved steps :)

flybayer:
This is looking good! I'll test more tomorrow

packages/generator/templates/page/__id__.tsx
```diff
@@ -6,23 +6,46 @@ import delete__ModelName__ from 'app/__modelNames__/mutations/delete__ModelName_
...
+
   return (
     <div>
       <h1>__ModelName__ {__modelName__.id}</h1>
       <pre>{JSON.stringify(__modelName__)}</pre>
 
-      <Link href="/__modelNames__/[id]/edit" as={`/__modelNames__/${__modelName__.id}/edit`}>
-        <a>Edit</a>
-      </Link>
+      {editLink}
```
flybayer, line 34: Don't have this before merging this, but it'd be really nice if we can somehow support having conditional components here without the separate `editLink` component. Because this results in a weird output in the generated file (a separate `editLink` component for no reason)

aem: yeah, I didn't love this, the problem is it's a limitation of the JSX syntax (can't put JS in JSX without `{}`). We could eventually build some handling for jsx conditionals with something like this:

```xml
<if condition="someTemplateValue">
 <Link />
  <else>
    <OtherLink />
  </else>
</if>
```

but that's a pretty significant change that we should split into another issue

aem:
@flybayer now **all** generated pages will put the model name in the path param rather than just parent pages:

```
CREATE app/tasks/pages/tasks/[taskId]/edit.tsx
CREATE app/tasks/pages/tasks/[taskId].tsx
CREATE app/tasks/pages/tasks/index.tsx
CREATE app/tasks/pages/tasks/new.tsx
```

```
CREATE app/tasks/pages/projects/[projectId]/tasks/[taskId]/edit.tsx
CREATE app/tasks/pages/projects/[projectId]/tasks/[taskId].tsx
CREATE app/tasks/pages/projects/[projectId]/tasks/index.tsx
CREATE app/tasks/pages/projects/[projectId]/tasks/new.tsx
```

aem:
@flybayer it seems like an odd UX thing to me that I’d have to re-generate mutations in addition to generating pages here, what’s the reasoning there?

aem:
also, if there's gonna be a bunch of changes you want to make on PRs like this I wonder if it's worth opening them up as separate issues  so we can keep diffs small? the git history here is gonna be huge

flybayer:
@aem yeah 😬 I considered doing some stuff in another PR, but it's mostly all affected by adding the parent feature so I decided to just do it here. Probably should have separated some stuff though.

On changes to mutations with --parent: you're right it's not a perfect UX from a code scaffolding standpoint, but this is absolutely the right thing to do from a code perspective. 

Mutations are where business logic is ultimately enforced, so they will have to change when adding a parent to a model.

flybayer:
@aem I'm done here and am ready to merge. Thank you so much!!

aem:
> On changes to mutations with --parent: you're right it's not a perfect UX from a code scaffolding standpoint, but this is absolutely the right thing to do from a code perspective.

Super fair. I wonder if the "right" thing here is to create a new generator rather than a CLI flag to enforce that you generate new pages, queries, and mutations (and soon models)? Just log an error and exit early if you use `--parent` with anything other than `blitz g all`? Originally I had it just set for `pages` since I didn't think about adding the validating in the queries/mutations themselves, but now I think the opposite direction makes sense, just require them to generate the whole thing. We may come across some use cases where people only want to generate some of these, but for now it might make sense to try and force people into the easiest path?

flybayer:
@aem let's leave it like it is for now, because just generating queries/mutations with `--parent` is a very valid use case too.

But we can definitely change it at some point. I just don't think we know enough yet what something better would be (if any)

aem:
sounds good

aem:
i'm gonna pull this down, update the docs since they just mention pages right now, rebase against canary, and then this should be all set

aem:
@flybayer need another ✅so I can merge and we can get this in 🎉

Issue #401, opened 2020-05-02
`blitz db reset` failing with `cannot find module`


hauptrolle:
### What is the problem?

When running `blitz db reset` I get the following error: `Error: Cannot find module '/Users/user.name/workspace/private/blitz-project/db'`

### Steps to Reproduce:

1. Setup blitz project
2. Run `blitz db reset`

### Versions:

```
macOS Catalina darwin-x64 node-v12.14.1

blitz: 0.6.1 (global)
blitz: 0.8.1 (local)
```

Skn0tt:
Thanks for taking the time to report, it's very much appreciated ☺️

Skn0tt:
Thanks for taking the time to report, it's very much appreciated ☺️

johncantrell97:
Also seeing this, I'll take a stab at fixing it.

Skn0tt:
Awesome! Let me us know if there's something we can help you with.

PR #497, opened 2020-05-11
Fix 'blitz db reset' command

wKovacs64:
### Type: bug fix 

Closes: #401

- Updated database path regex to support Windows
- Instantiated own DB client rather than trying to require from `<projectRoot>/db/index.ts` (code is JS at runtime, can't require TS)
- Uncommented DB migration instruction for PostgreSQL reset

### Checklist

- [ ] Tests added for changes

### Breaking change: no 

N/A

### Other information

Before: `Error: Cannot find module 'C:/dev/test/myapp/db'`
After: `✔ Your database has been reset.`

Confirmed locally with PostgreSQL and SQLite on Windows 10 and macOS Catalina.

/cc @johncantrell97 just saw you were working on this - not sure if you had something in the works, but I can close this if so!


flybayer:
Yay!!

PR #508, opened 2020-05-14
Lazy load typescript in `server` to fix error when running outside TS app

flybayer:

`blitz new myapp` currently fails because it can't load the typescript module (not installed in the app yet).

`blitz start` in plain JS project would also fail in the same way.

This lazy loads TS only when it's actually needed.


flybayer:
Well this doesn't work. It doesn't fix plain JS apps because TS will be needed there to compile the TS templates we include. 

Will have to solve another way.

Issue #485, opened 2020-05-09
Database connections are not being closed: FATAL: sorry, too many clients already


flybayer:
### What is the problem?

Database connections are not being closed when the server is stopped.

```
Error querying the database: db error: FATAL: sorry, too many clients already
```

### Steps to Reproduce (maybe)

1. `blitz start`
2. Visit page that reads from the database
3. Make changes to your app source files and save to trigger hot reloading
4. Repeat many times

### Versions:

```
0.9.0
```

### How to Fix

I think the blitz server needs to add a listener for a `kill` signal. Then run `db.disconnect()` inside that. Note: this must be implemented in a way that it will not fail if the `db` module doesn't exist.

This should be for both `blitz start` and `blitz start --production`

abuuzayr:
thanks @flybayer for opening this issue! 

the steps should be: 

1. `blitz start`
1. Visit page that reads from the database
1. Make changes to your app source files and save to trigger hot reloading
1. Repeat many times

I hope the above makes sense. 

Also, as per your suggestion I added the `?connection_limit=1` to my DB URL in `.env`, but I still get this issue.

ivandevp:
May I help with this issue?

abuuzayr:
of course @ivandevp please feel free!

abuuzayr:
@ivandevp this might help: https://github.com/prisma/prisma/issues/1983

flybayer:
@abuuzayr nice!

Looks like the solution is this:
```ts
import { PrismaClient } from "@prisma/client"

let prisma

if (process.env.NODE_ENV === "production") {
  prisma = new PrismaClient()
} else {
  // Ensure the prisma instance is re-used during hot-reloading
  global.prisma = global.prisma || new PrismaClient()
  prisma = global.prisma
}

export default prisma
```

ivandevp:
Hey guys! Sorry for the delay. I've been trying to reproduce but haven't found a way to do it. I followed the steps but can't see the error in the current canary version, when I create a project with Blitz 0.9.0 I got an error related to React.Suspense but not sure what is happening. Are you still facing the error in the current canary version? The problem and solution makes sense to me, but can't see it reflected in my dev environment.

sbardian:
I am also unable to reproduce.  Tried 0.90 and 0.91.

abuuzayr:
@flybayer yup so far so good! 

[...]

PR #514, opened 2020-05-14
Ensure same prisma client is being used on hot reload

abuuzayr:
Closes: #485

While on development, modules are constantly hot reloaded and this may cause the spawning of multiple connections for a single application. This ensures that the same client is being used while on development. 

In production the behavior is unchanged.

### Checklist

- [ ] Tests added for changes

### Other information

I have some questions about this contribution:

1. This is my first time contributing to a project like this. I'd like to test the `blitz new app` command to ensure that my changes are working, but I'm not sure how to replace or link the global binary to my local repo. How do I go about doing that?

Otherwise, the code is tested and working fine in my other project folders. 

2. Should I also update the `db/index.ts` files in the example applications?


abuuzayr:
@flybayer The steps were right! I somehow jumped the gun i suppose. I was getting "MODULE NOT FOUND" errors, that `../dist/cli` was missing, but I realised the `yarn dev` command wasn't completed till i saw:

```
blitz: created dist/cli.js in 1.1s
```

flybayer:
@abuuzayr ah yes, yes, I get that sometimes too. There's probably a way we could make that experience nicer, like show a message during dev like "make sure "yarn build" or "yarn dev" has completed compiling"

abuuzayr:
will this work? 

in **packages/blitz/bin/blitz**:

```javascript
#!/usr/bin/env node

try {
  require('../dist/cli')
} catch (e) {
  console.error(`blitz is unavailable yet, make sure "yarn build" or "yarn dev" has completed compiling`)
}
```

flybayer:
@abuuzayr were you going to update the database files for the examples apps? :)

examples/store/db/index.ts
```diff
@@ -1,6 +1,23 @@
...
-const prisma = new PrismaClient()
+let prisma
+
+declare global {
+  namespace NodeJS {
+    interface Global {
+      prisma: PrismaClient
+    }
+  }
+}
```
flybayer, line 12: Thanks for these updates!

One thing, let's remove this because we don't want people to actually use `global.prisma`.

I think switching to `global["prisma"]` will fix the TS warning without needing the type declaration.

Probably also need `let prisma: PrismaClient`

abuuzayr: ok let me try that instead, thanks for the tips!

abuuzayr:
@flybayer i think the type issue is cleared! but it seems to be failing because of cypress. should i try to re-run the action?

flybayer:
Thank you @abuuzayr!

Issue #274, opened 2020-04-26
Architecture: Extract generator and console out of cli


ryardley:
### What do you want and why?

We should extract the generator and to a lesser extent the console out of the cli package to their own packages or at the very least folders

* Both represent pieces of code with very separate usage scenarios
* Both are currently mixed up with oclif and need not be
* Separating input argument parsing from task execution makes sense as we may want to expose a node API for either component in the future
* Both are sharing a utils folder containing context specific functionality that should be kept separate
* Groking the code it is hard to see what code belongs to which code area
* If we were to remove or swap out oclif at a later date it would be much easier should these be in separate packages.

### Possible implementation(s)

`@blitzjs/generator`
`@blitzjs/console`

aem:
I forgot this issue existed. `@blitzjs/generator` exists now, we still need someone to pull out the `console` code

peaonunes:
I'm happy to try pulling out the `console` got. Looking at the codebase its just one file, but it depends on `runPrismaGeneration` which is owned right now by the `/commands/db.ts` (and also used by the `start.ts` command). Not sure if it should be shared by those files, should we:
* move this function to a shared package before doing it?
* keep the function as it is and make `console` package get the function from `commands` package?
* duplicate the logic inside the `console` package
🤔

aem:
my personal thought is duplicate the logic for now. I assume at some point we'll want a shared `@blitz/prisma` package that we can use to centralize all of our prisma code, but I don't think we have enough information to know what that shared API needs to look like so better to duplicate rather than choosing the wrong abstraction.

@ryardley @flybayer thoughts?

flybayer:
The prisma code will be in a plugin like `blitz-plugin-prisma` or something.

Ok to duplicate for now, but add a comment both places clearly stating this.

peaonunes:
I've the same feeling about it.
I like the plugin idea, right now the `db` is tightly coupled with `prisma` which would be hard to switch/replace implementation if needed. I guess the ultimate goal would be to have an decoupled `db` module where you could plug in and out any database provider. Perhaps that could be another work front since it is a long way until having that abstraction.

PR #520, opened 2020-05-17
Extract CLI repl into its own package

peaonunes:
Closes: #274

This PR adds @blitzjs/console package which has all of the files needed for start and setup of REPL.

If this is not correctly done it can break the `console` functionality as well as impact other `cli` commands.

### Checklist
- [x] Tests added for changes

### Breaking change: yes/no
yes

### Other information

Hey @flybayer @aem @ryardley , here a couple of things I would like to clarify/ask:
1. I have not added any codeowner to this particular package, who should I add?
2. If the PR is too big to review I am happy to break it in two PRs (one for creation of new package, second for replacing `cli/commands/console` implementation). Just let me know.
3. I have only extracted the functions related with the REPL console from the `commands/console` therefore the REPL is lean and unaware of database setup or any other thing. I think this is an even better solution than discussed [here](https://github.com/blitz-js/blitz/issues/274#issuecomment-627205260). Any pre conditions to run the `console` can happen inside the console command.
4. Would be nice if someone else could download the branch and see if this is introducing any bugs. Since the REPL environment needs to be manually tested I believe would be nice to have another pair of eyes looking at it.


packages/console/src/console.ts
```diff
@@ -0,0 +1,95 @@
...
+
+const runConsole = async (replOptions: REPL.ReplOptions) => {
+  const repl = initializeRepl(replOptions)
+
+  const watchers = [
+    // watch('package.json').on('change', () => Console.loadDependencies(repl)),
+    watch(getBlitzModulePaths()).on('all', () => loadBlitzModules(repl)),
+  ]
+
+  repl.on('reset', () => loadModules(repl)).on('exit', () => watchers.forEach((watcher) => watcher.close()))
```
peaonunes, line 34: According to the [repl](https://nodejs.org/api/repl.html#repl_commands_and_special_keys) reference there is no `.reset` command. There is only `.clear` and maybe it is a different command rather than reset. Besides, we have added `.reload` in the custom commands to do the same. So maybe we could remove the `.on('reset')` here.

peaonunes:
Strangely I am getting `Internal Server Error` when running `blitz start`, I may have screwed something. However, all the other commands like generate, db migrate and console are working fine. In fact it is happening with the latest version of the package too.

flybayer:
@peaonunes thank you for this! I'll let @aem review.

To get past your internal server error, you could try running `yarn reset` at the blitz repo root.

packages/console/src/console.ts
aem: yeah we probably don't need `reset` here

packages/console/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "@blitzjs/console",
+  "version": "0.10.0",
```
flybayer, line 3: ``
```suggestion
  "version": "0.11.0",
```

peaonunes:
Hey, no worries! Thanks for updating the branch with canary.
Looking forward to contribute more.

ryardley:
One question: Should we call this package `repl` as there might be confusion between this and the idea of a logging package? (Such as the `log` component)

peaonunes:
Makes a lot of sense to me. Happy to do that if everyone agrees.

peaonunes:
What are your thoughts @aem @flybayer ?

aem:
I agree with `repl`, I didn't think about the potential confusion there

peaonunes:
The PR is ready to re-review. I've renamed the package and solved the conflicts with canary. I guess is good enough to go now.
cc: @aem @ryardley @flybayer

peaonunes:
Hey @flybayer I've done it here: https://github.com/blitz-js/blitz/pull/520/files#diff-7d95ce28d6b0598638ef2b376877e6c3R55
I've added as `devDependencies` on the `cli` package, would not that be enough? Because just the `cli` refers at the moment to `repl`.

Also, I do not get why the tests are failing on CI. Locally they all pass. Is there anyway that I can rerun them?

flybayer:
That's not enough because currently the repl package will not be linked in the production build. The repl package must be listed as a regular `dependency` somewhere. 

Probably the repl package dep could stay in the cli package.json, but be moved to a regular dependency.

peaonunes:
That makes a ton of sense. Sorry for that lol
1. Moved it to be a regular `dependency` in the `cli`.
2. Then I did `yarn reset`.
3. Then `yarn dev` & `yarn link-cli`.
5. Then `blitz new app`, run the migrations.
6. Finally `blitz console` and worked fine 🎉 .

Would it be enough to validate it? So we make sure that the production build of the `cli` resolves correctly the dependency.

flybayer:
Sweet, looks good! Thank you!

PR #522, opened 2020-05-17
Fix oclif hook id bug

Zeko369:
#408 # What are the changes and their implications?

Fix the way we access command id (name) in the init hook to check if we're in the correct directory or not


flybayer:
Thank you!

Issue #517, opened 2020-05-16
Extract `usePaginatedQuery` out of `useQuery` into a separate hook


flybayer:
### What do you want and why?

Currently the blitz `useQuery` hook wraps both `useQuery` and `usePaginatedQuery` from react-query. This has caused a hook violation (https://github.com/blitz-js/blitz/issues/334) so we need to switch to having a separate `usePaginatedQuery` hook, just like react-query.

### Possible implementation(s)

Start in this file: https://github.com/blitz-js/blitz/blob/canary/packages/core/src/use-query.ts

And separate out the paginated hook to `use-paginated-query.ts`

eliasjohansson:
I can work on this, do you mind if I do #239 at the same time? :)

flybayer:
@eliasjohansson  Cool! Yeah, that probably makes sense to do all three at once. You can assign these issues to yourself and mark as `status/assigned`

PR #523, opened 2020-05-18
Extract `usePaginatedQuery` out of `useQuery` into a separate hook

eliasjohansson:
Closes: #517 

Seperated pagination logic into its own hook. For pagination you should now use `usePaginatedQuery` instead of the `paginated` option for `useQuery`.

### Checklist

- [ ] Tests added for changes


flybayer:
Fantastic, thank you!!

PR #524, opened 2020-05-19
Adds a failure message if the cli compilation isn't complete

abuuzayr:
Closes: 519

Adds a failure message if the cli is unavailable via a simple try...catch

### Checklist

- [ ] Tests added for changes


flybayer:
Thank you @abuuzayr !!

PR #526, opened 2020-05-20
BREAKING: Change useQuery() to require second argument

flybayer:

Previously `useQuery` didn't require the second argument — this worked:

```tsx
const [products] useQuery(getProducts)
```

But this caused a problem where Typescript would compile fine with above code even if `getProducts` had a required argument. 

And for both TS and JS apps, the runtime error was confusing — something about `null`. 

The solution is to always require the second argument.

```tsx
const [products] useQuery(getProducts, null)
//or
const [products] useQuery(getProducts, {})
```

Now Typescript compile will fail correctly, and for JS apps, they will still get a `null` error but now they are the ones adding the `null` argument, not the framework.

### DIFF
--- a/packages/core/src/use-query.ts
+++ b/packages/core/src/use-query.ts
@@ -6,9 +6,19 @@ type RestQueryResult<T extends QueryFn> = Omit<QueryResult<PromiseReturnType<T>>
 
 export function useQuery<T extends QueryFn>(
   queryFn: T,
-  params?: InferUnaryParam<T>,
+  params: InferUnaryParam<T>,
   options?: QueryOptions<T>,
 ): [PromiseReturnType<T>, RestQueryResult<T>] {
+  if (typeof queryFn === 'undefined') {
+    throw new Error('useQuery is missing the first argument - it must be a query function')
+  }
+
+  if (typeof params === 'undefined') {
+    throw new Error(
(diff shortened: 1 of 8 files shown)


dajinchu:
I think we can use TS method overloading to allow the second argument to be excluded whenever the QueryFn has no parameters.

flybayer:
@dajinchu I thought about that, but that doesn't solve the runtime error issue. Almost always a query will have some sort of input argument, so I think the actual UX regression is negligible.

dajinchu:
oh gotcha - no good way to check the queryFn params at runtime

packages/core/src/use-query.ts
```diff
@@ -6,9 +6,19 @@ type RestQueryResult<T extends QueryFn> = Omit<QueryResult<PromiseReturnType<T>>
 
 export function useQuery<T extends QueryFn>(
   queryFn: T,
-  params?: InferUnaryParam<T>,
+  params: InferUnaryParam<T>,
   options?: QueryOptions<T>,
```
dajinchu, line 10: unrelated and I don't want to hold up this PR, but I dug into this for a sec and it seems like QueryOptions<T> expects T to be the type of the result of the queryFn, not the queryFn itself.

flybayer: You are right! I just pushed a fix for this

PR #532, opened 2020-05-21
Update Render deployment instructions in USER_GUIDE

johncantrell97:
Adds instructions to get blitz deploy with postgres db working on Render

### Checklist

- [ ] Tests added for changes

### Breaking change: yes/no
No

### DIFF
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -118,18 +118,17 @@ export default async function getProject({where, include}: GetProjectInput) {
 
 ```ts
 // app/products/mutations/createProduct.tsx
-import db, { ProjectCreateArgs } from "db"
+import db, {ProjectCreateArgs} from 'db'
 
 type CreateProjectInput = {
-  data: ProjectCreateArgs["data"]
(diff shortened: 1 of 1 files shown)


USER_GUIDE.md
```diff
@@ -297,8 +310,8 @@ services:
     name: myapp
     env: node
     plan: starter
-    buildCommand: yarn
-    startCommand: blitz start --production -H 0.0.0.0
+    buildCommand: yarn; blitz build
+    startCommand: yarn next start -p 80 -H 0.0.0.0
```
flybayer, line 314: I didn't think `-p 80` was required?

johncantrell97: I'm not sure if it is either, I sort of assumed it would default to port 3000 or 8000 or 8080 or some development port.  Felt like being explicit couldn't hurt.

flybayer: Ok yeah render will auto detect whatever port is being used, so we can leave this out of the guide.

Issue #533, opened 2020-05-21
Upgrade Next to 9.4.2


flybayer:
### What do you want and why?

Upgrade Next to 9.4.2 in all the package.json files in the repo

PR #534, opened 2020-05-21
Update Next to 9.4.2

Zeko369:
Closes: #533 

### Breaking change: No, for now

Could be breaking changes if we need to change react version


flybayer:
Thank you!

Issue #234, opened 2020-04-25
Add blitz generate model


flybayer:
**What do you want and why do you want it?.**

We want to add a model generator that:

- Adds a model & attributes to `schema.prisma`
- (Future) scaffolds model validation
- (Future) scaffolds model authorization

All of the commands listed here should result in a valid, working schema file.

```
blitz generate model task email:string:unique name:string? age:int
```

**The field arguments have this format: `fieldName:fieldType:fieldAttribute`**

- `fieldName`:
	- passed through exactly
	- can have `?` suffix that is passed through to schema file
- `fieldType`:
	- CLI will automatically PascalCase and singularize it
	- Ex: `string`, `boolean`, `int`, `float`, `dateTime`
    - Also supports relation models: `product`, `product[]`
- `fieldAttribute`:
    - Passed through and set as `@fieldAttribute` on the field
    - Can have optional parameter in brackets
    - Example: `int:unique` -> `Int @unique`
    - Example: `boolean:default[false]` -> `Boolean @default(false)`

**Relations:**

- One-to-one
```
blitz generate model chair name:string table:table
blitz generate model table name:string chair:chair
```

- One-to-many
```
blitz generate model project name:string tasks:task[]
blitz generate model task name:string project:project
```

- Many-to-many
```
blitz generate model project name:string categories:category[]
blitz generate model category name:string projects:project[]
```

These must all generate all the code needed for the relation, like:
```
  author    User?  @relation(fields: [authorId], references: [id])
  authorId  Int?
```

**These field arguments must also work with the higher level generators like this:**
```
blitz generate all task email:string:unique name:string? age:int
```

**Additional info**

- Prisma schema docs: https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-schema/prisma-schema-file
- Prisma schema spec: https://github.com/prisma/specs/tree/master/schema
- We need to use `[]` for the fieldAttribute parameter instead of `()` because `()` is for command substitution in the fish shell

aem:
This can probably be done incrementally to get users an MVP as quickly as possible, rather than trying to support all of this at once. My existing solution gets us close to the MVP. The one addition I'd like to make would be rather than parsing the values inline in the ModelGenerator, to have a `PrismaField` class that can take the colon-delimited string and parse it into a complete line of text for the prisma schema file. This gives a natural place to extend the existing functionality into supporting foreign keys and other directives.

Blanen:
[...]

Issue #259, opened 2020-04-26
Change `blitz generate pages` to use actual model attributes


flybayer:
### What do you want and why?

Currently the `blitz generate` command assumes the model has a `name` attribute. But we need to change this to use the actual attributes on the generated model.

When generating pages without passing model attributes, we should show an inquirer multi-select prompt with the current model attributes.
```
blitz generate pages task
```

When generating pages with passed model attributes, we use those instead showing a prompt
```
blitz generate pages task name email
```

When generating everything with `all`, the generated pages will include the attributes passed:
```
blitz generate all task email:string:unique name:string? age:int
```

Related: https://github.com/blitz-js/blitz/issues/234 probably want to wait on this issues until that is completed.

flybayer:
Closing in favor of https://github.com/blitz-js/blitz/issues/2038

PR #545, opened 2020-05-23
Add `blitz generate model` 🎉

aem:
Closes: https://github.com/blitz-js/blitz/issues/234

Adds new CLI command: `blitz generate model`

The command allows you to supply a series of `field:type` arguments to specify your model fields, and the output is a well-formed model added to `schema.prisma`. The `Field` and `Model` classes are based on Prisma's Data Model Metadata Format (DMMF) which gives us all of the metadata we need to reliably generate valid models. 

It also introduces a DSL for specifying relationships between models that we can easily translate into Prisma models. Today we support 1:1 and 1:Many generation via the `hasOne`, `hasMany`, and `belongsTo` directives (taking some inspiration from Rails in the naming here). In the future, we could expand this to include more Rails-like concepts like "hasManyThrough" and "hasManyAndBelongsToMany," either by generating a mapping table model + queries that filter based on the mapping table, or use prisma raw queries to achieve the same.

Lastly, we support model attributes such as `@unique`, `@default`, or `@updatedAt` via an optional third parameter in the field definition string, e.g. `updated:dateTime:udpatedAt` will output `updated DateTime @updatedAt`. For `@default`, you can use square brackets to denote defaults: `completed:boolean:default[false]`.

For all models, we auto-generate an ID column.

Knowing all of that, this means the following CLI command:

```bash
blitz generate model task \
  name:string \
  description:string? \
  completed:boolean:default[false] \
  belongsTo:project
```

...will generate the following model:

```
model Task {
  id          Int     @default(autoincrement()) @id
  name        String
  description String?
  completed   Boolean @default(false)
  project     Project @relation(fields: [projectId], references: [id])
  projectId   Int
}
```

Additionally, the field specifiers can be passed to any generator that generates models, so this:

```
blitz g all tasks \
  name:string \
  description:string? \
  completed:boolean:default[false] \
  belongsTo:project
```

Will still generate the `Task` model in addition to generating all of the associated pages + queries + mutations. Once I get #259 working this will be an incredibly powerful command - a one-liner to create a DB model, queries and mutations, and pages that read directly from that model. 

As a side note, this was also a great exercise to figure out what hooks database plugins will need to provide us. I think the fact that we already have Prisma's DMMF is a big leg up, if we use that spec then we can have all DB plugin authors conform to that model, provide hooks for reading+writing the schema, and then the rest of the logic is pretty much identical. Right now the serialization is specific to prisma, but it's also fully segregated and could easily be replaced with a call to a plugin. 

#### Known limitations:

* Relationships currently assume integer IDs. Once I tackle #259 we'll have the infrastructure to introspect the schema, and then I'll be able to reuse that logic to make ID generation dynamic. For now, users who customize their ID column in the generated models will have to update generated relationships manually.
* Currently you can't specify the ID field's format, we just use an autoincrementing `Int`. I'd like to add an `--id` field that can be used to set the ID generation scheme to a `cuid` or `uuid` `String`
* Interactive shells (such as `zsh` or `fish`) will parse `[]` or `?` as special characters since those have special meaning. To avoid this, users can either escape the special characters, wrap each field in a string, or pass all fields as a single string (e.g. `blitz g model task "name:string description:string?"`)
* When passing field arguments to the generators, we should check if the model exists, and if it does append the fields rather than creating a new model. Right now this is a hard problem to solve because we need to somehow parse out the schema file and then rewrite parts in the middle. Once #259 is complete and we can read from the existing schema, the answer might just be to blindly regenerate the full schema rather than trying to interpolate the new fields, which could be fragile. Too early to say.

### Checklist

- [x] Tests added for changes


flybayer:
YES! This is so cool! Good work! Few points:

- There's currently no output from `blitz generate model` besides the alpha warning message. We need to provide some type of nice feedback. Probably would be nice to print out the exact string we append to the schema.
- I think we should use cuid for our default ids. I [opened a discussion on this](https://github.com/blitz-js/blitz/discussions/554). But for now I think we can merge this with int until #259.
- Just fyi, I'm using fish shell and the brackets work fine without quoting them (`price:int:default[0]`). But I do have to quote the question mark (`'name:string?'`)

aem:
> There's currently no output from blitz generate model besides the alpha warning message

Yep, meant to tackle that. For now we can just print out the model

> We should also add createdAt and updatedAt by default for all models (like Rails)

Agreed. Good call

> I think we should use cuid for our default ids

Looking at that discussion it seems like lots of people like Ints, so I'm gonna leave it like this for now, with #259 we can revisit it

> Just fyi, I'm using fish shell and the brackets work fine without quoting them

Ah, good to know. Both fail in `zsh` and when the question mark didn't work I just assumed the brackets wouldn't as well

aem:
new output:

```
✔ Model for 'task' created successfully:

> model Task {
>   id        Int      @default(autoincrement()) @id
>   createdAt DateTime @default(now())
>   updatedAt DateTime @updatedAt
> }
```

packages/generator/src/prisma/model.ts
```diff
@@ -0,0 +1,77 @@
...
+      type: FieldType.DateTime,
+    })
+  }
+
+  private getFields() {
+    return stringifyFieldsForPrinting([
+      this.getIdField(),
+      ...this.fields,
+      this.getCreatedAtField(),
+      this.getUpdatedAtField(),
```
flybayer, line 68: What do you think about leaving id, createdAt, and updatedAt all at the beginning? I like this because then all the "automatic" ones are grouped together. And later when I come back to add a new field, I can just add it to the end. This would also help if we support adding fields via CLI.

```suggestion
      this.getCreatedAtField(),
      this.getUpdatedAtField(),
      ...this.fields,
```

aem: I thought about this, my concern was that it makes the studio a little more tedious to use - columns are ordered, so it led to the first 3 columns in the DB being `id`, `createdAt`, `updatedAt`, which means most of the UI you see isn't super useful. I think from a technical perspective it doesn't matter too much, but there's maybe some usability concerns there

flybayer: Yeah that's a good point. I pondered this for a bit. And I think I specifically like the dates at the beginning in a UI. Often the dates are one of the items you are looking at or sorting by.

aem: Interesting, I felt the opposite way but totally get the argument that you might want to sort by date. I'm gonna drop a quick poll in the slack and see what folks think

flybayer:
@aem Thanks! I found a few more issues:

1) `--dry-run` is not respected for the model generator
2) I originally thought to require a field type, but now I think we should default to `String` like Rails does.
3) For "combination generators" model generation should happen first. And if it fails, generation should abort. For the below example, currently pages and everything would be created even though model generation fails. And now that I'm thinking about it, I think all generators should run in a single "transaction". I.e. if anything fails, everything is aborted. This probably requires a significant refactor. If you agree, maybe you can open an issue so we track that.
4) The error message shown below can be improved. Maybe something more like `✕ Error: You need to provide a name for the model field attribute you supplied as ':Int' - Example: 'something:Int'`

```
~/c/b/e/store> blitz g all ticket :Int -d

Would create    app/tickets/pages/tickets/[ticketId]/edit.tsx
Would create    app/tickets/pages/tickets/[ticketId].tsx
Would create    app/tickets/pages/tickets/index.tsx
Would create    app/tickets/pages/tickets/new.tsx
Would create    app/tickets/components/TicketForm.tsx
Would create    app/tickets/queries/getTicket.ts
Would create    app/tickets/queries/getTickets.ts
Would create    app/tickets/mutations/createTicket.ts
Would create    app/tickets/mutations/deleteTicket.ts
Would create    app/tickets/mutations/updateTicket.ts
✕ Error: [PrismaField]: no field name supplied
```

5) The `-h` screen has an error. It says `blitz g pages` will also generate the model
```
  # Database models can also be generated directly from the CLI
  # Model fields can be specified with any generator that generates
  # a database model ("all", "model", "resource"). Both of the below
  # will generate the proper database model for a Task.
  > blitz generate model task \
       name:string \
       completed:boolean:default[false] \
       belongsTo:project?
  > blitz generate pages tasks \
       name:string \
       completed:boolean:default[false] \
       belongsTo:project?
```

aem:
> For "combination generators" model generation should happen first. And if it fails, generation should abort. For the below example, currently pages and everything would be created even though model generation fails. And now that I'm thinking about it, I think all generators should run in a single "transaction". I.e. if anything fails, everything is aborted. This probably requires a significant refactor. If you agree, maybe you can open an issue so we track that.

🤔We could just write generators as recipes once the framework is functional - it's effectively just a copy files step

packages/generator/src/prisma/field.ts
```diff
@@ -56,7 +56,7 @@ export class Field {
 
   // 'name:type?[]:attribute' => Field
   static parse(input: string): Field[] {
-    const [_fieldName, _fieldType, attribute] = input.split(':')
+    const [_fieldName, _fieldType = 'String', attribute] = input.split(':')
```
aem, line 59: just a heads up @flybayer, this deafult is already set down in the `Field` constructor rather than in `parse` so that directly constructed instances get this benefit, I think we can remove this

flybayer: Oh.. I didn't realize because it wasn't working. `_fieldType` must be set here because it's used just two lines below I believe.

Issue #237, opened 2020-04-25
Implement useQuery prefetching


flybayer:
Both of the following should cache the `getProduct` query, same as if it was used with `useQuery`

```ts
const product = await getProduct({where: {id: props.id}})
```

```tsx
<button onMouseEnter={() => getProduct({where: {id: productId}})}>
  View Product
</button>
```

- This functionality needs to be added in the `rpc.ts` file in the `core` package. 
- It needs to use the same computed cache key as used inside the blitz `useQuery`.
- It needs to use [`queryCache.setQueryData`](https://github.com/tannerlinsley/react-query#querycachesetquerydata) from react-query to actually set the cache

Then end-result is that once prefetched like shown above, then the next call of `useQuery(getProduct, {where: {id: productId}})` will return the cached data.

flybayer:
Let me know if you want to work on this but need more information about how to build it!

ivandevp:
Hey @flybayer ! I can work on this. I'll probably need more information but I can start by checking how `react-query` works for this case and understand how to add it to Blitz ⚡.

flybayer:
@ivandevp sounds good! :)

PR #546, opened 2020-05-23
Add ability to prefetch query data

ivandevp:
Closes: #237

### Checklist

- [ ] Tests added for changes


flybayer:
Thank you for your work on this @ivandevp!

There's one more change needed (it's not working as is). You need to also use the query parameters for the actual cache key. See [this line](https://github.com/blitz-js/blitz/blob/canary/packages/core/src/use-query.ts#L23) for the exact query cache key. It's probably worth breaking that out into a utility function that we can use anytime we need the full cache key.

ivandevp:
You're right @flybayer! Will make the change.

ivandevp:
Hey @flybayer ! I've updated the PR to use the params in the query cache key 😃

flybayer:
Awesome, thanks @ivandevp! I'll test this tomorrow :)

flybayer:
Ok, this works great! It was also making extra calls for normal `useQuery` usage, so I pushed a fix for that and also added this to the store example.

Thank you for your help on this!

Issue #540, opened 2020-05-21
Add ability to run blitz start production without building code


flybayer:
### What do you want and why?

Services like render have different memory limits for build vs run step. So we need build the code in the build step and then only run the code in the run step. But currently `blitz start --production` rebuilds the code even if the code is already built.

### Possible implementation(s)

We could add a new command or command argument.

Or

We can change `blitz start --production` to use the built assets if they exist. We would add some type of checksum that makes it easy to verify that the code has not changed since the code was built. If the code has changed, then we run a new build.

I really think we should do the later.

cc @ryardley thoughts on this?

ryardley:
Yep I agree. If the assets have been changed they should be rebuilt.

ryardley:
Should we make it so all builds are idempotent like this but only fire if source files have changed?

Issue #226, opened 2020-04-25
Do not clean dev cache on restart


ryardley:
**What do you want and why do you want it?.**
Currently running `blitz start` will blow away the file cache next uses to run and rebuild it. This means that on huge projects when you stop blitz and then restart it the whole project needs to be copied to the cache which can take quite a bit of time. We should work out a way to prevent this and only copy changed files.

**Possible implementation(s)**
There are some ideas fleshed out in the [synchronizer readme](https://github.com/blitz-js/blitz/blob/canary/packages/server/src/synchronizer/README.md#future-thinking)

Basically we finish the optimizer to detect file system changes based on a single file we write. This file links the changes we have made to the filesystem back to the original file that spurred those changes. Whenever blitz starts it reads that file and uses it to determine if the file needs processing. This should probably be disabled on `blitz build`.

PR #565, opened 2020-05-27
Use previous build for `blitz start --production` if up-to-date

ryardley:
Closes: #540

### Checklist

- [x] Tests added for changes

### Other information

- [x] Build is not happening when files are the same
- [x] Add testing

### DIFF
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -36,6 +36,7 @@
     "detect-port": "1.3.0",
     "fast-glob": "3.2.2",
     "flush-write-stream": "2.0.0",
+    "folder-hash": "3.3.1",
     "from2": "2.3.0",
     "gulp-if": "3.0.0",
     "lodash": "4.17.15",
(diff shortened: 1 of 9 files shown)


merelinguist:
Is [this](https://github.com/blitz-js/blitz/issues/226) within the pr scope?

ryardley:
> Is [this](https://github.com/blitz-js/blitz/issues/226) within the pr scope?

No it isn't. That is a different problem. 

This assumes builds are either complete or not. #226 is checking for partial builds and optimizing accordingly and was more valuable optimizing against large codebases.

ryardley:
So after having done this I am wondering perhaps the checksumming logic should occur within the build function itself? That means that running build twice without changing input will not do anything. Thoughts?

flybayer:
@ryardley we could, but I'm not sure the UX is great. Like, if you run `blitz build` you expect it to build something, but if instead it says, already built, no changes, then you may not trust the system and want a `--force` flag. 🤷

PR #567, opened 2020-05-27
Ensure Prisma client is generated on blitz build & blitz start

### LINKED ISSUE #542: Generate Prisma Client w/o Migrating
### What do you want and why?

Some way to generate the prisma client so that blitz apps can be fully built without forcing migrations to happen at build time.

If we want to avoid ORM-coupling, this is kind of tough. In my opinion, prisma client would be generated as part of `blitz build`.

---

## Edit by @flybayer 

[See this comment below](https://github.com/blitz-js/blitz/issues/542#issuecomment-633617006) on how exactly to do this.

ganeshmani:
Closes #542

This PR adds the feature of generating prisma client from `blitz start --production` and `blitz build` commands

 This PR adds the prisma generation method in `cli/start.ts` and `cli/build.ts`

Screenshot 2020-05-27 at 9 52 00 PM

Screenshot 2020-05-27 at 9 52 08 PM

It also changes the `server/build.ts`  and `server/prod.ts` methods to accept the prisma generation function as an argument.

 
### Checklist

- [x] Tests added for changes


MrLeebo:
@ganeshmani why does this delete the lockfile?

ganeshmani:
@MrLeebo it shouldn't. let me correct this.

ganeshmani:
@MrLeebo Fixed the yarn lock issue. can you check it once?

flybayer:
@ganeshmani thank you! I'll let @ryardley review this, but you do have a number of merge conflicts that need resolved before it can be merged :)

ganeshmani:
@flybayer Okay. let me try to rebase and resolve the conflicts.

ryardley:
Hey @flybayer what is your take on how folks are expected to use `blitz start --production`? 
I think this is all fine if it is meant to be used in development to test a site in 'production' mode. Personally for deployment I would run a build and then deploy the site as a next site running `next start`
If it is meant to be the command they run in production after a deployment then I am not so sure this stuff we are doing (including the new build checksum stuff) is such a great idea. 
What if there is an error in the app and the app was automatically restarted that would mean that this db migrate stuff and build checksum stuff would run everytime. If there was an error that surfaced it would run in an infinite loop with every run possibly affecting the production db (although the idempotency provided by prisma will probably prevent a catastrophic change). All I am saying is that this will all probably be ok but starting the production app if we are actually in a production situation, to me, should be something that happens as quickly as possible without much magic around it. If we are simply testing production locally then this stuff is useful.

flybayer:
@ryardley I definitely expect folks to use `blitz start --production` for real deployments. 

Here's the config for deploying to Render:

```yaml
buildCommand: yarn; blitz db migrate; blitz build
startCommand: blitz start --production -H 0.0.0.0
```

The `buildCommand` is only run one time. But the `startCommand` will possibly be ran multiple times, like if the server crashes and it's auto restarted.

So `db migrate` is not an issue here. And the build checksum inside `blitz start --production` will always match because it's using the same build bundle.

Make sense?

ryardley:
I think my point is all deployment platforms provide a production step to run migration scripts which are only run once. After deployed all that should happen is that the app is started. Let's say you have a huge app with hundreds of containers. What if you are connecting to a bunch of slave-replicated databases that are not yet consistent. There is a small chance that any single one could potentially trigger a migration corruption. At the same time we are slowing down application start times (not by heaps but it does increase it as the checks are run). Just not sure allowing this bug is a good tradeoff for production. It is not bad DX to have to migrate and build separately before starting production. It actually prevents errors.

ryardley:
Ahh right I thought that we were running a migrate in start which we are not!

packages/cli/src/commands/build.ts
```diff
@@ -1,7 +1,9 @@
 import {Command, flags} from '@oclif/command'
 import {build} from '@blitzjs/server'
+import {runPrismaGeneration} from './db'
 
 export class Build extends Command {
+  // eslint-disable-next-line import/no-default-export
```
ryardley, line 6: Any reason for this?

ganeshmani: Ah.Okay. it might occurred while i was resolving merge conflict. shall i remove this?

flybayer:
@ryardley yeah ideally this prisma client generation would only happen in the build phase, that way `blitz start --production` would not regenerate prisma client if `blitz build` was already run.

But I think this is good for now, we can optimize this when we move the prisma stuff into a plugin.

flybayer:
Thank you @ganeshmani!!

@all-contributors add @ganeshmani for code

Issue #530, opened 2020-05-20
Custom useParams hook


Zeko369:
### What do you want and why?

As mentioned in the [slack](https://blitzjs.slack.com/archives/CUM2C68T1/p1589975795454700), current way of getting route params isn't really as nice as it could be

```ts
const router = useRouter()
const id = parseInt(router?.query.id as string)
```

### Better API (similar to prisma client)

```ts
// returns ID as number
const {id} = useParams({id: true}); // Or maybe have 'number'?
// returns ID as string
const {id} = useParams({id: 'string'});
// Custom parser for hex strings
const {id} = useParams({id: (raw: string) => parseInt(raw, 16)});

// Multiple params
const {projectId, slug} = useParams({projectId: true, slug: 'string'});

// Maybe have a shorthand syntax with an array
const {id} = useParams(['id']); // id is number by default
```

Also if we have an object we can go overboard with TS definitions (keyof stuff) and have a lot nicer API for getting this stuff

I'd like to build this, but before I'd like to discuss maybe there's some other ideas on how to improve this

### Insipiration

In a CRA project of mine

```ts
// useParamsInt.ts
import { useParams } from 'react-router-dom';

const useParamsInt = (key = 'id') => {
  const params = useParams();
  return params[key] ? parseInt(params[key]) : null;
};

export default useParamsInt;

// component.tsx
const id = useParamsInt();
const chapter_id = useParamsInt('chapter_id');
const page_id = useParamsInt('page_id');
```

flybayer:
I totally agree we need this. I was just thinking about this the other day.

The whole number/string thing is a real downer lol.

I think the majority of folks will use all number ids or all string ids. So I think it should be super simple for either of these.

How about this?

```ts
// Always returns numbers - no option argument
const {taskId} = useParamsInt()

// Always returns strings - no option argument
const {taskId} = useParamsStr()

// Escape hatch for a mixed bag
const {taskId} = useParamsInt()
const router = useRouter()
const projectId = router?.query.projectId as string
```

Zeko369:
@flybayer I think having more powerful useParamsCustom() which would take a prisma query like object (key -> boolean | function, where a function would be a parser function and true would just parseInt) would be quite useful, just to make it a bit more readable for some stuff.

Also I'd change your example

```ts
// Always returns strings - no option argument
const {id} = useParamsStr()

// Always returns numbers - no option argument
const {id} = useParamsInt()

// Always returns numbers - no option argument
[...]

PR #574, opened 2020-05-28
Add `useParams` and `useRouterQuery` hooks

anteprimorac:
Related: #530 

This PR implements two hooks `useRouterParams` and `useRouterQuery`.

### Checklist

- [x] Tests added for changes

### Breaking change: yes/no
No.

### Other information
Example of usage:

```ts
// app/locations/pages/locations/[postalCode].tsx
import { useRouterParams, useRouterQuery, useQuery } from 'blitz';

export const Locations = () => {
  const { postalCode } = useRouterParams();
  const { order, orderby } = useRouterQuery();
  const [locations] = useQuery(getLocations, { where: { postalCode }, orderBy: { [orderby]: order } });
  return locations;
};
```


packages/core/src/use-router-query.ts
```diff
@@ -0,0 +1,8 @@
+import {useRouter} from 'next/router'
+import {parse} from 'url'
```
flybayer, line 2: Need to add `url` as a dependency of the `blitzjs/core` package

anteprimorac: `url` dependency is added. I've also rewritten the implementation of `useRouterParams` as you described, replaced `router.query` in the store example with new hooks and added integration test that tests if `/admin/products?order=asc` changes the order of products in DOM

> Lastly, before merging this we should resolve our discussion in #530 about number parsing. Maybe we can have `useParamsStr`, `useParamsInt`, and `useRouterQuery`

Let's discuss it in the issue.

flybayer:
@anteprimorac can you also add at least one unit test for `useRouterParams`?

Zeko369:
@anteprimorac Also rerun prettier with our config, it seems like some files weren't formatted correctly

anteprimorac:
> @anteprimorac can you also add at least one unit test for `useRouterParams`?

I've changed store example to use `useRouterParams` here https://github.com/blitz-js/blitz/pull/574/commits/ba730b17fd648418d26ae082064c2a5bd1b07aad#diff-1b842ad42068e396b7c21d9376384d87R8 and integration test is run over it. Is that enough?

It's hard to write unit tests because it depends on `useRouter` hook from next.js. Not sure if they provide a mock for it.

> @anteprimorac Also rerun prettier with our config, it seems like some files weren't formatted correctly

🤔 it looks like there is a separate prettier configuration for each example. Going to fix it now. Thanks for pointing this.

packages/core/src/use-router-params.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export function useRouterParams() {
+  const router = useRouter()
+  const query = useRouterQuery()
+
+  return Object.fromEntries(
+    Object.entries(router.query).filter(
+      ([key, value]) => typeof query[key] === 'undefined' || query[key] !== value,
+    ),
+  )
```
flybayer, line 12: This is the part that needs unit tests — you can extract it into a function outside useRouterParams, export it, and then test this.

anteprimorac: Yeah, for sure, that could be exported and tested. Thanks for pointing this out. I will make a unit test for that.

anteprimorac: @flybayer the test is added. Let's see does it pass 😄.

packages/core/package.json
```diff
@@ -46,7 +46,8 @@
   "dependencies": {
     "pretty-ms": "6.0.1",
     "react-query": "1.3.3",
-    "serialize-error": "6.0.0"
+    "serialize-error": "6.0.0",
+    "url": "^0.11.0"
```
flybayer, line 50: Change version to be pinned `"0.11.0"`

packages/core/src/use-router-params.ts
```diff
@@ -0,0 +1,46 @@
...
+      }
+    }
+
+    return true
+  }
+
+  return value1 === value2
+}
+
+export function exportRouterParams(routerQuery: ParsedUrlQuery, query: ParsedUrlQuery) {
```
flybayer, line 33: This is minor, but let's rename to "extract" which is a good bit more clear about what's happening :)

```suggestion
export function extractRouterParams(routerQuery: ParsedUrlQuery, query: ParsedUrlQuery) {
```

packages/core/package.json
flybayer, line 50: Change version to be pinned "0.11.0"

packages/core/src/use-router-params.ts
flybayer, line 33: This is minor, but let's rename to "extract" which is a good bit more clear about what's happening :)

```suggestion
export function extractRouterParams(routerQuery: ParsedUrlQuery, query: ParsedUrlQuery) {
```

packages/core/test/use-router-params.test.ts
```diff
@@ -0,0 +1,29 @@
...
+        cat: 'category',
+        slug: ['example', 'multiple', 'slugs'],
+        empty: '',
+        queryArray: ['1', '123', ''],
+      }
+
+      const query = {
+        cat: 'somethingelse',
+        slug: ['query-slug'],
+        queryArray: ['1', '123', ''],
```
flybayer, line 17: There's one more case which isn't covered here which is the result `params` should not include a key that is in `query`, but not in `routerQuery`. So I think just add a key here to `query` and tests should still pass.

Makes sense? :)

README.md
```diff
@@ -270,6 +270,7 @@ Thanks to these wonderful people ([emoji key](https://allcontributors.org/docs/e
 
 <!-- markdownlint-enable -->
 <!-- prettier-ignore-end -->
+
```
anteprimorac, line 273: @merelinguist this change was introduced by husky pre-commit hook

anteprimorac:
@flybayer if you agree, we can merge this as it is and gather more info on what to provide regarding validation, sanitization and casting of params and queries.

flybayer:
@anteprimorac let's wait until we resolve our discussion in #530. I just made another comment.

And if we add `useParamNumber()` and friends, I think it'd be better to use `useParams()` instead of `useRouterParams()` so all the param hooks are the same, and it's only query that is different. Additionally, all (or most) of other routers have `useParams()` so this will be more natural/intuitive for many folks.

flybayer:
@anteprimorac what do you think about also changing `useRouter()` to have `params` and `query` to align with these new hooks?

**Current:**
```tsx
// Everything lumped in query
const {query} = useRouter()
```

**Idea:**
```tsx
// Separate route params from query string items
const {params, query} = useRouter()
```

This would be a breaking change for existing blitz apps

anteprimorac:
@flybayer that’s an excellent idea. We could provide `useRouter` as a wrapper around the hook from next.js

flybayer:
Yes exactly. I think we should do that

PR #580, opened 2020-05-29
Fix `examples/store` code style issues

anteprimorac:
The files in `examples/store` directory were styled with the wrong prettier config. This PR applies the right config on those files, and it will make future PR's cleaner.

### Breaking change: yes/no
No.


flybayer:
Thank you @anteprimorac!!

@all-contributors add @anteprimorac for code

Issue #352, opened 2020-04-30
Workout better file system mocking for tests


ryardley:
### What do you want and why?

We now have mock-fs handling our in memory tests but there are issues with it in the broader sense. 

https://github.com/blitz-js/blitz/pull/347#pullrequestreview-403152892

* Needs to work on windows
* Act as a fs overlay
* Can be reset per test

### Possible implementation(s)

unionfs/memfs

Skn0tt:
Also, mock-fs doesn't work with oclif.

ryardley:
@Skn0tt fair point if we want something that will be universal.

PR #584, opened 2020-05-30
Tidy up tests in server

ryardley:
Closes: #352 

Found a workaround for mock-fs to resolve the issue with `console.log`.

This creates a new `multiMock` function that enables clearer mocking syntax for tests.

### Checklist

- [x] Tests added for changes

### Other information

Tried various approaches including unionfs with memfs but found issues with vinyl compatibility. Investigated using a jest native approach but centralizing `mock-fs` and fixing the issue with console.log appears to be the easiest win here.


flybayer:
cc @pixelmord @dajinchu 👆

Issue #579, opened 2020-05-29
Long unresponsive delay after starting `blitz console`


flybayer:
### What is the problem?

There's a several second delay after starting `blitz console` where it is unresponsive.

The loading spinner appears saying "loading code" which is fine and correct, but after that finishes and the prompt appears, it's frozen.

Maybe there's something being loaded after the spinner is marked as success?

### Steps to Reproduce

1. `blitz console`
2. Try to type stuff as soon as the prompt appears

### Versions

```
0.12.0
```

jportela:
I'll tackle this 👀

alkamin:
On `0.17.1-canary.4` I'm still experiencing this.

flybayer:
Someone needs to dig into this and figure out exactly what is causing the delay. Then we can figure out the proper fix.

flybayer:
Duplicate of https://github.com/blitz-js/blitz/issues/1098

PR #592, opened 2020-05-31
Fix `blitz console` loading freeze

jportela:
Closes: https://github.com/blitz-js/blitz/issues/579

I've identified two issues, that I addressed:

* The slowdown was caused by [`globby.sync`](https://github.com/jportela/blitz/commit/8abfd55faa73a2780a88ac5b7dccd681532d8c06#diff-55211f3adece9b5e1a19c1252a19369dL8-R20) on [`loadBlitz`](https://github.com/jportela/blitz/commit/8abfd55faa73a2780a88ac5b7dccd681532d8c06#diff-72f8daee91d07356f0891c0979903245L57). This operation takes ~300ms on my machine and it freezes input on the console. I've changed that to an async equivalent
* I've also set [`ignoreInitial: true`](https://github.com/jportela/blitz/commit/8abfd55faa73a2780a88ac5b7dccd681532d8c06#diff-72f8daee91d07356f0891c0979903245R58) for `chokidar`, since it was triggering `add` and `addDir` events for directories that it discovered while it loaded, which is not necessary in this case, since we already load the modules before watching (at least this was my interpretation of https://github.com/paulmillr/chokidar#path-filtering )

Feel free to suggest changes to this approach. We could, for example, block the console interaction while loading Blitz modules (and show a Spinner while doing it), but I don't think that's necessary at all.

### Checklist

- [ ] Tests added for changes

### DIFF
--- a/packages/repl/src/repl.ts
+++ b/packages/repl/src/repl.ts
@@ -25,13 +25,7 @@ const commands = {
 
 const runRepl = async (replOptions: REPL.ReplOptions) => {
   const repl = initializeRepl(replOptions)
-
-  const watchers = [
-    // watch('package.json').on('change', () => Console.loadDependencies(repl)),
-    watch(getBlitzModulePaths()).on('all', () => loadBlitzModules(repl)),
-  ]
-
-  repl.on('reset', () => loadModules(repl)).on('exit', () => watchers.forEach((watcher) => watcher.close()))
+  await setupFileWatchers(repl)
 }
 
 const initializeRepl = (replOptions: REPL.ReplOptions) => {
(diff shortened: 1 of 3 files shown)


flybayer:
@jportela awesome, thank you for your work on this! :)

One question:

> We could, for example, block the console interaction while loading Blitz modules (and show a Spinner while doing it), but I don't think that's necessary at all.

From reading and using this code, it seems the behavior is the same as before, where all files are loaded before the actual repl is started. But is this not the case?

jportela:
@flybayer yes, the startup behavior is essentially the same, since the modules are loaded before we show the REPL prompt. What will change is the reloading of the modules, whenever there’s a file change. Instead of blocking the input, users will still be able to use the terminal, before the new changes come through, which IMO provides a better UX.

flybayer:
@jportela ah ok, yeah that makes sense and I agree. Thanks!

Issue #572, opened 2020-05-28
DB is imported in generated api routes, even if it isn’t used


merelinguist:
### What is the problem?

Even if I’m not using a database, Blitz still expects it to be there to use queries. 

### Steps to Reproduce

1. Create a project

2. Delete the `db` directory

3. Create a query

4. Output looks like this:

```
import {rpcHandler} from '@blitzjs/server'
import resolver from 'app/_rpc/queries/scrapeRecipe'
import db from 'db'
export default rpcHandler('query', 'scrapeRecipe', resolver, () => db.connect())
```

### Versions

```
debug: local
debug: pkgPath: /Users/dylan/dev/recipes/node_modules/@blitzjs/cli 

macOS Catalina | darwin-x64 | Node: v14.0.0

blitz: 0.11.0 (global)
blitz: 0.11.0 (local)
```

flybayer:
Ah yeah! This will move into the prisma plugin eventually, but for now we can probably just wrap a try/catch around `require(db)` and also change to `db && db.connect()`

MykalMachon:
I'd like to take a stab at this if no one else is working on it! 
seems like a good place to get started with contributing 😃

ivandevp:
Go ahead @MykalMachon! 👍

PR #593, opened 2020-05-31
Fix queries when Prisma isn't being used

MykalMachon:
this should make it possible to use queries in blitz, **without** using blitz's inbuilt database system.

Closes: #572 

- I changed how `db` was being imported from an `import db from 'db'` statement to an undefined db declaration followed by a `require('db')` statement within a try/catch block. 
- I also altered the exported `rpcHandler` function call to ensure `db` has been defined before running `db.connect()`. 

This should allow users to fully remove the database portion of blitz if they're not planning on using it while still using queries with an external database or API. 

### Checklist

- [x] Tests added for changes _(N/A as far as I can tell)_
- [x] User facing changes documented _(N/A from a user perspective, from a maintainer perspective I referenced the issue and why changes were made in a comment above the `rpcHandlerTemplate` function)_ 

### Breaking change: yes/no
no

### Other information
This is my first contribution to blitz, and one of my first contributions to OS in general, I know it's a simple one but any feedback would be greatly appreciated. **Thank you so much for the opportunity to help out 🙏**

### DIFF
--- a/packages/server/src/synchronizer/pipeline/rules/rpc/index.ts
+++ b/packages/server/src/synchronizer/pipeline/rules/rpc/index.ts
@@ -62,11 +62,15 @@ import resolver from '${resolverPath}'
 export default isomorphicRpc(resolver, '${resolverPath}') as typeof resolver
 `
 
+// Clarification: try/catch around db is to prevent query errors when not using blitz's inbuilt database (See #572)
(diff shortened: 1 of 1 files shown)


flybayer:
@MykalMachon awesome, great job all around! :) Thank you so much!

@all-contributors add @MykalMachon for code 🎉

Issue #582, opened 2020-05-30
Implement the `mutate` function for `useQuery`


flybayer:
### What do you want and why?

We need to add a `mutate` function that's returned from `useQuery`. This allows you to manually update the query cache without refetching from the network.

- [ ] `mutate` must be fully typed for Typescript usage. The input type is the same as the query function return type.
- [ ] After being called, `mutate` must automatically trigger a refetch for the initial query to ensure it has the correct data

```tsx
export default function (props}) {
  const [product, {mutate}] = useQuery(getProduct, {where: {id: props.id}})

  return (
    <Formik
      initialValues={product}
      onSubmit={async (values) => {
        try {
          const product = await updateProduct(values)
          mutate(product)
        } catch (error) {
          alert("Error saving product")
        }
      }}>
      {/* ... */}
    </Formik>
  )
}
```

eliasjohansson:
I'll have a go at this today :)

PR #596, opened 2020-05-31
Add `useQuery` mutate function for updating query cache

eliasjohansson:
Closes: #582 

- Added mutate function to all query hooks second return value object. Updates queryCache with provided value and then performs refetch to get the actual updated value.

Can be called in two different ways:

As a function that has access to the previous/current value
```js
mutate((oldData) => ({
    ...oldData,
    handle: "mutated-apple",
  })
)
```

Or just set the value directly
```js
mutate({
  description: "Mutated Apple",
  handle: "mutated-apple",
  id: 1,
  name: "Mutated Apple",
  price: 123,
})
```

### Checklist

- [ ] Tests added for changes

### Other information
Should we add an example for this anywhere? If so, where?


flybayer:
Beautiful!! Thank you!!

While you're at it, could you also add mutate usage in these two places?

1. In the store example in `app/admin/pages/admin/products/[id].tsx` in the `onSuccess` callback
2. To the page generator template in `packages/generator/templates/page/__modelIdParam__/edit.tsx`

eliasjohansson:
@flybayer Added, hopefully that was what you meant :)

flybayer:
Awesome, thanks @eliasjohansson! `mutate` wasn't typed properly, so I pushed a fix for that.

Issue #598, opened 2020-06-01
withBlitz does not account for `function` next config


ryardley:
### What is the problem?

Next configs can be a function: 

```js
module.exports = (phase, { defaultConfig }) => {
  return {
    /* config options here */
  }
}
```

See docs: https://nextjs.org/docs/api-reference/next.config.js/introduction

The current `withBlitz` implementation does not allow for functional next configs to be passed to it. 

We need to check that the config is a function and wrap it accordingly.

This is important for people wanting to use things like https://github.com/cyrilwanner/next-compose-plugins with their configuration. 

### Steps to Reproduce

1. Use withBlitz with a function config

### Versions

```
debug: local
debug: pkgPath: /Users/rudiyardley/projects/yatou/yatou/node_modules/@blitzjs/cli 

macOS Catalina | darwin-x64 | Node: v12.16.1

blitz: 0.12.0 (global)
blitz: 0.12.0 (local)
```

### Other

Please include applicable logs and screenshots that show your problem.

PR #599, opened 2020-06-01
Allow function config in withBlitz

ryardley:
Closes: #598 

* Users can export Next.js function configs in `blitz.config.js`

### Checklist

- [x] Tests added for changes

### DIFF
--- a/packages/server/test/with-blitz.test.ts
+++ b/packages/server/test/with-blitz.test.ts
@@ -2,11 +2,29 @@ import {withBlitz} from '../src'
 
 describe('withBlitz', () => {
   it('alters the webpack config as expected', () => {
-    const nextConfigFn = withBlitz({})
+    const nextConfigFn = withBlitz({
+      foo: 'bar',
+    })
     const newNext = nextConfigFn('', {defaultConfig: {}})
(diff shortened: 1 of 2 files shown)


packages/server/test/with-blitz.test.ts
```diff
@@ -2,11 +2,29 @@ import {withBlitz} from '../src'
...
+    const nextConfigFn = withBlitz((phase: any) => ({phase, foo: 'bar'}))
+    const newNext = nextConfigFn('foo', {defaultConfig: {}})
+    const newNextWithoutWebpack = Object.assign({}, newNext, {webpack: null})
+
+    expect(newNextWithoutWebpack).toStrictEqual({
+      phase: 'foo',
+      foo: 'bar',
       experimental: {
         reactMode: 'concurrent',
       },
```
flybayer, line 30: Should `withBlitz` return a function here since the input is a function?

ryardley: `withBlitz` always returns a function because it uses `withPlugins` when you call the function you flatten the config to an object. So from `newNext` onwards we have an object.

Issue #597, opened 2020-06-01
Blitz warns about api routes in global `/pages` folder


ryardley:
### What is the problem?

We want to make upgrading a next app to blitz super easy. Currently adding blitz to a next app warns about having `/api` routes in the global pages folder. 

This should only warn if you add an api folder to a pages folder within `/app`

### Steps to Reproduce

1. In a blitz app create a global `/pages` folder and add and `/pages/api` route to it. 
1. Try to start Blitz this should not show a warning. 

Note warnings should show if you attempt to add `api` routes to `pages` folders nested under `/app`

### Versions

```
debug: local
debug: pkgPath: /Users/rudiyardley/projects/yatou/yatou/node_modules/@blitzjs/cli 

macOS Catalina | darwin-x64 | Node: v12.16.1

blitz: 0.12.0 (global)
blitz: 0.12.0 (local)
```

### Other
![image](https://user-images.githubusercontent.com/1256409/83370707-84515000-a403-11ea-883d-9c5d78680b1f.png)

flybayer:
As you might have guessed, I don't think we should do this 😅 

Like we discussed in #471, I totally agree on making it easy to migrate Next apps to Blitz apps. However, I don't think we should introduce inconsistencies in Blitz features based on Next-to-blitz migrations.

Supporting `/pages/api` but not `/app/pages/api` is going to be confusing to people who start using Blitz without any knowledge of Next.

We can do all of the migration stuff with a Recipe. We can support running this inside any Next.js project. One command to do everything for you.

```
npx blitz install blitz
```

ryardley:
What is wrong with being compatible with next? If we do this we get huge adoption wins. Next compatibility is pretty much the #1 thing this project has going for it. If you throw it out you loose users. 

Try and put yourself in the shoes of someone who has a large next codebase and want to try blitz. 

Relying on an installer is a really bad idea because: 
* People want to try blitz because it is Next++ 
* It becomes more and more likely you will diverge from an easy manual next migration - we will end up with more and more steps to convert a next app to a blitz app - this is definitely a bad thing!
* If I have to maintain a large codebase the last thing I want to do is run a magic installer on it just to try some of blitz's features.
* People cannot try blitz and change their minds. How do you reverse the installer?
* People cannot adopt blitz incrementally and slowly migrate their routes. Incremental adoption is really important.

There is no technical reason why we can't easily ensure Next compatibility and I am convinced not doing so is a mistake. 

[...]

PR #601, opened 2020-06-01
Do not warn on nested route errors

ryardley:
Closes: #597 

* Do not warn on nested route errors
* Sibling Api folders are still compiled to pages/api like before
* Added missing unit test on duplicate route function
* Added unit test on entire page stage at the stream level
* Fixed bug in displaying code that was displaying too many matches.

### Checklist

- [x] Tests added for changes


ryardley:
There is contention here in #597 so we can’t merge this. Good to have eyes on it.

wKovacs64:
Yep, I'm following that. Just fulfilling my review request in the event you convince Brandon. 😉

ryardley:
Ok adjusted the PR based on #597 to remove the Nested Route check. Please review and I will merge.

ryardley:
ooh @flybayer - you know what I am not sure maybe I will add a test case for that here.

ryardley:
Extra testing has been added. Found a bug where we were displaying all the routes multiple times depending on how many matches we were getting so I rewrote the duplication checking logic and added appropriate tests. Just checking all the windows stuff works then should be ready to merge.

packages/server/src/stages/pages/index.ts
```diff
@@ -17,11 +78,14 @@ export const createStagePages: Stage = ({config, bus, getInputCache}) => {
...
 
-  const stream = through.obj((file: File, _, next) => {
+  const stream: NodeJS.ReadWriteStream = through.obj((file: File, _, next) => {
     const entries = getInputCache().toPaths()
 
+    const duplicates = findDuplicates(entries, fullTransformer)
+
     // Check for duplicate pages entries
-    const duplicatePages = getDuplicatePaths(entries, 'pages')
+    const duplicatePages = filterBy(duplicates, 'pages', 'api')
```
merelinguist, line 87: Nice refactoring!

PR #608, opened 2020-06-02
Rewrite recipe installer framework with Ink

aem:

Rewrites installer architecture to use `ink` to render as a React app, adds better progress messages and interactivity, enforces a propose/commit framework to give users a chance to bail out, renames "installers" to "recipes." 

The previous installer framework worked well, but was a little clunky, and adding new functionality was very tedious. In the new framework, we enforce a common propose/commit framework which makes enabling new recipe executors is a one-line change. This also allows our "UI" to be declarative, just like the rest of Blitz, so the behavior is much more predictable. Eventually I expect to be able to create a common library of `ink` components in `@blitzjs/display` and we could convert our more complex CLI commands to React for added functionality.

As a more minor change, we rename what we used to refer to as `installers` to be called `recipes` so it's a bit more user-friendly, and also disambiguates some terminology in the code. The idea of an `installer` exists, but we now use it just to refer to the Executor that wraps a Recipe. Essentially, you construct an Installer instance for a Recipe. 

Here's the new UI: https://www.dropbox.com/s/udmhpgand5gzy1l/new-installer-renderer.mp4?dl=0

And it leaves your console nice and clean after:
image

I have not yet been able to track down the bug where the last success log is duplicated, but I've been sitting on this for over a week now so I'd love to get it into canary and then we can track that down later

### Checklist

- [ ] Tests added for changes


aem:
Both the UI and the architecture here were **heavily** inspired by the Gatsby team, very appreciative that they took time to sit down with us

flybayer:
Awesome!! Ship it! (once tests pass)

ryardley:
Can't wait to get Ink in this project. :)

Issue #607, opened 2020-06-01
Fix glob in GUI for Windows


merelinguist:
### What is the problem?

The [glob regex](https://github.com/blitz-js/blitz/blob/f53a560b84bea40f9c4e6f0c99a423b4dc10b0de/packages/gui/utils/getDirMtime.ts#L7) I’ve written for calculating directory age is not cross-platform. More on that [here](https://github.com/mrmlnc/fast-glob#how-to-write-patterns-on-windows).

MykalMachon:
I would love to pick this up if no one else is working on it!

merelinguist:
@MykalMachon Super! Go right ahead. 🎆

PR #609, opened 2020-06-02
Fix Glob in GUI for Windows

MykalMachon:
Closes: #607 

updated the getDirMTime function in blitz/gui with a windows specific glob regex, making it functional on Mac & Windows

### Checklist

- [x] Tests added for changes (N/a as this was a bug-fix)
- [x] User facing changes documented (N/a as this was a bug-fix)

### Breaking change: yes/no
no

### Other information
Not anything I can think of! This was a pretty simple bugfix.
I just added a special case for when `process.platform === 'win32'` to update the path/regex string to be windows compatible.


Zeko369:
@MykalMachon Why not use path.resolve here?

MykalMachon:
@Zeko369 just looked into that, and that definitely seems to be a more concise solution as it doesn't introduce platform specific cases. **I'm not super familiar with node's path module; thanks for the tip!**

The reason I used the solution I did is that it was referenced in the initial issue as a solution to the cross platform issue (See the linked glob docs in #607). 

I'll make another commit after work today to resolve this :smile:

MykalMachon:
So after some further testing, it seems that **neither `path.resolve` or `path.normalize` worked on windows as solutions.** 

However, my previous solution that utilizes javascript's `replace()` function did work on both my windows machine and my Linux VM, so I removed the platform-specific check but kept my previous solution. 

I might be doing something wrong here, as you would think the inbuilt node APIs would account for handling the path string properly but it seems that's not the case? **would be more than open to some more feedback here if this still isn't a solid enough solution 😃** 

### Test Output For Reference
Here are the results of the testing I did across approaches and platforms, I generated the glob string on both Windows and Linux and logged the resulting path string and whether it worked or not. 

#### `replace()` function
Windows
` C:/Users/mykal/blitz-projects/tester/{,!(node_modules|.blitz|.git)/**/}*.* (WORKED)` ✔️ 

Linux
` C:\Users\mykal\blitz-projects\tester\{,!(node_modules|.blitz|.git)\**\}*.* (WORKED)` ✔️ 

#### `path.normalize` function
Windows
`C:\Users\mykal\blitz-projects\tester\{,!(node_modules|.blitz|.git)\**\}*.* (DID NOT WORK)` ❌ 

Linux
`C:\Users\mykal\blitz-projects\tester\{,!(node_modules|.blitz|.git)\**\}*.* (WORKED)` ✔️ 

#### `path.resolve` function
Windows
`C:\Users\mykal\blitz-projects\tester\{,!(node_modules|.blitz|.git)**}*.* (DID NOT WORK)` ❌ 

Linux
`C:\Users\mykal\blitz-projects\tester\{,!(node_modules|.blitz|.git)\**\}*.* (WORKED)` ✔️

merelinguist:
Thanks for all the detail @MykalMachon! I just tested this on my Mac, and it worked perfectly too! Great job. 🎆

Issue #610, opened 2020-06-03
Add unit tests on helpers in file-pipeline


ryardley:
### What do you want and why?

We don't have enough testing around the file pipeline. We should add unit tests for each helper in the file pipeline. 

Part of this task is to make it so we can also test the transformFiles function by substituting the fs dependent stages out for mock streams.

Issue #602, opened 2020-06-01
Need a universal stream abstraction for file-pipeline stages


ryardley:
### What do you want and why?

In the file-pipeline when we deal with Stages there is heaps of boiler plate around setting up file streams. It would be far more approachable if we were to abstract away much of the boilerplate

```ts
const stream = through({objectMode: true}, function (file, _, next) {
  if (!File.isVinyl(file) || typeof file === 'string') return next(null, file)

  // stream logic goes here

  next(null, file)
}
```

Secondly this would allow us an easy point at which to monitor and debug the pipeline and create good processes for testing the pipeline.

### Possible implementation(s)

In object streams we pretty much never need to use encoding and there is also the issue of automatic backpressure handling which so far has not surfaced as too much of an issue but is lurking in there when we deal with slow stream transforms. We will likely want to abstract this away so we never have to think about it.

Perhaps we could use something like this:

```ts
const stream = transform((file, {push, next}) => {
  push(file) // same as this.push(file) in a through stream
  next(new Error) // next is the same
  return file // same as next(null, file)
})
```

Maybe there is something simpler.

ryardley:
Thinking maybe we can use an early return:

```ts
```

ryardley:
Maybe add a second parameter to set options:

```ts
const stream = transform((file) => {
  // 
}, {
  objectMode: true,
  highWatermark: 50,
})
```

PR #611, opened 2020-06-03
Add extra unit tests around stage helpers within file-pipeline 

ryardley:
Under construction... ⛑🚧👻👷‍♀️ 

Closes: #610 
Closes: #602 

* File pipeline was missing unit tests to test each component this provides test for:
 
  * Transform Files function  test 
  * Agnostic source
  * Idle handler
  * Work optimizer
  * Writer

* Created stream test utility `testStreamItems`
* Created `transform` abstraction for streams - solving #602 with tests

This also fixes a bug:

* Ready handler was misfiring under certain conditions eg when chokidar and vfs was running slowly / taxing resources. (Still needs more work to be rock solid as it may fire before all files have completed writing to disk although in practice this is not going to be much of a problem because of the way next and webpack work and we can solve this to be rock solid at a later point by introducing better queueing practices such as ordering events)
Prior to this update the ready event was firing once there had been no activity for a certain amount of time. This doesn't work well on filetree/gitignore combinations that slowed down chokidar/vinyl-fs globbing. (possibly symlinks with a gitignore?). To fix this the agnostic source stream was altered so that when the crossover between vinyl file system globbing and chokidar  happened the stream sends down an event which signifies elsewhere in the chain to fire the ready event after the writer.

### Checklist

- [x] Tests added for changes


ryardley:
Gonna try and split this one up as its is a bit much in a single PR

Issue #612, opened 2020-06-03
[cli] `blitz help` fails outside of Blitz projects


wKovacs64:
### What is the problem?

Running `blitz help` (or `-h`, `--help`) outside of a Blitz project tells you to run `blitz help`.

### Steps to Reproduce

1. Run `blitz help` outside of a Blitz project.

### Versions

```
debug: global-linked
debug: pkgPath: C:\dev\contrib\blitz\packages\cli

Windows 10 | win32-x64 | Node: v12.17.0

blitz: 0.12.0 (global)
```

### Other

```
❯ blitz help
You are using alpha software - if you have any problems, please open an issue here:
  https://github.com/blitz-js/blitz/issues/new/choose

 »   Error: You are not inside a Blitz project, so this command won't work.
 »   You can create a new app with blitz new myapp or see help with blitz help
```

PR #613, opened 2020-06-03
Fix 'blitz help' not working from global cli

wKovacs64:
Closes: #612

Adds `help`, `-h`, and `--help` to the universally allowed global commands so you can access the help regardless of whether you're in a Blitz project directory or not.

### Checklist

- [ ] Tests added for changes

### Breaking change: yes/no

no

### Other information

I don't know oclif, so maybe there is a slicker way to include the aliases (`-h`, `--help`) without explicitly listing them?


flybayer:
Thank you @wKovacs64!!

PR #615, opened 2020-06-04
Add universal stream abstraction

ryardley:
Closes: #602 

* This adds a transform creator that simplifies the API for creating streams in the file-pipeline. 
* Adds the ability to choose to make a file transform `transform.file()` that only transforms file objects or a normal transforms that transforms everything. This is required as mostly we only want to transform files. 
* Simplify writer to correctly report that a file has been deleted
* Allows the write stream to be injected to the writer so we can more easily test the stream

### Checklist

- [x] Tests added for changes


merelinguist:
This is awesome!

ryardley:
You happy I merge this @merelinguist ?

Issue #614, opened 2020-06-04
Crashing because files are not being copied to dev folder before next starts under certain conditions


ryardley:
### What is the problem?

Ready handler is misfiring under certain conditions eg when chokidar and vfs was running slowly / taxing resources. As a hack by design the ready callback was firing once there had been no activity for a certain amount of time. This doesn't work well on filetree/gitignore combinations that slowed down chokidar/vinyl-fs globbing. This can happen when using symlinks or big gitignore files which adds to the globbing that has to be done on each file in the folder.

To fix this the agnostic source stream should be altered so that when the crossover between vinyl file system globbing and chokidar happens the stream sends down an event which signifies elsewhere in the chain to fire the ready event after the writer.

For now we can probably get away with just sending a "ready" string down the chain and ensuring it can be passed through but because of the way streams work we may want to do something fancier at a later date.

PR #616, opened 2020-06-04
Fix ready event bug

ryardley:
Closes: #614 

* Ready event was firing too early in certain situations when chokidar and vfs globbing was too slow. (symlinks/large gitignore/low memory/low cpu)
* Refactors the way the ready event works. The ready event is fired when the agnosticinput stream switches from using the vinyl-fs glob to the chokidar watcher instead of after a arbitrary length of pause time. 
* Adds universal transform helper to more stages
* This is still not perfect until we add ordering to the queue so that the event only fires after the previous event has actually been written (the nature of streams means events can be processed out of order). We can do that in a later PR as this gets us close enough for most cases because of the way webpack and next work.

### Checklist

- [x] Tests added for changes


packages/file-pipeline/src/helpers/agnostic-source/agnostic-source.test.ts
```diff
@@ -0,0 +1,49 @@
...
+      if (data === 'ready') {
+        stream.end()
+        expect(log).toEqual(expected)
+        return done()
+      }
+      log.push(data.path)
+    })
+  })
+
+  test.skip('when watching = true', async () => {
```
Skn0tt, line 36: Is there a specific reason this test is skipped?

ryardley: Nice catch!

ryardley:
So in order to run a test that actually uses chokidar (despite closing the watcher properly) it seems we have to increase the available file system watchers in linux. So I have done that here. Please review again. If you are happy I can merge. @flybayer @merelinguist @Skn0tt

Skn0tt:
LGTM, provided that we investigate number of watchers in the future.

PR #631, opened 2020-06-06
Fix `blitz start -p 1234` not throwing an error if port 1234 is already in use

### LINKED ISSUE #624: Remove `--hostname` and `--port` from `blitz build`
### What is the problem?

The `blitz build` command current accepts `--hostname` and `--port` but these are invalid because blitz build does not start a server. It only compiles code.

Here's the file where these need to be removed from: https://github.com/blitz-js/blitz/blob/canary/packages/cli/src/commands/build.ts

### LINKED ISSUE #630: When specifying a server port that's already in use, blitz will pick another port instead of throwing an error
### What is the problem?

When specifying a server port like `blitz start -p 4000`, blitz should throw an error instead of picking another port.

The automatic available port detection should only be used if a specific port is not specified by the user.

The offending line of code is here: https://github.com/blitz-js/blitz/blob/canary/packages/server/src/next-utils.ts#L40

This will require a change to how we handle the `-p` argument which happens here: https://github.com/blitz-js/blitz/blob/canary/packages/cli/src/commands/start.ts#L17

Currently, `config.port` is always set, so `next-utils.ts` has no way to know whether to use automatic port detection or not.

So maybe we need to manually parse the input arguments here to see if the user specified a port or not: https://github.com/blitz-js/blitz/blob/canary/packages/cli/src/commands/start.ts#L31

### Steps to Reproduce

1. Start a server on port 3000
2. `blitz start -p 3000` -> server started on 3001 instead of throwing an error

### Versions

```
0.12.0
```

ganeshmani:
This PR closes the issue of having `port` and `hostname` flags in `build` command. 

Closes #624 
Closes #630 

 it changes the build command in `cli/build`  by removing `port` and `hostname` flags. it also changes the `next-utils` in `server` config to accept `undefined` values for flags `port` and `hostname`, adding `non null assertion`

### Checklist

- [x] Tests added for changes

### DIFF
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -1,4 +1,4 @@
-import {Command, flags} from "@oclif/command"
+import {Command} from "@oclif/command"
 import {build} from "@blitzjs/server"
 import {runPrismaGeneration} from "./db"
 
@@ -6,26 +6,11 @@ export class Build extends Command {
   static description = "Create a production build"
   static aliases = ["b"]
 
-  static flags = {
-    port: flags.integer({
-      char: "p",
-      description: "Set port number",
-      default: 3000,
-    }),
-    hostname: flags.string({
-      char: "H",
-      description: "Set server hostname",
-      default: "localhost",
-    }),
-  }
-
   async run() {
-    const {flags} = this.parse(Build)
+    // const {flags} = this.parse(Build)
 
     const config = {
       rootFolder: process.cwd(),
-      port: flags.port,
-      hostname: flags.hostname,
     }
 
     try {
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -7,8 +7,6 @@ export class Build extends Command {
   static aliases = ["b"]
 
   async run() {
-    // const {flags} = this.parse(Build)
-
     const config = {
       rootFolder: process.cwd(),
     }
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -7,6 +7,8 @@ export class Build extends Command {
   static aliases = ["b"]
 
   async run() {
+    // const {flags} = this.parse(Build)
+
     const config = {
       rootFolder: process.cwd(),
     }
--- a/packages/server/src/next-utils.ts
+++ b/packages/server/src/next-utils.ts
@@ -35,7 +35,7 @@ async function createCommandAndPort(config: ServerConfig, command: string) {
 
   if (config.port) {
     availablePort = await detect({port: config.port})
-    spawnCommand = spawnCommand.concat(["-p", `${config.port}`])
+    spawnCommand = spawnCommand.concat(["-p", `${availablePort}`])
   }
   if (config.hostname) {
     spawnCommand = spawnCommand.concat(["-H", `${config.hostname}`])
@@ -52,21 +52,17 @@ export async function nextStartDev(
   config: ServerConfig,
 ) {
   const transform = createOutputTransformer(manifest, devFolder).stream
-  const {spawnCommand, availablePort} = await createCommandAndPort(config, "dev")
+  const {spawnCommand} = await createCommandAndPort(config, "dev")
 
   return new Promise((res, rej) => {
-    if (availablePort && availablePort !== config.port) {
-      rej(`Couldn't start server on port ${config.port}::port already in use`)
-    } else {
-      spawn(nextBin, spawnCommand, {
-        cwd,
-        stdio: [process.stdin, transform.pipe(process.stdout), transform.pipe(process.stderr)],
+    spawn(nextBin, spawnCommand, {
+      cwd,
+      stdio: [process.stdin, transform.pipe(process.stdout), transform.pipe(process.stderr)],
+    })
+      .on("exit", (code: number) => {
+        code === 0 ? res() : rej(`'next dev' failed with status code: ${code}`)
       })
-        .on("exit", (code: number) => {
-          code === 0 ? res() : rej(`'next dev' failed with status code: ${code}`)
-        })
-        .on("error", rej)
-    }
+      .on("error", rej)
   })
 }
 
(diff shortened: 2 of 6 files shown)


packages/cli/src/commands/build.ts
```diff
@@ -1,31 +1,14 @@
...
-      description: 'Set port number',
-      default: 3000,
-    }),
-    hostname: flags.string({
-      char: 'H',
-      description: 'Set server hostname',
-      default: 'localhost',
-    }),
-  }
-
```
flybayer, line 21: Good, this needed removed (unused)

packages/server/src/next-utils.ts
```diff
@@ -37,10 +37,10 @@ export async function nextStartDev(
   config: ServerConfig,
 ) {
   const transform = createOutputTransformer(manifest, devFolder).stream
-  const availablePort = await detect({port: config.port, hostname: config.hostname})
+  const availablePort = await detect({port: config.port!, hostname: config.hostname!})
```
ryardley, line 40: We don't know if port or hostname is set here. Probably a bit dangerous to use the non-null assertion operator (`!`). Be best if we check for undefined values and adjust accordingly or possibly tidy up the config to abstract away the differences between dev and build. Same as below

ganeshmani: @ryardley got it. it makes sense it add a conditional declaration rather than using an assertion operator. I will make the change. Thanks for the feedback

ganeshmani:
> Thanks for working on this!
> 
> However, even with this change, `config.port` will always be set because of `default: 3000` in `commands/start.ts`. And so automatic port detection still being used even when the user specifies a port.
> 
> So you need to make a change to `start.ts`, similar to how I describe in the original issue.
> 
> And I think also, the `availablePort` code should only be conditionally used if a port isn't specified.

@flybayer what happens if we run `blitz start` command without specifying port and hostname. for that, don't we need a default port and hostname for that?

flybayer:
@ganeshmani we can simply let Next.js handle the defaults, so we don't have to worry about it

packages/server/src/next-utils.ts
```diff
@@ -65,13 +77,31 @@ export function nextBuild(nextBin: string, cwd: string) {
...
+  return new Promise((res, rej) => {
+    if (availablePort && availablePort !== config.port) {
+      rej(`Couldn't start server on port ${config.port} ::port already in use`)
+    } else {
+      spawn(nextBin, spawnCommand, {
+        cwd,
+        stdio: 'inherit',
+      })
+        .on('exit', (code: number) => {
+          code === 0 ? res() : rej(`'next build' failed with status code: ${code}`)
```
ryardley, line 99: This probably could be changed from `next build` to `next start`

ryardley, line 99: Nice catch to add an exit handler.

packages/server/src/next-utils.ts
```diff
@@ -65,13 +77,31 @@ export function nextBuild(nextBin: string, cwd: string) {
...
+
+  return new Promise((res, rej) => {
+    if (availablePort && availablePort !== config.port) {
+      rej(`Couldn't start server on port ${config.port} ::port already in use`)
+    } else {
+      spawn(nextBin, spawnCommand, {
+        cwd,
+        stdio: 'inherit',
+      })
+        .on('exit', (code: number) => {
```
ryardley, line 98: Nice catch to add an exit hander!

packages/server/src/next-utils.ts
```diff
@@ -65,13 +77,31 @@ export function nextBuild(nextBin: string, cwd: string) {
...
-  )
+  let spawnCommand: string[] = ['start']
+  let availablePort: number
+  if (config.port) {
+    availablePort = await detect({port: config.port})
+    spawnCommand = spawnCommand.concat(['-p', `${config.port}`])
+  }
+  if (config.hostname) {
+    spawnCommand = spawnCommand.concat(['-H', `${config.hostname}`])
+  }
```
ryardley, line 88: There is an optional opportunity here to consolidate adding the port and hostname into a function that is shared between the start and dev functions.

flybayer:
@ganeshmani awesome, this is looking good!

One last issue I see is that the availablePort is not being used.

1. Start a server on port 3000
2. Run `blitz start` and it will try to use port 3000 and then fail.

What should happen is this:

1. Start a server on port 3000
2. Run `blitz start` 
3. Blitz starts on next available port (3001 for example)

and this:

1. Start a server on port 3000
2. Run `blitz start -p 3000` 
3. Blitz fails to start

ganeshmani:
> @ganeshmani awesome, this is looking good!
> 
> One last issue I see is that the availablePort is not being used.
> 
> 1. Start a server on port 3000
> 2. Run `blitz start` and it will try to use port 3000 and then fail.
> 
> What should happen is this:
> 
> 1. Start a server on port 3000
> 2. Run `blitz start`
> 3. Blitz starts on next available port (3001 for example)
> 
> and this:
> 
> 1. Start a server on port 3000
> 2. Run `blitz start -p 3000`
> 3. Blitz fails to start

Ah Okay. i thought we needed to fix that. https://github.com/blitz-js/blitz/issues/630

flybayer:
`blitz start` should never throw an error.

`blitz start -p 3000` should throw an error if port 3000 is already in use

The original issue is that `blitz start -p 3000` did not throw an error.

That issue is fixed, but currently there's a regression with `blitz start` :)

ganeshmani:
Got it. Let me fix that one. :-)

ganeshmani:
@flybayer  Fixed. let me know if there is anything can be improved

flybayer:
@ganeshmani sorry for the delay in testing this!

Unfortunately it's still not working right. Two problems:

1. Start another server on port 3000
2. Run `blitz start`.
3. ❌ What happens: Blitz starts on port 3000 (should start on port 3001)

.

1. Start another server on port 3000
2. Run `blitz start -p 3000`.
3. ❌ What happens: Blitz starts on port 3001 (should throw an error that port is in use)

flybayer:
Thank you so much @ganeshmani!!

PR #633, opened 2020-06-06
Do some yarn workspace maintenance

flybayer:
This cleans up a few things.

- Move all devDependencies to root
- Removes unused git hooks from sub-packages
- Specifies lint directory to remove tsdx warnings


merelinguist:
Nice! I wonder if there’s an opportunity here too for standardising some of the pkg.json fields?

flybayer:
@merelinguist sure, what were you thinking?

merelinguist:
Okay admittedly this isn’t as bad as I thought! Never mind about it

Issue #634, opened 2020-06-06
Re-export some missing items from Next


flybayer:
### What do you want and why?

We need to re-export `Error` from `next/error`. This got overlooked previously.

Also need to re-export `AppProps`, `InferGetStaticPropsType` and `InferGetServerSidePropsType` from `next`

### Possible implementation(s)

Add the re-exports in this file: https://github.com/blitz-js/blitz/blob/canary/packages/core/src/index.ts

---

This is a very simple change, so Blitz contributor veterans, please leave this for someone new to blitz or new to open-source contributing :)

gvasquez11:
Hey there I would love to take this issue if possible :)

flybayer:
@gvasquez11 yes, for sure! Let me know if you need help :)

gvasquez11:
@flybayer I looked at the 'The Contributing Guide' in order to try and find how to test on my local machine since I had some failed cases :( but the link didn't work. Any advice would be much appreciated :)

flybayer:
@gvasquez11 oops, here it is: https://blitzjs.com/docs/contributing

PR #636, opened 2020-06-06
Add missing re-export items from Next.js

gvasquez11:
Closes: #634 

re-export Error from next/error
re-export AppProps, InferGetStaticPropsType and InferGetServerSidePropsType from next
### Checklist

- [ ] Tests added for changes


packages/core/src/index.ts
```diff
@@ -34,3 +34,11 @@ export {
...
 
 export {default as dynamic} from 'next/dynamic'
+
+export {Error} from 'next/error'
+
+export {AppProps} from 'next'
+
+export {InferGetStaticPropsType} from 'next'
+
+export {InferGetServerSidePropsType} from 'next'
```
flybayer, line 44: Great, looking good!

One small change that would be good is to move these exports to around line 15 where other items are being exported from `next`. That way `from 'next'` only appears in this file once. Make sense?

gvasquez11: yeah make sense, got it :) will get this done, thank you!

gvasquez11:
Currently the only error I'm receiving is 

> semantic error TS2300: Duplicate identifier 'AppProps' 

Do you still want me to add that in the export from 'next' as mentioned before, I noticed that on line 21 it's also used but its being exported from 'next/app'

flybayer:
Oh yes, you're right! I overlooked that — AppProps is already being exported, so don't add it a second time :)

gvasquez11:
Awesome! 
One last thing, when trying to re-export 'Error' from 'error/next' I receive:

> semantic error TS2614:  has no exported member 'Error'

I suspect that 'Error' may not be reading from the .next file properly, unless I made a silly typo.

packages/core/src/index.ts
```diff
@@ -34,3 +36,5 @@ export {
 } from 'next/document'
 
 export {default as dynamic} from 'next/dynamic'
+
+export {Error} from 'next/error'
```
flybayer, line 40: This should fix it
```suggestion
export {default as Error} from 'next/error'
```

flybayer:
Looks like something goofed with your editor. Lint is failing because it's missing a new line character at the end.

Adding a new line at the end of the file should fix that

flybayer:
Woohoo!

@all-contributors add @gvasquez11 for code

gvasquez11:
> Woohoo!
> 
> @all-contributors add @gvasquez11 for code

Yay! Finally got it,
Thank you so much for letting me contribute even if it was a small part :)

Issue #620, opened 2020-06-04
Open the GUI with a CLI command


merelinguist:
### What do you want and why?

We need to bundle the GUI with the CLI, and provide a command to open it with `blitz gui`.

cc/ @peaonunes

peaonunes:
Nice to have it as a separate issue! Can you put me as assignee ? I'm afraid I cannot assign myself.

peaonunes:
I think we can charge it to be not ready to work. Right?

flybayer:
Yes, icebox :)

PR #644, opened 2020-06-07
Start gui cli command

peaonunes:
Closes: #620

Adds a `cli` command to start the Blitz `gui` app.

### Checklist
- [x] Tests added for changes

### Question

Added the new command and I can run `blitz gui` using the linked cli. However, after I do `blitz new` it will not make that new command available in the project just created. I believe that's because the project will use the npm `cli` version, right?

Not sure how to workaround this problem. I am currently not able to test `blitz gui` inside a newly created project because of this limitation so I cannot guarantee it is working fine 🤔 .

Curious on everyone's thoughts/tips on that.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -31,6 +31,7 @@
     "/lib"
   ],
   "dependencies": {
+    "@blitzjs/gui": "0.13.0",
     "@blitzjs/display": "0.13.0",
     "@blitzjs/repl": "0.13.0",
     "@oclif/command": "1.5.20",
--- a/packages/cli/src/check-before-running.ts
+++ b/packages/cli/src/check-before-running.ts
@@ -3,11 +3,11 @@ import chalk from 'chalk'
 
 import {isBlitzRoot, IsBlitzRootError} from './utils/is-blitz-root'
 
-const whitelistGlobal = ['-h', '--help', 'help', 'new']
+const globalCommands = ['-h', '--help', 'help', 'new', 'gui']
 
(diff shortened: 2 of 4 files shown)


packages/cli/package.json
```diff
@@ -31,6 +31,7 @@
     "/lib"
   ],
   "dependencies": {
+    "@blitzjs/gui": "0.13.0",
```
peaonunes, line 34: Even though I added it as a cli's dependency it is not working yet. When I run `blitz gui` the `spawn` is unable to find the package, it should be carried over during installation.
```bash
blitz gui                 
You started the Blitz GUI
{
  error: Error: spawnSync @blitzjs/gui ENOENT
  ...
```
So there must be something else I'm missing here 🤔

flybayer:
@peaonunes here's how I test something like this:

1. In blitz repo root run `yarn blitz new test-app`
2. `mv test-app examples/test-app`
3. `cd examples/test-app`
4. `yarn`
5. `blitz start`

We should add that as dev script or something.

Also, we need to ensure the `gui` command is hidden from `blitz help` until the gui is ready for people to try using it.

peaonunes:
Hey @flybayer thanks for the tip. Gonna try that tomorrow. 

I thought the gui was already in a stage we could show up, I’ll change it to be a hidden command then!

merelinguist:
@peaonunes Sorry, definitely not! It’s still very experimental.

peaonunes:
We are delaying this work until we have version ready to publish.

PR #651, opened 2020-06-07
Fix pipeline build error

merelinguist:
Screenshot 2020-06-07 at 20 18 55

See also https://github.com/Microsoft/TypeScript/issues/30128


merelinguist:
@ryardley Suddenly this is all working for me again. Not sure what’s going on but sorry about the noise

ryardley:
Yeah I always thought it’s funny how it’s not actually part of JavaScript. Maybe it is better to just use an any type in this case? Saves some head scratching/keystrokes later potentially?

PR #652, opened 2020-06-08
Add HTTP Middleware 🎉

flybayer:

Add HTTP Middleware!! 🎉  [Read the new docs](https://blitzjs-com-git-middleware.blitzjs.vercel.app/docs/middleware)

[Originally I thought](https://github.com/blitz-js/blitz/discussions/326#discussioncomment-23149) I could support both the connect/express api **and** the async `next()` function like Koa. However, I spent way too long trying to make that work before finally giving up 😅 

So now connect compatibility is provided via the `connectMiddleware` wrapper. I still think it's somehow possible to not need this wrapper. Maybe someone else can investigate. The core issue is that many connect middleware don't `return next()` and so then you're left with a dangling promise that resolves early and short circuits the chain. You can wrap everything in a promise, but then you have a promise inside a promise which I couldn't get to work either.

Also:

- [x] Refactor the server rpc stage with updated naming and change the api route to use the isomorphic client
- [x] Rename the special `_rpc` folder inside `.blitz` to `_resolvers` for accuracy
- [x] Enhance isomorphic rpc client with name, type, path, and of course middleware
- [x] Fix issue in dev about `installer` package types not found

### Future Considerations

I think we should also add this middleware support to API routes, but someone else can add that later.

### Checklist

- [x] Tests added for changes
- [x] PR submitted to [blitzjs.com](https://github.com/blitz-js/blitzjs.com) for any user facing changes: https://github.com/blitz-js/blitzjs.com/pull/80


flybayer:
@all-contributors add @exclipy for docs

MrLeebo:
I find the `finishRequest` option alarming. I would expect that if the middleware chain doesn't explicitly end the response then that would be a programming error and `finishRequest: true` is just going to shadow that mistake. Am I misreading that?

flybayer:
@MrLeebo yes I think you are right. I originally erroneously believed that plain api routes were automatically ended for you, but that's not true.

I'll remove the `finishRequest` option.

packages/core/src/ssr-query.ts
```diff
@@ -1,20 +1,41 @@
...
 ): Promise<ReturnType<T>> {
-  const data = await queryFn(params)
+  const handler = (queryFn as unknown) as EnhancedResolverModule
 
-  return data
+  const middleware = getAllMiddlewareForModule(handler)
+
+  middleware.push(async (_req, res, next) => {
+    const logPrefix = `${handler._meta.name}`
+    console.log('') // New line
```
MrLeebo, line 25: Should be @blitzjs/display log(''), right?

packages/server/src/rpc.ts
```diff
@@ -25,31 +40,36 @@ export function rpcHandler(
       if (typeof req.body.params === 'undefined') {
         const error = {message: 'Request body is missing the `params` key'}
         log.error(`${logPrefix} failed: ${JSON.stringify(error)}\n`)
-        return res.status(400).json({
+        res.status(400).json({
           result: null,
           error,
         })
+        return next()
```
MrLeebo, line 47: Should these `next()` calls contain the `{result, error}` parameter in case the RPC has custom middleware that needs to react to the result of the RPC call?

packages/core/src/middleware.ts
```diff
@@ -0,0 +1,177 @@
...
+    middleware.push(...resolverModule.middleware)
+  }
+  return middleware
+}
+
+export async function handleRequestWithMiddleware(
+  req: BlitzApiRequest | IncomingMessage,
+  res: BlitzApiResponse | ServerResponse,
+  middleware: Middleware | Middleware[],
+  opts: {finishRequest: boolean} = {finishRequest: true},
```
MrLeebo, line 52: See my earlier comment about this.

examples/store/app/products/queries/getProducts.ts
```diff
@@ -27,3 +29,12 @@ export default async function getProducts({
...
 }
+
+export const middleware: Middleware[] = [
+  async (req, res, next) => {
+    await next()
+    if (req.method !== "HEAD" && Array.isArray(res.blitzResult)) {
+      console.log("[Middleware] Total product count:", res.blitzResult.length, "\n")
+    }
+  },
+]
```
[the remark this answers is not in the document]
flybayer: @MrLeebo did you see [the new docs](https://blitzjs-com-git-middleware.blitzjs.vercel.app/docs/middleware)? Almost this exact code is included there.

If you did see it, then maybe the docs need improved?

packages/server/src/rpc.ts
flybayer: Also on this, did you see [the docs](https://blitzjs-com-git-middleware.blitzjs.vercel.app/docs/middleware)?

`result` is available via `res.blitzResult`

packages/core/src/middleware.ts
flybayer: fixed

packages/core/src/ssr-query.ts
flybayer: fixed

packages/core/src/middleware.ts
```diff
@@ -0,0 +1,177 @@
...
+    await handler(req as MiddlewareRequest, res as MiddlewareResponse, (error) => {
+      if (error) {
+        throw error
+      }
+    })
+    if (opts.finishRequest && !res.writableEnded) {
+      res.end()
+    }
+  } catch (error) {
+    console.log('') // new line
```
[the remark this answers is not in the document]
flybayer: fixed

examples/store/app/products/queries/getProducts.ts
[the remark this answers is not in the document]
MrLeebo: I didn't see it, thanks for the link. I was just worried that the Store example was turning into a scratch pad for new Blitz features whether they belong there or not

Issue #332, opened 2020-04-29
Create `@blitzjs/config` package


ryardley:
### What do you want and why?

We need a central projectwide place to load user configuration. 

Basically when we start any cli process our code should have access to project configuration inside the blitz.config.js 

If using an npm module helps set us up to have multiple ways to define our configuration then that makes sense as a future step but it should be limited to `blitz.config.js` initially as there are ramifications for expectations in the build rules in the server compiler.

It should also provide a way to set defaults within the package

### Possible implementation(s)

Ideal API might be something like this: 

```ts
import {serverless} from '@blitzjs/config'
```

But this may not be possible so adding some kind of asynchronicity is fair. 

```ts
import {getConfig} from '@blitz/config'
const {serverless} = await getConfig();
```

https://github.com/davidtheclark/cosmiconfig

NOTE: We might be edging towards avoiding tsdx and using tsc to build out packages. see [discussion](https://github.com/blitz-js/blitz/issues/306).

Initially this can probably be a node only package

Currently config has been accessed here: https://github.com/blitz-js/blitz/pull/159/files/c6d8afd222a6a0b15aa317fe4a74e156fedf0e04#diff-3ee0da0e1fb5e753344f3e2a579587bfR17 and if that is merged before this is done we need to ensure this uses the same mechanism.

Zeko369:
I can add this and then merge #159

PR #653, opened 2020-06-09
Add @blitzjs/config package for internal use

Zeko369:
Closes: #332

Add getConfig function to pull `next.config.js` and `blitz.config.js` and cache them (so you only import them once)

```ts
import {getConfig} from '@blitz/config'
const {serverless} = await getConfig();
```

### TODO:
- [ ] Add tests


flybayer:
Thank you @Zeko369 !!

Issue #647, opened 2020-06-07
Prisma isn't being set to the latest version during `blitz new`


flybayer:
### What is the problem?

During `blitz new` we fetch the latest version for all dependencies and update package.json to match that.

However, this is broken for Prisma.

### Steps to Reproduce

1. `blitz new myapp`
2. Open `myapp/package.json`
3. See this the below which is not the latest version

```
    "@prisma/cli": "2.0.0-beta.3",
    "@prisma/client": "2.0.0-beta.3",
```

### Versions

```
0.13.0
```

### Extra  Info

The issue is this line: https://github.com/blitz-js/blitz/blob/canary/packages/generator/src/utils/fetch-latest-version-for.ts#L22

We can allow an exception for that regex for `@prisma/*`

After you make the change, you can test the cli locally like this:

1. `yarn global remove blitz`
2. Run inside blitz repo root: `yarn link-cli`
3. Run somewhere else outside the blitz repo: `blitz new myapp`

josemiguelo:
Hi @flybayer.

Is the following statement appropriate for the purpose of the fix?

``` javascript
if (!version.match(/\d.x/) && !dep.match(/@prisma\/\w+/)) skipFetch = true
```

flybayer:
@josemiguelo I think so! You can test it to make sure :)

josemiguelo:
I already did so :) Just making sure it does make sense to put both regex on the same statement.
Can I be assigned to this? I'd love to contribute

flybayer:
Ok awesome. And yes for sure! You can submit a PR for this. Let me know if you need any help with that :)

flybayer:
Prisma just released [`2.0.0`](https://github.com/prisma/prisma/releases/tag/2.0.0) which totally eliminates this problem 🤷 :)

PR #654, opened 2020-06-09
Fix latest version prisma when running blitz new

josemiguelo:
Closes:  #647 

A new regex's been added in such a way that when running the generator using blitz new, new remote versions are fetched for prisma's packages.

When fetching latest versions of dependencies, major versions of Prisma's packages were not  detected properly since passed "template version" didn't match `/\d.x/` but `2.0.0-beta-*` instead. 

### Checklist

- [ ] Tests added for changes

### DIFF
--- a/packages/generator/src/utils/fetch-latest-version-for.ts
+++ b/packages/generator/src/utils/fetch-latest-version-for.ts
@@ -19,7 +19,7 @@ export const fetchLatestVersionsFor = async <T extends Record<string, string>>(
       async ([dep, version]): Promise<[string, string]> => {
         let skipFetch = false
 
-        if (!version.match(/\d.x/)) skipFetch = true
+        if (!version.match(/\d.x/) && !dep.match(/@prisma\/\w+/)) skipFetch = true
 
         // We pin experimental versions to ensure they work, so don't auto update experimental
         if (version.match(/experimental/)) skipFetch = true
--- a/packages/generator/src/utils/get-latest-version.ts
+++ b/packages/generator/src/utils/get-latest-version.ts
@@ -17,7 +17,8 @@ export const getLatestVersion = async (
   dependency: string,
   templateVersion: string = '',
 ): Promise<Fallbackable<string>> => {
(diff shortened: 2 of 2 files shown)


osirvent:
Right now Prisma is beta9. At least on [Beta 7](https://github.com/prisma/prisma/releases/tag/2.0.0-beta.7) there is a change that gonna break scaffolding, the Client pagination.

packages/generator/src/utils/get-latest-version.ts
```diff
@@ -17,7 +17,8 @@ export const getLatestVersion = async (
   dependency: string,
   templateVersion: string = '',
 ): Promise<Fallbackable<string>> => {
-  const major = templateVersion.replace('.x', '')
+  const toBeReplaced = templateVersion.substring(templateVersion.indexOf('.'))
```
donni106, line 20: could this be renamed to something like `minorPatch`, which says more about the value and fits more to the next line, where the major version is extracted from the whole version string: `const major = templateVersion.replace(minorPatch, '')`?

packages/generator/src/utils/fetch-latest-version-for.ts
```diff
@@ -19,7 +19,7 @@ export const fetchLatestVersionsFor = async <T extends Record<string, string>>(
       async ([dep, version]): Promise<[string, string]> => {
         let skipFetch = false
 
-        if (!version.match(/\d.x/)) skipFetch = true
+        if (!version.match(/\d.x/) && !dep.match(/@prisma\/\w+/)) skipFetch = true
```
donni106, line 22: could this check be described by a short comment, saying why `@prisma` stuff is handled explicitly?

flybayer:
@josemiguelo thank you for this!

But guess what... Prisma just released [`2.0.0`](https://github.com/prisma/prisma/releases/tag/2.0.0) so we don't need this anymore 😅

flybayer:
@all-contributors add @josemiguelo for code

PR #655, opened 2020-06-09
Update query/mutation templates for Prisma 2.0.0

merelinguist:

Updates the generator code to use the new style of pagination in Prisma.


flybayer:
Thanks Dylan!!

PR #243, opened 2020-04-25
Fix top level await in console cross platform

Zeko369:
### Pull Request Type

- [x] Feature
- [x] Bug fix
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Tests related changes
- [ ] Other (please describe): Dependency updates

### Checklist
- [x] Get it working on Windows
- [x] Test on real linux (not WSL)
- [x] Update CLI text then -> await, remove await missing line

### Works on
- [x] macOS Catalina
- [x] Windows 10 Pro 1903 - native (powershell and node.exe binary)
- [x] Windows 10 Pro 1903 - git bash 
- [x[ Windows 10 Pro 1903 - WSL (ubuntu, Linux DESKTOP-R287K1V 4.4.0-18362-Microsoft #476-Microsoft)
- [x] Ubuntu 19.04

Closes: #230

Issue #230, opened 2020-04-25
Add top level await support for blitz console


flybayer:
**What do you want and why do you want it?.**

We were enabling top level await in our binaries using this:
```
#!/usr/bin/env node --experimental-repl-await
```

But that [doesn't work on Linux](https://github.com/nxtep-io/ts-framework-cli/issues/1), which also breaks production deployments, like to Vercel/Zeit. So I had to remove it for now.

We need to add this back in some other fully cross platform way.

**Possible implementation(s)**

I already tried the same approach as https://github.com/nxtep-io/ts-framework-cli/pull/10 . This works good on macOS and Linux, but breaks on windows since windows doesn't have bash.

So I'm sort of stumped at how to do this. Maybe create a separate bin for the repl that we can spawn from the console command?

Zeko369:
Looking into this

flybayer:
Someone else is welcome to try and add this cross-platform.

PR #659, opened 2020-06-09
Fix top level await in `blitz console` 🎉

wKovacs64:
Closes: #230
Closes: #243

Spawns a separate `node` process for CLI commands, explicitly passing `--experimental-repl-await` to the new child process.

This currently spawns a new process for every command, when it might only be useful for `console` (at this time). We could add some logic that checks to see if the command is `console` and only spawn the new process in that case, but I'm not sure if the added noise is worth it.

### Checklist

- [ ] Tests added for changes

I've manually tested this with Node 12 on macOS Catalina, Windows 10, and Linux Mint Tessa.

I can open a PR to the docs site to replace all the `.then` references with `await` syntax if we decide to merge this.


flybayer:
Awesome!!! Looks great. I can't test it at the moment, but I'll try to a bit later. Maybe @aem can verify.

aem:
@flybayer I can't merge since this touches a file outside of the packages I'm codeowner for, but it's good to go

wKovacs64:
Cool. I'll work on updating the examples in the docs tomorrow.

flybayer:
@wKovacs64 that'd be great, thank you!

Issue #600, opened 2020-06-01
BUG on generated routes with --context


osirvent:
### What is the problem?
bug on import generated route for queries and mutations when `--context` is used

### Steps to Reproduce

1. `blitz g all project --context=marketing`
2. 
<img width="641" alt="imagen" src="https://user-images.githubusercontent.com/5927133/83393819-bbb9ff80-a3f7-11ea-895f-762e56b8e3fd.png">
3.
<img width="609" alt="imagen" src="https://user-images.githubusercontent.com/5927133/83394022-1fdcc380-a3f8-11ea-93de-2b5d6de865a0.png">

### Versions

```
debug: local
debug: pkgPath: /Users/osirvent/dev/blitzjs/tests/test2/node_modules/@blitzjs/cli 

macOS Catalina | darwin-x64 | Node: v12.17.0

blitz: 0.12.0 (global)
blitz: 0.12.0 (local)
```

### Other

Please include applicable logs and screenshots that show your problem.

flybayer:
Uh oh! Thank you for opening an issue! :)

peaonunes:
Happy to took at it.

flybayer:
@peaonunes you bet!

PR #669, opened 2020-06-12
Fix page generator with custom context

peaonunes:
Closes: #600 

Fixes the page templates and the generations.
The paths on templates should be different from the `modelNames` since the generated `queries` and `mutations` may have additional path of `context` included.
Added a new template value in the pages to refer to import path of the model considering the context.

### Checklist
- [ ] Tests added for changes


packages/generator/src/generators/page-generator.ts
```diff
@@ -41,12 +41,17 @@ export class PageGenerator extends Generator<PageGeneratorOptions> {
       modelNames: this.options.modelNames,
       ModelName: this.options.ModelName,
       ModelNames: this.options.ModelNames,
+      importModelNames: this.getPathWithContext(),
```
peaonunes, line 44: Happy to change to a better name could not think about any other atm.

packages/cli/src/commands/generate.ts
```diff
@@ -173,18 +173,27 @@ export class Generate extends Command {
     }
   }
 
-  getModelNameAndContext(modelName: string, context?: string): {model: string; context: string} {
+  getModelNameAndContext(modelName: string, context?: string): {model: string; context?: string} {
```
peaonunes, line 176: Found a bug here while testing the different input values.
Passing not setting the `context` was leading to `contextSegments.= ''` here. However, if we call `path` with empty string we gotta a wrong path:
```ts
console.log(path.join(''))
> .
```
So I noticed a wrong it would generate the wrong import on the newly `__importModelNames` I've added. Since the `options.context` passed down to `generators` would be `.`. That potentially would affect other parts of the code as well.
Screen Shot 2020-06-12 at 20 49 15

packages/generator/src/generators/page-generator.ts
flybayer: What about `modelNamesPath`?

flybayer:
Awesome, thank you so much!! :) I'll let @aem do a full review

packages/generator/src/generators/page-generator.ts
peaonunes: Changed here 783027e

packages/cli/src/commands/generate.ts
```diff
@@ -173,18 +173,27 @@ export class Generate extends Command {
...
-    const contextSegments = (context || '').split(/[\\/]/)
+
     if (modelSegments.length > 1) {
       return {
         model: modelSegments[modelSegments.length - 1],
         context: path.join(...modelSegments.slice(0, modelSegments.length - 1)),
       }
     }
+
+    if (!!context && context !== '') {
```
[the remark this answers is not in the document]
peaonunes: TS seems to think `Boolean` constructor is not equivalent to use `!!`. It's complaining.
image
So I changed to be `!= null and != undefined` explicitly here 4f973554395b6d5e5574a1ac90098763863dd019

packages/cli/src/commands/generate.ts
```diff
@@ -173,18 +173,27 @@ export class Generate extends Command {
...
-    const contextSegments = (context || '').split(/[\\/]/)
+
     if (modelSegments.length > 1) {
       return {
         model: modelSegments[modelSegments.length - 1],
         context: path.join(...modelSegments.slice(0, modelSegments.length - 1)),
       }
     }
+
+    if (context !== undefined && context !== null && context !== '') {
```
aem, line 186: This is good if you're happy with this solution! Two other options that are a little shorter:

1. `if (context != null && context !== '')`
2. `if (Boolean(context)) (context as string).split(/[\\/]/)`

We can either take advantage of `!= null` checking for both `null` and `undefined` but no other falsy values, or the fact that we're a little smarter than the TS compiler and just do a manual cast to `string` since we know the value can't be `undefined` at that point

peaonunes: Good point! Happy to go with 2. Done here a8ca8fd

Issue #672, opened 2020-06-12
Add a spinner during `blitz new` when prettier is running


flybayer:
### What do you want and why?

Currently there's a awkward pause during `blitz new` after dependencies are installed. It makes the user thing something is broken.

However, it's just prettier running. So we need to add a spinner so the user knows what is happening.

### Possible implementation(s)

[Right here is the code where prettier runs](https://github.com/blitz-js/blitz/blob/canary/packages/generator/src/generators/app-generator.ts#L152)

You need to add something like this:

```ts
const spinner = log.spinner(log.withBrand('Formatting your code')).start()

// on success
spinner.succeed()

// on error
spinner.fail(chalk.yellow.bold("We had an error running Prettier, but don't worry your app will still run fine :)"))
```

---

Veterans, please leave this for someone new to Blitz or open-source :)

fnoah:
@flybayer I'd like to work on this as a first issue in order to get to know the Blitz codebase.

flybayer:
@fnoah awesome, it's all yours! :)

PR #680, opened 2020-06-13
Add spinner for Prettier formatting during 'blitz new'

fnoah:
Closes: #672

Added a new spinner for prettier formatting on `blitz new`. I also removed the error that was previously thrown which prevented the `Your new Blitz app is ready! Next steps:` message from appearing. I think it makes sense to still show this message since a prettier failure doesn't prevent the app from running.

Example of a successful run:
image

I used the error message copy which @flybayer suggested in the issue.  I wasn't sure which styling to use for the error message (since it's an error, but nothing fatal). The current commit uses the first one of the following examples, but I'm interested in what others think is best.

image
image
image

### Checklist

- [ ] Tests added for changes

### DIFF
--- a/packages/generator/src/generators/app-generator.ts
+++ b/packages/generator/src/generators/app-generator.ts
@@ -149,9 +149,16 @@ export class AppGenerator extends Generator<AppGeneratorOptions> {
       }
 
       // Ensure the generated files are formatted with the installed prettier version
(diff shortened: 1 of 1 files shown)


flybayer:
@fnoah awesome this looks perfect! Thank you so much!

@all-contributors add @fnoah for code

Issue #670, opened 2020-06-12
Prevent `blitz db migrate` from attempting to create new migration files in production.


flybayer:
### What is the problem?

Currently `blitz db migrate` always runs `prisma migrate save` and `prisma migrate up`.

`migrate up` will create new migration files if needed and `migrate up` will apply existing migration files.

For safety, `migrate save` should only run if `NODE_ENV !== 'production'`. But `migrate up` should always run.

Here's the line of code where this happens: https://github.com/blitz-js/blitz/blob/canary/packages/cli/src/commands/db.ts#L43

toshi1127:
I want to work in this, Can I take this issue?

flybayer:
@toshi1127 yes!! :)

flybayer:
Closed in #683

PR #683, opened 2020-06-14
Prevent `blitz db migrate` from attempting to create new migration files in production

toshi1127:
This PR closes the issue of Prevent `blitz db migrate` from attempting to create new migration files in production. #670 

- Prevent `blitz db migrate` from attempting to create new migration files in production.
- Added test code for Db command.

### Checklist

- [x] Tests added for changes

### DIFF
--- a/packages/cli/src/commands/db.ts
+++ b/packages/cli/src/commands/db.ts
@@ -36,30 +36,37 @@ export const runPrismaGeneration = async ({silent = false} = {}) => {
   }
 }
 
+const runMigrateUp = (prismaBin: string, resolve: (value?: unknown) => void) => {
+  const cp = spawn(prismaBin, ['migrate', 'up', schemaArg, '--create-db', '--experimental'], {
+    stdio: 'inherit',
+  })
+  cp.on('exit', async (code) => {
+    if (code === 0) {
+      await runPrismaGeneration()
+      resolve()
+    } else {
+      process.exit(1)
+    }
+  })
+}
+
 export const runMigrate = async () => {
   const prismaBin = await getPrismaBin()
-
   return new Promise((resolve) => {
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/db.ts
```diff
@@ -36,30 +36,37 @@ export const runPrismaGeneration = async ({silent = false} = {}) => {
...
+      cp.on('exit', (code) => {
+        if (code === 0) {
+          runMigrateUp(prismaBin, resolve)
+        } else {
+          process.exit(1)
+        }
+      })
+    } else {
+      runMigrateUp(prismaBin, resolve)
+    }
```
flybayer, line 69: This logic is backward. In production only `up` should run. Otherwise both `save` and `up` should run.

toshi1127: Fixed.

toshi1127:
@flybayer 
Thanks for the review !!

I thought it might be a good idea to edit the [issue](https://github.com/blitz-js/blitz/issues/670) to avoid confusion later on.

1. 
```
For safety, migrate save should only run if NODE_ENV === 'production'. 
```

2.
```
migrate up will create new migration files
```

flybayer:
Woohoo! Thanks :)

@all-contributors add @toshi1127 for code

packages/cli/src/commands/db.ts
```diff
@@ -36,30 +36,37 @@ export const runPrismaGeneration = async ({silent = false} = {}) => {
...
-          } else {
-            process.exit(1)
-          }
-        })
-      } else {
-        process.exit(1)
-      }
-    })
+    if (process.env.NODE_ENV === 'production') {
+      runMigrateUp(prismaBin, resolve)
```
BjoernRave, line 57: aren't you running `migrate up` twice in prod now ?

Issue #689, opened 2020-06-16
Upgrade react-query to latest version


flybayer:
### What do you want and why?

Upgrade react-query to the latest version.

It's specified in `packages/core/package.json`

---

Veterans, please leave for someone new to Blitz/open-source

simonedelmann:
Am I wrong or is this only changing a version number in one file? If yes, I can do that ;)

flybayer:
@simonedelmann yes it should be that easy :)

PR #692, opened 2020-06-16
Update react-query to 1.5.8

simonedelmann:
Closes: https://github.com/blitz-js/blitz/issues/689

Update react-query from 1.3.3 to 1.5.8. According to their changelog there are no breaking changes.

### Checklist

- [ ] Tests added for changes

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -43,7 +43,7 @@
     "@blitzjs/config": "0.15.0",
     "@blitzjs/display": "0.15.0",
     "pretty-ms": "6.0.1",
-    "react-query": "1.3.3",
+    "react-query": "1.5.8",
     "serialize-error": "6.0.0",
     "url": "0.11.0"
   },


packages/core/src/utils/query-cache.ts
```diff
@@ -1,7 +1,7 @@
 import {queryCache} from 'react-query'
 
 export interface QueryCacheFunctions<T> {
-  mutate: (newData: T | ((oldData: T) => T)) => void
+  mutate: (newData: T | ((oldData: T | undefined) => T | undefined)) => void
```
flybayer, line 4: I think this function should still return `T` instead of undefined, even if `oldData` is undefined. Right?
```suggestion
  mutate: (newData: T | ((oldData: T | undefined) => T)) => void
```

simonedelmann: We can try. If I did't miss something, this will throw a TypeScript error. Still I do agree, this *should* still return `T`.

simonedelmann: Okay, I was wrong. Thanks for watching out!

Issue #577, opened 2020-05-29
Implement session management


flybayer:
Here's the full session management RFC: #475

@rishabhpoddar has finished the pseudo code for this part of the **Essential** method: https://github.com/blitz-js/drafts/pull/10

Here's the [rendered version of his pseudo code](https://github.com/blitz-js/drafts/blob/session-pseudo-code/session-pseudo-code.md).

rishabhpoddar:
An important point: The expiry of the session is updated on use of the session. This does not happen for each request since it would mean that we would have to update the set-cookie + the public data token on the frontend for each request which is inefficient (we will have to do a localstorage write for each request to save the public token). Instead, it happens for all non-GET api requests that happen anytime after 1/4th the session expiry time has passed already.

The reason for doing it for non-GET API requests is that we cannot update the public token info on the frontend's localstorage if the user has done a browser level navigation to the site. If we did do it for GET APIs there is a good chance that it might lead to the frontend code seeing that the session exists even though it actually doesn't. 

I am happy to change this behaviour or go into a more detailed explanation.

PR #694, opened 2020-06-16
🎉Add supertokens session management package

flybayer:
Closes: #577

Adds the built in Blitz session management!

For more info on what we are doing, see [the session management RFC](https://github.com/blitz-js/blitz/pull/475)

### Checklist

- [x] Tests added for changes
- [x] PR submitted to [blitzjs.com](https://github.com/blitz-js/blitzjs.com) for any user facing changes: https://github.com/blitz-js/blitzjs.com/pull/112

---
### WIP

- [x] Server code for essential method
- [x] Add couple basic integration tests
- [x] Finish proper creation of the two custom error objects
- [x] Review maybe using advanced by default?
- [x] Write frontend code
- [x] (in progress) Get everything working in the app
- [x] Add this helper: `ctx?.session?.authorize('admin')` - also accept array
- [x] Figure out best way to configure auth from apps, especially DB access
- [x] Anonymous sessions
- [x] Convert session context to a class? So that `ctx.session.handle` will be updated after a call to `ctx.session.create()`
- [x] useSession subscribe to storage event so all tabs update
- [x] Lowercase header check in rpc
- [x] Password hashing
- [x] Switch to with Zod
- [x] Move authn to experimental function
- [x] Namespace JWT payload + add `iss`

Later:
- [ ] (in progress) Social login - Passportjs integration 
- [ ] Add e2e tests with chrome to test cookies
- [ ] Add unit tests
- [ ] Figure out way to make Authentication/Authorization errors have stacktrace to call site of `authorize()`
- [ ] Explore a more permissions based authz

Anyone is welcome to open PRs to this branch for any remaining items!


flybayer:
@rishabhpoddar here's my WIP PR on this! I have probably 2/3 of your pseudo code in here currently.

flybayer:
@rishabhpoddar ok all your pseudo code is in here now. I'm still cleaning up and productionizing it, but I think you can start reviewing it and at least check for anything obvious.

flybayer:
Ok @rishabhpoddar, ready for an in-depth review! 

I have a lot of TODO comments with questions for you.

I just remembered that I think we were not going to have anonymous session at the start, but I've coded up most of what I think is needed for that.

examples/auth/app/users/mutations/login.ts
```diff
@@ -0,0 +1,23 @@
...
+  email: string
+  password: string
+}
+
+type Ctx = {session?: SessionContext}
+
+export default async function login({email, password}: LoginInput, ctx: Ctx = {}) {
+  const user = await db.user.findOne({where: {email}})
+
+  if (!user || password !== user.hashedPassword) {
```
garrisons, line 15: @flybayer what's the plan for unhashing the password here? Would it be hashed on the client before it's passed as a param in mutation? This is looking great, btw. Authentication & Authorization is something that always seems to drive me crazy.

flybayer: Hashing will be done right here. Just haven't got to that part yet :D

merelinguist:
@flybayer Can you elaborate on “permissions-based authz”?

flybayer:
@merelinguist so the idea is that roles are groups of permissions. And you enforce permissions, not roles.

So you centralize the definition of roles with something like this:
```ts
const roles = {
  admin: ['product:read', 'product:create', 'product:update', 'product:delete'],
  manager: ['product:update'],
}
```

And then in your code you enforce a permission.

```ts
export default authorize('product:update', async function updateProduct() { ... })
```

This allows you to add/remove/refactor your roles without also having to change a ton of other files.

But I'm looking to use something more full featured like https://github.com/onury/accesscontrol

goldo:
maybe `social logins` can be added later ? I think it would be nice releasing this PR early, in order to iterate quicker, and developer to start using it sooner

flybayer:
@goldo Yeah it definitely can be! The current blocker is code review from @rishabhpoddar

goldo:
congrats @flybayer 🎉 great milestone ! Thanks a lot! 
Here are my quick feedbacks:
- User model create a table `User` in pg, which is a reserved keyword and thus I have to add quote each time (or use `public.`): `select * from "User";`
- in the documentation, it doesn't mention you have to `blitz generate db` (after changing the schema)
- in the document, at the first mention of the model `User` in the schema, it misses the `roles` part (and you don't talk about it later, I just found out in the "email/password" part.
- The status code return by the `  ctx.session.authorize()` function is always `200`, even when it returns an object like this: 
```js
{
  "statusCode": 401,
  "name": "AuthenticationError",
  "message": "",
  "stack": "..."
} 
```

- maybe provide the request as curl in order to create a user quickly in DB ? (without having to create the front-end part): or maybe you are planning on using "generate" command ?

I'll keep you posted in case I find some more stuff 👍 thanks again 👏

flybayer:
Thanks for the feedback @goldo!

> User model create a table User in pg, which is a reserved keyword and thus I have to add quote each time (or use public.): select * from "User";

You're welcome to change the model name and provide custom db hooks to the sessionManagement middleware. I think `User` will be best for most people, but we can change it if enough people want to.

> in the documentation, it doesn't mention you have to blitz generate db (after changing the schema)

Fixed

> in the document, at the first mention of the model User in the schema, it misses the roles part (and you don't talk about it later, I just found out in the "email/password" part.

Fixed

> The status code return by the ctx.session.authorize() function is always 200, even when it returns an object like this

This is by design because it's an RPC api, not a REST api. For more details see the spec: https://blitzjs.com/docs/rpc-specification

> maybe provide the request as curl in order to create a user quickly in DB ? (without having to create the front-end part): or maybe you are planning on using "generate" command ?

The best way to add a user is to use your `signup()` mutation via the blitz console:

1. Run `blitz console`
2. Run `await signup({email: 'me@example.com', password: 'password123'})`

goldo:
Thanks for the explanations ! 👍

PR #695, opened 2020-06-16
[Page template] Forward submit event from form to parent event handler

wKovacs64:
Closes: N/A

This updates the model form component template to forward the submit event from the generated `onSubmit` event handler to the parent so the parent has access to the event.

I ran into needing the event in the parent while updating the tutorial and rather than document how to do this part, I figured it would be better to just update the template since needing access to the event in the user's `onSubmit` handler is probably a common scenario.

### Checklist

- [ ] Tests added for changes


flybayer:
Great, thank you!

PR #705, opened 2020-06-20
Fix broken cli when using npm where global blitz version used instead of local

flybayer:

This fixes a bug where `blitz -vV` will return `debug: global` when run inside a blitz project. This means it's using the globally installed blitz package.

This also was causing the following error when running `blitz generate`:

```
✕ Error generating __modelIdParam__/edit.tsx
✕ Error: Couldn't resolve parser "babel-ts"
```

So I also added a try/catch around prettier in the generator to ensure that a prettier failure doesn't block file generation.

### Checklist

- [ ] Tests added for changes

### DIFF
--- a/packages/blitz/src/bin/cli.ts
+++ b/packages/blitz/src/bin/cli.ts
@@ -14,7 +14,8 @@ console.log(
 
 let usageType: "local" | "monorepo" | "global" | "global-linked"
 
-const localCLIPkgPath = path.resolve(process.cwd(), "node_modules", "@blitzjs/cli")
+const localCLIPkgPath1 = path.resolve(process.cwd(), "node_modules", "@blitzjs/cli")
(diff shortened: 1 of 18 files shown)


packages/blitz/src/bin/cli.ts
```diff
@@ -14,7 +14,8 @@ console.log(
 
 let usageType: "local" | "monorepo" | "global" | "global-linked"
 
-const localCLIPkgPath = path.resolve(process.cwd(), "node_modules", "@blitzjs/cli")
+const localCLIPkgPath1 = path.resolve(process.cwd(), "node_modules", "@blitzjs/cli")
+const localCLIPkgPath2 = path.resolve(process.cwd(), "node_modules/blitz/node_modules/@blitzjs/cli")
```
flybayer, line 18: Actual path of the local installed module can be one of these two paths.

Issue #714, opened 2020-06-25
blitz db reset broken in Prisma 2 (need to change `raw` to `rawQuery`)


johncantrell97:
### What is the problem?
It looks like the code to reset prisma database is using PrismaClient's raw() method.  Looking at the docs for the official Prisma 2.0+ release I think they might have changed this to a method called rawQuery() instead.

I downgraded back to Prisma 2.0.0-beta6 and it was working again for me.

I was getting this error:

Resetting the database has failed with an error from the database:  TypeError: db.raw is not a function

### Steps to Reproduce

1. Add prisma with a non-beta version of Prisma 2
2. Run `blitz db reset`

### Versions

macOS Catalina | darwin-x64 | Node: v12.13.1

blitz: 0.8.1 (global)
blitz: 0.15.3 (local)

SigurdMW:
I could look into this one. Is it just to replace the calls to `db.raw` with `db.rawQuery` in the files where it appears like this one: /packages/cli/src/commands/db.ts ?

PR #716, opened 2020-06-25
Fix `blitz db reset` for Prisma >=2.0.0-beta.8

SigurdMW:
Closes: https://github.com/blitz-js/blitz/issues/714

Changed `db.raw` to `db.rawQuery` that was needed due to Prisma 2.0

### Checklist

- [ ] Tests added for changes

### DIFF
--- a/packages/cli/src/commands/db.ts
+++ b/packages/cli/src/commands/db.ts
@@ -79,15 +79,15 @@ export async function resetPostgres(connectionString: string, db: any): Promise<
   const dbName: string = getDbName(connectionString)
   try {
     // close all other connections
-    await db.raw(
+    await db.rawQuery(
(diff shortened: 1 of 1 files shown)


SigurdMW:
So I noted that the types for db in the method `resetPostgres` is set to any, should I try to fix it so that it's at least typed to the `PrismaClient` type even though that's also any at this point?  Could be that the type changes in the future

SigurdMW:
@flybayer does it look better now? Did run successfully when symlinking and testing locally 👍

flybayer:
Beautiful!! Thank you so much :) 

@all-contributors add @SigurdMW for code

johncantrell97:
fyi -- this means it's now broken if you try to use older versions of prisma :)

Issue #712, opened 2020-06-23
Upgrade React Query to v2


flybayer:
### What do you want and why?

[React Query v2](https://github.com/tannerlinsley/react-query/releases/tag/v2.0.0) was just released and has some breaking changes.

We need to upgrade to this version, update our usage of it as needed, and update our docs to reflect the changes.

toshi1127:
I'd like to tackle this issue.

flybayer:
Sweet :)

flybayer:
Closed in #720

PR #720, opened 2020-06-29
Breaking: Upgrade React Query to v2

toshi1127:
This PR closes the issue of Upgrade React Query to v2. #712 

- To update React Query, I have fixed the implementation to accommodate some breaking changes.

**Affected Change List**
1. The new way to do dependent queries to instead use the enabled config flag. You should move your conditions in your key to the enabled option
    - Before
    ```useQuery(ready && queryKey, queryFn) ```
    - After
    ```useQuery(queryKey, queryFn, { enabled: ready })```
2. If you use functions as query keys that can potentially throw, you should migrate their logic to the enabled config option and use optional chaining to cast them to a boolean
    - Before
    ```useQuery(() => ['user', user.id], queryFn) ```
    - After
    ```useQuery(['user', user?.id], queryFn, { enabled: user?.id })```
3. `refetchQueries` has been renamed to `invalidateQueries`. You will need make this rename change for your app to continue working propertly. The name change comes due to some differences in what the function does.
    - Before, any queries that matched the `queryKey` used in `refetchQueries(queryKey)` and were also stale, would be refetched... even queries that were inactive and not rendered on the screen. This resulted in quite a few queries being refetched regardless of their immediate necessity.
    - Now, with invalidateQueries, only queries that are actively rendered will be refetched, while any other matching queries will forcefully be marked as stale.
4. The type of `isFetchingMore` in `InfiniteQueryResult` has been changed from `boolean` to `false | 'previous' | 'next'`.

### Checklist

- [ ] Tests added for changes
- [x] PR submitted to [blitzjs.com](https://github.com/blitz-js/blitzjs.com) for any user facing changes: https://github.com/blitz-js/blitzjs.com/pull/108

### DIFF
--- a/examples/store/app/products/pages/products/infinite.tsx
+++ b/examples/store/app/products/pages/products/infinite.tsx
@@ -22,7 +22,7 @@ const Products = () => {
       ))}
 
       <div>
-        <button onClick={() => fetchMore()} disabled={!canFetchMore || isFetchingMore}>
+        <button onClick={() => fetchMore()} disabled={!canFetchMore || !!isFetchingMore}>
           {isFetchingMore ? "Loading more..." : canFetchMore ? "Load More" : "Nothing more to load"}
         </button>
       </div>
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -43,7 +43,7 @@
     "@blitzjs/config": "0.15.3",
     "@blitzjs/display": "0.15.3",
     "pretty-ms": "6.0.1",
-    "react-query": "^2.4.4",
+    "react-query": "2.4.4",
     "serialize-error": "6.0.0",
     "url": "0.11.0"
   },
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -43,7 +43,7 @@
     "@blitzjs/config": "0.15.3",
     "@blitzjs/display": "0.15.3",
     "pretty-ms": "6.0.1",
-    "react-query": "1.5.8",
+    "react-query": "^2.4.4",
     "serialize-error": "6.0.0",
     "url": "0.11.0"
   },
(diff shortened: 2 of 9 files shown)


examples/store/app/products/pages/products/infinite.tsx
```diff
@@ -22,7 +22,7 @@ const Products = () => {
       ))}
 
       <div>
-        <button onClick={() => fetchMore()} disabled={!canFetchMore || isFetchingMore}>
+        <button onClick={() => fetchMore()} disabled={!canFetchMore || !!isFetchingMore}>
```
flybayer, line 25: Why this change?

packages/core/package.json
```diff
@@ -43,7 +43,7 @@
     "@blitzjs/config": "0.15.3",
     "@blitzjs/display": "0.15.3",
     "pretty-ms": "6.0.1",
-    "react-query": "1.5.8",
+    "react-query": "^2.4.4",
```
flybayer, line 46: This needs pinned to an exact version :)
```suggestion
    "react-query": "2.4.4",
```

flybayer:
By the way, always make sure your PRs have `Closes #712` instead of just `#712`. It has to have `closes` or else the issue won't automatically be closed when the PR is merged :)

examples/store/app/products/pages/products/infinite.tsx
toshi1127: This is because the type of `disabled` is a boolean but the type of `isFetchingMore` has been changed from `boolean` to `false | 'previous' | 'next'`.

packages/core/package.json
toshi1127: Fixed

Issue #410, opened 2020-05-03
Bug: can't override experimental flags in `withBlitz` - needed for Material UI


JoseRFelix:
### What is the problem?

This is related to #374 

Whenever `withBlitz` is used on `next.config.js` it disables the ripple of MUI button. Maybe there is a conflict with React concurrent mode?  

### Steps to Reproduce: 

1. Run `blitz start`
2. Test button

### Steps for it to work: 

1. Head to `.blitz/caches/dev`
2. Comment out config in `next.config.js`
3. Run `yarn next dev`
4. Test button

[Source](https://github.com/Jfelix61/blitz/tree/example/material-ui)

### Versions:

```
debug: local
debug: pkgPath: /home/jfelix/Documents/programming/open-source/blitz/node_modules/@blitzjs/cli 

Linux 5.3 | linux-x64 | Node: v12.16.1

blitz: 0.8.1 (global)
blitz: 0.8.1 (local)
```

### Supporting Documentation

[`withBlitz`](https://github.com/blitz-js/blitz/blob/canary/packages/server/src/withBlitz.ts) plugin.

merelinguist:
Thanks for the issue @Jfelix61 ! I’m going to investigate this now. My main question is whether MUI and concurrent is the issue here, or if it’s something else...

Jfelix61:
> My main question is whether MUI and concurrent is the issue here, or if it’s something else...

Yes... Another thing I could think is that it is maybe related to module resolution.

merelinguist:
As I suspected, concurrent mode works fine with MUI. I think I remember seeing an announcement from the MUI team for concurrent mode support, and this simple sandbox proves that it works fine: https://codesandbox.io/s/new-dawn-5tvpb?file=/next.config.js

merelinguist:
@Jfelix61 I’ve managed to create a basic working example with Blitz + MUI. 

Please see code here: https://codesandbox.io/s/blitz-hello-world-j1o7g?file=/app/pages/index.js

And the demo here: https://j1o7g.sse.codesandbox.io/

flybayer:
Does MUI require SSR? If not, it would be nice to also have an MUI example that doesn't use getInitialProps, because using getInitialProps in _document forces **all** pages to be SSR, if I remember correctly.

merelinguist:
Does this help: https://github.com/zeit/next.js/discussions/11073#discussioncomment-1440?

flybayer:
[...]

PR #721, opened 2020-06-30
Fix experimental flags in blitz.config.js not being used

ganeshmani:
Closes: #410

It adds an option to override the blitz default with `blitz.config.js`  experimental variables in the config. 

### Checklist

- [x] Tests added for changes

### DIFF
--- a/packages/server/src/with-blitz.ts
+++ b/packages/server/src/with-blitz.ts
@@ -18,6 +18,7 @@ export function withBlitz(nextConfig: any) {
       Object.assign({}, normalizedConfig, {
         experimental: {
           reactMode: "concurrent",
+          ...(nextConfig.experimental || {}),
         },
         webpack(config: any, options: Record<any, any>) {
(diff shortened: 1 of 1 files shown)


packages/server/src/with-blitz.ts
```diff
@@ -18,6 +18,7 @@ export function withBlitz(nextConfig: any) {
       Object.assign({}, normalizedConfig, {
         experimental: {
           reactMode: "concurrent",
+          ...(nextConfig.experimental || {}),
```
flybayer, line 21: I didn't catch this earlier, but we need to use the normalized version. `nextConfig` can be a function.
```suggestion
          ...(normalizedConfig.experimental || {}),
```

flybayer:
Thanks @ganeshmani!!

By the way, to link a PR to the issue, here's what to write in markdown: `Closes: #410`. You don't need to manually link it. And also "closes" must be next to the issue number without any other words in the middle.

Issue #415, opened 2020-05-04
Can't use non-js/ts files inside `pages/` folders


gielcobben:
### What is the problem?

I want to use CSS Modules with Blitz so I followed this guide: https://nextjs.org/docs/basic-features/built-in-css-support. I also installed `npm i sass` so it supports `scss` files. Everything seems to work fine except when I'm using CSS Modules in /pages:

```
app/pages/index.tsx
app/pages/index.module.scss

app/pages/projects/index.tsx
app/pages/projects/index.module.scss
```

_Note: This error only occurs in `/pages`, so CSS Modules work perfectly fine in `/components`_

### Steps to Reproduce: 

1. Create a file: `/app/pages/index.module.scss`
1. Import the file: `import styles from "./index.module.scss"`
1. Add a style to a html element: `<main className={styles.container}>`
1. Get a `Module not found` error.

### Versions:

```
debug: local
debug: pkgPath: /test/node_modules/@blitzjs/cli

macOS Catalina | darwin-x64 | Node: v12.16.1

blitz: 0.8.1 (global)
blitz: 0.8.1 (local)
```

### Supporting Documentation

```
./pages/index.tsx
Module not found: Can't resolve 'app/pages/index.module.scss' in '/test/.blitz/caches/dev/pages'
```

---

### Edit from @flybayer 

The problem is that our compiler is only using js/ts files inside `pages/`. We need to change this so that all files inside `pages/` get processed. Our conflict detection must also detect conflicts for these files.

simonpeterdebbarma:
Hello @gielcobben, thank you so much for reporting this issue!

flybayer:
@gielcobben the problem is that all files inside `pages/` must be a React component that will be rendered as a page. 

What you are trying to do doesn't work with Next.js.

The solution is to put your scss files in another folder outside of `pages/`.

Try this, and let us know if it works for you!

gielcobben:
Well, it works perfectly fine in all my next projects. It's just not working properly in a blitz project.
This is my file structure in a next project:

![image](https://user-images.githubusercontent.com/2663212/81101944-25d1a880-8f0f-11ea-98dc-85dacb9fea07.png)

flybayer:
@gielcobben oh wow, I didn't know that would work!

We'll get this fixed

emgoto:
[...]

PR #733, opened 2020-07-08
Fix issue with using non-page files inside `pages` folders

ryardley:
Closes: #415 

* Only copy typescript/javascript files to the pages folder - leave other extensions where they are.
* Allow imports to be relative transformed when import does not have an import specifier.

### Checklist

- [x] Tests added for changes


ryardley:
Just realised we are not doing path transformation when imports don't use the `from` syntax. So adding it to this PR.

Issue #292, opened 2020-04-27
[Feature] Create alpha example with knex or pg driver instead of prisma


Johnson444:
Thank you very much for Blitz and also that you don't focus on GraphQL. I share your opinion that GraphQL is great but might be to heavy for indie developers.

### What do you want and why?

Following up on your twitter tweet it would be cool to add an example to the alpha that explains how to use blitz with native db driver instead of prisma.

I am interested mostly how to share the db connection between files. Is it enough to connect to the database at start up of blitz and import / export it to all javascript files, or do I have to do more to support a shared connection.

### Possible implementation(s)

I feel knex implementation would be best as it's widely used and support multiple databases. Alternative would be pg driver.

### How might we do this?

Implement simple CRUD app with at least two pages.

### Additional context

https://twitter.com/blitz_js/status/1254706828014911488
http://knexjs.org
https://github.com/brianc/node-postgres

aem:
Hm. So I like the idea of letting people work in the environment they work best, but my concern about not being opinionated about Prisma is that we lose some control over the ecosystem and lose the ability to have a truly batteries-included implementation. For example, it would be quite challenging to truly support an arbitrary database abstraction with `blitz generate`, if we make it easy to not use Prisma then I think we'll find a lot of users upset that they don't have access to some of the CLI features until after they're too invested in their new DB engine.

A good corollary is Rails and ActiveRecord. Rails is _very_ opinionated about ActiveRecord (while still letting you use whatever DB you'd like under the hood, mysql, pg, etc.) which means that, yes, you're stuck with ActiveRecord even if you don't like it, but the `rails` CLI becomes infinitely more useful. While the scaffolding + generation part of Rails is nice, what truly makes it **magical** is the data layer, and I anticipate the same for Blitz. If we can effectively use Prisma to generate most of the code for our users so they don't have to touch anything in the data layer, Blitz's value will be obvious and immediate. If we allow a user to swap out their data layer then Blitz becomes not much more than a light wrapper around Next.js.

ryardley:
I think it would be fine so long as we provided a plugin architecture around it. We should be developing in such a way to enable this.

aem:
I think that's fair, but currently we're fully baked in on Prisma right now. I think we should put this on the backlog until #242 is fleshed out. Until then We don't really know how to architect the CLI to support this. Maybe we should have an architecture call about #242 soon?

flybayer:
[...]

PR #739, opened 2020-07-12
Allow custom DB connection logic and add example for using DB clients other than Prisma

Skn0tt:
Closes: #292

- Added a minimal example on how to use DB clients other than Prisma
- To make that work: Changed API handler template to allow custom connection logic

### Checklist

- [ ] Tests added for changes


Skn0tt:
Alright! Back working on Blitz after quite a while, decided to pick #292 :D The example isn't very beautiful right now and will be changed before I un-draft this, but I wanted to talk to y'all about the changes I introduced in https://github.com/blitz-js/blitz/pull/739/commits/e1825ccf384f9b717d09043e94700a9a921a5f80. It allows users to override the DB connection creation for clients other than Prisma like this:

```ts
// db/index.ts
// [...]

export default myDbClient

export async function connect() {
  await myDbClient.establishConnection()
  await myDbClient.runTransactions()
}
```

This could be a way of integrating between Blitz and non-Prisma DB clients until we land dedicated database plugins.

@aem @flybayer @ryardley

flybayer:
Nice, this looks good!

Issue #741, opened 2020-07-13
First render on server is missing route id


benjick:
### What is the problem?

On SSR the first render won't have a route id

### Steps to Reproduce

1. `blitz new myApp`
2. `cd myApp`
3. `blitz generate all project name:string`
4. `blitz db migrate`
5. `blitz start`
6. Go to http://localhost:3000/projects and create a project
7. End up on http://localhost:3000/projects/1 and reload the page
8. Check terminal for error:

<details>
  <summary><code>(node:10863) UnhandledPromiseRejectionWarning: Error:</code></summary>

 ```
(node:10863) UnhandledPromiseRejectionWarning: Error:
Invalid `prisma.project.findOne()` invocation in
webpack-internal:///./app/_resolvers/projects/queries/getProject.ts:10:76

{
  where: {
    id: null
        ~~~~
  }
}

Argument id for where.id must not be null. Please use undefined instead.


    at Document.validate (/Users/admin/dev/hikyaku/node_modules/@prisma/client/runtime/index.js:1:49485)
    at Object.s [as Project] (/Users/admin/dev/hikyaku/node_modules/@prisma/client/runtime/index.js:1:208862)
    at Object.n.<computed> [as findOne] (/Users/admin/dev/hikyaku/node_modules/@prisma/client/runtime/index.js:1:210789)
    at getProject (webpack-internal:///./app/_resolvers/projects/queries/getProject.ts:10:76)
    at queryFn (/Users/admin/dev/hikyaku/node_modules/@blitzjs/core/dist/core.cjs.development.js:133:14)
    at /Users/admin/dev/hikyaku/node_modules/react-query/dist/react-query.development.js:804:26
    at _catch$1 (/Users/admin/dev/hikyaku/node_modules/react-query/dist/react-query.development.js:377:20)
    at /Users/admin/dev/hikyaku/node_modules/react-query/dist/react-query.development.js:802:14
    at /Users/admin/dev/hikyaku/node_modules/react-query/dist/react-query.development.js:406:34
    at /Users/admin/dev/hikyaku/node_modules/react-query/dist/react-query.development.js:975:40
    at _catch$1 (/Users/admin/dev/hikyaku/node_modules/react-query/dist/react-query.development.js:377:20)
    at /Users/admin/dev/hikyaku/node_modules/react-query/dist/react-query.development.js:969:18
    at /Users/admin/dev/hikyaku/node_modules/react-query/dist/react-query.development.js:406:34
[...]

PR #761, opened 2020-07-17
Fix error during static pre-rendering in development

flybayer:
Closes: #741

Unfortunately Next does static pre-rendering in development in which all route parameters are empty. This works great for the production build because it statically generates the loading screens.

But in development, this results in an error in our terminal logs because the query input is null.

This PR adds some smart detection to `useQuery`, `useInfiniteQuery` and `usePaginatedQuery` that detects if it is being called during prerender. If yes, then it returns a never resolving promise.

Related react-query issue: https://github.com/tannerlinsley/react-query/issues/744

### Checklist

- [ ] Tests added for changes


flybayer:
This fix is not ideal, but going to go ahead and add it for now. Hopefully Next.js or react-query comes up with a solution and then we can remove this.

PR #767, opened 2020-07-18
For comparison, convert auth example to use react-hook-form

flybayer:
This converts the auth example to use react-hook-form so we can have a comparison between this and react-final-form.

Change the diff mode to side-by-side for better comparison:
https://github.com/blitz-js/blitz/pull/767/files?diff=split


examples/auth/app/auth/components/LoginForm.tsx
```diff
@@ -1,69 +1,79 @@
...
 import login from "app/auth/mutations/login"
 
 type LoginFormProps = {
   onSuccess?: () => void
 }
 
 export const LoginForm = (props: LoginFormProps) => {
+  const methods = useForm({
+    mode: "onBlur",
+    async resolver(values) {
```
jay-jlm, line 13: Alternatively, if you are interested in steering users towards using more idiomatic JS:
```
async resolver(value) {
    return Object.keys(values).reduce((acc, key) => {
        if(requiredFields.includes(key) && !isRequiredValid(_values[key])) acc[key] = "Required"
        return acc;
    }, {})
}
```

examples/auth/app/auth/components/LoginForm.tsx
```diff
@@ -1,69 +1,79 @@
...
-            await login({email: values.email, password: values.password})
-            props.onSuccess && props.onSuccess()
-          } catch (error) {
-            if (error.name === "AuthenticationError") {
-              return {[FORM_ERROR]: "Sorry, those credentials are invalid"}
-            } else {
-              return {
-                [FORM_ERROR]:
-                  "Sorry, we had an unexpected error. Please try again. - " + error.toString(),
+      <FormProvider {...methods}>
```
jay-jlm, line 29: I see you are following some of the examples in from the lib's own docs here, but this object does not contain only methods, so it's a bit of a misnomer, ihmo.

examples/auth/app/auth/components/LoginForm.tsx
```diff
@@ -1,69 +1,79 @@
...
+        <form
+          onSubmit={methods.handleSubmit(async (values) => {
+            try {
```
jay-jlm, line 32: Extract to a function?

jay-jlm:
So happy with this pull request! 
I'm big fan of react-hook-form.

examples/auth/app/auth/components/LoginForm.tsx
flybayer: Yeah I agree

examples/auth/app/auth/components/LoginForm.tsx
flybayer: I find `reduce` very hard to read for working with objects :)

jay-jlm: I'd say it's more a case of getting used to it but you're not alone, for sure!

flybayer:
Hook form option added in #917

PR #776, opened 2020-07-21
Add Blitz Recipe for Tailwind CSS

aem:

Introduces the first Blitz recipe! Continuing with the canonical example that we've been using for the duration of the recipes discussion, our first recipe will be for Tailwind CSS. This recipe will install the appropriate tailwind dependencies, add the base stylesheets, and import them into the target application. The only piece that's missing for this to be 100% prod-ready is that we're missing full dependency resolution (cc @merelinguist I know you've been poking at that). 

Once this is merged, it'll be accessible via `blitz install tailwind` (the recipe name for the CLI is based on the directory name under `/recipes`).

Working on writing up docs now!

### DIFF
--- a/packages/installer/src/recipe-builder.ts
+++ b/packages/installer/src/recipe-builder.ts
@@ -3,7 +3,18 @@ import * as AddDependencyExecutor from "./executors/add-dependency-executor"
 import * as NewFileExecutor from "./executors/new-file-executor"
 import * as TransformFileExecutor from "./executors/file-transform-executor"
 
-export function RecipeBuilder() {
+export interface RecipeBuilder {
+  setName(name: string): RecipeBuilder
+  setDescription(description: string): RecipeBuilder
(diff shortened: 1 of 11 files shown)


packages/installer/src/recipe-builder.ts
```diff
@@ -3,7 +3,18 @@ import * as AddDependencyExecutor from "./executors/add-dependency-executor"
 import * as NewFileExecutor from "./executors/new-file-executor"
 import * as TransformFileExecutor from "./executors/file-transform-executor"
 
-export function RecipeBuilder() {
+export interface RecipeBuilder {
```
aem, line 6: externally, TS wasn't able to infer the `RecipeBuilder` type so I went ahead and added an explicit interface declaration. initially I tried converting it to a class instead of a plain JS object but the intellisense/tsserver experience was slightly worse and had less intelligent autocomplete so I went with this instead

Issue #788, opened 2020-07-23
Remove `delete data.id` from  Generated update mutations


simonedelmann:
### What do you want and why?

Currently all generated mutations include this code for deleting the id.
```
// Don't allow updating
delete data.id
```
This is not necessary because prisma's generated `UpdateModelInput` does not contain the `@id` field and prisma will throw an error (`Unknown arg `id` in data.id`) if it is provided accidentally. 
Furthermore as `id` does not exist on prisma's generated update types, the code above will cause a typescript warning.

### Possible implementation(s)

Simply remove that 2 lines from the generated mutations code.

flybayer:
Here's the line that needs removed: https://github.com/blitz-js/blitz/blob/canary/packages/generator/templates/mutation/update__ModelName__.ts#L21

aem:
I'm not totally sure we want to do this. With JS-based projects we won't get the benefit of the TypeScript warning, so it's possible that a user with limited Prisma experience might try to send up the ID with the request. If that happens, their code will fail at runtime and they probably won't have a great idea why. This might be unnecessary, but I think it provides a nice protection against a footgun for users not using TS.

flybayer:
Related: https://github.com/blitz-js/blitz/discussions/782#discussioncomment-40553. We need to add full input validation to queries & mutations.

I think it's fine to remove for now to prevent the typescript build error.

simonedelmann:
@aem I wasn't thinking about JS users. You're right, they might get stuck there. 
But if they do, there is a big red error on the console:
```
✕ updateProject failed: Error: 
Invalid `prisma.project.update()` invocation in
webpack-internal:///./app/_resolvers/projects/mutations/updateProject.ts:16:76

{
  where: {
    id: 1
  },
  data: {
    name: 'Lorem Ipsum',
    description: '',
    start: '2020-09-19T00:00:00.000Z',
    end: '2020-09-28T00:00:00.000Z',
    id: 100
    ~~
  }
}

Unknown arg `id` in data.id for type ProjectUpdateInput. Did you mean `end`? Available args:
type ProjectUpdateInput {
  createdAt?: DateTime
  updatedAt?: DateTime
  name?: String
  description?: String
  start?: DateTime
  end?: DateTime
}
```
So maybe they will notice not to update the id?

PR #789, opened 2020-07-23
Remove `delete data.id` from Generated update mutations

simonedelmann:
Closes: #788 

See #788


packages/generator/templates/mutation/update__ModelName__.ts
```diff
@@ -17,8 +17,6 @@ export default async function update__ModelName__(
   {where, data}: Update__ModelName__Input,
   ctx: Record<any, any> = {},
 ) {
-  // Don't allow updating
```
flybayer, line 20: Can you keep this comment, but move it inside `if (process.env.parentModel)` ?

simonedelmann: Done!

Issue #800, opened 2020-07-25
extraArgs.flatMap is not a function


IzumiSy:
### What is the problem?
I got an error saying `extraArgs.flatMap is not a function`

### Steps to Reproduce

1. npm install -g blitz
2. blitz new new-blitz-app
3. cd new-blitz-app
4. blitz start (and stop it with Ctrl-C)
5. blitz generate all project name:string

### Versions

```
$ blitz --version --verbose
You are using alpha software - if you have any problems, please open an issue here:
  https://github.com/blitz-js/blitz/issues/new/choose

debug: blitzPkgPath: /Users/seiya/Devs/new-blitz-app/node_modules/blitz/dist/index.js
debug: cliPkgPath: /Users/seiya/Devs/new-blitz-app/node_modules/blitz/node_modules/@blitzjs/cli/lib/src/index.js

macOS Catalina | darwin-x64 | Node: v10.16.0

blitz: 0.16.4 (global)
blitz: 0.16.4 (local)
```

### Other

Please include applicable logs and screenshots that show your problem.

flybayer:
Hi @IzumiSy! That's because of Node v10. Is it possible for you to upgrade to v12?

IzumiSy:
Yes, I can! Then, this issue is kind of version incompatibility?

flybayer:
Yep! Probably we should add a warning/error if users are < Node v12

IzumiSy:
Thank you 👍  That would help people like me.

PR #804, opened 2020-07-25
Add Node version check and warning to CLI startup routine

Skn0tt:
Closes: https://github.com/blitz-js/blitz/issues/800#issuecomment-663810776

Check Node version at startup and warn if it's older than v12.

### Checklist

- [x] Tests added for changes


Skn0tt:
@flybayer I'm a bit uncertain about wether we should write the small Semver parsing logic ourselves or pull in the `semver` package instead - what do you think? :D

flybayer:
Awesome, thanks @Skn0tt!! Looks good.

PR #805, opened 2020-07-25
Add Blitz Recipe for Emotion

wKovacs64:
Closes: N/A

Adds a Blitz recipe to install and configure the Emotion CSS-in-JS library.

### Checklist

- [ ] Tests added for changes

### DIFF
--- a/recipes/emotion/index.ts
+++ b/recipes/emotion/index.ts
@@ -9,13 +9,8 @@ function wrapComponentWithCacheProvider(ast: ASTNode, b: builders, t: NamedTypes
   if (!t.File.check(ast)) return
 
   visit(ast, {
-    visitFunction(path) {
-      const {node} = path
-      // TODO: need a better way to detect the custom App function
-      if (t.Identifier.check(node.id) && node.id.name === "MyApp") {
-        return this.traverse(path)
-      }
-      return false
+    visitExportDefaultDeclaration(path) {
+      return this.traverse(path)
     },
     visitJSXElement(path) {
       const {node} = path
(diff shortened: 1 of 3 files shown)


wKovacs64:
@aem OK Adam, I have no idea what I'm doing here (no experience with AST, etc.) so a lot of this might be pretty bad. Couldn't even find any docs for `recast` and friends, so was mainly just poking around source and hoping for the best. Consider yourself warned!

#### Outstanding issues/edge-cases that have crossed my mind:

1. Finding/identifying nodes to modify makes some huge assumptions like the name of the node (in the case of the custom `MyApp` function and its `Component` prop) or the base class (in the case of the custom `MyDocument` class). This feels awful and extremely brittle - hoping you have a better idea.
1. I blindly insert my own `getInitialProps` method at the top of the custom `Document` (basically assuming a fresh Blitz project). It would probably be better to see if there's an existing one and modify that if so - or are we not going to be that concerned with collisions this early in the recipe ecosystem?
1. [Inserting a fragment doesn't work](https://github.com/blitz-js/blitz/pull/805/files#diff-92b00441769684e4f18acaa83cc85efcR224-R230) (results in a weird error I couldn't find anything on, aside from some other poor soul who also experienced it) so I had to fake it out by inserting a `JSXElement` with an empty element identifier. It works, but it should be wrapped in parentheses and I couldn't figure out how to do that nicely (fixable with Prettier, though).
1. Adding an import via the existing `addImport` utility function adds a new import statement, disregarding any existing import from the same package. Possibly an enhancement to be made to `addImport`?

Probably other issues I'm forgetting...

Issue #748, opened 2020-07-14
Add envinfo to `blitz -v` output


flybayer:
### What do you want and why?

I want to add some [`envinfo`](https://github.com/tabrindle/envinfo) output to the `blitz -v` command. envinfo is a sweet utility to get lots of debugging data easily.

Also, I would like to add a line that indicates if the package manager is npm or yarn. We already have the `has-yarn` dependency installed which can be used for that.

### Possible implementation(s)

I think the basics that we need is the output of this:

```
npx envinfo --system --binaries --npmPackages blitz,typescript,react,react-dom,@prisma/cli,@prisma/client
```

So I picture the output of `blitz -v` looking like this:

```
/t/a1> blitz -v
You are using alpha software - if you have any problems, please open an issue here:
  https://github.com/blitz-js/blitz/issues/new/choose

blitz: 0.16.4 (global)
blitz: 0.16.4 (local)

  Package manager: npm        (<-- NOTE - this added, but doesn't come from envinfo)
  System:
    OS: macOS 10.15.5
    CPU: (8) x64 Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz
    Memory: 1.37 GB / 16.00 GB
    Shell: 3.0.2 - /usr/local/bin/fish
  Binaries:
    Node: 12.16.3 - /var/folders/dn/t7l73r1d08q_0djz0hz06bt80000gn/T/fnm-shell-7340312/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.4 - /var/folders/dn/t7l73r1d08q_0djz0hz06bt80000gn/T/fnm-shell-7340312/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  npmPackages:
    @prisma/cli: 2.2.1 => 2.2.1
    @prisma/client: 2.2.1 => 2.2.1
    blitz: 0.16.4 => 0.16.4
    react: 0.0.0-experimental-33c3af284 => 0.0.0-experimental-33c3af284
    react-dom: 0.0.0-experimental-33c3af284 => 0.0.0-experimental-33c3af284
    typescript: 3.9.6 => 3.9.6
```

[Here's the file that where this change needs to be made](https://github.com/blitz-js/blitz/blob/canary/packages/blitz/src/bin/cli.ts#L41)

robertgrzonka:
🙋🏼‍♂️

chris-tse:
[...]

PR #816, opened 2020-07-31
Add extra debugging info to `blitz -v`

chris-tse:
Closes: #748 

This adds a call to `npx envinfo` via an `exec` call. Whenever a user runs `blitz -v` it will show whether they are using npm or yarn along with any relevant data from the output of `envinfo`.

#### Regarding Tests

The test file in this package was empty, so I wasn't sure how to proceed with this. I'm not too familiar with back-end testing (I'm definitely more of a front-end person). It seems like this is the "entry point" for the CLI and there's no real "command" for the `-v` flag. 

The canary branch also seemed to be having a few failing tests in other branches at the time I opened this, so if those get fixed I'll pull them in.

### Checklist

- [ ] Tests added for changes


flybayer:
Awesome @chris-tse!! Sorry for the delay in responding here. This looks really great!

Just one change, let's add envinfo as a dependency of the `blitz` package instead of using npx there. My usage of `npx` in the original issue was just so you could easily see the output of it :)

chris-tse:
Thanks for the feedback!

I had originally wanted to do that, but I had errors after installing and importing it as a dependency somewhere in the JS output, but I might have installed it in the wrong place. It did end up working now, I'll push my changes so you can see what you think.

packages/blitz/src/bin/cli.ts
```diff
@@ -2,72 +2,102 @@ import * as path from "path"
...
 
-console.log(
-  chalk.yellow(
-    `You are using alpha software - if you have any problems, please open an issue here:
-  https://github.com/blitz-js/blitz/issues/new/choose\n`,
-  ),
-)
 
-if (parseSemver(process.version).major < 12) {
+async function main() {
```
chris-tse, line 10: All the lines added here were due to the indent with wrapping in a async function. Reason detailed below.

packages/blitz/src/bin/cli.ts
```diff
@@ -2,72 +2,102 @@ import * as path from "path"
...
-    console.log("") // One last new line
-  } catch (e) {
-    console.log("blitz error", e)
+      await printEnvInfo()
+
+      console.log("") // One last new line
+    } catch (e) {
+      throw new Error(`Blitz Error: ${e}`)
+    }
+    process.exit(0)
```
chris-tse, line 76: Since the call to `envinfo` is async, this line is causing the script to exit before it can finish, leading to no output being printed. 

An alternative would be to move process.exit into the callback of envinfo? Though if other async stuff needs to be added down the road we might run into the same problem again.

flybayer:
Thanks @chris-tse! 

I fixed the following:
- `envinfo` wasn't in package.json
- Whitelisted the `blitz/src/bin` folder for prettier
- Lazy loaded envinfo and hasYarn so it's only loaded when `-v` is passed.

chris-tse:
Thanks for fixing it up @flybayer ! Good thinking with the lazy loading!

Weird that it wasn't in the package.json, because it was definitely there in this commit here: https://github.com/blitz-js/blitz/blob/aeea3c3b2b07a737fef4ccd23ed373611197a851/packages/blitz/package.json

Could have been accidentally removed when I merged from canary? Not sure.

flybayer:
Yeah could have been lost in a merge conflict :) 

Thanks again for this!

@all-contributors add @chris-tse for code

PR #828, opened 2020-08-04
Remove check for Vercel

Skn0tt:
This shouldn't be required since Blitz.js is now supported natively by Vercel, right?

Don't allow `next.config.js` to be present, even during builds on Vercel.

### Checklist

- [ ] Tests added for changes


flybayer:
Nope, we still need that :) Because we still use the `@vercel/next` builder and runtime. The first class support just configures that for you.

PR #830, opened 2020-08-04
Add Blitz recipe for Chakra

Zeko369:

Add recipe for installing chakra ui. @flybayer you can continue here, I got tangled in trying to figure out if the function is defaultly exported or not, besides that everything should work

@aem Some notes and improvements for the installers in general
- Most installers will interact with pages / _app / _document, and since they're are `export default`, there should be a super simple and reusable function to find that AST node
- Also most installers will wrap App in a provider (Random stores, Themes...), so we should also have a reusable function to do so

### DIFF
--- a/recipes/chakra/index.ts
+++ b/recipes/chakra/index.ts
@@ -1,0 +1,82 @@
+import {RecipeBuilder, paths, addImport} from "@blitzjs/installer"
+import {builders} from "ast-types/gen/builders"
+import {ASTNode} from "ast-types/lib/types"
+import {NamedTypes} from "ast-types/gen/namedTypes"
+import {visit} from "ast-types"
+
+// Copied from https://github.com/blitz-js/blitz/pull/805, let's add this to the @blitzjs/installer
+function wrapComponentWithCacheProvider(ast: ASTNode, b: builders, t: NamedTypes) {
+  if (!t.File.check(ast)) return
+
+  visit(ast, {
+    visitFunction(path) {
+      // const {node} = path
+      // TODO: need a better way to detect the custom App function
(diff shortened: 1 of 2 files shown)


recipes/chakra/index.ts
```diff
@@ -0,0 +1,82 @@
...
+      return false
+    },
+    visitJSXElement(path) {
+      const {node} = path
+      path.replace(
+        b.jsxElement(
+          b.jsxOpeningElement(b.jsxIdentifier("ThemeProvider")),
+          b.jsxClosingElement(b.jsxIdentifier("CacheProvider")),
+          [
+            b.literal("\n  \t  "),
```
aem, line 28: hm. shouldn't need to do spacing here, `recast` knows about `prettier` so it should auto-format everything. are you seeing otherwise?

aem:
> Also most installers will wrap App in a provider (Random stores, Themes...), so we should also have a reusable function to do so

Good thought. We can add this to the installer package to make that more generic

recipes/chakra/index.ts
flybayer: I can confirm that auto formatting is not working.

flybayer:
I have this almost finished

flybayer:
Finished!! Thank you so much @Zeko369!!

recipes/chakra/index.ts
aem: oof. i'll look into this. recast should be picking up the prettier config

aem: oh. i bet it's because we put the prettier config in `package.json` instead of a separate `.prettier.config.js`

PR #835, opened 2020-08-04
Add Blitz Recipe for Render.com

flybayer:

Adds `blitz install render` which will add a pre-configured `render.yaml` to your project


recipes/render/index.ts
```diff
@@ -0,0 +1,17 @@
...
+  .setDescription("")
+  .setOwner("b@bayer.ws")
+  .setRepoLink("https://github.com/blitz-js/blitz")
+  .addNewFilesStep({
+    stepId: "addRenderConfig",
+    stepName: "Add render.yaml",
+    explanation: `NOTE: Your app must be configured to use Postgres for this render.yaml config`,
+    targetDirectory: ".",
+    templatePath: join(__dirname, "templates"),
+    templateValues: {},
```
aem, line 15: so the API currently supports additional CLI args, as a stopgap until we support interactive prompts we could use CLI args to populate the template values for the app name and the connection string:

```js
{
  templateValues(cliArgs: any) {
    if (!cliArgs.connectionString || !cliArgs.dbName) {
      // make this error msg better pls
      throw new Error('Must supply connectionString and dbName args');
    }
    return cliArgs;
  }
}
```

and then in the .yaml template:

```diff
-          name: myapp-db
+          name: __dbName__
-          property: connectionString
+          property: __connectionString__
```

flybayer: Ok yeah good idea for the app and db name. However `connectionString` doesn't change. It's a special reference to the connectionString of the DB.

aem: oh right right, that just links to the environment variable. been a bit since I set up render

Issue #838, opened 2020-08-05
Update RPC generation with new Prisma connect function


merelinguist:
### What do you want and why?

Prisma [`2.4` ](https://github.com/prisma/prisma/releases/tag/2.4.0) deprecated several top-level client methods, including one which we use, `prisma.connect`. This means that using any Blitz queries or mutations will result in a console warning about the deprecated usage.

### Possible implementation(s)

This is the offending line: https://github.com/blitz-js/blitz/blob/canary/packages/server/src/stages/rpc/index.ts#L93

`db.connect` needs to be swapped for `db.$connect`.

### Additional context

Note that this is sort of a breaking change, since Prisma would have to be upgraded by the user for this to work, but it would be better to fix it earlier rather than later, imo.

flybayer:
For backward comparability, if `.$connect` exists, use it, otherwise use `.connect`

agoxlea:
I'm new here. Can I take this one?

PR #839, opened 2020-08-06
Fix `prisma.connect` warning with prisma 2.4.0

agoxlea:
Closes: #838 

Uses the new Prisma $connect() method if available rather than connect() which was deprecated in [Prisma 2.4.0](https://github.com/prisma/prisma/releases/tag/2.4.0)

### Checklist

- [ ] Tests added for changes

I was able to reproduce this by updating examples/auth to use Prisma 2.4.0. It seemed like bumping the Prisma version for the examples was out of scope for this PR, but let me know if it makes sense to bump those as part of this.

### DIFF
--- a/packages/server/src/stages/rpc/index.ts
+++ b/packages/server/src/stages/rpc/index.ts
@@ -90,7 +90,7 @@ let db
 let connect
 try {
   db = require('db').default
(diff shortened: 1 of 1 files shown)


flybayer:
Thank you @agoxlea, looks great!

@all-contributors add @agoxlea for code

PR #844, opened 2020-08-08
typo on tailwind recipe

osirvent:

Just a very small typo


flybayer:
Thank you @osirvent!

PR #810, opened 2020-07-28
Add Jest and React Testing Library to new apps

peaonunes:
Closes: #277

### What are the changes and their implications?

Adds jest and RTL setup and an example test. The setup is very inspired on what `create-react-app` setups by default with fewer configurations on the config file.

I've added both `jest` which is the test runner and the [`react-testing-library`](https://testing-library.com/) which is the default and recommended approach to write tests by the `create-react-app`.

image

### Checklist

- [ ] Tests added for changes
- [x] PR submitted to [blitzjs.com](https://github.com/blitz-js/blitzjs.com) for any user facing changes

PR #847, opened 2020-08-08
Finish default testing setup for new apps

flybayer:

This builds on the work started in #810 

- [x] Add top level `test` folder with `utils.ts` which includes custom `render` and `renderHook` test functions which include a router mock
- [x] Add `@testing-libary/react-hooks`
- [x] Clean up and improve `jest.config.js`
- [x] Move `app/pages/__tests__/index.test.ts` to `app/pages/index.test.ts` (co-location)
- [x] `skip` the default test because every person is going to break the test as soon as the replace the default content which is really annoying.
- [x] Add typechecking to test files with `ts-jest`
- [x] Parse test files with babel so jsx works
- [x] Make TS absolute paths and path aliases work in jest

### Checklist


packages/core/src/with-router.tsx
```diff
@@ -4,6 +4,8 @@ import {WithRouterProps as WithNextRouterProps} from "next/dist/client/with-rout
 import {useParams, extractRouterParams} from "./use-params"
 import {useRouterQuery} from "./use-router-query"
 
+export {RouterContext} from "next/dist/next-server/lib/router-context"
```
aem, line 7: is this something we should write a test for in case this file moves?

packages/generator/templates/app/test/setup.ts
```diff
@@ -2,4 +2,4 @@
 // allows you to do things like:
 // expect(element).toHaveTextContent(/react/i)
 // learn more: https://github.com/testing-library/jest-dom
-import '@testing-library/jest-dom/extend-expect';
+import "@testing-library/jest-dom/extend-expect"
```
peaonunes, line 5: **question**: Do we have a prettier config for the template files? It's funny to see a lot of changes in the style on this PR. I am not sure what should be the one to follow.

packages/generator/templates/app/jest.config.js
```diff
@@ -1,37 +1,29 @@
...
-    "app/**/*.{js,jsx,ts,tsx}",
-    "!app/**/*.d.ts"
-  ],
-  testMatch: [
-    "<rootDir>/app/**/__tests__/**/*.{js,jsx,ts,tsx}",
-    "<rootDir>/app/**/*.{spec,test}.{js,jsx,ts,tsx}"
-  ],
-  testEnvironment: "jest-environment-jsdom-fourteen",
-  modulePaths: [],
-  moduleFileExtensions: [
```
peaonunes, line 21: **question**: Why did you remove the file extensions to watch? Is that being replace by any combination of the rules you have added or we will track any file extension?

packages/generator/templates/app/test/utils.tsx
```diff
@@ -0,0 +1,93 @@
+import React from "react"
```
peaonunes, line 1: These utils are really nice 😄

packages/core/src/with-router.tsx
flybayer: The build will fail if it moves

packages/generator/templates/app/test/setup.ts
flybayer: It should be using the monorepo config. Not sure how the wrong format got in originally.

packages/generator/templates/app/jest.config.js
flybayer: The default includes js, jsx, ts, tsx, and node. And the extra entries with `web` are not needed from what I can tell.

peaonunes: Awesome!

packages/core/src/with-router.tsx
aem: 🤦‍♂️ it's been a long day

Issue #821, opened 2020-08-02
Upgrade everything to eslint v7 - internal code and in the new app template


flybayer:
### What do you want and why?

`eslint-config-react-app` [relatively recently added support for eslint v7](https://github.com/facebook/create-react-app/pull/8978) so we can upgrade all our internal tooling to to the latest eslint packages and we can upgrade all the eslint packages [in the new blitz app template](https://github.com/blitz-js/blitz/blob/canary/packages/generator/templates/app/package.json)

I think this task should be fairly straightforward. Updating the major version numbers in the new app template will be manual, but all the other eslint packages in the repo should be able to be upgraded via the `yarn upgrade-interactive --latest` command.

nettofarah:
hey, @flybayer. My friend @pragmaticivan and I wanted to try and tackle this issue together.
Here's pt1 https://github.com/blitz-js/blitz/pull/852

Thanks!

PR #853, opened 2020-08-09
Update eslint to v7 in new app template and the blitz repo

pragmaticivan:
Closes: #821
Address the root migration part of #821.

This PR updates eslint to v7.x. 
Follows new recommendation for V7 migration: https://eslint.org/docs/user-guide/migrating-to-7.0.0#default-ignore-patterns
It currently ignores dotfiles but not .eslintrc.js

### Checklist

- [x] Tests added for changes


pragmaticivan:
cc @nettofarah

.eslintrc.js
```diff
@@ -25,7 +25,7 @@ module.exports = {
     ],
     "@typescript-eslint/no-floating-promises": "error",
   },
-  ignorePatterns: ["packages/cli/", "packages/generator/templates"],
+  ignorePatterns: ["packages/cli/", "packages/generator/templates", ".eslintrc.js"],
```
flybayer, line 28: I'm just curious, what was the problem that we need to add this?

nettofarah: Eslint would fail to start linting the codebase unless `.eslintrc.js` was marked as ignored.
I think the error has something to do with this breaking change: https://eslint.org/docs/user-guide/migrating-to-7.0.0#runtime-deprecation-warnings-for-eslintrc-config-files

flybayer:
Ok now there's a yarn.lock conflict that needs resolved, and then we can merge!

Note: this doesn't fully close #821, the one remaining item is to update the eslint and plugin versions [in the new app template](https://github.com/blitz-js/blitz/blob/canary/packages/generator/templates/app/package.json?rgh-link-date=2020-08-02T19%3A13%3A47Z). If you want to tackle that issue too, you can add that to this PR or do it as a separate PR. Up to you. And of course you can leave it for someone else to do. :)

pragmaticivan:
hi @flybayer, I just fixed the conflicts, rebased the PR, and added the update on template app.

flybayer:
@pragmaticivan awesome! One final thing which I think you might have overlooked in my last comment, is to also update the eslint plugins in the new app template package.json.

packages/generator/templates/app/package.json
```diff
@@ -47,7 +47,7 @@
     "@typescript-eslint/eslint-plugin": "2.x",
     "@typescript-eslint/parser": "2.x",
     "babel-eslint": "10.x",
-    "eslint": "6.x",
+    "eslint": "7.x",
```
pragmaticivan, line 50: hi @flybayer you meant this change?

flybayer: Yes, this is good, but we also need to update the other `eslint-*` packages in this file

pragmaticivan: Do you think it would be also worth updating?:
```
 @typescript-eslint/eslint-plugin                           2.x  →      3.x
 @typescript-eslint/parser                                  2.x  →      3.x
```

flybayer: yes! All the eslint related packages

pragmaticivan: done, thanks!

flybayer:
Awesome, I tested and it works great. Thank you so much!!

@all-contributors add @pragmaticivan for code

PR #868, opened 2020-08-12
Prevents secure cookies when running on localhost

### LINKED ISSUE #855: Automatically detect http/https on localhost and use secure cookies accordingly
### What do you want and why?

Currently [we are setting cookies as `secure`](https://github.com/blitz-js/blitz/blob/canary/packages/server/src/supertokens.ts#L699) (which requires HTTPS) if `NODE_ENV === 'production'`. This causes an issue if you run `blitz start --production` on your local machine because the url will be `http://localhost:3000` so cookies won't work because it's not https.

So instead we also want to check and only enable `secure` if request is not from `localhost`

**Edit**: updated to remove https localhost check

toddgeist:
Closes: #855

### Checklist

- [ ] Tests added for changes


flybayer:
Thanks Todd! You also need to add this in 4 places in this server supertokens file:

https://github.com/blitz-js/blitz/blob/canary/packages/server/src/supertokens.ts#L699

toddgeist:
@flybayer  Can I import code from one Package into another?  

I was thinking of moving the isLocalhost function to core/utils/index.ts and importing into server.

flybayer:
Yes you can import code from `core` into `server`.

toddgeist:
Ok I have another issue :-(

the four places in supertokens.js where this logic needs to go do not have access to the Req object.  They only have access to the Res Object.  I don't

flybayer:
Add `req` to the parameters of those and pass it in, same as `res`. Pass `req` down as many functions as you need to.

toddgeist:
Appreciate the patience.  This is my first time in a large typescript project and first time with lerna/packages approach.

Worked on adding the code to supertokens.js.  I think I got it all.  Local testing of each package passed.  Un sure if I need to add new test for this feature or not.

flybayer:
Great! One question, did you forget to `git push` your latest changes? :) 

And you don't have to worry about a test for this for now.

toddgeist:
ha - sure did forget :-)

toddgeist:
t

toddgeist:
Ok so I think it is ready now.  But you'll have to update this PR again I think.  Sorry for the confusion

flybayer:
No problem! It's up to date with canary, but I still see only the code for the passport adapter. Click here to see the diff between canary and this branch: https://github.com/blitz-js/blitz/pull/868/files

toddgeist:
I clearly don't understand this process. :-(

toddgeist:
I blew this. don't know how.  I am just going to close this PR and start over.

flybayer:
@toddgeist Sorry :( I'm more than happy to jump on a quick call or something and walk through it with you. You can ping me on slack any time

PR #876, opened 2020-08-13
Change new app template to set database through .env

dwightwatson:
As floated on Slack, this takes advantage of a [newer feature available in Prisma](https://github.com/prisma/prisma/issues/1487#issuecomment-635999202) which makes configuring databases easier. Instead of needing to tinker with your `schema.prisma` file you can control your database through environment variables.

Primsa will inspect the `DATABASE_URL` and determine the first provider from the array that supports it.

This lets you use SQLite in development without having to change the schema file before deploying. It also makes it easy to switch to Postgres when ready with a one-line change.

### Checklist

- [ ] Tests added for changes


flybayer:
Awesome, thank you!!

Issue #864, opened 2020-08-12
Getting error of unwanted types in vanilla js app


svobik7:
### What is the problem?
After creation of new blitz app using `--js` flag and following main three steps from app homepage I am getting error `Syntax error: Unexpected token, expected ";"`

### Steps to Reproduce

1. `blitz new myApp --js`
2. `blitz generate all project name:string`
3. `blitz db migrate`
4. `blitz start`
4. Go to `http://localhost:3000/projects`

### Versions

```
macOS Catalina | darwin-x64 | Node: v12.14.1

blitz: 0.17.0 (global)
blitz: 0.17.0 (local)
```

### Other

<img width="953" alt="Screenshot 2020-08-12 at 14 50 02" src="https://user-images.githubusercontent.com/761766/90017414-c9b7b680-dcab-11ea-8ab7-fbbb9582aa20.png">

Seems this line cause the error https://github.com/blitz-js/blitz/blob/90a762448959dadb4513eea8ba78047fd9174d8f/packages/server/src/stages/rpc/index.ts#L72.

I think this "as typeof" statement could be removed completely from `isomorphicHandlerTemplate` as it is used only internally by blitz app. 

Or am I missing something? I am ready to PR.

flybayer:
Thanks for reporting this! I think we can remove that entire `as` statement, so we just need to try that and see.

You're welcome to submit a PR!

PR #878, opened 2020-08-13
Fix queries and mutations for plain JS projects

svobik7:
Closes: #864

Removed TS types from RPC templates to be able to run blitz app with `--js`

### DIFF
--- a/packages/cli/src/commands/start.ts
+++ b/packages/cli/src/commands/start.ts
@@ -1,7 +1,12 @@
-import {Command, flags} from "@oclif/command"
 import {dev, prod} from "@blitzjs/server"
-
+import {Command, flags} from "@oclif/command"
+import fs from "fs"
+import path from "path"
+import pkgDir from "pkg-dir"
 import {runPrismaGeneration} from "./db"
+
+const projectRoot = pkgDir.sync() || process.cwd()
+const isTypescript = fs.existsSync(path.join(projectRoot, "tsconfig.json"))
 
 export class Start extends Command {
   static description = "Start a development server"
@@ -22,12 +27,14 @@ export class Start extends Command {
   }
 
   async run() {
+    
     const {flags} = this.parse(Start)
 
     const config = {
       rootFolder: process.cwd(),
       port: flags.port,
       hostname: flags.hostname,
+      isTypescript,
     }
 
     try {
--- a/packages/server/src/stages/rpc/index.ts
+++ b/packages/server/src/stages/rpc/index.ts
@@ -2,12 +2,12 @@ import {Stage, transform} from "@blitzjs/file-pipeline"
 import {relative} from "path"
 import slash from "slash"
 import File from "vinyl"
-import {absolutePathTransform, isTypescript} from "../utils"
+import {absolutePathTransform} from "../utils"
 
 /**
  * Returns a Stage that manages generating the internal RPC commands and handlers
  */
-export const createStageRpc: Stage = function configure({config: {src}}) {
+export const createStageRpc: Stage = function configure({config: {src, isTypescript = true}}) {
   const fileTransformer = absolutePathTransform(src)
 
   const getResolverPath = fileTransformer(resolverPath)
@@ -43,7 +43,7 @@ export const createStageRpc: Stage = function configure({config: {src}}) {
     // Isomorphic client
     const isomorphicHandlerFile = file.clone()
     isomorphicHandlerFile.contents = Buffer.from(
-      isomorhicHandlerTemplate(resolverImportPath, resolverName, resolverType),
+      isomorhicHandlerTemplate(resolverImportPath, resolverName, resolverType, isTypescript),
     )
     push(isomorphicHandlerFile)
 
@@ -61,6 +61,7 @@ const isomorhicHandlerTemplate = (
   resolverPath: string,
   resolverName: string,
   resolverType: string,
+  useTypes: boolean = true,
 ) => `
 import {getIsomorphicRpcHandler} from '@blitzjs/core'
 const resolverModule = require('${resolverPath}')
@@ -69,7 +70,7 @@ export default getIsomorphicRpcHandler(
   '${resolverPath}',
   '${resolverName}',
   '${resolverType}',
-) ${isTypescript() ? "as typeof resolverModule.default" : ""}
+) ${useTypes ? "as typeof resolverModule.default" : ""}
 `
 
 // Clarification: try/catch around db is to prevent query errors when not using blitz's inbuilt database (See #572)
--- a/packages/server/src/stages/rpc/index.ts
+++ b/packages/server/src/stages/rpc/index.ts
@@ -1,8 +1,8 @@
-import File from "vinyl"
-import slash from "slash"
-import {absolutePathTransform} from "../utils"
-import {relative} from "path"
 import {Stage, transform} from "@blitzjs/file-pipeline"
+import {relative} from "path"
+import slash from "slash"
+import File from "vinyl"
+import {absolutePathTransform, isTypescript} from "../utils"
 
 /**
  * Returns a Stage that manages generating the internal RPC commands and handlers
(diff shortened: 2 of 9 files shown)


ryardley:
Hi @svobik7 hey thanks heaps for the PR! Looks like there is a little more type work to do as those `as` types were masking those coming from getIsomorphicRpcHandler which don't seem to be enhancing the types correctly.

svobik7:
Hi @ryardley so the better approach would be to strip this `as types` only when app is in `--js` mode and keep them in `ts` mode?

flybayer:
Yeah looks like that is the case.

We could maintain two templates, one for JS and one for TS, or we can transform the template with babel like we do with the CLI code generator.

svobik7:
Ok, I will take a look at templates more closely lately today.

svobik7:
Hi there, I chose to use babel transform in same way as it is used in generator. Now `--js` mode do not throws any type error. 

But there is one strange thing I noticed (also noted in codebase). If no filename is provided in `babel.transform` following error is thrown during `blitz start`:

```
/.eslintrc.js(node:32835) UnhandledPromiseRejectionWarning: Error: [BABEL] unknown: Preset /* your preset */ requires a filename to be set when babel is called directly,

babel.transform(code, { filename: 'file.ts', presets: [/* your preset */] });

See https://babeljs.io/docs/en/options#filename for more information.
    at validateIfOptionNeedsFilename (/Users/jirkasvoboda/Development/packages/blitzjs/node_modules/@babel/core/lib/config/full.js:274:11)
    at /Users/jirkasvoboda/Development/packages/blitzjs/node_modules/@babel/core/lib/config/full.js:286:52
    at Array.forEach (<anonymous>)
    at validatePreset (/Users/jirkasvoboda/Development/packages/blitzjs/node_modules/@babel/core/lib/config/full.js:286:25)
    at loadPresetDescriptor (/Users/jirkasvoboda/Development/packages/blitzjs/node_modules/@babel/core/lib/config/full.js:293:3)
    at loadPresetDescriptor.next (<anonymous>)
    at recurseDescriptors (/Users/jirkasvoboda/Development/packages/blitzjs/node_modules/@babel/core/lib/config/full.js:107:30)
    at recurseDescriptors.next (<anonymous>)
    at loadFullConfig (/Users/jirkasvoboda/Development/packages/blitzjs/node_modules/@babel/core/lib/config/full.js:142:6)
    at loadFullConfig.next (<anonymous>)
```

Any ideas why babel requires filename in this case?

svobik7:
Hi @flybayer, @ryardley you are right, also for me now it seems like have-weight solution for striping one line of TS. Let's do it the other way by ternary strip unwanted types. I am going to update PR today.

packages/server/src/stages/rpc/index.ts
```diff
@@ -1,8 +1,13 @@
 import File from "vinyl"
 import slash from "slash"
 import {absolutePathTransform} from "../utils"
-import {relative} from "path"
+import path, {relative} from "path"
 import {Stage, transform} from "@blitzjs/file-pipeline"
+import * as babel from "@babel/core"
+// @ts-ignore TS wants types for this module but none exist
```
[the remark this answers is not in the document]
svobik7: Agree. I just copy-paste these lines from generator.  For now we are moving back to non-babel solution for resolving this PR/issue so these lines will be removed (see https://github.com/blitz-js/blitz/pull/878#issuecomment-678931088).

packages/server/src/stages/rpc/index.ts
```diff
@@ -130,3 +136,19 @@ function resolverPath(path: string) {
 function apiHandlerPath(path: string) {
   return path.replace(/^app/, "pages/api")
 }
+
+function transformTemplate(template: string) {
+  const projectRoot = pkgDir.sync() || process.cwd()
+  const isTypescript = fs.existsSync(path.join(projectRoot, "tsconfig.json"))
```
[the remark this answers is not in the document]
svobik7: As I am removing babel from this PR (see https://github.com/blitz-js/blitz/pull/878#issuecomment-678931088) than `transformTemplate` will be useful just for one case https://github.com/blitz-js/blitz/blob/679efef951cb8fbec22605d05ed78ab9ff4eb544/packages/generator/src/generator.ts#L202 but I think it became out of the scope of this PR.

But I am about to extract `const isTypescript = fs.existsSync(path.join(projectRoot, "tsconfig.json"))` as helper for detecting active TS.

svobik7:
@ryardley could you please review my last push? I used the ternary approach to strip types in --js mode. Also I squeezed all commits to single one to keep just relevant git logs.

packages/server/src/stages/rpc/index.ts
```diff
@@ -69,7 +69,7 @@ export default getIsomorphicRpcHandler(
   '${resolverPath}',
   '${resolverName}',
   '${resolverType}',
-) as typeof resolverModule.default
+) ${isTypescript() ? "as typeof resolverModule.default" : ""}
```
ryardley, line 72: We are probably going to want to cache isTypeScript outside the hot path as this won't change on the fly - especially as it runs a fs.existsSync call. If we can call it in the creation stage and hold on to the output then that should be fine.

ryardley (changes requested):
Looks good just we probably want to be careful with making sync fs calls.

packages/server/src/stages/rpc/index.ts
ryardley: An ideal solution would calculate this once and pass this in as config but that is not 100% necessary.

svobik7: Ok, I cached isTypescript on `blitz start` and pass down as config var so it can be used anywhere else in future. Please see my last push.

ryardley: Looks as though there are some tests that also need to be adjusted. I was hesitant about doing this as we should probably refactor the way config is injected here (my fault honestly) but this looks all good if you can get tests to pass.

svobik7: No problem, I'll take a look tomorrow.

svobik7: @ryardley finally it was easy task to get tests passed so I did it immediately. I just added `isTypescript` to mocked stage args inside test. PR is ready to final check and merge I think :-)

ryardley (approved):

packages/cli/src/commands/start.ts
```diff
@@ -1,8 +1,13 @@
...
 import {dev, prod} from "@blitzjs/server"
-
+import {Command, flags} from "@oclif/command"
+import fs from "fs"
+import path from "path"
+import pkgDir from "pkg-dir"
 import {runPrismaGeneration} from "./db"
 
+const projectRoot = pkgDir.sync() || process.cwd()
+const isTypescript = fs.existsSync(path.join(projectRoot, "tsconfig.json"))
```
ryardley, line 9: This is probably fine for now but we should look at how this sort of config stuff is managed in a different PR.

svobik7: What is your idea here? Should we create new discussion thread about config stuff here on GH?

packages/server/src/stages/rpc/index.ts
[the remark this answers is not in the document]
Skn0tt: Sounds good.

Skn0tt (approved):

packages/cli/src/commands/start.ts
ryardley: It would be good not to have to drill config down so many levels deep. Ideally there is a thing that determines information about the environment that is always run. We kind of have a place that might be more natural for this stuff perhaps [here](https://github.com/blitz-js/blitz/blob/canary/packages/server/src/config.ts) rather than have to remember to run that check everytime.

svobik7: I think I get your point. Typescript checking in cli level really seems unnecessary. It should be moved to `server/config.ts`. 

I opened new discussion about drilling config down here where I am also including ts checking move: https://github.com/blitz-js/blitz/discussions/941 so we can continue in discussion there.

flybayer:
Woohoo!

@all-contributors add @svobik7 for code

PR #880, opened 2020-08-13
Remove icon from scaffolded templates

dwightwatson:

All the generated template files include the favicon in the head - which seems overkill. Out of the box it's in `app/pages/index.tsx`. If you were seriously wanting to configure your favicon it would probably make sense to do once in the `_document.tsx` so you don't need to repeat yourself.

### Checklist

- [ ] Tests added for changes

### DIFF
--- a/packages/generator/templates/app/app/auth/pages/login.tsx
+++ b/packages/generator/templates/app/app/auth/pages/login.tsx
@@ -9,7 +9,6 @@ const SignupPage: BlitzPage = () => {
     <>
       <Head>
         <title>Log In</title>
-        <link rel="icon" href="/favicon.ico" />
       </Head>
 
       <div>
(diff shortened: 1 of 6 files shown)


flybayer:
Thanks Dwight! (sorry I fat-fingered the close button)

You are right this isn't ideal. I think the best solution is to add a Layout component that sets the favicon in there. 

However, until we get that added, I think we should leave the favicon in there. I.e. something is better than nothing. With it in there, people can see at least one way to do it.

What do you think?

dwightwatson:
My gut feeling is no, simply because it's left in `app/pages/index.tsx` as an example that people can extrapolate from there, but I appreciate where you're coming from.

Is there anything particular you had in mind for a base layout component - I could have a crack at putting that together instead? I've used `app/Layouts/Layout.tsx` in my project but I'm not really sure what the intended convention is.

flybayer:
Yeah, that file but used with option #4 [from this post](https://adamwathan.me/2019/10/17/persistent-layout-patterns-in-nextjs/) where you add a `getLayout` property on pages (also need to add a call to getLayout in `_app.tsx`.

flybayer:
I made an issue for it: https://github.com/blitz-js/blitz/issues/891

flybayer:
Closing this because we fixed in #897 instead

Issue #858, opened 2020-08-11
usePaginatedQuery docs example incorrect


scisteffan:
### What is the problem?

When I use the example usePaginatedQuery code in the docs (https://blitzjs.com/docs/use-paginated-query) the next button is not appearing / disappearing correctly.

### Steps to Reproduce

1. Go to https://blitzjs.com/docs/use-paginated-query and implement the code
2. Ensure the maximum number of items on the page is set to 3
3. Add 3 items to the database - *next button does not disappear*

### Versions

```
You are using alpha software - if you have any problems, please open an issue here:
    https://github.com/blitz-js/blitz/issues/new/choose

debug: blitzPkgPath: /Users/steffanharris/Documents/Projects/BlitzJS/usepaginatedquery-test/node_modules/blitz/dist/index.js
debug: cliPkgPath: /Users/steffanharris/Documents/Projects/BlitzJS/usepaginatedquery-test/node_modules/@blitzjs/cli/lib/src/index.js 

macOS Catalina | darwin-x64 | Node: v12.16.3

blitz: 0.17.1-canary.3 (global)
blitz: 0.17.1-canary.3 (local)

  Package manager: yarn 
  System:
    OS: macOS 10.15.5
    CPU: (16) x64 Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz
    Memory: 1.86 GB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 12.16.3 - ~/.nvm/versions/node/v12.16.3/bin/node
    Yarn: 1.22.4 - ~/.yarn/bin/yarn
    npm: 6.14.7 - ~/.nvm/versions/node/v12.16.3/bin/npm
    Watchman: Not Found
  npmPackages:
    @prisma/cli: 2.4.1 => 2.4.1 
    @prisma/client: 2.4.1 => 2.4.1 
    blitz: 0.17.1-canary.3 => 0.17.1-canary.3 
    react: 0.0.0-experimental-33c3af284 => 0.0.0-experimental-33c3af284 
    react-dom: 0.0.0-experimental-33c3af284 => 0.0.0-experimental-33c3af284 
    typescript: 3.9.7 => 3.9.7 
```

### Other

<img width="432" alt="Screenshot 2020-08-11 at 16 16 38" src="https://user-images.githubusercontent.com/2676185/89915336-22287e80-dbee-11ea-92ae-0c34b06bde59.png">

[...]

PR #881, opened 2020-08-13
Change store example to use single query for both pagination and infinite loading

scisteffan:
Closes: #858

Updated the paginated product list in the store example to calculate hasMore logic on the server, to fix a bug when the Next button would be incorrectly disabled.

I added strick null check = true to tsconfig.json since this is enabled by default on new blitz projects, and fixed compilation errors.

### Checklist

- [ ] Tests added for changes


flybayer:
Great work @scisteffan!!

After seeing all this, I think we should consolidate `getProducts`, `getProductsPaginated`, and `getProductsInfinite` all into a single `getProducts`. 

Are you up for making that change?

And then as a separate task, I think we should update the default query generator to match this because everyone is going to need pagination of some sort.

flybayer:
@all-contributors add @scisteffan for code,doc,financial

flybayer:
@all-contributors add @scisteffan for code, doc, financial

scisteffan:
Yep, happy to take this on board!

scisteffan:
There's an issue when using the same query for both the paginated and infinite listings. The results of the query are cached. In the case of usePaginatedQuery, the result is an object type, whereas useInfiniteQuery it's an array. So when navigating between the two using client side navigation, the cached value is used which is of the wrong type.

Do you have any suggestions on how to fix this?

flybayer:
Hmmm that's weird. I'll look into that.

flybayer:
@scisteffan I found the bug! Fixed in #963 

Again thanks for your work on this!!

PR #895, opened 2020-08-16
Fix to not use secure cookies on localhost

toddgeist:
Closes: #855
Added checks for serving from localhost to the cookie handling. Prevents secure cookies from being used while on localhost.


flybayer:
@all-contributors add @toddgeist for code

flybayer:
Thanks Todd!! I fixed the build issue (have to import from the package in a monorepo instead of a relative file path).

toddgeist:
ok good. now I know how to do that :-)

Issue #891, opened 2020-08-16
Add Persistant Layout to New App Template


flybayer:
### What do you want and why?

In my experience, most apps will need or want persistent layout in their apps. By default with Next.js, you don't have any persistent layouts. If you simply wrap each page with `<Layout>`, Layout will unmount and remount for every page.

[This post](https://adamwathan.me/2019/10/17/persistent-layout-patterns-in-nextjs/) goes into great detail on various ways to accomplish this. Option 4 in the post is the best option. I use it to great success and from what I've seen many others are using this approach too.

### Possible implementation(s)

1. Move the `<Head>...</Head>` stuff from [`app/pages/index.tsx`](https://github.com/blitz-js/blitz/blob/canary/packages/generator/templates/app/app/pages/index.tsx) in the new app template to `app/layouts/Layout.tsx`
2. Add a `title` prop to Layout so pages can easily set page title
3. Add call to `getLayout()` in `app/pages/_app.tsx` as shown in the post above (and properly type via Typescript)
4. Add `getLayout` in `app/pages/index.tsx` as shown in the post above
4. Add `getLayout` in the [page generator templates](https://github.com/blitz-js/blitz/tree/canary/packages/generator/templates/page) as shown in the post above

PR #897, opened 2020-08-16
Add Persistent layout to new app template with `<Layout>` and `getLayout()`

dwightwatson:
Closes: #891

This introduces a new `Layout` component as suggested in #891. I've attempted to follow the suggestion in the linked blog post as well as the guide from the issue. Unfortunately my TypeScript isn't too hot, so I'm not sure where to take it from here. But hoping this may serve as a good starting point.

### Checklist

- [ ] Tests added for changes


flybayer:
Great, thank you!! I pushed a commit that adds all the Typescript stuff

Issue #698, opened 2020-06-17
Generated pages include embedded expression braces from templates


wKovacs64:
### What is the problem?

Super minor, but I noticed the generated pages include braces `{ }` from [inline conditional rendering logic](https://github.com/blitz-js/blitz/blob/b8e2e10288cefd637678556a80d897fd26ed74ac/packages/generator/templates/page/new.tsx#L47-L57) in the corresponding templates, like so:

```jsx
<p>
  {
    <Link href="/projects">
      <a>Projects</a>
    </Link>
  }
</p>
```

When I expected:

```jsx
<p>
  <Link href="/projects">
    <a>Projects</a>
  </Link>
</p>
```

This is not a huge deal as it still renders correctly, but I could see it surprising someone.

### Steps to Reproduce

1. `blitz generate pages project`
1. Examine any of the generated pages and find certain elements surrounded by braces

### Versions

```
debug: monorepo
debug: pkgPath: C:\dev\contrib\blitz\node_modules\@blitzjs\cli

Windows 10 | win32-x64 | Node: v12.17.0

blitz: 0.15.0 (global)
blitz: 0.15.0 (local)
```

### Other

N/A

peaonunes:
Happy to have a look at it.

flybayer:
@peaonunes @aem has some thoughts on this. It's not an easy thing to solve 😬

peaonunes:
😬 oops. happy to delay that if we do not have a goto approach. I can switch collaborate to something with a higher priority.

PR #898, opened 2020-08-17
Migrate Recipes to jscodeshift and Support conditional JSX expressions

aem:
Closes: https://github.com/blitz-js/blitz/issues/698 (and a little extra 😬)

In terms of functional changes, there's only one: now we natively support conditionals in JSX (`<if condition="something"><A /><else><B /></else></if>`) rather than needing to embed JS code (`{process.env.something ? <A /> : <B />}`). This solves the problem of the nested expression output in #698 

Under the hood, we made significant changes to the generator infrastructure to support this. Rather than using Recast's verbose and undocumented API, this moves us to `jscodeshift`. While still mostly undocumented, the TS support for `jscodeshift` via their DefinitelyTyped types is far ahead of Recast's, making the development experience much more pleasant. The API is also much nicer - functional and declarative rather than imperative and mutative. This also simplifies client code modifying files - you don't need a bunch of nested imports from both `recast` and `ast-types` anymore, `jscodeshift` is self-contained, so just `import j from 'jscodeshift'` and you're off and running!

In order to do this migration, we unfortunately had to migrate everything at once, and ⚠️ **the changes are breaking** ⚠️ - users wanting to install these recipes will need to upgrade to the latest Blitz release once this is merged (and we should cut a release right after merging this) since the `transform` and `addImports` APIs are not backwards compatible. This should be the only time this happens - I expect future iterations to be backwards compatible, it just isn't possible for this migration because we're switching libraries entirely. 

Ultimately, I'm hoping that this newer, easier API will make it much more reasonable for newcomers to get up and running with the AST code, and will allow us to more easily write generic function implementations so that transforms can truly just be composition of pre-built functions.

I tested all of the migrated recipes locally, but extra eyes are always appreciated :) 

### Checklist

- [x] Tests added for changes
- [x] PR submitted to [blitzjs.com](https://github.com/blitz-js/blitzjs.com) for any user facing changes https://github.com/blitz-js/blitzjs.com/pull/132


aem:
ugh. silly windows line endings.

aem:
@wKovacs64 can you see what's going on with the windows tests? I assumed explicitly setting the end of line character would've fixed that test but it's still broken and when I tried to fire up my windows VM it appears to be corrupted :/

wKovacs64:
@aem Fails locally with the same issue as CI. I've never seen these "dots" that are rendering for new lines. EOL definitely seems to be involved, but changing the setting doesn't seem to matter (`lf` is Prettier default anyway).

```
@blitzjs/generator:     - Expected
@blitzjs/generator:     + Received
@blitzjs/generator:     -
@blitzjs/generator:     + ·  <----------------------------------------- what the heck is this?
@blitzjs/generator:     -         <if condition="condition">
@blitzjs/generator:     +         <if condition="condition">· <-------- and this
@blitzjs/generator:     -           <div>true</div>
@blitzjs/generator:     +           <div>true</div>· <----------------- and this, etc.
```

I'll have to play around more later. If we want to just get it passing, we could probably drop back to `.toMatchInlineSnapshot()` (even though the snapshot will look a little weird) for now? 🤷‍♂️  Or if the formatting doesn't really matter, we could also wrap `statement` with `oneLineTrim` from [`common-tags`](https://www.npmjs.com/package/common-tags#onelinetrim). 😉 Cheating, I know...

aem:
i guess in the test i could just do a regex replace on CRLF and replace with regular LF. feels hacky, but idk how else to fix it.

I'd like to preserve formatting just because that's kinda the point of using `jscodeshift` rather than a babel plugin or something

wKovacs64:
> I'd like to preserve formatting just because that's kinda the point of using `jscodeshift` rather than a babel plugin or something

Yeah good point.

flybayer:
Awesome! The current blocker for releasing a new stable version is to add a prompt for selecting your form library during `blitz new`. I'll start work on this within a day or so. I want to get a stable release out within a few days if possible.

PR #903, opened 2020-08-17
Show `blitz install` command in `blitz help` output

dwightwatson:
### What are the changes and their implications.
As mentioned on Slack last night, this is the only command out of the box that is `hidden`. I assume it may have been hidden while in early development, but as `install` is now listed in the docs we may as well make it visible.

### DIFF
--- a/packages/cli/src/commands/install.ts
+++ b/packages/cli/src/commands/install.ts
@@ -47,7 +47,7 @@ interface RecipeMeta {
 }
 
 export class Install extends Command {
-  static description = "Install a third-party package into your Blitz app"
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/install.ts
```diff
@@ -50,7 +50,6 @@ export class Install extends Command {
   static description = "Install a third-party package into your Blitz app"
```
flybayer, line 50:
```suggestion
  static description = "Install a Recipe into your Blitz app"
```

flybayer:
Thanks @dwightwatson!!

Issue #890, opened 2020-08-16
Add conditional check to blitz webpack config to fix usage without `db/index.ts` file


Weilbyte:
### What do you want and why?

If running `blitz start` without a `db` folder, Blitz will error out saying that it cannot resolve `./db/index`. However, if running with a blank `db/index` file then it's all okay. 

It would be nice if Blitz could catch the `Can't resolve` and warn the user that it's running without a database. 

I know Blitz is meant for database applications but I think handling that error would do some good. 

### Possible implementation(s)

Catch resolve error on start and run without a database (warn user that its running without a database as well), similar to how it runs with an empty `db/index` file.

flybayer:
Yes, that's a bug! [This line is where the problem is](https://github.com/blitz-js/blitz/blob/canary/packages/server/src/with-blitz.ts#L18).

We need to conditionally add that line only if the `db/index.*` file exists.

Weilbyte:
Will attempt to tackle this soon.

Weilbyte:
PR up => https://github.com/blitz-js/blitz/pull/906

PR #906, opened 2020-08-18
Fix to allow usage without `db/index.(ts|js)` file

Weilbyte:
Closes: #890 

Adds the `db` module to webpack only if it exists, allowing for usage without it. 
It checks if the `db` folder exists within the `.blitz` cache.

Works when running with dev, will test soon with prod.

### DIFF
--- a/packages/server/src/with-blitz.ts
+++ b/packages/server/src/with-blitz.ts
@@ -1,3 +1,5 @@
+const fs = require("fs")
+
 export function withBlitz(nextConfig: any) {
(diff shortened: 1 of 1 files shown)


flybayer:
Thanks @Weilbyte!

What you have may work, but I think we should change it to use the same method we have been using for everything else because we know for sure this works:

```ts
import pkgDir from "pkg-dir"
const projectRoot = pkgDir.sync() || process.cwd()
fs.existsSync()
```

Weilbyte:
Ah that's great, will make it use that method!

Issue #886, opened 2020-08-15
useQuery retry on error wanted?


simonedelmann:
### What do you want and why?

When using useQuery, there are 3 retries by default (react-query). In case the query returns an error, the request will be repeated three times with growing delay in between. 
In blitz we already limited the retries to 2. In case we detect a known error (AuthenticationError, NotFoundError, ...) retry will be disabled entirely. 
But this does not work for custom errors. As even 2 retries takes 14s and no error page is shown during this time, I wonder if we do need retries at all. 
If not, I would suggest to disable them by default. 

### Possible implementation(s)

Disable retries in useQuery.

flybayer:
So I think maybe we can only retry if the error is a network error vs a server response error. We are using plain `fetch`, so would just need to figure out how to detect network error properly.

simonedelmann:
Seems like this guy has put a lot of research in detecting network errors with `fetch`. Should be easy to implement. 

https://medium.com/to-err-is-aaron/detect-network-failures-when-using-fetch-40a53d56e36

flybayer:
Perfect I think we should go with that

PR #908, opened 2020-08-18
Change `useQuery` to only retry on network errors by default

simonedelmann:
Closes: https://github.com/blitz-js/blitz/issues/886

Changed the `retryFunction` to detect and retry only network errors (in production).


flybayer:
Awesome, thank you!!

PR #916, opened 2020-08-18
Some improvements to auth in new app template

flybayer:

1. Rename `app/auth/index.ts` to `app/auth/auth-utils.ts` for clarity
2. Add `getCurrentUser` query
3. Add `useCurrentUser` hook

This also adds the above changes to the auth example.

### Checklist


dwightwatson:
These are great DX improvements.

Semi-related, I wonder if it's worthwhile providing some sort of CurrentUser context that can be used through the app, so that `getCurrentUser` will only be called once rather than by each component that uses it. Or does `useQuery` automatically prevent itself from making the same request multiple times?

flybayer:
@dwightwatson useQuery automatically de-dups requests :)

Issue #851, opened 2020-08-08
Auto run db migration during new app setup


flybayer:
### What do you want and why?

The new blitz app template on canary includes authentication and the database schema needed for that. Currently users have to manually run `blitz db migrate` after running `blitz new`. But we can automate that, so we should.

### Possible implementation(s)

Basically we need to add code to spawn the prisma CLI to the [app generator](https://github.com/blitz-js/blitz/blob/canary/packages/generator/src/generators/app-generator.ts#L163). You can see how we are already doing this in [db.ts](https://github.com/blitz-js/blitz/blob/canary/packages/cli/src/commands/db.ts)

You will want to pass the `--auto-approve` and `--name` arguments to prisma so that it doesn't prompt the user to enter a name for the migration.

Also, you'll want to use a `silent` spawn so users don't see the normal migration logging. Instead show a spinner like the rest of the new app setup.

PR #927, opened 2020-08-21
Automatically run initial db migrations during new app setup

karaggeorge:
Closes: #851 

This PR adds a few things:
- Refactors the db command to use `p-event` and abstract the calls to prisma. The logic with multiple nested callback funcs was a bit hard to follow. This way, using async/await I feel it's more readable.
- Add hidden option `name` to the db command which passes it to `prisma migrate save` call
- Adds the initial migration to the `new` command
- Adds the `yarn` or `npm install` step to the new command if the run is dry or `--skip-install` is passed
- Adds/fixes some tests in the db command
- Adds logic for testing `new` command stdout and instruction steps
- Adds a full install test for the `new` command

I'm not sure about the copy in many places, I can update if needed.

For implementation, since the recommendation is to run `blitz db migrate`, I figured it should *actually* call that command underneath. This way, if that command changes in the future, we don't have to duplicate the logic and maintain it both places. Only difference in this is that if something goes wrong, instead of hard exit, it'll just print the message including the step to call the `db migrate`.

For the spinner, I'm not sure about the behavior if something goes wrong. For now I just stopped/hid it. But we could also mark it as `fail()`

### Checklist

- [x] Tests added for changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -123,6 +123,7 @@
     "rollup-plugin-node-resolve": "5.2.0",
     "rollup-plugin-peer-deps-external": "2.2.2",
     "semver": "7.3.2",
+    "stdout-stderr": "0.1.13",
     "test-listen": "1.1.0",
     "ts-jest": "24.3.0",
     "tsdx": "0.13.1",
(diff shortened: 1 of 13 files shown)


flybayer:
@karaggeorge ok yeah idk! I added it for you

PR #936, opened 2020-08-23
Add `isLoading` property to `useSession()` hook result

alan2207:

I am adding the `isReady` piece of state in the `useSession` hook in order to check if the session has been initialised in the component. When using the `useSession` hook and checking for `session.userId`,  I am noticing a flash of unauthenticated view for a couple of milliseconds until the session data becomes available (until useEffect runs and gets the session data from `publicDataStore`). 

 It would be nice to be able to check if the session data is ready and display a loading spinner until the session data is actually available in the hook and decide what to render conditionally afterwards. I have added the usage in the auth example as well.

Dosc PR: https://github.com/blitz-js/blitzjs.com/pull/142

### DIFF
--- a/packages/core/src/supertokens.ts
+++ b/packages/core/src/supertokens.ts
@@ -1,5 +1,6 @@
-import {useState, useEffect} from "react"
+import {useState} from "react"
 import BadBehavior from "bad-behavior"
+import {useIsomorphicLayoutEffect} from "./utils/hooks"
 
 export const TOKEN_SEPARATOR = ";"
 export const HANDLE_SEPARATOR = ":"
@@ -140,15 +141,17 @@ publicDataStore.initialize()
 
 export const useSession = () => {
   const [publicData, setPublicData] = useState(emptyPublicData)
+  const [isLoading, setIsLoading] = useState(true)
 
-  useEffect(() => {
+  useIsomorphicLayoutEffect(() => {
     // Initialize on mount
     setPublicData(publicDataStore.getData())
+    setIsLoading(false)
     const subscription = publicDataStore.observable.subscribe(setPublicData)
     return subscription.unsubscribe
   }, [])
 
-  return publicData
+  return {...publicData, isLoading}
 }
 
 // Taken from https://github.com/HenrikJoreteg/cookie-getter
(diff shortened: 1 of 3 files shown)


flybayer:
Hey @alan2207 thank so you much for this! I was busy today, but I should be able to review tomorrow.

flybayer:
Ok @alan2207 yeah this looks good with one change. Let's rename `isReady` to `isLoading` so that's it's more consistent with what people expect like with Apollo, etc. It also is better if we ever decide to make an API will inside useSession.

packages/core/src/supertokens.ts
```diff
@@ -140,15 +140,17 @@ publicDataStore.initialize()
 
 export const useSession = () => {
   const [publicData, setPublicData] = useState(emptyPublicData)
+  const [isReady, setIsReady] = useState(false)
 
   useEffect(() => {
```
flybayer, line 145: Also as a small optimization, let's add this

```ts
const IS_SERVER = typeof window === 'undefined'

// React currently throws a warning when using useLayoutEffect on the server.
// To get around it, we can conditionally useEffect on the server (no-op) and
// useLayoutEffect in the browser.
const useIsomorphicLayoutEffect = IS_SERVER ? useEffect : useLayoutEffect
```

and then this:

```suggestion
  useIsomorphicLayoutEffect(() => {
```

alan2207: hey @flybayer , sounds good. How about we define `useIsomorphicLayoutEffect` inside `/utils/hooks.ts` and export it from there, in case it needs to be used somewhere else? Or would you rather just colocate it next to the `useSession` definition?

flybayer: Yeah that's fine to put it in utils/hooks :)

alan2207:
hey @flybayer , I think this is ready now. Thanks for the suggestions 👍

flybayer:
@alan2207 awesome!! One last thing, can you open a PR to add this to the auth docs for `useSession` on this page: https://blitzjs.com/docs/auth ? I should have mentioned that earlier

alan2207:
> @alan2207 awesome!! One last thing, can you open a PR to add this to the auth docs for `useSession` on this page: https://blitzjs.com/docs/auth ? I should have mentioned that earlier

Here you go: https://github.com/blitz-js/blitzjs.com/pull/142

flybayer:
Thanks!

@all-contributors add @alan2207 for code, docs

Issue #957, opened 2020-08-29
Return 404 when there's no default export from a query/mutation file


cherta:
### What do you want and why?

I think sometimes it's useful having code that's not automatically exposed as an endpoint, but it's sitting there in a query or a mutation file. For example, when doing trunk based development you want the code in place, but not yet exposed to the world.

A common way to do this are feature flags, but those carry their own issues, another common approach I take is to code the feature, but hide the endpoint. In rails this is easy to do because a controller needs to be connected on the routes file.

### Possible implementation(s)

I think this should be as simple as returning 404 if there's no default export on a query or a mutation. Currently it tries to apply a middleware to an undefined object and it fails.

### Additional context

N/A

flybayer:
Thanks @cherta! Your request definitely makes sense. My only thought is if there's some way we can distinguish between when there's intentionally no default export vs when it's accidental.

But probably either way a 404 or 500 error will cause the dev who did it accidentally to figure it out 🤔 

I say let's switch this behavior unless someone has more compelling reason not to.

cherta:
I see your point @flybayer, but for me the error of not exporting something should not be a runtime concern, but a build time one. I see this more as a linter / compile step issue than an execution problem.

Do you see this as a potential good first PR, or it's too complex for me to try solving it?

flybayer:
Yeah I think you could do it as a first PR.

Depending on where the actual runtime error orginates from, you'll probably want to [make the change here](https://github.com/blitz-js/blitz/blob/canary/packages/server/src/stages/rpc/index.ts#L96) or [here](https://github.com/blitz-js/blitz/blob/canary/packages/server/src/stages/rpc/index.ts#L68). And if default export isn't there, then use a default 404 handler code, something like this:

```
export default (req, res) => res.status(404).end()
```

PR #976, opened 2020-08-31
Return 404 if no default export is found on query or migrations

cherta:
Closes: #957

This PR changes the default behavior of the `isomorhicHandlerTemplate` and  `apiHandlerTemplate` to return 404 when there's no default export from a query or a mutation.

### Checklist

- [ ] Tests added for changes
- [X] PR submitted to [blitzjs.com](https://github.com/blitz-js/blitzjs.com) for any user facing changes


packages/server/src/stages/rpc/index.ts
```diff
@@ -65,12 +65,14 @@ const isomorhicHandlerTemplate = (
...
-  '${resolverType}',
-) ${useTypes ? "as typeof resolverModule.default" : ""}
+export default resolverModule.default
+  ? getIsomorphicRpcHandler(
+      resolverModule,
+      '${resolverPath}',
+      '${resolverName}',
+      '${resolverType}',
+    ) ${useTypes ? "as typeof resolverModule.default" : ""}
+  : undefined
```
yuta0801, line 75: I don't know why, but it seems `resolverModule` is an empty object on the front side so it doesn't work.

```suggestion
export default getIsomorphicRpcHandler(
  esolverModule,
  '${resolverPath}',
  '${resolverName}',
  '${resolverType}',
) ${useTypes ? "as typeof resolverModule.default" : ""}
```

yuta0801 (changes requested):

flybayer:
Hey @cherta I'm closing this from inactivity. You're welcome to re-open if you want to continue with it.

Issue #975, opened 2020-08-31
`session.isAuthorized` throws an error instead of returning a boolean


flybayer:
### What is the problem?

If a user is not logged in, `session.isAuthorized` throws an error instead of returning a boolean.

The problem [is this line of code](https://github.com/blitz-js/blitz/blob/canary/packages/server/src/supertokens.ts#L194).

We need to make this change:

```diff
-if (!this.userId) throw new AuthenticationError()
+if (!this.userId) return false
```

PR #977, opened 2020-08-31
Fix `session.isAuthorized` to not throw an error if user not logged in

dillonraphael:
Closes: #975

- Patch that fixes isAuthorized


flybayer:
@all-contributors add @dillonraphael for code

PR #980, opened 2020-09-01
fix(app-generator): init git repo prior to install

### LINKED ISSUE #973: Husky git hooks are not set up properly in new apps
### What is the problem?

Husky git hooks do not work right away after creating a new blitz app. You have to first run `npm rebuild husky`.

### Steps to Reproduce

1. `blitz new myapp`
2. See that `husky.sh` is missing from `myapp/.git/hooks/`

### Versions

```
0.18.0
```

### Other

I don't know 100% why this is happening, but probably it's because we install yarn dependencies before initializing the git repo. So we could either initialize the git repo before installing deps (but still wait to commit changes until the end). Or we can run `npm rebuild husky` after setting up the git repo.

[Here's the code where we initialize the git repo](https://github.com/blitz-js/blitz/blob/canary/packages/generator/src/generators/app-generator.ts#L214).

bpas247:
Initializes the git repo before it installs the dependencies,
ensuring that the husky hooks get properly set.

Fixes #973

### DIFF
--- a/packages/generator/src/generators/app-generator.ts
+++ b/packages/generator/src/generators/app-generator.ts
@@ -88,6 +88,17 @@ export class AppGenerator extends Generator<AppGeneratorOptions> {
   }
 
   async postWrite() {
+    let initResult
+    if (!this.options.skipGit) {
+      initResult = spawn.sync("git", ["init"], {
+        stdio: "ignore",
+      })
+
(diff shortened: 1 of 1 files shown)


flybayer (dismissed):

flybayer:
@bpas247 awesome, thank you!!

Issue #573, opened 2020-05-28
blitz db migrate should accept arguments to pass to prisma migrate


abuuzayr:
### What do you want and why?

`prisma migrate save` current accepts these parameters:

      --help / -h
      --name / -n
      --preview / -p
      --create-db / -c
      --experimental
      --schema

`blitz db migrate` runs `prisma migrate save` with the following parameters for us:

      --create-db 
      --experimental
      --schema

For most times, this is sufficient. But sometimes, we might want a bit more control especially to define the migration name via `--name`. Also, if prisma adds more parameters in future we are covered. 

This request is so that we can run `blitz db migrate` with additional parameters which would be passed to `prisma migrate save`

### Possible implementation(s)

Make changes here to accept args and send with the `spawn` function:

https://github.com/blitz-js/blitz/blob/dc9c9756ae85323db8d500d68d3f40286d27c018/packages/cli/src/commands/db.ts#L42

### Additional context

- What happens if use parameters that are already defined, like `--schema`?
- Should we allow to override `--schema`?

flybayer:
Great, thank you!

I think we should allow overriding any parameter that blitz includes by default.

kevotovar:
I want to work in this, Can I take this issue?

flybayer:
@kevotovar you bet!

flybayer:
Hey @kevotovar did you make any progress on this?

nitaking:
@flybayer 
Can I work on this?

flybayer:
@nitaking yes! Let me know if you need any help!

nitaking:
Even if the args are unsupported, you can run run runPrisma once and then run
Is it okay to go in the direction of showing "Uh oh, Blitz does not support that command." in case of an error?
(The rest of the logic remains the same.)

nitaking:
Supported arguments: Execute (as existing)
Not supported arguments : Execute runPrisma for now
　-> "not support" if an error occurs.

This is what I wanted to convey…😇
(I must learn more English!)

flybayer:
Ok yes I think that'll be fine. I can review for sure once you open a PR. By the way, you can open a PR now with what you have and click the button that marks it as "draft". If it's marked as draft, then we know it's not finished. After you finish, then mark as "ready for review".

Here's the [github docs for that](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-requests#draft-pull-requests).

nitaking:
Thank you for being so gracious!

PR #1012, opened 2020-09-05
Enable passing extra prisma args to `blitz db migrate`

nitaking:
Closes: https://github.com/blitz-js/blitz/issues/573

Use the db migrate flag for all cases where it is present.

### Checklist

- [x] Tests added for changes

### DIFF
--- a/packages/cli/src/commands/db.ts
+++ b/packages/cli/src/commands/db.ts
@@ -68,16 +68,17 @@ const runMigrateUp = async ({silent = false} = {}) => {
   return runPrismaGeneration({silent})
 }
 
-export const runMigrate = async (name?: string) => {
+export const runMigrate = async (flags: object = {}) => {
   if (process.env.NODE_ENV === "production") {
     return runMigrateUp()
   }
+  // @ts-ignore escape:TS7053
+  const nestedFlags = Object.keys(flags).map(key => [`--${key}`, flags[key]])
+  const options = ([] as string[]).concat(...nestedFlags)
 
-  const silent = Boolean(name)
-  const args = ["migrate", "save", schemaArg, "--create-db", "--experimental"]
-  if (name) {
-    args.push("--name", name)
-  }
+  const silent = options.includes('name')
+
+  const args = ["migrate", "save", schemaArg, "--create-db", "--experimental", ...options]
 
   const success = await runPrisma(args, silent)
 
@@ -182,6 +183,8 @@ ${chalk.bold("reset")}   Reset the database and run a fresh migration via Prisma
     name: flags.string({hidden: true}),
   }
 
+  static strict = false
+
   async run() {
     const {args, flags} = this.parse(Db)
     const command = args["command"]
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/db.ts
```diff
@@ -68,16 +68,17 @@ const runMigrateUp = async ({silent = false} = {}) => {
...
+  // @ts-ignore escape:TS7053
+  const nestedFlags = Object.keys(flags).map(key => [`--${key}`, flags[key]])
+  const options = ([] as string[]).concat(...nestedFlags)
 
-  const silent = Boolean(name)
-  const args = ["migrate", "save", schemaArg, "--create-db", "--experimental"]
-  if (name) {
-    args.push("--name", name)
-  }
+  const silent = options.includes('name')
```
nitaking, line 79: change code?

`options.length > 0`

packages/cli/src/commands/db.ts
```diff
@@ -68,16 +68,17 @@ const runMigrateUp = async ({silent = false} = {}) => {
...
 }
 
-export const runMigrate = async (name?: string) => {
+export const runMigrate = async (flags: object = {}) => {
   if (process.env.NODE_ENV === "production") {
     return runMigrateUp()
   }
+  // @ts-ignore escape:TS7053
+  const nestedFlags = Object.keys(flags).map(key => [`--${key}`, flags[key]])
+  const options = ([] as string[]).concat(...nestedFlags)
```
aem, line 77: shouldn't this do the same?

```suggestion
  const options = nestedFlags as string[]
```

packages/cli/src/commands/db.ts
aem: i think there's a usecase for wanting the wizard experience but also providing custom options, I'd leave it like this

aem (approved):
Thanks for tackling this! Just one suggestion

packages/cli/src/commands/db.ts
nitaking: Thank you reviewed!

Fixed it! https://github.com/blitz-js/blitz/pull/1012/commits/1e062d209612ed49fd64a3e4eb2838bc01a5f6c4 ⚡️

nitaking: I forgot to mention that it's a fix for making it a flat because it's a nest array.

Before:
```ts
  const nestedFlags = Object.keys(flags).map(key => [`--${key}`, flags[key]])
  const options = ([] as string[]).concat(...nestedFlags)
  
  console.debug('options', options)

=>     options [ '--name', 'name' ]
````

https://github.com/blitz-js/blitz/commit/1e062d209612ed49fd64a3e4eb2838bc01a5f6c4:

````ts
const formatFlags = Object.keys(flags).map(key => [`--${key}`, flags[key]])
  const options = formatFlags as unknown as string[]
  
  console.debug('options', options)

=>    options [  [ '--name', 'name' ] ] 
````

nitaking: I reverted.
Are there any corrections to this?

aem (approved):
looks good! @flybayer any concerns about passing all args through?

nitaking:
@flybayer Can you confirm this? :)

flybayer:
Sorry for the delay, looks good! Thank you!!

@all-contributors add @nitaking for code

nitaking:
Thanks!!!

PR #1014, opened 2020-09-06
refactor(server): Added isTypescript explicit flag, config refactored

svobik7:
This PR exists just because of initial idea (https://github.com/blitz-js/blitz/discussions/941) of refactoring server config and try to make them cacheable and in the same time drop the need of prop drilling. As initial idea of caching leads to no benefits this PR just brings some config refactors, code cleanup and small perf optimization flag in server package.

- keeps `isTypescript` prop just in server config and server package and removes it from file pipelines package
- removes waiting for both next bin and patched bin as we do not switch between dev/prod in runtime
- removes some unnecessary lines and clean up the config code


ryardley:
Hey @svobik7 Thanks for this! I will need a couple of days to look at this properly. Looks good so far.

packages/server/src/config.ts
```diff
@@ -48,27 +59,44 @@ const defaults = {
...
-  writeManifestFile: true,
 }
 
-export async function normalize(config: ServerConfig): Promise<NormalizedConfig> {
-  const nextBinOrig = await resolveBinAsync("next")
-  const nextBinPatched = await resolveBinAsync("@blitzjs/server", "next-patched")
-  const git = parseChokidarRulesFromGitignore(resolve(process.cwd(), config.rootFolder))
+export async function normalize(
+  config: ServerConfig,
+  dev: boolean = false,
```
ryardley, line 66: Any reason why not just add `dev` to `ServerConfig`?

packages/server/src/config.ts
```diff
@@ -48,27 +59,44 @@ const defaults = {
...
-    isTypescript: config.isTypescript ?? true,
+    // -
+    ignore: defaults.ignoredPaths.concat(git.ignoredPaths),
+    include: defaults.includePaths.concat(git.includePaths),
+    // -
+    nextBin: await getNextBin(rootFolder, dev),
   }
 }
+
+async function getNextBin(rootFolder: string, dev: boolean = false) {
```
ryardley, line 91: Good work putting these together

packages/cli/src/commands/start.ts
```diff
@@ -24,17 +18,19 @@ export class Start extends Command {
       char: "H",
       description: "Set server hostname",
     }),
+    ts: flags.boolean({
+      description: "Use typescript explicitly (just for performance optimization)",
+    }),
```
ryardley, line 23: Is this really actually necessary? Seems like a slightly redundant feature when we are only avoiding a fs read at the start of a cli process

packages/server/src/config.ts
```diff
@@ -48,27 +59,44 @@ const defaults = {
...
+  // do not await for both bin-pkg because just one is used at a time
+  const nextBinPkg = dev ? "@blitzjs/server" : "next"
+  const nextBinExec = dev ? "next-patched" : undefined
+  const nextBin = await resolveBinAsync(nextBinPkg, nextBinExec)
+  return resolve(rootFolder, nextBin)
+}
+
+async function getIsTypescript(rootFolder: string) {
+  const fs = await import("fs")
+  return fs.existsSync(join(rootFolder, "tsconfig.json"))
```
ryardley, line 101: This doesn't really matter as it is not in stream but why not use a promisified `exists` function like you might from using promisify with fs.exists. If we had to run concurrent processes later that needed this function it wouldn't block the thread that way also why not just import fs up top?

ryardley (commented):
I kind of question why we are adding a flag here and if that is really worth it from a devx perspective? Also it is kind of hard to explain what it is doing and I am not really convinced we need to do it.

packages/cli/src/commands/start.ts
svobik7: The main reason was to find a way to avoid `fs.existsSync` called in config.ts which just checks if `tsconfig.json` file exists. I was thinking like following:

- As a newcomer to blitz I want to create and start my application with almost zero-config so I run `blitz start` and everything works as expected. 
- Once I settled my initial code base (js, ts) I want to avoid any unnecessary "delays" in dev process. So I have got an option to tell blitz "hey I am using TS, do not check it on your own" by running `blitz start --ts`.

Also this flag can be used in different commands (generate, console) to avoid checking ts implicitly. But honestly I did not make any perf tests and cannot say now if it is worth. So I do not have any compelling reason to keep this flag right now.

packages/server/src/config.ts
svobik7: - Importing `fs` on top: this is related to `--ts` flag. Once I was thinking about avoiding `fs.existsSync` I was already thinking about `do not import anything unnecessary if you know typescript is set explicitly`. 
- existsSync: it could be definitely done with promisified `exists`

I see now `not importing fs` approach is not very benefical hera as `fs` imported already for other reasons. So I agree to use promisified check and import from fs directly would be better.

I am gonna update this, ok?

packages/server/src/config.ts
svobik7: Thank you 👍

packages/server/src/config.ts
svobik7: There is no special reason for this. It could definitely be inside `ServerConfig`. In that case I suggest to use it like following:

```js
export type ServerConfig = {
  env: "dev" | "prod"
  ...
}
```

While 'prod' is default. What do you think?

ryardley: Works for me

packages/cli/src/commands/start.ts
ryardley: I think there are other way more expensive things we are doing as the command starts up that are going to be slower than one file stats read especially if we do any file pipeline work.

packages/server/src/config.ts
svobik7: Done in current commit

packages/cli/src/commands/start.ts
svobik7: Removed in current commit

packages/server/src/config.ts
svobik7: Done in current commit

ryardley (approved):
Looks good to me. I removed the reference to --ts config option in the PR description.

Issue #236, opened 2020-04-25
Add custom templates support


flybayer:
**What do you want and why do you want it?.**

Blitz comes with default templates for `blitz generate`. This is a feature that seems really cool to start with but very quickly becomes just a toy because you can't use it in real projects. This was the downfall of `rails generate`. 

However, `blitz generate` can be super useful throughout the entire life of a project if we have an easy way to eject templates and let users customize them.

It's not in scope of this issue, but we'll also have ability for users or plugins to add generators to the `blitz generate` command.

**Implementation**

I don't really have any thoughts about this at the moment, except that I think we should probably eject to `/templates/*`.

I'd love if someone can think about how this should all work and then post a proposal here.

aem:
Ok I've done some thinking and here's how I'd like to go about this:

1. Add a new, meta `blitz generate template [type] [name]` command. It'll accept the same `type`s as `blitz generate`, and will generate a template called `[name]` in `/templates/[name]`, with files copied directly from our existing templates. From there, the user can modify the template however they want. These templates should also have some method of getting values into the template, maybe via a `.getTemplateValues.ts` that lives at the root of the template. We can enforce this file supplies the values needed to conform to the `Generator`'s interface.
2. Update `blitz generate [type] [name]` to also look for `[type]`s in `/templates/*`. Since we can read these values from the filesystem during CLI startup, we could even include these template values in the help menu for `blitz generate`. We'll coalesce the methods/fields provided in `.getTemplateValues.ts` with the abstract `Generator` class to dynamically create a generator and place its files where instructed based on the CLI command, just like a normal `generate` invocation.

flybayer:
Nice!

- Can we use "eject" since folks are so common with that
	- So like  `blitz generate eject ...args` 
- Why do we need both `type` and `name`?
- Are we ejecting the actual generator class too or only the template? I was initially thinking the later, but I think we should do the former. Then users can even create totally new generators like `component` for example. So like `blitz generate generator component`

ryardley:
Awesome feature idea but personally I would question if "eject" is the best terminology here. Eject to me feels like a permanent change where I as a dev have no further use for the cli since everything is now available as a script that I can fully customize.  Would "custom generators" be a better way to describe the feature?

flybayer:
Even if we make it clear you are "ejecting a template" (or generator) instead of "ejecting blitz cli"?

This would be on a doc page under advanced generator usage. So I don't think it would be confusing.

[...]

PR #1018, opened 2020-09-08
Custom template authoring

aem:
Closes: https://github.com/blitz-js/blitz/issues/236

Introduces custom template support into the `generate` command. There's a bit of necessary complexity here that I'll try to explain clearly. Still working on the docs PR.

* Templates are generated via a special generator type - `blitz g template my-template type:[type]`. Creating templates via a separate CLI command would've required duplicating a lot of the logic in `generate`, and given that a command like `blitz generate template` is still pretty self-explanatory, I decided to minimize complexity here. The `type` param (specified the same way we specify options in the `ModelGenerator`) can be any valid resource type (`page`, `queries`, `form`, etc.). 
* Templates are a folder of source files, and live in `project-root/templates`. Templates of a specific type (`page`, `query`, `mutation`, etc) live in an identically-named subdirectory. For example, a custom `page` template might exist at `my-app/templates/page/my-custom-page/__modelName__.ts`. This is required so the generator knows to place special files (pages, queries, mutations) in their "magic" subdirectories. Without the nesting we wouldn't have a way to tell different page types apart.
* All custom templates support the standard `--context` option, as well as any options specific to that type (e.g. all custom `pages` templates will be passed the metadata normally passed to the `PagesGenerator` when supplying an option like `--parent=SomeModel`).
* One custom templates are created, they are accessible just like any built-in template type: `blitz g my-custom-page adminHome`. **Custom templates override built-in templates.** For example, templates in `project-root/templates/pages/pages` would be used instead of the built-in templates when running `blitz g pages [model]`. This means the entire functionality of the generator could be totally customized by the host app. 

### Checklist

- [ ] Tests added for changes


flybayer:
Sweet! I like the `blitz g template` command.


The `templates/[templateType]/[templateName]` structure certainly works, but it does seem a but clunky. What about something like `templates/[templateType]__[templateName]`? And [templateName] can be optional.

```
templates/page/__modelName__.ts           <-- default name: blitz g page
templates/page__authPage/__modelName__.ts  <-- custom name: blitz g authPage
```

And another way would be to define the type with a config file somehow. I think I like this option the best. The folder structure is much less confusing!

```
templates/page/
   - config.json                <-- has {type: "page"}
   - __modelName__.ts           <-- default name: blitz g page
templates/authPage/
   - config.json                <-- has {type: "page"}
   - __modelName__.ts  <-- custom name: blitz g authPage
```

Also, I think we should change the `__modelName__` syntax because it doesn't make sense for custom templates where there isn't actually a "model" like a single query, single component, etc. Maybe something like `__itemName__`.

Lastly, something else to consider in this implementation is how can we add support for custom generators, not just custom templates? Couple use cases:

- How can I add a test file to the query generator?
- I want a way to customize `blitz g all modelName` so that it also adds a menu item to my main nav bar as "Model Names".

This is definitely getting super complex, but I think it's totally worth it! Without this type of power and customization, code scaffolding only stays in the realm of demos and prototypes.

aem:
@flybayer 

> The templates/[templateType]/[templateName] structure certainly works, but it does seem a but clunky. What about something like templates/[templateType]__[templateName]? And [templateName] can be optional.

What about doing it as a part of the folder name but with `.` instead of `__`? Just since `__` is already a significant operator in the templating language, I'd be concerned about overloading its meaning here. `templates/authPage.page/auth.tsx`? Or `page.authPage` which would let us group templates by type when looking at the folder structure, if that feels better to you

> And another way would be to define the type with a config file somehow

That's an interesting idea. Could we configure this mapping via `blitz.config.js`? 

One other thought I had was with a "magic file" - like an empty `.blitz-template-page` file at the top level of the template dir. That means we don't have to maintain a whole separate config setup for templates, we could just explicitly exclude "hidden" files from the `mem-fs-copy` command and then it's both very visible what kind of template it is, minimizes the amount of code we have to maintain for it, and is pretty much infinitely extensible.

> how can we add support for custom generators, not just custom templates?

I think we can do something similar to the "magic file" system here. If we go with a naming scheme like `templates/authPage.page/*` for the folder structure, then maybe the generators can just be sibling files named by convention?

```
templates
├── authPage.generator.ts
└── authPage.page
    └── __modelName__Auth.tsx
```

flybayer:
Using `.` in the folder name is a great idea!

And `authPage.generator.ts` is nice too!

> One other thought I had was with a "magic file" - like an empty `.blitz-template-page` file at the top level of the template dir. That means we don't have to maintain a whole separate config setup for templates, we could just explicitly exclude "hidden" files from the `mem-fs-copy` command and then it's both very visible what kind of template it is, minimizes the amount of code we have to maintain for it, and is pretty much infinitely extensible.

I'm not totally following what you are saying here 🤔

Issue #1016, opened 2020-09-07
Navigating to the `/login` route in a freshly bootstrapped application throws a warning with React Developer Tools in Firefox


madflow:
### What is the problem?

Navigating to the `/login` route in a freshly bootstrapped application throws a warning with React Developer Tools in Firefox.

It is just a "warning" - maybe intentional, dunno. Still wanted to let you know...

### Steps to Reproduce

1. blitz new blitz-app
   1. Choose `Formik` as the form library
1. cd blitz-app
1. blitz start
1. Navigate to /login.

### Versions

```
Linux 4.15 | linux-x64 | Node: v12.18.0

blitz: 0.21.1 (global)
blitz: 1.0.0 (local)

  Package manager: yarn 
  System:
    OS: Linux 4.15 Ubuntu 18.04.5 LTS (Bionic Beaver)
    CPU: (4) x64 Intel(R) Core(TM) i7-4600U CPU @ 2.10GHz
    Memory: 2.10 GB / 11.61 GB
    Shell: 5.4.2 - /usr/bin/zsh
  Binaries:
    Node: 12.18.0 - ~/.nvm/versions/node/v12.18.0/bin/node
    Yarn: 1.22.5 - /usr/bin/yarn
    npm: 6.14.4 - ~/.nvm/versions/node/v12.18.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  npmPackages:
    @prisma/cli: Not Found
    @prisma/client: Not Found
    blitz: Not Found
    react: Not Found
    react-dom: Not Found
    typescript: Not Found
```

### Other

```
Warning: A component is changing an uncontrolled input to be controlled. This is likely caused by the value changing from undefined to a defined value, which should not happen. Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components
input
label
div
_c<@webpack-internal:///./app/components/LabeledTextField.tsx:26:14
form
Formik@webpack-internal:///../../../node_modules/formik/dist/formik.esm.js:1136:28
Form@webpack-internal:///./app/components/Form.tsx:32:18
div
LoginForm
div
LoginPage@webpack-internal:///./pages/login.tsx:21:71
Layout@webpack-internal:///./app/layouts/Layout.tsx:14:15
ErrorBoundary@webpack-internal:///../../../node_modules/react-error-boundary/dist/react-error-boundary.umd.js:43:37
[...]

PR #1019, opened 2020-09-08
Fix react warnings from Formik in new apps

alan2207:
Closes: https://github.com/blitz-js/blitz/issues/1016

- added `initialValues` prop to both the login and the signin form - missing it caused the warning on both /login and /signup pages
- moved the sign in form into a component so it is consistent with the login component


flybayer:
Awesome, this is perfect, thank you!!

Issue #861, opened 2020-08-11
Blitz start --production rebuilds the app on Heroku


gabrielcolson:
### What is the problem?

The `blitz start --production` command rebuilds the app during Heroku deployment even if it was previously built during Heroku's build phase. The application timeout on start after 60 seconds.

### Steps to Reproduce

1. create a new project `blitz new blitz-heroku && cd blitz-heroku`
2. create a new heroku app `heroku create`
3. create a root `Procfile` containing `web: npx blitz start --production -p $PORT`
3. deploy to Heroku:
```
git add .
git commit -m "first commit"
git push heroku master
```
4. go to https://your-app-name.herokuapp.com
5. the app crash. You can see in the logs (`heroku logs`) that this is because of the start command timeout.

### Versions

```
debug: blitzPkgPath: /Users/gabrielcolson/Documents/blitz-heroku/node_modules/blitz/dist/index.js
debug: cliPkgPath: /Users/gabrielcolson/Documents/blitz-realworld/node_modules/blitz/node_modules/@blitzjs/cli/lib/src/index.js

macOS Catalina | darwin-x64 | Node: v12.16.3

blitz: 0.17.0 (global)
blitz: 0.17.0 (local)
```

### Other

The workaround I found was to replace the content of my Procfile by `web: npx next start -p $PORT`. It then worked perfectly.

flybayer:
Hey @gabrielcolson can you confirm that you ran `blitz build` during the Heroku build phase?

Assuming you did that, do you know how Heroku decides which files to copy to the runtime environment? I wonder if it's not copying both the `.next` and `.blitz` folders to the runtime image.

gabrielcolson:
Hey @flybayer 
Yes the `blitz build` is indeed executed we can see that in the logs.

I believe the output is copied "as is" to the runtime image after pruning  the dev dependencies

flybayer:
Ok cool, someone just needs to dig into this and figure out exactly why our diff algorithm thinks the files are changed. (`blitz start --production` will re-use the built files unless it detects a change)

ryardley:
Is there a different lock file or something like that being added to the folder as part of the dev dependency removal?

If any files change aside from the following, blitz will trigger a rebuild:

```
node_modules
.blitz-build
.blitz
cypress
.next
```

gabrielcolson:
@ryardley I couldn't say. I cannot ssh into the runtime container since it didn't even start.
[...]

PR #1023, opened 2020-09-08
Fix Heroku issue where `blitz start --production` always rebuilds even after running `blitz build`

anteprimorac:
Closes: #861 

Fixes heroku rebuilding app when running start command by ignoring `.heroku` and `.profile.d` folders in artefacts hash.


flybayer:
Awesome, thank you so much!!

PR #1034, opened 2020-09-09
Fix: export `AppProps` as a generic type

anteprimorac:
Exports AppProps type as a generic type.


flybayer:
Thank you!!

PR #1039, opened 2020-09-10
Add ability to properly type `session.userId` (in new apps by default, existing apps must opt-in)

flybayer:

This adds the ability to properly type `session.userId` so that it matches the actual type of the userId in your schema.

This is enabled by adding the following code to your app, like in `/types.ts`. This file is now in new apps by default. Existing apps will need to add this if you want to benefit from this. Without this file, `userId` remains as `any` like it has been for a good while.

```ts
import { DefaultAuthTypes } from "blitz"
import { User } from "db"

declare module "blitz" {
  export interface AuthTypes extends DefaultAuthTypes {
    userId: User["id"]
  }
}
```

### Checklist


flybayer:
Waiting to merge this until we rework ctx, after which we can make `session.authorize()`. Adding  this change without that assertion function type working will require doing `userId as number` which sort of defeats the purpose of it.

flybayer:
Succeeded by https://github.com/blitz-js/blitz/pull/1219

Issue #942, opened 2020-08-25
Blitz deploy fails on Render.com


SigurdMW:
### What is the problem?
Blitz fails when deploying to Render.com with the following `render.yaml` file:
```yml
services:
  - type: web
    name: someweb
    env: node
    plan: starter
    buildCommand: yarn; blitz db migrate; blitz build
    startCommand: blitz start --production -H 0.0.0.0
    envVars:
      - key: NODE_ENV
        value: production
      - key: DATABASE_URL
        fromDatabase:
          name: somedb
          property: connectionString
databases:
  - name: somedb
    plan: starter
```

### Steps to Reproduce

1. Install a brand new Blitz app
2. Add `render.yaml` file as above
3. See it fail :/ 

### Versions

happens both on version 0.18.0 and 0.17.1-canary.7

### Other
==> Starting service with 'blitz start --production -H 0.0.0.0'
Aug 25 08:08:24 PM  You are using alpha software - if you have any problems, please open an issue here:
Aug 25 08:08:24 PM      https://github.com/blitz-js/blitz/issues/new/choose
Aug 25 08:08:24 PM  
Aug 25 08:08:30 PM  - Preparing for launch
Aug 25 08:08:42 PM  ✔ Prepped for launch
Aug 25 08:08:45 PM  warn  - You have enabled experimental feature(s).
Aug 25 08:08:45 PM  warn  - Experimental features are not covered by semver, and may cause unexpected or broken application behavior. Use them at your own risk.
Aug 25 08:08:45 PM  
Aug 25 08:08:45 PM  warn  - No build cache found. Please configure build caching for faster rebuilds. Read more: https://err.sh/next.js/no-cache
Aug 25 08:08:45 PM  info  - Creating an optimized production build...
Aug 25 08:08:45 PM  Attention: Next.js now collects completely anonymous telemetry regarding usage.
Aug 25 08:08:45 PM  This information is used to shape Next.js' roadmap and prioritize features.
Aug 25 08:08:45 PM  You can learn more, including how to opt-out if you'd not like to participate in this anonymous program, by visiting the following URL:
Aug 25 08:08:45 PM  https://nextjs.org/telemetry
Aug 25 08:08:45 PM  
Aug 25 08:09:02 PM  Failed to compile.
Aug 25 08:09:02 PM  
Aug 25 08:09:02 PM  ./test/utils.tsx:3:41
[...]

PR #1045, opened 2020-09-11
Fix bug where test files would cause `blitz build` to fail

SigurdMW:
Update the default ignore path ref #942

Closes: #942

Ignoring test files so that build doesn't fail due to missing depts

### Checklist

- [ ] Tests added for changes


flybayer:
Sweet, thank you!!

Issue #1032, opened 2020-09-09
`blitz generate query` does not preserve plurality


flybayer:
### What is the problem?

`blitz g query getPrices` generates `app/queries/getPrice.ts` instead of `app/queries/getPrices.ts`. And the name of the query inside the file is `getPrice` instead of `getPrices`

### Steps to Reproduce

1. `blitz g query getPrices`

### Versions

```
0.21.1
```

### Other

The issue is that [this line](https://github.com/blitz-js/blitz/blob/canary/packages/generator/src/generators/query-generator.ts#L15) uses `modelName` which is always singular. Also, the template name is [`__modelName__.ts`](https://github.com/blitz-js/blitz/blob/canary/packages/generator/templates/query/__modelName__.ts) which is singular for the same reason.

We might have to add a `rawInputName` to the Generator parent class for things like this.

engelkes-finstreet:
I would like to look into this

flybayer:
@engelkes-finstreet ok, all yours!

PR #1083, opened 2020-09-14
Fix `blitz generate query` does not preserve plurality

engelkes-finstreet:
…use this rawInput as its templateValue.

Closes: https://github.com/blitz-js/blitz/issues/1032

The queryGenerator was using modelName which did not preserve plurality. `blitz generate query getProjects` created a query with the name `getProject.ts`. 

Using the rawInput creates the correct file- and functionName.


flybayer:
Thank you @engelkes-finstreet!

Issue #1022, opened 2020-09-08
`blitz generate pages` should kebab-case all url paths.


flybayer:
### What do you want and why?

Currently if you generate pages with two words, like `blitz generate testResult`, the resulting urls will have `/testResults`, but I want it to be `/test-results`.

### Possible implementation(s)

[Here's the place](https://github.com/blitz-js/blitz/blob/canary/packages/generator/src/generators/page-generator.ts#L57) where the pages file path is determined.

aem:
I actually noticed this with the custom generators as well and meant to open this issue 😅 - I think my plan is to change this code to line up with the model generation logic - don't change any casing at all. I think we still want to enforce pluralization in certain places, but at least let people stick with camel/kebab if they'd like

flybayer:
> don't change any casing at all... at least let people stick with camel/kebab if they'd like

Are you saying to not automatically kebab-case url paths?

aem:
> Are you saying to not automatically kebab-case url paths?

That's my gut reaction, to just not be opinionated about it - I definitely know people who prefer camelcase URL paths

cr101:
For SEO best practice use kebab-case url paths. Using a hyphen in your URLs is recommended by Google, because it makes your website easy to read for humans. As an end result, this means that your site will place better on search engines.

flybayer:
Looks like the consensus from slack is that people do what kebab case by default. If someone wants something else, they can use custom templates.

aem:
Sounds good to me. I'll try and pick this up once I clean up + merge the
custom templates PR, if no one else has

engelkes-finstreet:
Are you already working on this @aem? Else I could take this over.

flybayer:
@engelkes-finstreet it's all yours!

PR #1089, opened 2020-09-15
Fix `blitz generate pages` to use kebab-case for all url paths

engelkes-finstreet:
Closes: https://github.com/blitz-js/blitz/issues/1022

If you run `blitz generate pages testResult` right now the generated URLs for this would container `testResult`. For better SEO all urls should be kebab-case. Running the above command now generates the URLs in kebab-case form (`test-result`).


flybayer:
Sweet, thanks! Can you add some unit tests for `camelCaseToKebabCase`? That's a gnarly regex :D

flybayer:
Thank you!

Issue #1055, opened 2020-09-12
Add automatic query cancelation for when queries become inactive (like off screen)


flybayer:
### What do you want and why?

By default React-query let's requests finish if they become inactive. But it does allow you to configure query cancellation. This isn't a big deal, but it's an optimization we might as well make!

### Possible implementation(s)

Implement the [react-query cancellation](https://react-query.tanstack.com/docs/guides/query-cancellation) feature [at this line](https://github.com/blitz-js/blitz/blob/canary/packages/core/src/use-query.ts#L52).

Also need to add that in the `use-paginated-query.ts` and `use-infinite-query.ts` files.

hardfire:
@flybayer let me take this.

flybayer:
@hardfire All yours!

hardfire:
okay, this isn't as straightforward as i initially thought it would be. some observations
 - [here](https://github.com/blitz-js/blitz/blob/canary/packages/core/src/use-query.ts#L52) `queryRpcFn` should be the same as `queryFn` in browser context, but it isn't as [this](https://github.com/blitz-js/blitz/blob/canary/packages/core/src/rpc.ts#L145) should be executed at some point.
 - since the promise is never returned [here](https://github.com/blitz-js/blitz/blob/canary/packages/core/src/rpc.ts#L49) we will have to jump some more hoops so that react query can also call cancel on the promise.

flybayer:
[rpcFn](https://github.com/blitz-js/blitz/blob/canary/packages/core/src/rpc.ts#L152) is what is passed into our `useQuery`. Feel free to add more comments in there! And refactor anything that needs it.

I think what we can do is instantiate the abort controller at [this line](https://github.com/blitz-js/blitz/blob/canary/packages/core/src/use-query.ts?rgh-link-date=2020-09-12T20%3A06%3A29Z#L52) and pass `signal` down through `rpcFn` to `executeRpcCall` where you set it on `fetch`.

Something like this:

```ts
queryFn: (_apiUrl, params) => {
  // Create a new AbortController instance for this request
  const controller = new AbortController()
  // Get the abortController's signal
  const signal = controller.signal

  const promise = queryRpcFn(params, {fromQueryHook: true, signal})

  // Cancel the request if React Query calls the `promise.cancel` method
  promise.cancel = () => controller.abort()

  return promise
}
```

hardfire:
o yeah! i had something similar as well. 
[...]

PR #1092, opened 2020-09-16
Enable automatic network request cancelation for queries/mutations

hardfire:
Closes: https://github.com/blitz-js/blitz/issues/1055

I realized that instead of making changes in all the `use-*-query` hooks this could be handlled independently in the `executeRpcCall` . This does add a bit of complexity to `rpc.ts` but i'll take a stab to refactor it and still have to test it. 

### Checklist

- [ ] Tests added for changes


flybayer:
@hardfire I marked this PR as a "draft" instead of using "WIP" in the title. So when you are done, you can click the "Ready for review" button to take it out of "draft" status.

hardfire:
awesome! didn't know of the "draft" feature. Also, do review if the approach seems right.

flybayer:
@hardfire ah nice. Did you confirm that that works?

Let's change this code to only use one `.then()`. Make that handler async, and use `await` inside that. This will make it cleaner.

```ts
.then(async (result) => {
  // stuff
  try {
    const json = await result.json()
    // more stuff
  } catch(err) {
    //stuff
  }
})

hardfire:
> @hardfire ah nice. Did you confirm that that works?

Yes! tested it manually on one of the examples. 
Moved all of the then's to a single async function.

Its weird why the CI/Lint fails though.

flybayer (approved):

flybayer:
Thank you!!

@all-contributors add @hardfire for code

PR #1105, opened 2020-09-17
🐛 fix query + page generator for default setup

brunocrosier:
When running a fresh install of Blitz, the default homepage says to run:

`blitz generate all project name:string`
`blitz db migrate`
`blitz start`

and then to navigate to `/projects`

however, because the database is empty and the `getProjects.ts` file generated by Blitz is relying on `db.project.findMany` it causes an error because `db.project` is `undefined`

A similar thing happens with the `/pages/projects/index.tsx` file that is generated by Blitz, which has `projects.map` and errors out because `projects` is `undefined`.

Hopefully this should fix that!

but this results in 
### Checklist

- [ x] Tests added for changes

### DIFF
--- a/packages/generator/templates/page/index.tsx
+++ b/packages/generator/templates/page/index.tsx
@@ -27,7 +27,7 @@ export const __ModelNames__List = () => {
     return (
       <div>
         <ul>
-          {__modelNames__.map((__modelName__) => (
+          {__modelNames__?.map((__modelName__) => (
             <li key={__modelName__.id}>
               <Link
                 href="/__parentModels__/__parentModelParam__/__modelNames__/__modelIdParam__"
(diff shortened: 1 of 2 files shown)


flybayer:
Hey @brunocrosier! If `db.project` is undefined, then it means you probably ran `blitz db migrate` without restarting the blitz server afterwards. `db.project` will be defined, even when the db is empty.

Can you try again on a fresh app and make sure you restart the blitz server after `blitz db migrate`?

brunocrosier:
Aahhh that would make sense! Maybe worth mentioning this on the default homepage in that case so others don't get caught out by it?

flybayer:
@brunocrosier yes! Want to make a PR for that?

That template file is here: https://github.com/blitz-js/blitz/blob/canary/packages/generator/templates/app/app/pages/index.tsx

Issue #1056, opened 2020-09-13
`blitz start` is significantly slower to start than `next dev`


flybayer:
### What is the problem?

Next.js re-uses it's own build cache between `next dev` invocations so that subsequent runs of `next dev` start extremely fast.

But the problem is that we are deleting the entire `.blitz` cache folder, including the next.js cache. This means that `blitz start` is always slow to start.

### Steps to Reproduce

1. Run `blitz start` several times and notice the start time.
2. Run `yarn next .blitz/caches/dev` several times and notice the much faster start time

### Versions

```
0.22.0
```

cc @ryardley

ryardley:
Be happy jump on this

PR #1115, opened 2020-09-19
Move next folder outside of cache

ryardley:
Closes: #1056 

| Stage | First run | Second Run |
|------|-------|------|
| Original |   40s | 32s |
| Move `.next` outside cache folder |  36s | 18s|

### Checklist

- [ ] Tests added for changes

### DIFF
--- a/packages/server/src/build.ts
+++ b/packages/server/src/build.ts
@@ -35,7 +35,7 @@ export async function build(
   await nextBuild(nextBin, buildFolder)
 
   const rootNextFolder = resolve(rootFolder, ".next")
-  const buildNextFolder = resolve(buildFolder, ".next")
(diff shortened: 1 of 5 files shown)


flybayer:
Sweet! Is the `.next` folder now at the project root?

ryardley:
Nah gave up and as to go grab dinner. I was having a few issues working out what was happening around the supertoken `__db` hacks when you do this as there is an expectation that .next lives relative to the cache root. Also this doesn’t deliver as much performance as I was expecting. Today I want to attempt a different method where we don’t blow away the cache and keep track of work done as that should speed up things quite a bit.

flybayer:
Ok sounds good!

Issue #1094, opened 2020-09-16
ALPHA software warning is being added to .zshrc when setting up autocomplete


pbteja1998:
### What is the problem?
When trying to add `blitz autocomplete` with the given instructions, the following is also being appended to `.zshrc` since `blitz` command shows this every time.

```
You are using alpha software - if you have any problems, please open an issue here:
    https://github.com/blitz-js/blitz/issues/new/choose
```

### Steps to Reproduce

1. Run `printf "$(blitz autocomplete:script zsh)" >> ~/.zshrc; source ~/.zshrc` in terminal. 

### Versions

```
output of `blitz --version --verbose`
You are using alpha software - if you have any problems, please open an issue here:
    https://github.com/blitz-js/blitz/issues/new/choose

debug: blitzPkgPath: /Users/bhanuteja.p/.nvm/versions/node/v12.16.2/lib/node_modules/blitz/dist/index.js
debug: cliPkgPath: /Users/bhanuteja.p/.nvm/versions/node/v12.16.2/lib/node_modules/blitz/node_modules/@blitzjs/cli/lib/src/index.js

macOS Catalina | darwin-x64 | Node: v12.16.2

blitz: 0.22.1 (global)

  Package manager: yarn
  System:
    OS: macOS 10.15.6
    CPU: (8) x64 Intel(R) Core(TM) i5-8279U CPU @ 2.40GHz
    Memory: 274.18 MB / 8.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 12.16.2 - ~/.nvm/versions/node/v12.16.2/bin/node
    Yarn: 1.22.4 - ~/.yarn/bin/yarn
    npm: 6.14.4 - ~/.nvm/versions/node/v12.16.2/bin/npm
    Watchman: Not Found
  npmPackages:
    @prisma/cli: Not Found
    @prisma/client: Not Found
    blitz: Not Found
    react: Not Found
    react-dom: Not Found
    typescript: Not Found
```

### Other

Please include applicable logs and screenshots that show your problem.
NOT APPLICABLE

### Potential Solution

I think just changing the documentation should suffice here. If needed, I could change the documentation by adding this info.

flybayer:
Oops, thanks for reporting this! This is now marked available for anyone to work on.

pbteja1998:
Hey @flybayer I would like to work on this. Can you please assign it to me?

enricoschaaf:
[...]

PR #1122, opened 2020-09-19
Fix `blitz autocomplete` adding alpha warning message to the shell config

enricoschaaf:
Closes: #1094 

Check before printing Alpha software warning, if command is autocomplete:script without any flags, dont print the warning. That way the warning doesnt land in the config file of the terminal.

### Checklist

- [ ] Tests added for changes
No tests for this that I am aware of.
PR fixes a bug, so no update to the documentation necessary.

### DIFF
--- a/packages/blitz/src/bin/cli.ts
+++ b/packages/blitz/src/bin/cli.ts
@@ -5,12 +5,16 @@ import chalk from "chalk"
 import {parseSemver} from "../utils/parse-semver"
 
 async function main() {
-  console.log(
-    chalk.yellow(
-      `You are using alpha software - if you have any problems, please open an issue here:
-    https://github.com/blitz-js/blitz/issues/new/choose\n`,
-    ),
-  )
+  const options = require("minimist")(process.argv.slice(2))
+
+  if (options._[0] !== "autocomplete:script" || Object.keys(options).length > 1) {
+    console.log(
+      chalk.yellow(
+        `You are using alpha software - if you have any problems, please open an issue here:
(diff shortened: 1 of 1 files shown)


packages/blitz/src/bin/cli.ts
```diff
@@ -5,12 +5,16 @@ import chalk from "chalk"
...
 async function main() {
-  console.log(
-    chalk.yellow(
-      `You are using alpha software - if you have any problems, please open an issue here:
-    https://github.com/blitz-js/blitz/issues/new/choose\n`,
-    ),
-  )
+  const options = require("minimist")(process.argv.slice(2))
+
+  if (options._[0] !== "autocomplete:script" || Object.keys(options).length > 1) {
```
flybayer, line 10:
```suggestion
  if (options._[0] !== "autocomplete:script") {
```

Unless I'm totally missing something, this second condition is not needed.

enricoschaaf: I wanted to only disable the alpha software message on the command itself, the second argument ensures that when you type `blitz --help autocomplete` or `blitz autocomplete --help` or any other autocomplete command with a flag the warning is still printed.

pbteja1998:
@flybayer Why can't we add `#` before the warning? That way, even when it is added to `.zshrc`, it becomes a comment. Will there be any problems if we do this?

flybayer:
@pbteja1998 let's just go with the fix in this PR because it's not a big deal, this message will be removed relatively soon anyways :)

PR #1127, opened 2020-09-20
🎉 Add Forgot Password code to new app template

flybayer:

This adds all the code for the Forgot Password flow to the new app template.

I've tried to comment it very well so it's easy to understand what the generated code is doing and why it is doing it.

This should meet all the [OWASP forgot password guidelines](https://cheatsheetseries.owasp.org/cheatsheets/Forgot_Password_Cheat_Sheet.html)

Todo before merge:

- [x] Add jest tests for forgotPassword and resetPassword mutations (tests will be included in new app template)

**Please review and suggest any improvements!**


wKovacs64:
Haven't had a chance to try it, but gave it a read-through and it looks sweet!

dillonraphael:
Looks good! But what I'm really excited for is seeing the start of mailers/integrations.

packages/generator/templates/app/app/auth/mutations/forgotPassword.ts
```diff
@@ -0,0 +1,51 @@
...
+  _ctx: { session?: SessionContext } = {}
+) {
+  const { email } = ForgotPasswordInput.parse(input)
+
+  // 1. Get the user
+  const user = await db.user.findOne({ where: { email: email.toLowerCase() } })
+
+  // 2. Generate the token and expiration date.
+  // We use encodeURIComponent(token) since it will be used in the URL
+  const token = generateToken()
```
clgeoio, line 22: suggestion: Consider putting all this computation in the `if(user)` block, no real need to calculate hashes etc if they will not be used.

packages/generator/templates/app/app/auth/mutations/forgotPassword.ts
```diff
@@ -0,0 +1,51 @@
...
+        expiresAt,
+        hashedToken,
+        sentTo: user.email,
+      },
+    })
+    // 6. Send the email
+    await forgotPasswordMailer({ to: user.email, token: urlSafeToken }).send()
+  } else {
+    // 7. If no user found wait the same time so attackers can't tell the difference
+    await new Promise((resolve) => setTimeout(resolve, 750))
```
clgeoio, line 46: suggestion: Perhaps a bit of random offset here would be useful. An attacker could determine if the user fetch failed if the request returned after 750ms. 
Spreading the range of timeout values would make it much harder to determine a success/failure case

packages/generator/templates/app/app/auth/pages/forgot-password.tsx
```diff
@@ -0,0 +1,118 @@
...
+        </div>
+      ) : (
+        <Form<ResetPasswordInputType>
+          submitText="Reset Password"
+          schema={ResetPasswordInput}
+          initialValues={{ password: "", passwordConfirmation: "", token: query.token as string }}
+          onSubmit={async (values) => {
+            try {
+              await resetPassword(values)
+              setSuccess(true)
```
johncantrell97, line 84: suggestion: why not just create a new session for the user at the end of resetPassword mutation and then redirect  to whatever the default logged in page is.  Saves the user from having to click login link and then login manually.  I know a lot of apps use the flow you have here but I can't really think of a benefit of making the user go through these extra steps.

packages/generator/templates/app/app/auth/pages/forgot-password.tsx
```diff
@@ -0,0 +1,118 @@
...
+          <h2>Password Reset Successfully</h2>
+          <p>
+            You may now <Link href="/login">log in with your new password</Link>
+          </p>
+        </div>
+      ) : (
+        <Form<ResetPasswordInputType>
+          submitText="Reset Password"
+          schema={ResetPasswordInput}
+          initialValues={{ password: "", passwordConfirmation: "", token: query.token as string }}
```
johncantrell97, line 80: suggestion: maybe just use useParam("token","string") instead of useRouterQuery() to save yourself this cast?

flybayer: Because the query parameters aren't in the params object.

packages/generator/templates/app/app/auth/pages/forgot-password.tsx
flybayer: Because of this [from here](https://cheatsheetseries.owasp.org/cheatsheets/Forgot_Password_Cheat_Sheet.html).
> Once they have set their new password, the user should then login through the usual mechanism. Don't automatically log the user in, as this introduces additional complexity to the authentication and session handling code, and increases the likelihood of introducing vulnerabilities.

But now that I'm thinking about it more, I think we can do it automatically since our code for creating sessions is so simple (1 line)

packages/generator/templates/app/app/auth/pages/forgot-password.tsx
johncantrell97: whoops, too tired to be reviewing code :|

packages/generator/templates/app/app/auth/pages/forgot-password.tsx
johncantrell97: Ah ok hadn't read that.  It does seem fairly low risk and offers imo a better end-user experience 🤷‍♂️

packages/generator/templates/app/app/auth/mutations/resetPassword.ts
```diff
@@ -0,0 +1,51 @@
...
+  _ctx: { session?: SessionContext } = {}
+) {
+  const { password, token } = ResetPasswordInput.parse(input)
+
+  // 1. Try to find this token in the database
+  // Have to first decodeURIComponent(token) because encodeURIComponent
+  // was used in forgotPassword.ts when sending to the browser
+  const hashedToken = hashToken(decodeURIComponent(token))
+  const possibleToken = await db.token.findOne({
+    where: { hashedToken, type: "RESET_PASSSWORD" },
```
aaronfulkerson, line 22: This line should be changed to account for the fact that there's a unique index on hashedToken and type:
`    where: { hashedToken_type: { hashedToken, type: 'RESET_PASSSWORD' } },`

aaronfulkerson, line 22: This line should be changed to account for the fact that there's a unique index on hashedToken and type:
`    where: { hashedToken_type: { hashedToken, type: 'RESET_PASSSWORD' } },`

aaronfulkerson (changes requested):

packages/generator/templates/app/app/auth/mutations/resetPassword.ts
flybayer: @aaronfulkerson I'm not understanding that change? The code in the PR is working without this change.

aaronfulkerson: Screen Shot 2020-10-10 at 2 24 16 PM

That's strange. According to the typings and the way Prisma works, you can only pass unique fields to findOne.

flybayer: Oh, you're right! My production code is different than this. I think the real fix is to change `findOne` to `findFirst` and then the code in the PR should work.

aem (dismissed):
Code looks great here. Haven't had a chance to pull it down locally but architecture looks solid

packages/generator/templates/app/app/auth/components/LoginForm.tsx
```diff
@@ -36,6 +36,9 @@ export const LoginForm = (props: LoginFormProps) => {
       >
         <LabeledTextField name="email" label="Email" placeholder="Email" />
         <LabeledTextField name="password" label="Password" placeholder="Password" type="password" />
+        <div>
+          <Link href="/forgot-password">Forgot your password?</Link>
```
rockymeza, line 40: Should this be?

```suggestion
          <Link href="/forgot-password"><a>Forgot your password?</a></Link>
```

flybayer: You only need the a tag if you want to add classes to it. But probably makes sense to add it.

packages/generator/templates/app/app/auth/pages/forgot-password.tsx
```diff
@@ -0,0 +1,118 @@
...
+          }}
+        >
+          <LabeledTextField name="email" label="Email" placeholder="Email" />
+        </Form>
+      )}
+    </div>
+  )
+}
+
+const ResetPasswordForm = () => {
```
thomasbroderick, line 62: I think the SSR might be getting confused by this page.  When I actually click the link from the reset email, I get this console error on the Set a New Password page: 

react-dom.development.js?2961:67 Warning: Text content did not match. Server: "Forgot your password?" Client: "Set a New Password"
    at h1
    at div
    at ResetPasswordForm (webpack-internal:///./pages/forgot-password.tsx:151:75)
    at ForgotPasswordPage (webpack-internal:///./pages/forgot-password.tsx:288:75)

packages/generator/templates/app/app/auth/auth-utils.ts
```diff
@@ -1,9 +1,13 @@
 import { AuthenticationError } from "blitz"
 import SecurePassword from "secure-password"
+import hasha, { HashaInput } from "hasha"
+import { nanoid } from "nanoid"
 import db from "db"
 
 const SP = new SecurePassword()
 
+export const RESET_PASSSWORD_TOKEN_EXPIRATION_IN_HOURS = 4
```
MrLeebo, line 9: 🐍 Too many S's 🐍

packages/generator/templates/app/app/auth/pages/forgot-password.tsx
```diff
@@ -0,0 +1,118 @@
...
+} from "app/auth/validations"
+import forgotPassword from "app/auth/mutations/forgotPassword"
+import resetPassword from "app/auth/mutations/resetPassword"
+
+/*
+ * By default this page renders the ForgotPasswordForm form which gets the user
+ * email and sends password reset instructions.
+ *
+ * When they click the link in that email, they will be sent back to this same
+ * page but with the token set as a query paramater like `?token=XXXXX`.
```
MrLeebo, line 20:
```suggestion
 * page but with the token set as a query parameter like `?token=XXXXX`.
```

And in other places

packages/generator/templates/app/app/auth/mutations/resetPassword.ts
```diff
@@ -0,0 +1,51 @@
...
+export default async function resetPassword(
+  input: ResetPasswordInputType,
+  _ctx: { session?: SessionContext } = {}
+) {
+  const { password, token } = ResetPasswordInput.parse(input)
+
+  // 1. Try to find this token in the database
+  // Have to first decodeURIComponent(token) because encodeURIComponent
+  // was used in forgotPassword.ts when sending to the browser
+  const hashedToken = hashToken(decodeURIComponent(token))
```
MrLeebo, line 20: Shouldn't `useRouterQuery()` have decoded this?

MrLeebo (changes requested):

packages/generator/templates/app/app/auth/mutations/forgotPassword.ts
flybayer: Yeah that would be slightly better, but for most folks this is fine

packages/generator/templates/app/app/auth/pages/forgot-password.tsx
flybayer: I'm changing it to automatically log you in

packages/generator/templates/app/app/auth/pages/forgot-password.tsx
flybayer: Fixing this by separating this page into two pages

Issue #1125, opened 2020-09-20
Add lint rule to prevent Object.fromEntries() in client code.


flybayer:
### What do you want and why?

https://github.com/blitz-js/blitz/pull/1124 adds some lint rules to keep us from breaking ES5 compatibility in the `core` package.

However, there's still one big missing lint rule which is to prevent `Object.fromEntries`.

### Possible implementation(s)

[This lint rule](https://eslint-plugin-es.mysticatea.dev/rules/no-object-fromentries.html) should do exactly this, but I was unable to get it to work. Hopefully you can!

The place to add the rule is `packages/core/.eslintrc.js`

hardfire:
got this to work! apparently the default npm package doesnt have this. So had to use from github (maybe we should pin it to a specific commit). Sending a Pull Request. 
![image](https://user-images.githubusercontent.com/513457/93705080-004de000-fb3a-11ea-91e8-27e43142f025.png)

PR #1129, opened 2020-09-20
add lint rule to prevent Object.fromEntries in core package

hardfire:
Closes: ??

Fixes https://github.com/blitz-js/blitz/issues/1125


flybayer:
Sweet, thanks!

PR #1134, opened 2020-09-21
[RFC] Add useAuthorize() hook

engelkes-finstreet:
The useEnsureHasRole hook prevents users without a given role to access a page.

Documentation PR: https://github.com/blitz-js/blitzjs.com/pull/169


flybayer:
I just had possibly an even better idea. Check this out:

```ts
import createItem from 'app/mutations/createItem'

const NewPage = () => {
  return (
    <Authorize resolver={createItem} loading={<LoadingSpinner />}>
      <NewItemForm />
    </Authorize>
  )
}
```

This is great because now you only specify the required role in the resolver instead of both resolver and component.

Implementation wise, <Authorize> would make an empty network request to the `createItem` mutation and see if it throws `AuthenticationError` or `AuthorizationError`. If it doesn't, then it will render it's children.

(I definitely should have had you post this first as an RFC, sorry about that!)

flybayer:
And here's yet another option. We can use `useAuthorize` by making it use suspense. This will force users to handle the loading state and eliminate my concern about user's not doing that.

```ts
import createItem from 'app/mutations/createItem'

const NewItemForm = () => {
  useAuthorize(createItem)

  return /* form stuff */
}

const NewPage = () => {
  return (
    <Suspense fallback={<LoadingSpinner />}>
      <NewItemForm />
    </Suspense>
  )
}
```

Zeko369:
@flybayer I think that we should have both `useAuthorize` and `<Authorize>` because it would make it easier for people to authorize only certain parts of the page.

```tsx
return (
  <div>
	<Public/>
	<Authorize {...authorizeProps}>
		<AdminHeader/>
		<Admin/>
	</Authorize/>
  </div>
);
```

engelkes-finstreet:
@flybayer I think your last suggestions with using suspense is the best one! Will refactor the code during the day!

engelkes-finstreet:
It did not feel right to use this for admin stuff in my app. Maybe if you have some things that only USER A can see and an additional thing for USER B this might be a good solution but not for hiding all ADMIN related tasks since you have this `Authorize` component all throughout your app. For the admin problem I proposed the following RFC https://github.com/blitz-js/blitz/discussions/1166

flybayer:
@engelkes-finstreet ok. I think it's still very useful for role-based authorization throughout the app, right?

engelkes-finstreet:
Yes, thats right! I did not have the time to look into react-query and their suspense method. Will do it during the week and add both the `Authorize` component and `useAuthorize()` hook

engelkes-finstreet:
@flybayer I implemented both the component and the hook inside the core package and tried to use them inside one of the examples but got the following error:

`Module parse failed: Unexpected token (4:22)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
| import {useAuthorize} from "./use-authorize";
| 
> export const Authorize: React.FC<{resolver: any}> = ({resolver, children}) => {
|     useAuthorize(resolver);
|
`

Is the core package the correct place for the `Authorize` component? If so, what would we have to change to make this work?

flybayer:
@engelkes-finstreet yeah in core. If you go ahead and open the PR as a draft, I can take a look and see if anything obvious is wrong.

flybayer:
Superseded by https://github.com/blitz-js/blitz/pull/1900/

PR #1139, opened 2020-09-22
Update tailwind recipe to install tailwindcss@latest instead of tailwindcss@1

enricoschaaf:
Closes: ??

Install the latest version of tailwindcss instead of version 1.
### Checklist

- [ ] Tests added for changes


flybayer:
Hey @enricoschaaf was you having a problem with this set to `1`? The reason we set it to `1` and not `latest` is that version 2 could have breaking changes that require updating the recipe before it will work with version 2.

enricoschaaf:
Sry, somehow I thought, that "1" would install 1.0, but that's not the case. 🙈

PR #1144, opened 2020-09-23
(newapp) Improve Form typing: remove need to pass schema type

flybayer:

Improve Form typing for new apps: remove need to pass schema type

### DIFF
--- a/examples/auth/app/components/Form.tsx
+++ b/examples/auth/app/components/Form.tsx
@@ -32,13 +32,7 @@ export function Form<S extends z.ZodType<any, any>>({
           return error.formErrors.fieldErrors
         }
       }}
-      onSubmit={async (values, form) => {
-        const result = await onSubmit(values, form)
-        if (result === undefined) {
-          form.reset()
-        }
-        return result
-      }}
+      onSubmit={onSubmit}
       render={({handleSubmit, submitting, submitError}) => (
         <form onSubmit={handleSubmit} className="form" {...props}>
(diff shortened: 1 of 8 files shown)


examples/auth/app/components/Form.tsx
```diff
@@ -3,26 +3,26 @@ import {Form as FinalForm, FormProps as FinalFormProps} from "react-final-form"
...
+  onSubmit: FinalFormProps<z.infer<S>>["onSubmit"]
+  initialValues?: FinalFormProps<z.infer<S>>["initialValues"]
+  schema?: S
 } & Omit<PropsWithoutRef<JSX.IntrinsicElements["form"]>, "onSubmit">
 
-export function Form<FormValues extends Record<string, unknown>>({
+export function Form<S extends z.ZodType<any, any>>({
   children,
   submitText,
   schema,
```
nitaking, line 19: Which one is more type-checked?

`ZodType<any, any>` `Record<string, unknown>`


It seems that unknown is better than any

flybayer: Thanks for reviewing!

In this case `S` is **extending** `z.ZodType<any, any>` which means that it's up to the user to define what the actual type will be. So it doesn't matter what we use here.

nitaking: I see! I'll try it next time. Thanks for the comment!

nitaking:
LGTM!

Issue #940, opened 2020-08-24
Blitz doesn't ignore directories added to `.gitignore`


anteprimorac:
### What is the problem?
Blitz doesn't ignore directories added to `.gitignore`

### Steps to Reproduce

1. Add a directory to `.gitignore`, for example `/dir/`
2. Add some files to that directory
3. Run `blitz start`

### Workaround

Currently, you can work around this issue by adding `/dir/**/*` to `.gitignore`.

### Versions

```
macOS Catalina | darwin-x64 | Node: v12.16.2

blitz: 0.18.0 (global)
blitz: 0.18.0 (local)

  Package manager: npm
  System:
    OS: macOS 10.15.6
    CPU: (12) x64 Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
    Memory: 1.04 GB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 12.16.2 - ~/.nvm/versions/node/v12.16.2/bin/node
    Yarn: 1.22.0 - ~/.yarn/bin/yarn
    npm: 6.14.5 - /usr/local/bin/npm
    Watchman: Not Found
  npmPackages:
    @prisma/cli: 2.5.1 => 2.5.1
    @prisma/client: 2.5.1 => 2.5.1
    blitz: 0.18.0 => 0.18.0
    react: 0.0.0-experimental-33c3af284 => 0.0.0-experimental-33c3af284
    react-dom: 0.0.0-experimental-33c3af284 => 0.0.0-experimental-33c3af284
    typescript: 3.9.7 => 3.9.7
```

### Other

Please include applicable logs and screenshots that show your problem.

flybayer:
Hey @anteprimorac! When you say ignored by Blitz, what do you mean by that?

anteprimorac:
@flybayer it’s same as here https://github.com/blitz-js/blitz/issues/360

flybayer:
Ah, I forgot about that! 👍

svobik7:
Is this still relevant? I am ready to take a look.

anteprimorac:
@svobik7 yes, it is still relevant.

@flybayer does it make sense to apply the same logic to the build hash generation? With it, we could solve heroku issue in the user-land.

flybayer:
@svobik7 all yours!

And yes @anteprimorac I had the exact same thought

svobik7:
[...]

PR #1150, opened 2020-09-24
Fix .gitignore files not being ignored by blitz

ntgussoni:
Closes: #940 

The current canary was indeed using the .gitignore to filter out unwanted folders and files, however it didn't take `.git/info/excludes` file into account.

This PR adds the necessary changes to parse all exclude files in the project, including any git submodule in the tree.

### Checklist

- [x] Tests added for changes


flybayer:
You haven't added support for global gitignore to this, yet, right?

ntgussoni:
> You haven't added support for global gitignore to this, yet, right?

I have it on a branch, im adding some tests but i havent had the time this couple of days. I'll Try to push my changes soon

flybayer:
Cool, no problem! Mark it back as ready for review when finished.

ntgussoni:
@flybayer This should be enough for now. 

One thing to keep in mind is that you could have something like `history` in your .gitignore and git would ignore 
`history/foo/bar` as well.

Since we are piping these strings directly to a glob search, the glob would only match anything called `history` but not `history/**`.

For most cases we should be fine with what we have right now. But we could explore using `git check-ignore` instead of manually reading the files.

flybayer (approved):

flybayer:
@ntgussoni awesome, thanks!! Will let @ryardley do final review.

@all-contributors add @ntgussoni for code

flybayer:
@ryardley can you review this?

ryardley (approved):
Looks good to me

flybayer:
@all-contributors add @ntgussoni for code

flybayer:
@all-contributors add @ntgussoni for code

Issue #1155, opened 2020-09-24
useParams and useParam is changing on every render - needs to use useMemo


flybayer:
### What is the problem?

`const params = useParams()` returns a new object on every single render, causing bugs in apps that expect it to only change if the contents have changed.

Also have the same problem for `useParam()`

### How to fix

Here's the code for [useParams](https://github.com/blitz-js/blitz/blob/canary/packages/core/src/use-params.ts#L47) and [useParam](https://github.com/blitz-js/blitz/blob/canary/packages/core/src/use-params.ts#L90) that needs optimized to use `useMemo`.

bhatvikrant:
Hey @flybayer is this still available to work on?

flybayer:
@bhatvikrant no, PR is in review (you can tell because it's no longer labeled `status/ready-to-work-on`

bhatvikrant:
Oh okay, @flybayer thankyou

Issue #1154, opened 2020-09-24
Query object is changing on every render - needs to use useMemo


flybayer:
### What is the problem?

`const query = useRouterQuery()` returns a new object on every single render, causing bugs in apps that expect it to only change if the contents of query changed.

### How to fix

[Here's the line of code that needs changed to use `useMemo`](https://github.com/blitz-js/blitz/blob/canary/packages/core/src/use-router-query.ts#L6).

TatisLois:
Hey! I have some time and would like to take this up!

flybayer:
@TatisLois already closed in #1157! (notice that label was `status/in-review`. Issues that are open to work on are labeled `states/ready-to-work-on` :)

PR #1157, opened 2020-09-24
Reduce re-renders by memoizing useParams and useRouterQuery

sandulat:
Closes: https://github.com/blitz-js/blitz/issues/1155 & https://github.com/blitz-js/blitz/issues/1154

I've added memoization and deep comparison to the `useParams` hook. Had to use state instead of refs because refs can't be passed to the `useMemo` hook as a dependency.

No changes were done to the `useParam` hook since it relies on the `useParams` hook and is automatically memoized.

Added deep comparison only to the `useRouterQuery` since if we will add an `useMemo` to watch `router.asPath`, the query will trigger updates on URL change, even if the effective query from URL didn't change.

### Checklist

- [ ] Tests added for changes

### DIFF
--- a/packages/core/src/use-params.ts
+++ b/packages/core/src/use-params.ts
@@ -1,6 +1,7 @@
+import {useMemo, useState} from "react"
+import {fromPairs, isEqual} from "lodash"
 import {useRouter} from "next/router"
 import {useRouterQuery} from "./use-router-query"
-import {fromPairs} from "lodash"
 
 type ParsedUrlQueryValue = string | string[] | undefined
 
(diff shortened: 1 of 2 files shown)


JoseRFelix:
Thanks for the PR! I have a question. You say:

> router.asPath, the query will trigger updates on URL change, even if the effective query from URL didn't change.

Isn't `asPath` a string? Wouldn't React just check the content instead of reference?

sandulat:
@JoseRFelix You're right! But if you're for example on url `/products/page/1?sortBy=id` and navigate to `products/page/2?sortBy=id`, the `asPath` will trigger `useMemo`, however the query `{ sortBy: 'id' }` is still the same.

nitaking:
Nice work!

sandulat:
@nitaking Thank you very much!

packages/core/src/use-params.ts
```diff
@@ -47,40 +48,49 @@ export function useParams(returnType: "array"): Record<string, Array<string | un
 export function useParams(returnType?: "string" | "number" | "array") {
   const router = useRouter()
   const query = useRouterQuery()
+  const [lastRouterQuery, setLastRouterQuery] = useState(router.query)
 
-  const rawParams = extractRouterParams(router.query, query)
+  if (!isEqual(router.query, lastRouterQuery)) {
+    setLastRouterQuery(router.query)
+  }
```
flybayer, line 55: 1. We can't setState in the render method because render methods must be pure (especially for concurrent mode)
2. Can we just use `router.query` directly instead of `lastRouterQuery`?

packages/core/src/use-router-query.ts
```diff
@@ -1,8 +1,18 @@
...
   const {query} = parse(router.asPath, true)
-  return query
+
+  const lastQueryRef = useRef(query)
+
+  if (!isEqual(query, lastQueryRef.current)) {
+    lastQueryRef.current = query
+  }
+
+  return lastQueryRef.current
```
flybayer, line 17: I'm not confident that this is concurrent mode safe. I've asked for confirmation [on twitter](https://twitter.com/flybayer/status/1310599536964968454).

flybayer (commented):
Thank you @sandulat!!

Couple comments.

packages/core/src/use-router-query.ts
flybayer: No this isn't CM safe. https://twitter.com/adamrackis/status/1309665030967889920?s=21

Can only write to the ref inside useEffect

sandulat:
My bad @flybayer! I've exaggerated with the implementation a little bit. I forgot that people usually watch a specific param and not the entire params object. 

I've pushed a new commit, which includes just simple useMemo.

flybayer:
@sandulat no problem :) Thanks!

PR #1160, opened 2020-09-25
🔥 Add `useMutation()` and strong typing for `ctx` and `ctx.session`

flybayer:

This is the biggest change we've had to Blitz since the initial alpha release! This greatly cleans up your resolver code, and gives you nice features like access to `isLoading` for mutations to show busy indicators.

This adds a `useMutation` hook as decided in [this RFC](https://github.com/blitz-js/blitz/discussions/1052#discussioncomment-78091). Blitz `useMutation` is a very thin wrapper around react-query's `useMutation`.

It also adds ability to strongly type the `ctx` parameter to all queries/mutations.

### Checklist

- [x] Tests added for changes
- [x] PR submitted to [blitzjs.com](https://github.com/blitz-js/blitzjs.com) for any user facing changes: https://github.com/blitz-js/blitzjs.com/pull/177


Skn0tt:
In https://github.com/blitz-js/blitz/discussions/1052#discussioncomment-78114, you mentioned an API like the following:

```ts
const result = await execute(getUser, {where: /*...*/})
```

Is this planned to be part of this PR? I find that to be rather important ^^

flybayer:
@Skn0tt yes as a separate PR

PR #1168, opened 2020-09-25
[RFC] Implemented Mailer recipe

sandulat:

This is a recipe for a multi-driver mailing system.

### Drivers
For now, it can send emails via 3 drivers:
- SendGrid
- SMTP
- Preview (in the browser)

### Configuration
All the configuration is done through environment variables:
```
MAIL_FROM_EMAIL=
MAIL_FROM_NAME=

# Mail drivers: SENDGRID, SMTP, PREVIEW
MAIL_DRIVER=

# These variables are required only when MAIL_DRIVER is SMTP
MAIL_HOST=
MAIL_PORT=
MAIL_USER=
MAIL_PASS=

# This variable is required only when MAIL_DRIVER is SENDGRID
MAIL_SENDGRID_KEY=

```

### Templating
The system includes a templating engine powered by Handlebars. All templates are wrapped in a global layout. Variables can be passed to templates.

Template example:
```html
Foo variable equals: {{foo}}
```
Will be compiled into:
```
Foo variable equals: bar
```

### Extending drivers
Drivers are simple functions that must implement a specific interface. New drivers can be easily added to the drivers object:
```js
import {smtpDriver} from "./drivers/smtp"
import {sendgridDriver} from "./drivers/sendgrid"
import {previewDriver} from "./drivers/preview"

const mailDrivers = {
  SMTP: smtpDriver,
  SENDGRID: sendgridDriver,
  PREVIEW: previewDriver,
}
```

The keys of this object are used as values of the `MAIL_DRIVER` environment variable.

A new mail driver can be implemented as easy as:
```js
import {MailDriver} from "../mail"

export const customDriver: MailDriver = () => {
  // init configurations, singletons, etc.

  return {
    send: ({from: {email, name}, to, subject, html}) => {
      // my custom driver logic
    },
  }
}
```

### Usage example
```js
import {mail} from "mail/mail"

mail.send({
  subject: "Something cool",
  to: "email@example.org",
  view: "mail-example/template-example", // basically means app/mail-example/template-example.html
  variables: {foo: "bar"}, // variables that will be passed to the view
})
```

### Recipe steps:

1. Adds the following dependencies:
 - handlebars 4.7.6
 - @sendgrid/mail 7.2.6
 - nodemailer 6.4.11
 - preview-email 2.0.2
 - @types/handlebars 4.1.0 (dev dependecy)
 - @types/nodemailer 6.4.0 (dev dependecy)
 - @types/preview-email 2.0.0 (dev dependecy)

2. Copies the mailing system into the root folder `./mail`.

3. Adds a usage example to `./app/mail-example`.

4. Adds a temporary `.env-mail.example` file to the root folder to showcase the necessary environment variables.

### Checklist

- [ ] Tests added for changes

### DIFF
--- a/recipes/mailer/index.ts
+++ b/recipes/mailer/index.ts
@@ -1,0 +1,47 @@
+import {RecipeBuilder} from "@blitzjs/installer"
+import {join} from "path"
+
+export default RecipeBuilder()
+  .setName("Mailer")
+  .setDescription("This recipe will add a multi-driver mailing system.")
+  .setOwner("alex@sandulat.com")
+  .setRepoLink("https://github.com/blitz-js/blitz")
+  // .addAddDependenciesStep({
+  //   stepId: "addDeps",
+  //   stepName: "Add npm dependencies",
+  //   explanation: `This mailing system relies on a couple of dependencies. Handlebars will be installed as the templating engine. SendGrid, Nodemailer and Preview Email will be installed as mail drivers.`,
+  //   packages: [
+  //     {name: "handlebars", version: "4.7.6"},
+  //     {name: "@sendgrid/mail", version: "7.2.6"},
+  //     {name: "nodemailer", version: "6.4.11"},
+  //     {name: "preview-email", version: "2.0.2"},
+  //     {name: "@types/handlebars", version: "4.1.0", isDevDep: true},
+  //     {name: "@types/nodemailer", version: "6.4.0", isDevDep: true},
+  //     {name: "@types/preview-email", version: "2.0.0", isDevDep: true},
+  //   ],
+  // })
+  .addNewFilesStep({
+    stepId: "addMailer",
+    stepName: "Add the mailing system and all its drivers",
+    explanation: `We are simply copying the mailing system into your app. Feel free to extend it with your own drivers.`,
+    targetDirectory: "./mail",
+    templatePath: join(__dirname, "templates/mail"),
+    templateValues: {},
+  })
+  .addNewFilesStep({
+    stepId: "addUsageExample",
+    stepName: "Add an example of mailing system usage",
+    explanation: `We will add a mailing system usage example.`,
+    targetDirectory: "./app/mail-example",
+    templatePath: join(__dirname, "templates", "mail-example"),
+    templateValues: {},
+  })
+  .addNewFilesStep({
+    stepId: "addEnvExample",
+    stepName: "Add an example of environment variables",
+    explanation: `We will add an environment configuration example, which will contain all the necessary variables for the mailing system to work. Do not forget to move these variables into your own .env and delete this example file.`,
+    targetDirectory: ".",
+    templatePath: join(__dirname, "templates", "config"),
+    templateValues: {},
+  })
+  .build()
--- a/recipes/mailer/index.ts
+++ b/recipes/mailer/index.ts
@@ -6,20 +6,20 @@ export default RecipeBuilder()
   .setDescription("This recipe will add a multi-driver mailing system.")
   .setOwner("alex@sandulat.com")
   .setRepoLink("https://github.com/blitz-js/blitz")
-  // .addAddDependenciesStep({
-  //   stepId: "addDeps",
-  //   stepName: "Add npm dependencies",
-  //   explanation: `This mailing system relies on a couple of dependencies. Handlebars will be installed as the templating engine. SendGrid, Nodemailer and Preview Email will be installed as mail drivers.`,
-  //   packages: [
-  //     {name: "handlebars", version: "4.7.6"},
-  //     {name: "@sendgrid/mail", version: "7.2.6"},
-  //     {name: "nodemailer", version: "6.4.11"},
-  //     {name: "preview-email", version: "2.0.2"},
-  //     {name: "@types/handlebars", version: "4.1.0", isDevDep: true},
-  //     {name: "@types/nodemailer", version: "6.4.0", isDevDep: true},
-  //     {name: "@types/preview-email", version: "2.0.0", isDevDep: true},
-  //   ],
-  // })
+  .addAddDependenciesStep({
+    stepId: "addDeps",
+    stepName: "Add npm dependencies",
+    explanation: `This mailing system relies on a couple of dependencies. Handlebars will be installed as the templating engine. SendGrid, Nodemailer and Preview Email will be installed as mail drivers.`,
+    packages: [
+      {name: "handlebars", version: "4.7.6"},
+      {name: "@sendgrid/mail", version: "7.2.6"},
+      {name: "nodemailer", version: "6.4.11"},
+      {name: "preview-email", version: "2.0.2"},
+      {name: "@types/handlebars", version: "4.1.0", isDevDep: true},
+      {name: "@types/nodemailer", version: "6.4.0", isDevDep: true},
+      {name: "@types/preview-email", version: "2.0.0", isDevDep: true},
+    ],
+  })
   .addNewFilesStep({
     stepId: "addMailer",
     stepName: "Add the mailing system and all its drivers",
--- a/recipes/mailer/package.json
+++ b/recipes/mailer/package.json
@@ -1,0 +1,27 @@
+{
+  "name": "@blitzjs/mailer",
+  "private": true,
+  "version": "0.23.1-canary.0",
+  "description": "The Blitz Recipe for adding a multi-driver mailing system",
+  "main": "index.ts",
+  "scripts": {
+    "test": "echo \"No tests yet\""
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/blitz-js/blitz.git"
+  },
+  "keywords": [
+    "blitz",
+    "blitzjs"
+  ],
+  "author": "Alexandru Stratulat <alex@sandulat.com>",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/blitz-js/blitz/issues"
+  },
+  "homepage": "https://github.com/blitz-js/blitz#readme",
+  "dependencies": {
+    "@blitzjs/installer": "0.23.1-canary.0"
+  }
+}
--- a/recipes/mailer/templates/config/.env-mail.example
+++ b/recipes/mailer/templates/config/.env-mail.example
@@ -1,0 +1,14 @@
+MAIL_FROM_EMAIL=
+MAIL_FROM_NAME=
+
+# Mail drivers: SENDGRID, SMTP, PREVIEW
+MAIL_DRIVER=
+
+# These variables are required only when MAIL_DRIVER is SMTP
+MAIL_HOST=
+MAIL_PORT=
+MAIL_USER=
+MAIL_PASS=
+
+# This variable is required only when MAIL_DRIVER is SENDGRID
+MAIL_SENDGRID_KEY=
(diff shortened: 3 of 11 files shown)


nahue:
What if we let the user choose its own driver? We could add Sendgrid as an example in the docs.
Lets say i am a user of Mailgun, i wont appreciate a framework pushing a dependency that i wont be using.

sandulat:
@nahue That's a perfectly valid point! As far as I know, unfortunately, Blitz's recipe system doesn't yet provide a way to ask users to choose from a list of items. Correct me if I'm wrong.

IMHO, until then, we can stick to 1 third party driver (in our case it's SendGrid) to showcase the developer how third-party drivers can be easily added. The developer can delete the SendGrid dependency and driver in a matter of seconds, and switch to Mailgun's official SDK.

nahue:
Oh i dont think we will need a list of items.. just a doc that describes how the dev needs to write a driver that complies with the MailDriver type (or maybe it could be a interface). Lets wait for opinions on integrating Sendgrid as part of the recipe

nahue:
I propose we consider this as a starting point for a mailing system in blitz core

Zeko369:
I agree with @nahue, we should try adding this to blitz core (we'll need this for password-reset-flow)

MrLeebo:
It's a good idea to have these settings configurable by environment variable, but I think that a recipe should also initialize a layer of JS so that your application has some control over whether/how the settings are used. I don't know if *blitz.config.js* is the best choice, but I think it would be better to have the settings there or in another config file e.g. 

```js
// blitz.config.js with basic mailer configuration
module.exports = {
  mailer: {
    driver: process.env.MAIL_DRIVER,
    host: process.env.MAIL_HOST,
    from: process.env.MAIL_FROM || 'noreply@blitz.com' // easy to add fallbacks
  }
}
```

And the recipe logic can access `config.mailer.driver` instead of reading the environment variable directly.

That way you have somewhere to put customized logic without contorting your environment variables or core mailer logic:

```js
// blitz.config.js with additional mailer logic
module.exports = phase => {
  // in development, replace the mailer with mailtrap or perhaps a local preview driver (email-preview)
  if (phase === DEVSERVER) {
    return {
      mailer: {
        driver: 'smtp',
        host: process.env.MAILTRAP_HOST
      }
    }
  }

  // use the real credentials in prod
  return { 
    mailer: {
      driver: "sendgrid",
      sendgrid_key: process.env.SENDGRID_KEY
    }
  }
}
```

flybayer:
Finally reviewing this! Thanks @sandulat!!

I'm good to add this to core.

- I think we should not use any env variables for config. Instead, use `blitz.config.js`. Users can still use an env variable for something if they want to.
- In terms of file structure, can we do this?
```
mailers/
  - forgotPasswordMailer/
    - index.ts
    - template.html.hbs
```
- I think by default the preview mailer should be used during development.

And how to package all this?
- One way is to only include the preview driver in core, and then have separate packages for other drivers like `@blitzjs/sendgrid-stmp`. These separate packages can bundle any specific dependencies needed like `@sendgrid/mail`.
- The other way is to include all drivers in core, but then require users to manually install any needed deps like `@sendgrid/mail`.

Another question I have is how to also seamlessly support using third-party templates. For example, here's how I'm currently using Postmark.

```ts
// mailers/forgotPasswordMailer.ts
import postmark from "integrations/postmark"
import previewEmail from "preview-email"
import config from "./config"
import { serverEnv } from "config/serverEnv"
import { logger } from "utils/logger"

type ResetPasswordMailer = {
  to: string
  token: string
}

export function forgotPasswordMailer({ to, token }: ResetPasswordMailer) {
  const log = logger.getChildLogger({ prefix: ["forgotPasswordMailer()"] })

  const resetUrl = `${serverEnv.APP_ORIGIN}/forgot-password?token=${token}`
  log.debug("Sending email...", { to, resetUrl })

  const msg = {
    From: config.from,
    To: to,
    TemplateId: 20265735,
    TemplateModel: {
      resetUrl,
    },
  }

  return {
    async send() {
      if (serverEnv.isProduction) {
        await postmark.sendEmailWithTemplate(msg)
      } else {
        // Preview email data in the browser
        await previewEmail({
          ...msg,
          html: `
            <h1>Development Only Preview</h1>
            <h2>Reset Your Password</h2>

            <a href="${msg.TemplateModel.resetUrl}">
              Click here to set a new password
            </a>
          `,
        })
      }
    },
  }
}
```

nahue:
- I vote for having the preview mailer as part of the core, and other mail backends/services as external addons with their respective dependencies.
- Maybe for templating we can go with something more email-oriented templating framework like https://mjml.io/ , what do you think about this one?

flybayer:
Hey @sandulat, any update on this?

BjoernRave:
I used this code as the starting point for my mailingsystem ,but adapted it to be able to use react for mailing. Not sure if this is a better approach, but I personally prefer it. 


I have two different layouts in `layouts/...`, which are just html files. Then for each email I create a react component, in which i use styled components. the styles get later extracted and inlined and rendered with `ReactDOM.renderToString()`



```
//driver.ts

import SES from 'aws-sdk/clients/ses'
import nodemailer from 'nodemailer'
import previewEmail from 'preview-email'
import { MailDriver } from '.'
import { REGION } from '../lib/fileUpload'

export const previewDriver: MailDriver = () => ({
  send: ({ from, ...options }) =>
    previewEmail({
      ...options,
      from: `"${from.name}" ${from.email}`,
    }),
})

export const sesDriver: MailDriver = () => {
  const transporter = nodemailer.createTransport({
    SES: new SES({
      apiVersion: '2010-12-01',
      region: REGION,
      credentials: {
        accessKeyId: process.env.AWS_SERVER_ACCESS_KEY_ID,
        secretAccessKey: process.env.AWS_SERVER_SECRET_ACCESS_KEY,
      },
    }),
  })

  return {
    send: ({ from, ...options }) =>
      transporter.sendMail({
        ...options,
        from: `"${from.name}" ${from.email}`,
      }),
  }
}
```

```
//index.ts

import Mail from 'nodemailer/lib/mailer'
import { previewDriver, sesDriver } from './driver'
import { compileView } from './view'

export * from './views/emails'

export const layouts = {
  default: 'prisma/mail/layouts/defaultLayout.html',
  blank: 'prisma/mail/layouts/blankLayout.html',
}

const senderEmail = 'info@inventhora.com'

const mailDrivers = {
  SES: sesDriver,
  PREVIEW: previewDriver,
}

const mailDriver =
  process.env.NODE_ENV === 'development'
    ? mailDrivers.PREVIEW()
    : mailDrivers.SES()

export type MailDriver = () => {
  send: (options: {
    from: {
      email: string
      name: string
    }
    to: string
    subject: string
    html: string
    attachments?: Mail.Attachment[]
  }) => Promise<any>
}

export const mail = {
  send: async ({
    subject,
    to,
    attachments,
    content,
    layout,
  }: {
    subject: string
    to: string
    attachments?: Mail.Attachment[]
    content: Promise<string>
    layout: keyof typeof layouts
  }) => {
    try {
      return await mailDriver.send({
        from: {
          email: senderEmail,
          name: process.env.MAIL_FROM_NAME!,
        },
        to,
        subject,
        attachments,
        html: compileView({
          subject,
          content: await content,
          layout: layouts[layout],
        }),
      })
    } catch (e) {
      console.error(e?.response?.body || e)
    }
  },
}
```

```
//views.ts

import fs from 'fs'
import Handlebars from 'handlebars'
import path from 'path'

const resolveView = (viewPath: string) => {
  const viewResolvedPath = path.resolve(viewPath)

  if (!fs.existsSync(viewResolvedPath)) {
    throw new Error(`Template "${viewResolvedPath}" doesn't exist.`)
  }

  return fs.readFileSync(viewResolvedPath, 'utf8')
}

export const compileView = ({
  subject,
  content,
  layout,
}: {
  subject: string
  content: string
  layout: string
}) => {
  const resolvedLayout = resolveView(layout)

  return Handlebars.compile(resolvedLayout)({
    subject,
    content,
  })
}
```

```
views/index.ts

import inlineCSS from 'inline-css'
import React from 'react'
import ReactDOMServer from 'react-dom/server'
import { Email } from 'react-html-email'
import { ServerStyleSheet, StyleSheetManager } from 'styled-components'

export const renderEmail = async (emailComponent) => {
  const doctype =
    '<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">'

  const sheet = new ServerStyleSheet()

  const html = ReactDOMServer.renderToStaticMarkup(
    <StyleSheetManager sheet={sheet.instance}>
      {emailComponent}
    </StyleSheetManager>
  )

  const styleTag = sheet.getStyleTags()

  const cleanedStyles =
    styleTag ===
    '<style data-styled="true" data-styled-version="5.2.1"></style>'
      ? ''
      : styleTag

  const inlinedHtml = await inlineCSS(doctype + cleanedStyles + html, {
    url: 'noUrl',
  })

  return inlinedHtml
}

export const EmailWrapper = ({ title, children }) => (
  <Email title={title}>
    <h1>{title}</h1>
    {children}
  </Email>
)

export const Space = ({ pixel }: { pixel: string }) => (
  <tr>
    <td colSpan={3} height={pixel}></td>
  </tr>
)
```
//views/emails.ts

import React from 'react'
import { Box, Item } from 'react-html-email'
import styled from 'styled-components'
import { EmailWrapper, renderEmail, Space } from '.'

const TableName = styled.td`
  font-weight: bold;
`

export const renderSupportEmail = ({
  message,
  userName,
  subDomain,
  email,
  t,
}) =>
  renderEmail(
    <EmailWrapper title='A user requested support'>
      <Item>{t('message')}</Item>
      <Space pixel='20' />
      <Item>{message}: </Item>
      <Space pixel='50' />
      <Item />
      <Box>
        <tr>
          <TableName>{t('name')}: </TableName>
          <td>{userName}</td>
        </tr>
        <tr>
          <td>{t('subDomain')}: </td>
          <td>{subDomain}</td>
        </tr>
        <tr>
          <td>{t('email')}: </td>
          <td>{email}</td>
        </tr>
      </Box>
    </EmailWrapper>
  )

...
```

**Comment by MrLeebo** [comment]
I agree, I think JSX + ReactDOMServer.renderToStaticMarkup() is more than an adequate default for templates. Way more expressive and flexible than Handlebars and doesn't add to node_modules.

**Review by MrLeebo — CHANGES_REQUESTED** [review]
1. It sounds like there's consensus around having a simple mailer system in core, though perhaps specific integrations such as sendgrid can be recipes. I think that's a good pattern because there will probably be a lot of recipes that would like to interact with mailers.
2. I recommend ditching the handlebars dependency and assume JSX + renderToStaticMarkup() as the default template format.

**Comment by MrLeebo** [comment]
I think this PR has gone stale. If you're interested in running this recipe in your app, you can try to install it with the command below, but the branch is out-of-date with the latest versions of blitz.

```bash
blitz install https://github.com/sandulat/blitz/tree/mailer-recipe/recipes/mailer
```

Jarrodsz:
Any updates on this?
Exactly what I need + MJML support

```
import {mail} from "mail/mail"

mail.send({
  subject: "Something cool",
  to: "email@example.org",
  view: "mail-example/template-example.MJML", // basically means app/mail-example/template-example.html
  variables: {foo: "bar"}, // variables that will be passed to the view
})
```

PR #1169, opened 2020-09-26
Implemented Seeder recipe

sandulat:

This is a recipe for a database seeder system. Might be very useful to seed data in a test environment or in tests.

### Usage example
This is a seeder for the Prisma entity "user":
```js
import faker from "faker"
import {seeder} from "./seeder"

export const userSeeder = seeder("user", () => {
  return {
    data: {
      email: faker.internet.exampleEmail(),
    },
  }
})
```

> Note that the first argument of the seeder() function is fully typed based on your Prisma schema.

The user seeder can be easily run like that:
```js
await userSeeder.seed()
```

The method `.seed()` can optionally receive one object argument with the following structure:
| Attribute | Type               | Optional | Description                                                                                                                                |
|-----------|--------------------|----------|--------------------------------------------------------------------------------------------------------------------------------------------|
| amount    | number             | true     | Amount of data to seed. Defaults to 1.                                                                                                     |
| options   | any                | true     | Custom options to be passed to the callback of the seeder. Typings are inherited from the interface defined in the callback of the seeder. |
| merge     | "Entity"CreateArgs | true     | Data to be overwritten when seeding. Might be very useful when you need specific values in some columns when seeding in some test.         |

Example:
```js
import faker from "faker"
import { seeder } from "./seeder"

export const userSeeder = seeder("user", (options: { activeSubscription: boolean }) => {
  return {
    data: {
      email: faker.internet.exampleEmail(),
      name: faker.name.firstName,
      subscribedAt: options.activeSubscription ? new Date() : null,
      subscriptionExpiresAt: options.activeSubscription ? new Date() : null,
      subscriptionActive: options.activeSubscription,
    },
  }
})

userSeeder.seed({
  amount: 500,
  options: { activeSubscription: true },
  merge: {
    data: {
      name: "Custom Name",
    }
  },
})
```

The example above will seed 500 users with a custom name and an active subscription.

### Running the seeder manually
A seeder can be run using `blitz console` like that:

```js
await require('./seeder/userSeeder').userSeeder.seed()
```

### Recipe steps:

1. Adds the following dependencies:
 - lodash.merge 4.6.2
 - faker 5.1.0 (dev dependecy)
 - @types/faker 5.1.2 (dev dependecy)

2. Copies the seeder system along with a simple user seeder example into the root folder `./seeder`.

### Checklist

- [ ] Tests added for changes

### DIFF
--- a/recipes/seeder/index.ts
+++ b/recipes/seeder/index.ts
@@ -1,0 +1,27 @@
+import {RecipeBuilder} from "@blitzjs/installer"
+import {join} from "path"
+
+export default RecipeBuilder()
+  .setName("Seeder")
+  .setDescription("This recipe will add a database seeder system.")
+  .setOwner("alex@sandulat.com")
+  .setRepoLink("https://github.com/blitz-js/blitz")
+  .addAddDependenciesStep({
+    stepId: "addDeps",
+    stepName: "Add npm dependencies",
+    explanation: `We will just install faker.js for generating fake data.`,
+    packages: [
+      {name: "lodash.merge", version: "4.6.2"},
+      {name: "faker", version: "5.1.0", isDevDep: true},
+      {name: "@types/faker", version: "5.1.2", isDevDep: true},
+    ],
+  })
+  .addNewFilesStep({
+    stepId: "addSeeder",
+    stepName: "Add the seeder system and an example.",
+    explanation: `We are simply copying the seeder system into your app with a simple user seeder example.`,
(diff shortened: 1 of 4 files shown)


flybayer:
what do you all think about adding this to core instead of as a recipe?

sandulat:
@flybayer please see https://github.com/blitz-js/blitz/pull/1199.

flybayer:
@sandulat we can close this in favor of #1199, right?

PR #1182, opened 2020-09-28
Add `useRedirectLoggedInUserTo()` hook

engelkes-finstreet:
Some users were asking how to redirect from login page if user is already logged in. Small change to redirect to the root page if a users is logged in from `signup` / `login` page.

### DIFF
--- a/packages/generator/templates/app/app/auth/pages/login.tsx
+++ b/packages/generator/templates/app/app/auth/pages/login.tsx
@@ -2,8 +2,10 @@ import React from "react"
 import { useRouter, BlitzPage } from "blitz"
 import Layout from "app/layouts/Layout"
 import { LoginForm } from "app/auth/components/LoginForm"
+import { useRedirectLoggedInUser } from "../../hooks/redirectLoggedInUser"
 
 const LoginPage: BlitzPage = () => {
+  useRedirectLoggedInUser("/")
   const router = useRouter()
 
   return (
--- a/packages/generator/templates/app/app/auth/pages/login.tsx
+++ b/packages/generator/templates/app/app/auth/pages/login.tsx
@@ -2,10 +2,10 @@ import React from "react"
 import { useRouter, BlitzPage } from "blitz"
 import Layout from "app/layouts/Layout"
 import { LoginForm } from "app/auth/components/LoginForm"
-import { useRedirectLoggedInUser } from "../../hooks/redirectLoggedInUser"
+import { useRedirectLoggedInUserTo } from "app/hooks/redirectLoggedInUser"
 
 const LoginPage: BlitzPage = () => {
-  useRedirectLoggedInUser("/")
+  useRedirectLoggedInUserTo("/")
   const router = useRouter()
 
   return (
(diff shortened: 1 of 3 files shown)


packages/generator/templates/app/app/hooks/redirectLoggedInUser.ts
```diff
@@ -0,0 +1,13 @@
+import { useRouter, useSession } from "@blitzjs/core"
+import { useEffect } from "react"
+
+export const useRedirectLoggedInUser = (url: string) => {
+  const session = useSession()
+  const router = useRouter()
+
+  useEffect(() => {
+    if (session.userId) {
+      router.push(url, url)
```
flybayer, line 10:
```suggestion
      router.push(url)
```

packages/generator/templates/app/app/auth/pages/signup.tsx
```diff
@@ -2,8 +2,10 @@ import React from "react"
 import { useRouter, BlitzPage } from "blitz"
 import Layout from "app/layouts/Layout"
 import { SignupForm } from "app/auth/components/SignupForm"
+import { useRedirectLoggedInUser } from "../../hooks/redirectLoggedInUser"
```
flybayer, line 5:
```suggestion
import { useRedirectLoggedInUser } from "app/hooks/redirectLoggedInUser"
```

packages/generator/templates/app/app/auth/pages/login.tsx
```diff
@@ -2,8 +2,10 @@ import React from "react"
 import { useRouter, BlitzPage } from "blitz"
 import Layout from "app/layouts/Layout"
 import { LoginForm } from "app/auth/components/LoginForm"
+import { useRedirectLoggedInUser } from "../../hooks/redirectLoggedInUser"
```
flybayer, line 5:
```suggestion
import { useRedirectLoggedInUser } from "app/hooks/redirectLoggedInUser"
```

flybayer (commented):
Awesome, thanks!!

What do you think about changing name from `redirectLoggedInUser()` to `redirectLoggedInUserTo()`? I feel like the later reads a bit better and is more clear about what the function argument is for.

engelkes-finstreet:
> Awesome, thanks!!
> 
> What do you think about changing name from `redirectLoggedInUser()` to `redirectLoggedInUserTo()`? I feel like the later reads a bit better and is more clear about what the function argument is for.

Way better name 👍  You are right!

flybayer:
Ack, I don't know why I didn't think about this before, but let's just add this hook to blitz/core! And then use that inside the templates.

engelkes-finstreet:
I was thinking about this as well, but I do not like the flickering to be honest. To use this for the special case that a user loggen in it is fine but I think we should provide a better solution for different cases.

engelkes-finstreet:
@flybayer To get this PR closed - Your call if I should move it inside core package or not.

flybayer:
Well, since we aren't super happy with the UX of it, let's close this PR and instead add the hook source to the docs as an example for people. Later we can revisit adding something like it to core.


To: https://blitzjs.com/docs/authorization

PR #1193, opened 2020-09-30
Change all generated file paths to be kebab-case

### LINKED ISSUE #1185: Change all generated file paths to be kebab-case
### What do you want and why?

Currently for a model of `TestResult` you end up with a path of
```
app/testsResults/pages/test-results/index.tsx
```

And I'm finding that I keep expecting it to instead be

```
app/test-results/pages/test-results/index.tsx
```

I think it's better for everything to be consistent.

We would need to change [this method](https://github.com/blitz-js/blitz/blob/canary/packages/generator/src/generators/page-generator.ts#L49) in all the generators.

brunocrosier:
fix #1185

Closes: #1185 

Unable to test this properly locally because `yarn link-cli` isn't working for me, but I think this should work?

### Checklist

- [ ] Tests added for changes

### DIFF
--- a/packages/generator/src/generators/page-generator.ts
+++ b/packages/generator/src/generators/page-generator.ts
@@ -47,7 +47,8 @@ export class PageGenerator extends Generator<PageGeneratorOptions> {
   }
 
   getModelNamesPath() {
-    const context = this.options.context ? `${this.options.context}/` : ""
+    const kebabCaseContext = camelCaseToKebabCase(this.options.context)
+    const context = this.options.context ? `${kebabCaseContext}/` : ""
     return context + this.options.modelNames
   }
 


packages/generator/src/generators/page-generator.ts
```diff
@@ -47,7 +47,8 @@ export class PageGenerator extends Generator<PageGeneratorOptions> {
   }
 
   getModelNamesPath() {
-    const context = this.options.context ? `${this.options.context}/` : ""
+    const kebabCaseContext = camelCaseToKebabCase(this.options.context)
+    const context = this.options.context ? `${kebabCaseContext}/` : ""
     return context + this.options.modelNames
```
flybayer, line 52: You also need to kebab case `this.options.modelNames`

flybayer (commented):
@brunocrosier to test this, test it inside one of the apps inside `examples/`. (store or auth example). Those will use your local dev version of blitz automatically.

CI is failing because there's some typescript errors with your code.

packages/generator/src/generators/page-generator.ts
brunocrosier: ah gotcha, created a new PR now. Hopefully that solves it!

PR #1199, opened 2020-09-30
Add `paginate` resolver utility and add it to `blitz generate` template

sandulat:
# Note by @flybayer on Feb 5, 2021

Now that Prisma has a `db.model.createMany()` API, this seeder utility is no longer needed. So this PR **no longer adds a seeder utility**

---

Implemented new DB agnostic utils: `paginate` and` seeder`.

Initially, the implementation of these 2 utils started with a heavy dependency on Prisma. I've battled Prisma's generated types for a lot of hours, and it turned out to be very hard to write complex generics for Prisma's types, with a lot of pitfalls and edge cases.

I thought that we could abstract these utils to be DB agnostic (since Prisma is very hard to parse), and as a huge bonus, we don't have to publish a separate package with Prisma as a dependency. This way, anyone who uses Blitz can leverage these utils.

In the end, the DB agnostic solution turned out to be even more scalable than the Prisma one.

### Paginate

`paginate` is a simple function that receives one object parameter with the following structure:

| Name    | Type                                                      | Optional | Description                                                                                                                     |
|---------|-----------------------------------------------------------|----------|---------------------------------------------------------------------------------------------------------------------------------|
| skip    | `number`                                                    | false    | How many rows to skip.                                                                                                          |
| take    | `number`                                                    | false    | How many rows to take.                                                                                                          |
| takeMax | `number`                                                    | true     | The maximum value allowed for `take`. It protects your DB when "take" is sent by the user. Defaults to 100.                                        |
| count   | `() => Promise <number>`                               | false    | A function that returns a promise which resolves to a number.                                                                   |
| query   | `(payload: { skip: number; take: number }) => T` | false    | A function that receives pagination payload and returns anything (type of return will be inherited by the pagination result). |

`paginate` returns the following object:

| Name       | Type                     | Description                                                |
|------------|--------------------------|------------------------------------------------------------|
| items      | `T`                        | Data resolved by the `query` parameter.                    |
| nextPage   | `{ skip: number, take: number } \| null` | Next page pagination payload.                              |
| hasMore    | `boolean`                  | True if there are more items to load.                      |
| itemsCount | `number`                   | Total number of items. Resolved by the  `count` parameter. |

Example of usage:
```js
const { items, hasMore, nextPage, count } = await paginate({
  skip: 0,
  take: 10,
  count: db.articles.count,
  query: db.articles.findMany,
})
```

Example of usage with a custom query:
```js
const query = { where: { published: true } };

const { items, hasMore, nextPage, count } = await paginate({
  skip: 0,
  take: 10,
  count: () => db.articles.count(query),
  query: (payload) => db.articles.findMany({ ...payload, ...query }),
})
```

Example of `take` validation:
```js
// Will throw PaginationPayloadInvalidError if request.take > 10
const { items, hasMore, nextPage, count } = await paginate({
  skip: request.skip,
  take: request.take, 
  takeMax: 10,
  count: db.articles.count,
  query: db.articles.findMany,
})
```

### Seeder

`seeder` is a simple function that receives one callback, which must return a promise. Example:

```js
import faker from 'faker'

const userSeeder = seeder(() => {
  return db.user.create({
    data: {
      email: faker.internet.email(),
    },
  })
})
```

`userSeeder` can be run like that:

```js
// Will seed one user with a random generated email.
const [user] = await userSeeder.seed()
```

The `.seed()` method receives one optional object parameter with the following structure:

| Name    | Type     | Optional | Description                                         |
|---------|----------|----------|-----------------------------------------------------|
| amount  | `number` | true     | Amount of data to seed.                             |
| options | `any`    | true     | Options to be passed to the callback of the seeder. |

Example of seeding more items at once:

```js
// Will seed 10 users with random generated emails.
const users = await userSeeder.seed({ amount: 10 })
```

Example of passing options to seeder's callback:

```js
import faker from 'faker'

const userSeeder = seeder((index, options?: { subscriptionActive: boolean }) => {
  return db.user.create({
    data: {
      email: faker.internet.email(),
      subscriptionActive: options?.subscriptionActive,
      subscriptionExpiresAt: options?.subscriptionActive ? new Date() : null,
    },
  })
})

// Will seed 10 users with random generated emails and active subscriptions.
const users = await userSeeder.seed({ amount: 10, options: { subscriptionActive: true } })
```

> Did you notice that the first argument of seeder's callback is `index`? It's the number of iteration from the seeder's loop. Might be useful if somebody want's to populate some fields with incremental computed data. 

> Note that `faker` is not a dependency. Developers are free to use whatever they want.

### Checklist
- [x] Tests added for changes


sandulat:
@flybayer I've ditched the `ensureFieldUnique` and `ensureFieldExists` utils because their logic can be achieved as easy as:

```js
if (db.user.count({ take: 1, where: { email: request.email } })) {
  throw new EmailUsedError();
}
```

So such utils might be redundant to use.

sandulat:
The tests have failed but it seems like it has nothing to do with this PR.

packages/core/src/db-utils.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  const itemsCount = await count()
+  const hasMore = skip + take < itemsCount
+  const nextPage = hasMore ? {take, skip: skip + take} : null
+
+  return {
+    items: await query({skip, take}),
+    nextPage,
+    hasMore,
+    itemsCount,
```
flybayer, line 53: Can we do this?
```suggestion
    count,
```

packages/core/src/db-utils.ts
```diff
@@ -0,0 +1,55 @@
...
+  takeMax?: number
+  count: () => Promise<number>
+  query: (payload: {skip: number; take: number}) => T
+}) => {
+  const skipValid = typeof skip === "number" && skip % 1 === 0 && skip >= 0
+  const takeValid = typeof take === "number" && take % 1 === 0 && take > 0 && take <= takeMax
+
+  if (!skipValid || !takeValid) {
+    throw new PaginationPayloadInvalidError()
+  }
```
flybayer, line 43: Can you split this out and throw a specific error for each one instead of combining them? You can use the same error class, but change the message for each.

packages/core/src/db-utils.ts
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    return seeds
+  },
+})
+
+export const paginate = async <T extends Promise<object[]>>({
+  skip,
+  take,
+  takeMax = 100,
```
flybayer, line 28: What do you think about increasing this default to maybe 500? 100 seems lower than necessary and could be annoying when people try to use more than 100 in their code.

packages/core/src/db-utils.ts
```diff
@@ -0,0 +1,55 @@
...
+  query: (payload: {skip: number; take: number}) => T
+}) => {
+  const skipValid = typeof skip === "number" && skip % 1 === 0 && skip >= 0
+  const takeValid = typeof take === "number" && take % 1 === 0 && take > 0 && take <= takeMax
+
+  if (!skipValid || !takeValid) {
+    throw new PaginationPayloadInvalidError()
+  }
+
+  const itemsCount = await count()
```
flybayer, line 45: Did you investigate to see if it's feasible to do `query()` and `count()` in parallel to save a bit of time?

packages/core/src/errors.ts
```diff
@@ -25,3 +25,11 @@ export class NotFoundError extends Error {
     super(message)
   }
 }
+
+export class PaginationPayloadInvalidError extends Error {
+  name = "PaginationPayloadInvalidError"
```
flybayer, line 30:
```suggestion
export class PaginationArgumentError extends Error {
  name = "PaginationArgumentError"
```

packages/core/src/errors.ts
```diff
@@ -25,3 +25,11 @@ export class NotFoundError extends Error {
...
+  statusCode = 422
+  constructor(message = "The skip or take parameters are invalid") {
```
flybayer, line 32:
```suggestion
  constructor(message = "The pagination arguments are invalid") {
```

Remove the specific fields here so that the base is generic, and then you can add a specific message when you throw it.

packages/core/src/index.ts
```diff
@@ -13,6 +13,7 @@ export * from "./middleware"
 export * from "./types"
 export * from "./supertokens"
 export * from "./passport-adapter"
+export * from "./db-utils"
```
flybayer, line 16: Change this to explicitly re-export the items that need exported (this index file is refactored on master to follow this pattern)

flybayer (commented):
Awesome, this is really great!!

In addition to my other comments, two other things:

- [x] Open PR to docs repo to document all of this
- [ ] Can you also update all the templates and our examples to use the paginate utility?

sandulat:
@flybayer I've implemented all the improvements you've suggested above. Also, the `seeder` util had some troubles when seeding larger amounts of data. For example, the seeder is freezing when I try to seed 1500 records. I've refactored it to use async batches, after which I was able to seed 50k records in around 30 secs. The size of one batch defaults to 500 items, and the interval between batches defaults to 10ms. These values can be adjusted individually, so everyone can adjust  the threshold themselves to match their needs/machine specs: `userSeeder.seed({ amount: 50000, batchSize: 1000, batchInterval: 100 })`. Could you please take a look?

flybayer:
Sweet, looking good! Then what's left is the two todo items in my previous comment

flybayer:
@sandulat have you tested this seeder util with sqlite? Looks like folks are having problems with sqlite and Promise.all for seeding: https://github.com/prisma/prisma/issues/2955#issuecomment-713371921

Issue #1198, opened 2020-09-30
Creating new session after revoking current one does not work as expected


johncantrell97:
### What is the problem?

As part of my application there is a flow where a user can be invited to join other users team.  One possible edge-case of this flow is that the user who is invited already has an account using a different email address then the one that was invited.  This means when they go to accept the invitation to create an account under this team they might actually be logged in already under their other account.

In an attempt to handle this I wanted to log them out of their current session and then log them into the new one under the new account.  If I use 

`await session.revoke()`
`await session.create(data)`

in a mutation back-to-back I notice a few things.  The main problem is that on the client side if I try to `useSession` the data that is returned is not the new session data (it's empty). 

### Steps to Reproduce

1. revoke session
2. create new one in same mutation

### Versions

blitz: 0.23.2-canary.2 (global)
blitz: 0.23.2-canary.3 (local)

johncantrell97:
Looks like this is the problem:

// This is used on the frontend to clear localstorage
setHeader(res, HEADER_SESSION_REVOKED, "true")


When revoke is called it sets a response header to clear localStorage data.  The new session that gets created must get wiped out on the front-end by the code that handles this header.

Fix is either to clear this header when createSession is run or update the logic that handles this header to clear the storage first before setting the new data (or let it set new data, depending on how this is implemented, haven't looked at it yet).

johncantrell97:
Not 100% sure but I think maybe just swapping the order of these two ifs in handling headers might fix it:

https://github.com/blitz-js/blitz/blob/90df4e840922ea22e3d7e70e0ac90c4a0db64819/packages/core/src/rpc.ts#L74

Seems like updates public data first and then handles the clear.  If we clear then handle public data it should fix it.  Unless in the case you really want to clear the update would still do something unintended.

johncantrell97:
not as simple as I had hoped as the clear() also deletes the cookie so the update that happens next won't have any data to load into storage.  Is it safe to clear the header during createSession as that seems like the next simplest approach

PR #1200, opened 2020-09-30
Fix bug with creating new session after revoking current one

johncantrell97:
Closes: https://github.com/blitz-js/blitz/issues/1198

This clears the HEADER_SESSION_REVOKED header from the response when a new session is created.  This allows someone revoke() and then create() within the same request without the new data being lost.

I tested this change in the examples/auth app by adding a session.revoke() call as part of the login mutation.  Before the fix I was experiencing the same issue I described in 1198. After this fix it works as expected.


flybayer:
Thanks!!

PR #1203, opened 2020-10-01
remove result logging for useQuery/useMutation

hardfire:

This change removes the annoying result log for useQuery/useMutation endpoints

### DIFF
--- a/packages/core/src/ssr-query.ts
+++ b/packages/core/src/ssr-query.ts
@@ -27,7 +27,6 @@ export async function ssrQuery<TInput, TResult>(
     try {
       log.progress(`Running ${logPrefix}(${JSON.stringify(params, null, 2)})`)
       const result = await enhancedResolver(params, res.blitzCtx)
-      log.success(`${logPrefix} returned ${log.variable(JSON.stringify(result, null, 2))}\n`)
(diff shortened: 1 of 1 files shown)


Zeko369 (changes requested):
While this is a great idea, this should be passed via a config from CLI.
Suggestions / ideas
- DISABLE_LOGS env var or --no-log param in `blitz start`
- Ability to ignore logging of some queries
- Ability to filter some data in logs (similar to how rails does it)
- Also logging to an external JSON file should be setup

@flybayer @ryardley any suggestions/ideas?

hardfire:
@Zeko369 Yes, that was the initial idea to control it using an env. config but somehow decided that this can be removed altogether. 

Since there will be work on the whole logging system in the future, i think (or rather hope) then it would be easy to enable/disable specific logs (or group of logs probably)

Since this discussion was on slack, ill put the screenshot here

image

Zeko369 (approved):
Then just keep it as is, removed :)

flybayer:
😬 I just now saw this. I overhauled all the logs yesterday in #1212 which already removed logging result data.

Thanks anyways @hardfire!

Issue #757, opened 2020-07-16
Add tests to session management


flybayer:
### What do you want and why?

We need to add comprehensive unit tests to our session management.

Feel free to submit PRs for single or few tests. And feel free to ask clarification questions.

Source file: https://github.com/blitz-js/blitz/blob/canary/packages/core/src/server/auth/sessions.ts
Unit tests: https://github.com/blitz-js/blitz/blob/canary/packages/core/src/server/auth/sessions.test.ts
Integration tests: https://github.com/blitz-js/blitz/tree/canary/test/integration/auth

cbadger85:
I'd like to try and tackle this issue! I haven't contributed to open source before, but I'm pretty handy with jest.

flybayer:
@cbadger85 that'd be awesome! Here's our [contributing docs](https://blitzjs.com/docs/contributing). And here's a great free egghead video on the basics of open source like forking, submitting a PR, etc: https://egghead.io/courses/how-to-contribute-to-an-open-source-project-on-github

Let me know if you have any questions!

geeklisted:
Is this issue still ready to work?  Seems there have been a couple of merges.

flybayer:
@geeklisted yeah, still plenty to tests to add! :)

geeklisted:
I hope to try and contribute some tests over the course of the week.  Excited to see a project heading back to the good ole' days of Web development.

clgeoio:
@geeklisted, I am probably going to get stuck in to the server/supertokens file soon. We should coordinate our efforts :)

swinner2:
Looks like these files have been moved around @flybayer

flybayer:
@swinner2 original issue updated!

flybayer:
Sufficient tests for now

PR #1204, opened 2020-10-01
Refactor and add tests for public data store

clgeoio:
Extracts out public datastore to a separate class to aid testability.  Uses `class` to avoid the need for an init method and uses the ctor instead. I made the emptyPublicData a property of the public data store because it's the only place that really uses it and it should be responsible for defining what this is.

 - Adds some testing to help close out: https://github.com/blitz-js/blitz/issues/757
- Adds an optional param to `updateStatus` to stop a cyclic call stack.

I've extracted out the `*Cookie` methods to their own utility file, I have not written test for them yet, but they can come soon.

This PR does do a decent amount of refactor work, so I would be happy to have a second reviewer.

### Checklist

- [x] Tests added for changes

### DIFF
--- a/packages/core/src/public-data-store.test.ts
+++ b/packages/core/src/public-data-store.test.ts
@@ -1,0 +1,22 @@
+import {publicDataStore} from "./public-data-store"
+import {getPublicDataToken} from "./supertokens"
+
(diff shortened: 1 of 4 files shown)


flybayer:
Nice, thank you!!

PR #1207, opened 2020-10-01
Fix husky pre-push script for new apps

flybayer:

Fix husky pre-push script for new apps


lukebennett:
Ah doh sorry, this was my bad - I have it right in my actual blitz app and somehow had a copy and paste fail 🙄 🤦‍♂️

flybayer:
No problem :)

Issue #1208, opened 2020-10-01
AssertionError after form prompt during `blitz new ... --js`


Weilbyte:
### What is the problem?
My environment is Ubuntu 20.04 inside WSL2. I am running `blitz new` with `--js` and after the `Hang tight while we set up your new Blitz app!` - it crashes with an error like below: 
```js
AssertionError [ERR_ASSERTION]: Trying to copy from a source that does not exist:
    /home/weil/projects/asdf/_forms/finalform/Form.tsx
    Code: ERR_ASSERTION
```

### Steps to Reproduce

1. Be on Ubuntu 20.04 on WSL2 (? not sure if this is contributing to the failure ?)
2. Run `blitz new aaagh --js --no-git`
3. Select a form thingy
4. Error

### Versions

```
debug: blitzPkgPath: /home/weil/projects/node_modules/blitz/dist/index.js
debug: cliPkgPath: /home/weil/projects/node_modules/blitz/node_modules/@blitzjs/cli/lib/src/index.js

Linux 4.19 | linux-x64 | Node: v14.13.0

blitz: 0.23.2 (global)
blitz: 0.23.2 (local)

  Package manager: npm
  System:
    OS: Linux 4.19 Ubuntu 20.04.1 LTS (Focal Fossa)
    CPU: (4) x64 AMD Ryzen 3 1200 Quad-Core Processor
    Memory: 10.84 GB / 12.44 GB
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 14.13.0 - /usr/bin/node
    Yarn: 1.22.5 - /mnt/c/Users/Weil/AppData/Roaming/npm/yarn
    npm: 6.14.8 - /usr/bin/npm
    Watchman: Not Found
  npmPackages:
    @prisma/cli: Not Found
    @prisma/client: Not Found
    blitz: Not Found
    react: Not Found
    react-dom: Not Found
    typescript: Not Found
```

### Other

The Typescript version does not raise an error.

flybayer:
Yep, that's a bug!

Weilbyte:
Will work on it tomorrow :)

PR #1211, opened 2020-10-02
Fix `blitz new --js` #1208

Weilbyte:
Closes: #1208 

This PR fixes the AssertionError that occurs after selecting a form to use for the template when creating a new Blitz JavaScript app. Adds in a simple ternary operator to the bit that moves the form files to make sure proper file extension is used.


flybayer:
Thanks @Weilbyte! This should work, but I wonder if you can clean it up a bit by doing something like this instead of all the ternaries?

```ts
const ext = useTs ? '.tsx' : '.js'
this.destinationPath("_forms/formik/LabeledTextField" + ext)
```

Weilbyte:
@flybayer Thanks! I added that in and also removed some other unnecessary repetition by simply having the form type in a variable for later.

flybayer:
Sweet, thank you!

Issue #856, opened 2020-08-11
Prevent Prisma from logging secrets in queries


dwightwatson:
### What do you want and why?

I noticed that the Blitz logs include what appear to be output from Prisma when queries are run. This is great, except for when potentially sensitive information is shown in these queries.

```
> Running login({
  "email": "dwightwatson@github.com",
  "password": "secretpassword"
})
```

Other frameworks have features that help protect sensitive parameters. Rails has `filter_parameter_logging` and Laravel prevents this arguments from being re-flashed. In both these examples it defaults to an array that contains "password", giving developers the ability to filter additional parameters if required.

Especially with Blitz supporting auth out of the box it probably makes sense that we filter these secret parameters if possible.

### Possible implementation(s)

If I'm correct in thinking this logging is powered by Prisma we'll probably need to [switch to event-based logging](https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-client/logging) and provide it to the Prisma client.

flybayer:
@dwightwatson you're absolutely right. This is logging by Blitz. 

I've had this on my todo list for awhile. I'll do this as part of a logging system overhaul I have planned.

PR #1212, opened 2020-10-02
Lots of logging & error improvements!! Remove result logs, redact passwords, etc

flybayer:
Closes: #856 

- [x] 🎉 Prevent the annoying "Unhandled error" popup for AuthenticationError and AuthorizationError
- [x] Stop logging the result of queries/mutations
- [x] Filter out `password` and `passwordConfirmation` fields. Later we can add a config item for users to extend this list
- [x] Overall nicer and more useful logging. Not super happy with it, but definitely better than what we have now.
- [x] Eliminate the gross prisma errors showing in the browser
- [x] Stop transmitting server error stack trace to client

---

image


ntgussoni:
Lovely! Awesome work

Issue #1162, opened 2020-09-25
Enforce the required node version to the user using the blitz cli (including blitz new, blitz install, ...)


jbFeDev:
### What is the problem?

If you're on a older Node version (< 12) you'll see what appears like a Node version warning ("You are using an unsupported version of Node.js. **Consider** switching to v12 or newer."), but actually is more required than optional. I could successfully scaffold out the application and build it, even though I'm on an unsupported Node version. The errors appear afterwards when trying to signup or create new models.

### Steps to Reproduce

1. Install/use Node 10.19
2. blitz new blitz-app (you'll see the message)
3. signup

### Versions

```
macOS Catalina | darwin-x64 | Node: v10.19.0

blitz: 0.23.0 (global)
blitz: 0.23.0 (local)

  Package manager: yarn 
  System:
    OS: macOS 10.15.5
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 497.67 MB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 10.19.0 - ~/.nvm/versions/node/v10.19.0/bin/node
    Yarn: 1.19.1 - ~/.yarn/bin/yarn
    npm: 6.14.8 - ~/.nvm/versions/node/v10.19.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  npmPackages:
    @prisma/cli: 2.7.1 => 2.7.1 
    @prisma/client: 2.7.1 => 2.7.1 
    blitz: 0.23.0 => 0.23.0 
    react: 0.0.0-experimental-7f28234f8 => 0.0.0-experimental-7f28234f8 
    react-dom: 0.0.0-experimental-7f28234f8 => 0.0.0-experimental-7f28234f8 
    typescript: 4.0.3 => 4.0.3 
```

nitaking:
@jbFeDev Thanks for creating the issue!

PR #1213, opened 2020-10-02
Enforce NodeJS Version >= 12 for CLI

phillippschmedt:
Closes: https://github.com/blitz-js/blitz/issues/1162

Blitz officially requires NodeJS Version >= 12. Issues show that running it with nodejs below that version will result in problems. Therefore we enforce the version and exit the CLI if the nodejs version is lower than 12. This will save users from headaches.

### DIFF
--- a/packages/blitz/src/bin/cli.ts
+++ b/packages/blitz/src/bin/cli.ts
@@ -22,7 +22,7 @@ async function main() {
         `You are using an unsupported version of Node.js. Please switch to v12 or newer.\n`,
       ),
     )
-    process.exit(1)
+    process.exit()
   }
 
   const globalBlitzPath = resolveFrom(__dirname, "blitz")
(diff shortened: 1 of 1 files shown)


phillippschmedt:
This is my first pull request here. How do I deal with the tests and checklist? Currently, there is no test for the nodejs version warning. I am not sure how I could write a test for this. I would probably not worry for a test about this and also not edit any docs because the docs already say that blitz requires version > 12.

engelkes-finstreet (approved):
Looks good 🚀

engelkes-finstreet:
> This is my first pull request here. How do I deal with the tests and checklist? Currently, there is no test for the nodejs version warning. I am not sure how I could write a test for this. I would probably not worry for a test about this and also not edit any docs because the docs already say that blitz requires nodejs >= 12.

It is fine as it is @phillippschmedt. Asking for tests is just a reminder in the template for PullRequests. This fix does not need any tests.

nitaking:
Nice work, @phillippschmedt!

flybayer:
Thank you!!

@all-contributors add @phillippschmedt for code

PR #1219, opened 2020-10-02
Add ability to strongly type `PublicData`!

flybayer:

Add ability to strongly type `PublicData`. This also strongly types `session.userId`

You do this the same way you strongly type Ctx. Here's what your `types.ts` file should look like this.

```ts
import {DefaultCtx, SessionContext, DefaultPublicData} from "blitz"
import {User} from "db"

declare module "blitz" {
  export interface Ctx extends DefaultCtx {
    session: SessionContext
  }
  export interface PublicData extends DefaultPublicData {
    userId: User["id"]
  }
}
```

And you can add extra items to PublicData, and it will be properly typed for `session.create()`, `useSession`, etc.

### Checklist

- [x] Tests added for changes


hasparus:
I think I have a guess how to fix it. It seems that there is more than one PublicData loaded, so our global augmentation doesn't really click. I have some friends over at my place in a moment, so I'll have get back to it later, but this is definitely solvable.

flybayer:
@hasparus ok thank you so much!! 

Something that could be complicating things is our package setup. `PublicData` initially defined in `@blitzjs/core` and then re-exported from `blitz`. And using declare module on `blitz`. I tried declare module on `@blitzjs/core` but that didn't help.

hasparus:
I'll describe my reasoning. For the record — I have no clear answer, just the road I took to derive a hypothesis/guess. It's been a long time since I read TypeScript compiler, so I have only a vague idea what's happening.

---

I made SessionContext generic to start small and tackle one problem at once.

```
- export interface SessionContext {
+ export interface SessionContext<T extends PublicData = PublicData> {
```

The fact that is works suggests that the module augmentation didn't really click.  

Declaration merging is like mutating a global variable on type level. It's a little less obnoxious, because there's no time in the equation. With a global state, _when_ is super important and if we change the order of operations.

Let's try inspecting the keys of `session` returned from `useSession`.

```tsx
  const _keyOfSession = "roles" as keyof typeof session
  const _debug: never = _keyOfSession
```
```
error TS2322: Type '"userId" | "isLoading" | "roles"' is not assignable to type 'never'.
  Type '"userId"' is not assignable to type 'never'.

29   const _debug: never = _keyOfSession
           ~~~~~~
```

What's interesting, TypeScript 4.0.2 in my VSCode doesn't show this error. It simplifies that union into a string. Did the type-checking algorithm change between TS 3.8.3 installed in the workspace? Or is it just a change in the error formatter?

```
  Type 'string' is not assignable to type 'never'.ts(2322)
```

In TS 4.0.2, I need an arbitrary string literal instead of `never` there to show all keys of `session`.

```
error TS2322: Type '"userId" | "isLoading" | "roles"' is not assignable to type '"That is the most extreme form of nihilism: nothingness (the 'meaningless') eternally!"'.
```

This lead me to an explanation to why the problem occurs in `tsc`, but not in VSCode. Newest VSCode comes with TS 4.0.2, and TS 3.8.3 is installed in node_modules. When I change used TypeScript version in VSCode (bottom-right), the error appears.

I'm not satisfied with the fact that the problem doesn't appear in a newer TypeScript version.
Let us continue.

Let's go back to our _global variable_ mental model. What else can go wrong with global state?
Well... What if someone clones it, and we end up with two copies?

Could this be what's happening here?
Is this why accepting a generic parameter in `SessionContext` solved these other problems?
Do we have both `.tsx?` sources and built `.d.ts` declaration files.
We also reexport `interface PublicData` from `@blitzjs/core` to `blitz`.
Not only that, in `@blitzjs/core` we reexport it from `"./supertokens"`.

Where is `PublicData` truly defined? Is reexporting not _sharing_, but _copying_?

I don't know answers to this questions, and while I find them quite intriguing, I already had a little too much wine this night.

I don't know where `PublicData` is, but I can say where I want `PublicData` to be.
I want it to be in `"blitz"`, because we're using declaration merging on `PublicData` from `"blitz"` in user's code.

In `supertokens.ts`, I wrote:

```
- export interface PublicData extends DefaultPublicData {}
+ declare module "blitz" {
+  export interface PublicData extends DefaultPublicData {}
+ }
+ import { PublicData } from "blitz";
```

It worked 🤷

---

**Disclaimer:** For anybody reading this who is just learning TypeScript: Don't worry. You don't need these hacks.
Never teach this to newcomers, you'll scare them. You certainly don't need this to build solid apps. Frameworks may sometimes employ unorthodox methods to achieve their goals (React throws Promises y'all).

flybayer:
Wow amazing, thank you so much! You never cease to amaze me lol. Slowly I'm learning advanced TS concepts

flybayer:
@hasparus the blitz build is failing with

```
@blitzjs/core: (typescript) Error: /Users/b/c/blitz/packages/core/src/supertokens.ts(35,16): semantic error TS2664: Invalid module name in augmentation, module 'blitz' cannot be found.
```

So I'm going to upgrade the monorepo TS to 4 and see how that goes

hasparus:
Oh, that looks like the problem I mentioned in https://github.com/blitz-js/blitz/pull/1232#issuecomment-702987836

hasparus:
Updating to TS4 sounds good. Probably not worth to add this @blitz/types package, as it should be totally okay for "alpha software" framework to require latest language major version.

hasparus:
> Wow amazing, thank you so much! You never cease to amaze me lol. Slowly I'm learning advanced TS concepts

I'd like to notice it didn't even work on fresh build (can't augment module which isn't already built), so I clearly have no idea what I'm doing :D

hasparus:
Ha! Reexporting seems to be the culprit. I think we'd have to read TypeScript's changelog or commit history to figure out why it works in 4.0.

image

I have no idea why it works with `Ctx` though.
As a side-note: Bundling emitted .d.ts to one file with [Microsoft API Extractor](https://api-extractor.com/pages/overview/demo_rollup/) could be another workaround.

flybayer:
@hasparus it works with Ctx because in the app you explicitly import Ctx. Whereas PublicData is never explicitly imported in the app

Issue #1153, opened 2020-09-24
Prefix session cookies with unique identifier


flybayer:
### What do you want and why?

Currently all our session cookies are always named the same like `sSessionToken`. This causes a problem when you are running different blitz apps on the same domain, like `localhost` for example. 

This means you can only be logged in to one blitz app on a domain at a time. We should fix this.

### Possible implementation(s)

We want to use some type of unique identifier for the app. For example we could use the name field in package.json by default and also expose a config item in session middleware to set it manually.

One tricky thing is that we also need to know the prefix in the browser to read the right cookies. So somehow this config item will need to be available there.

sandulat:
@flybayer The only solution that comes to my mind is to use the `publicRuntimeConfig` option in the config, instead of passing options to the session middleware.
https://nextjs.org/docs/api-reference/next.config.js/runtime-configuration

If that's ok, I could work on this.

PR #1229, opened 2020-10-02
Prefix session cookies with app name

sandulat:
Closes: https://github.com/blitz-js/blitz/issues/1153

I've tried to automatically prefix the cookies based on the app's name, using `process.env.npm_package_name`, however, this variable is not available on the front-end.

In my implementation, the session can be prefixed by setting the optional `SESSION_PREFIX` env variable. This way, the prefix is available both on the server and the client.

Example:
```
SESSION_PREFIX=myAppName
```

### Checklist

- [ ] Tests added for changes


packages/core/src/supertokens.ts
```diff
@@ -2,18 +2,23 @@ import {useState} from "react"
 import BadBehavior from "bad-behavior"
 import {useIsomorphicLayoutEffect} from "./utils/hooks"
 import {queryCache} from "react-query"
+import getConfig from "next/config"
+
+const {publicRuntimeConfig = {}} = getConfig() || {}
```
sandulat, line 7: Next's `getConfig` will return the config on the client only.

packages/core/src/supertokens.ts
```diff
@@ -2,18 +2,23 @@ import {useState} from "react"
...
+
+const sessionPrefix = publicRuntimeConfig.sessionPrefix || process.env.SESSION_PREFIX || "s"
```
sandulat, line 9: `process.env.SESSION_PREFIX` will be available on the server only.

flybayer:
Thanks @sandulat! Let's find a way to make it work automatically by default.

Perhaps in withBlitz we can do this? Anything in `env` will be available in the frontend via `process.env.sessionPrefix`

```
env: {
  sessionPrefix: process.env.npm_package_name || process.env.SESSION_PREFIX
}
```

But we'd also need to merge in `env` from the user's blitz.config.js

sandulat:
Great idea @flybayer! I've implemented it. The only thing is that when the app's name in `package.json` equals for example `@examples/auth`, the session is being set, however, the authentication doesn't effectively happen. It feels like the server can't match the cookies. When the name is just `auth`, everything works perfectly. It looks to me that one of the `@` `/` symbols are somewhere stripped, but I don't really understand where, and which is the pattern.

flybayer:
Cool! I think we should replace any non-url safe characters with and underscore. So `@examples/auth` would turn to `_examples_auth`.

Also, I had a new concern about the process.env thing. Can you try setting `isInternalBlitzDevelopment` to `false` in `withBlitz` and see if it still works? The way the env values are inlined via webpack is a bit goofy, so I want to make sure it works without the transpile modules plugin we have there in the monorepo.

sandulat:
@flybayer I've set isInternalBlitzDevelopment to false, and everything seems to work just fine. I've also pushed the replacement of non-url safe characters in the prefix.

flybayer:
@sandulat Sweet, thanks! I'll go ahead and merge, but please also open a PR to the docs site to document this new functionality.

PR #1232, opened 2020-10-02
Make strongly typed PublicData "global" (in "blitz" module) 

hasparus:
Described in https://github.com/blitz-js/blitz/pull/1219#issuecomment-702983497

cc @flybayer

### DIFF
--- a/packages/core/src/supertokens.ts
+++ b/packages/core/src/supertokens.ts
@@ -32,7 +32,10 @@ export interface DefaultPublicData {
   roles: string[]
 }
 
-export interface PublicData extends DefaultPublicData {}
+declare module "blitz" {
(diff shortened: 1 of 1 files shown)


flybayer:
@all-contributors add @hasparus for code

hasparus:
I wanted to mention, this is a hack. Not only does it make "@blitjs/core" depend on "blitz" module, which looks like a circular dependency, I'm a bit worried it may make trouble during trouble build process.

This guides us to a solution, though.

If my worries turn up to be true problems (I didn't test them), and reexported types prove to be hard to extend using module augmentation, we could try to create single file "@blitz/types" package for all interfaces which are meant to be extensible.

Issue #756, opened 2020-07-16
Add advanced RBAC roles + permissions based authorization option


flybayer:
### What do you want and why?

Not sure on exactly what is best, but here's one idea.

The idea is that roles are groups of permissions. And you enforce permissions, not roles.

So you centralize the definition of roles with something like this:
```ts
const roles = {
  admin: ['product:read', 'product:create', 'product:update', 'product:delete'],
  manager: ['product:update'],
}
```

And then in your code you enforce a permission.

```ts
export default async function updateProduct(input, ctx) {
  ctx.session.authorize('product:update')
}
```

This allows you to add/remove/refactor your roles without also having to change a ton of other files.

This will be added as an experimental authorization implementation.

---

Another idea is use something like https://github.com/onury/accesscontrol

SharadKumar:
Your statement about roles being permission groups is spot on. Similarly, permissions for an entity are just a group of permissions that you need per property.

Field or Column level permissions are often needed, in any application. Hasura does it well, where they track that information per field in its metadata tables.

flybayer:
Here's some noodling from today on ways to integrate authorization at the prisma level.

Have some builder pattern to build up authn rules. The second argument is a prisma `Where` statement or a function that takes context object and returns a `Where` statement.

```ts
const secureDb = new SecuredPrisma(prismaClient)

secureDb.grant('anonymous')
    .read('post', {published: {equals: true}})
  .grant('user')
    .manage('project', ctx => ({owner: {id: ctx.userId}}))
    .read('project', ctx => ({creator: {id: ctx.userId}}))
    .create('task', ctx => ({project: {connect: {organization: {id: ctx.orgId}}}}))
  .grant('superadmin')
    .manageAll('project')
    .manageAll('task')

export secureDb
```

And then use that like this. It will automatically throw AuthenticationError or AuthorizationError based on who the current user is and what they are trying to access. No need to manually call `session.authorize()`

```ts
import secureDb from "db";

export default async function getProject(args, { session }) {
  const project = await secureDb
    .ctx(session)
    .project.findMany({ where: { id: args.id } }).one()

  if (!project) throw new NotFoundError();

  return project;
}
```

Zeko369:
@flybayer Really love the idea but have a different API idea, something closer to prisma itself. Also I think we should see if Prisma Team has any ideas about this @janpio @nikolasburk

```ts
// anonymous users can only select certain columns
const anonymousGrant = new Grant({
  roles: null,
[...]

PR #1234, opened 2020-10-02
Add `protect` decorator function + add it to query/mutation templates

flybayer:

Adds `protect()` decorator function for queries and mutations.

`protect` can do the following:

- Ensure user is logged in (does this by default)
- Ensure user has certain role (only allows `superadmin` or `SUPERADMIN` roles by default)
- If given a zod schema, it will use it to parse and validate the input. And the input arguments will be automatically typed.

--

- [ ] **Remaining thing to figure out:** the idea is to fully lock everything down by default (only `superadmin`), but this means when the developer signs up as a new user, they won't have the superadmin role. One solution is to create a superuser for them during new app creation. However this does have the possibility of failing, but I suppose we can tell them the steps to do it manually.
- [ ] I still need to test all templates to ensure they work

### Checklist

- [ ] Tests added for changes


packages/core/src/authorization.ts
```diff
@@ -0,0 +1,28 @@
...
+) => {
+  return (rawInput: U, ctx: Ctx) => {
+    if (authorize) {
+      const authorizeInput: any[] = ["superadmin", "SUPERADMIN"]
+      if (Array.isArray(authorize)) {
+        authorizeInput.push(...authorize)
+      } else if (typeof authorize !== "boolean") {
+        authorizeInput.push(authorize)
+      }
+      ;(ctx as any).session.authorize(authorizeInput)
```
johncantrell97, line 23: This doesn't leave anyway to call session.authorize() with no params (ensure logged in user without caring about what role they are).

packages/core/src/authorization.ts
```diff
@@ -0,0 +1,28 @@
+import {Ctx} from "./middleware"
+import {AuthenticatedSessionContext} from "./supertokens"
+import {ZodSchema, infer as zInfer} from "zod"
+
+export type ProtectArgs<T> = {schema?: T; authorize?: boolean | unknown}
```
johncantrell97, line 5: is typing authorize as boolean | unknown correct when the implementation allows it to be an array or string as well? fwiw, I ended up splitting it into two args (authorize: boolean, roles: string[] | string)

examples/auth/db/schema.prisma
```diff
@@ -43,3 +43,27 @@ model Session {
...
+  id        Int      @default(autoincrement()) @id
+  createdAt DateTime @default(now())
+  updatedAt DateTime @updatedAt
+}
+
+model User {
+  id        Int      @default(autoincrement()) @id
+  createdAt DateTime @default(now())
+  updatedAt DateTime @updatedAt
+}
```
johncantrell97, line 69: lots of duplicate User models :) I guess some kind of generator issue?

examples/auth/app/users/mutations/updateUser.ts
```diff
@@ -0,0 +1,17 @@
+import {protect} from "blitz"
+import db from "db"
+import * as z from "zod"
+
+export default protect(
+  {
+    schema: z.object({
```
johncantrell97, line 7: should the example really put the schemas inline like this? I like extracting and exporting it so I can re-use in my forms for validation.

packages/core/src/authorization.ts
```diff
@@ -0,0 +1,28 @@
...
+
+interface AuthenticatedCtx extends Ctx {
+  session: AuthenticatedSessionContext
+}
+
+export const protect = <T extends ZodSchema<any, any>, U = zInfer<T>>(
+  {schema, authorize = true}: ProtectArgs<T>,
+  resolver: (args: U, ctx: AuthenticatedCtx) => any,
+) => {
+  return (rawInput: U, ctx: Ctx) => {
```
johncantrell97, line 15: Did you not run into typing issues client-side without typing the return value? I had issues leaving it like this and had to introduce another generic type V that extends the resolver function type and then typed the return value of the returned function as:

return (rawInput: U, ctx: Ctx): ReturnType<V> => {

packages/core/src/authorization.ts
johncantrell97: I guess it sounds like you're solving for usage to be that you need to explicitly list all roles you want to allow instead of the opposite (what I'm suggesting) where the default is empty authorize() call and only add roles if you need it.

Not clear to me what the better developer experience is but I guess your direction is safer in the sense that everything is locked down by default.  It does mean if I add a new role I then have to go back and add it everywhere I use protect.  That should happen pretty rarely though so not a big deal.  I guess you can also just have a roles util where you export constants like ALL_ROLES, ALL_ADMIN_ROLES, MANAGEMENT_ROLES, etc and just protect using those and then can just update the actual roles in one place.

flybayer: Yeah, and this again points to why we really need a more permissions-based model: https://github.com/blitz-js/blitz/issues/756

aem (commented):
Obviously a little out of date on the current auth setup, but overall structure looks good. Let me know once you've got the last todos figured out and I'll take another pass

KATT:
I like the ideas and I definitely think authorization and data validation should be included in blitz.

Some thoughts:

- Authorization and data validation/parsing are two separate concerns. Maybe start with tackling one and refactor to get the other one in nicely.
- Would be great if we'd pick a validation-lib like zod that works out of the box and could be included. At the same time, it'll off-load core security concerns to out of the scope of blitz - what if there's a bug in the lib that doesn't validate properly and hence makes it possible to do db-injections? 
- I think data validation should be a first-class citizen and be **mandatory**. 

My idea of usage is something like this

```ts
import {protect, InputValidationError, Authorization, endpoint} from "blitz"
import db from "db"
import * as z from "zod"

// could be imported in react maybe? or put somewhere else so it's reused without transform magic
export const schema = z.object({
  id: z.number(),
})

export default endpoint({
  /**
   * mandatory, returns typed object used in resolver
   */
  async input(input: unknown, ctx) {
    const obj = schema.parse(input)
    if (obj.id < 1) {
      throw InputValidationError([
        {
          path: ["id"],
          message: "id cant be lower than 1",
        },
      ])
    }
    return obj
  },
  /**
   * Maybe? not convinced this would be needed
   * the benefit of doing this straight in the resolver is
   * that you get type inference on stuff like ctx.session, etc
   */
  async authorize(input /* inferred type from schema */, ctx): 'unauthorized' | 'forbidden' | 'ok' {
    if (!ctx.session) {
      return 'unauthorized' // -> 401
    }

    // receive the parsed input from `input`
    if (ctx.session.role === "admin" || ctx.session.userId === input.userId) {
      return 'ok;
    }
    return 'forbidden' // -> 403
  },
  async resolver({id} /* inferred type from schema */, ctx) {
    const user = await db.user.delete({where: {id}})

    return user
  },
})
```

I have a feeling that (only glanced at the transformation code so might be wrong) that always exporting an `endpoint`-function could simplify/reduce some of the codegen for rpc-handlers etc

flybayer:
For anyone following this PR, I just made an alternative proposal here: https://github.com/blitz-js/blitz/discussions/1787

flybayer:
Closing in favor of https://github.com/blitz-js/blitz/pull/1801

PR #1243, opened 2020-10-03
[RFC] change generated pages/[id].tsx to pages/[id]/index.tsx

mizchi:

current generated `pages` are below.

```
- [id]/edit.tsx
- [id].tsx
- index.js
- new.js
```

My proposal 

```
- [id]/edit.tsx
- [id]/index.tsx
index.js
new.js
```

## What is problem?

In my opinion, having the `[id]` directory and `[id].tsx` side by side is a bit confusing.

The `[id]/index.tsx` solves this problem.

### Checklist

- [x] Tests added for changes


nitaking:
+1 I agree!

flybayer:
Ok @mizchi everyone seems to like this, so we are good to merge. Just need to resolve the merge conflict and we're good to go.

mizchi:
ok! I'll fix it soon.

MrLeebo:
One drawback to the refactor is that if you don't need edit.tsx after generating,  it is one action to delete the folder. With this change, you also have to move the [id].tsx file into the parent folder and rename it. All other things being equal, that seems like a net productivity loss for apps with a lot of readonly models.

mizchi:
In my opinion, resource/xxx is an action to the plural and resource/[id]/xxx is an action to the singular.
CRUD is just a common boilerplate, and the actual implementation with multiple verbs would be listed here.

MrLeebo:
CRUD may be how you organize your API endpoints, but I'm not convinced that you should restrain your pages and UI to the same pattern. Delete is rarely its own page, and you can often combine Create and Update together.

flybayer:
The main reason I originally implemented like it currently is, is to have less files named `index.tsx`. But I think it's fine to change, since majority seems to want it. 

This could be a config option at some point.

flybayer:
Going to lay this aside for now with reason being desire to have as few "index" files as possible.

Issue #1214, opened 2020-10-02
window.requestIdleCallback is not supported on Safari


jan-wilhelm:
### What is the problem?
The fix that @flybayer implemented in #1174 does not work on Safari. Safari throws an error:
```
TypeError: window.requestIdleCallback is not a function. (In 'window.requestIdleCallback(function () {
            res(result);
          })', 'window.requestIdleCallback' is undefined)
```

According to [CanIUse](https://caniuse.com/requestidlecallback), Safari does not support window.requestIdleCallback on MacOS and iOS.

### Steps to Reproduce
Run the following code somewhere:

```typescript
const [data, {mutate}] = useQuery(...);

mutate({...data, test: "test" /*this is a new object*/});
```
### Versions

```
macOS Catalina | darwin-x64 | Node: v12.16.1

blitz: 0.18.0 (global)
blitz: 0.23.2-canary.3 (local)

  Package manager: yarn 
  System:
    OS: macOS 10.15.7
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 19.32 MB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 12.16.1 - /usr/local/bin/node
    Yarn: 1.22.5 - /usr/local/bin/yarn
    npm: 6.14.7 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  npmPackages:
    @prisma/cli: 2.8.0 => 2.8.0 
    @prisma/client: 2.8.0 => 2.8.0 
    blitz: v0.23.2-canary.3 => 0.23.2-canary.3 
    react: 0.0.0-experimental-33c3af284 => 0.0.0-experimental-33c3af284 
    react-dom: 0.0.0-experimental-33c3af284 => 0.0.0-experimental-33c3af284 
    typescript: 4.0.3 => 4.0.3 
```

flybayer:
Hmm, probably need to use `setTimeout()` instead maybe

jamiedavenport:
Could I pick this up? 🥺

PR #1247, opened 2020-10-04
Fix Safari compatibility: shim requestIdleCallback

jamiedavenport:
Closes: #1214 

Adds a shim for `requestIdleCallback` when undefined. `requestIdleCallback` is not supported on every browser, see https://caniuse.com/requestidlecallback.

### Checklist

- [x] Tests added for changes

### DIFF
--- a/packages/core/src/utils/react-query-utils.ts
+++ b/packages/core/src/utils/react-query-utils.ts
@@ -2,6 +2,7 @@ import {queryCache, QueryKey} from "react-query"
 import {serialize} from "superjson"
 import {Resolver, EnhancedResolverRpcClient, QueryFn} from "../types"
 import {isServer, isClient} from "."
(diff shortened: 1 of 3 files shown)


packages/core/src/utils/react-query-utils.ts
```diff
@@ -20,6 +20,21 @@ export interface QueryCacheFunctions<T> {
...
+        return setTimeout(function () {
+          cb({
+            didTimeout: false,
+            timeRemaining: function () {
+              return Math.max(0, 50 - (Date.now() - start))
+            },
+          })
+        }, 1)
+      }
+
```
flybayer, line 37: Awesome, thank you @jamiedavenport! Just one thing, can you move this shim out of `getQueryCacheFunctions` up to the top level?

jamiedavenport: Updated

Issue #1210, opened 2020-10-01
Improve `prisma generate` behaviour for updating a model that already exists in the schema 


brunocrosier:
### What do you want and why?

When I want to update a model in the schema, if I run `blitz generate all projects` and then run `blitz generate all projects title:string` it will add another `Project` model to the `schema.prisma` file, which will then fail when you run `blitz db migrate`

The aim should be to make it easy for developers to _update_ their already existing models, from the CLI.

Although both `blitz generate all` and `blitz generate resource` actually also update the queries and mutations, I think this should be made much clearer in the docs.

### Possible implementation(s)

How might we do this?

When `prisma generate all` or `prisma generate resource` is run, we should check whether the model already exists in the schema. If it does, we should just update the model rather than creating a new one. If it doesn't already, we can create it from scratch (which it already does).

From what I can tell, this is happening [here](https://github.com/blitz-js/blitz/blob/763252a5ed1e142b11f0afd621ce4c17d9c57090/packages/generator/src/generators/model-generator.ts#L12)

PR #1251, opened 2020-10-05
Fix `blitz generate model` to add fields to existing model instead of duplicating

engelkes-finstreet:
Updates the existing model only with missing fields.

Closes: #1210

Right now if a user generates a model that already exists a new model is added to the `schema.prisma`. This changes the behavior to update the existing model only with the fields that do not exist yet.

### Checklist

- [X ] Tests added for changes
- [X ] PR submitted to [blitzjs.com](https://github.com/blitz-js/blitzjs.com) for any user facing changes - [PR](https://github.com/blitz-js/blitzjs.com/pull/214)

### DIFF
--- a/packages/generator/src/generators/model-generator.ts
+++ b/packages/generator/src/generators/model-generator.ts
@@ -3,6 +3,7 @@ import path from "path"
 import {Model} from "../prisma/model"
 import {Field} from "../prisma/field"
 import {log} from "@blitzjs/display"
+import {matchBetween} from "../utils/match-between"
 
 export interface ModelGeneratorOptions extends GeneratorOptions {
   modelName: string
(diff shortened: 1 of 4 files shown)


packages/generator/src/generators/model-generator.ts
```diff
@@ -38,7 +39,26 @@ export class ModelGenerator extends Generator<ModelGeneratorOptions> {
...
+
+            //add new fields to the selected model
+            const newModel = model.replace("}", `${modelDefinition.getNewFields()}}`)
+
+            //replace all content with the newly added fields for the already existing model
+            this.fs.write(path.resolve("db/schema.prisma"), schema.replace(model, newModel))
+          }
+        }
       }
       log.success(
```
flybayer, line 63: Awesome @engelkes-finstreet!! Works great!

One more tweak, let's switch this success message do say "created" or "updated" depending on whether it added the whole model or just updated it.

PR #1259, opened 2020-10-06
Add tests for add-dependency-executor

Dohxis:
Added tests for `add-dependency-executor`. I decided to only test the actual functions that are doing real work(for example, spawning commands) instead of testing `Propose` and `Commit` components. The consequence of this is that I needed to export those functions, but I don't think it is a bad thing as I saw this in other places as well.

### Checklist

- [x] Tests added for changes
- [ ] ~~PR submitted to [blitzjs.com](https://github.com/blitz-js/blitzjs.com) for any user facing changes~~

### DIFF
--- a/packages/installer/jest.config.js
+++ b/packages/installer/jest.config.js
@@ -1,7 +1,7 @@
 module.exports = {
   preset: "ts-jest",
   testEnvironment: "node",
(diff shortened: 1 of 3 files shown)


flybayer:
Thank you!!

@all-contributors add @Dohxis for test, code

PR #1262, opened 2020-10-06
Fix broken session cookies prefix

flybayer:

Fix broken session cookies prefix which was added in `v0.24.0-canary.3` via https://github.com/blitz-js/blitz/pull/1229

Closes: https://github.com/blitz-js/blitz/issues/1153

### DIFF
--- a/packages/config/src/index.ts
+++ b/packages/config/src/index.ts
@@ -3,7 +3,17 @@ import {join} from "path"
 import {existsSync} from "fs"
 import {PHASE_DEVELOPMENT_SERVER, PHASE_PRODUCTION_SERVER} from "next/constants"
 
+export const getProjectRoot = () => {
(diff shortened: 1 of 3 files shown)


flybayer:
@sandulat I'm thinking maybe we should use https://github.com/kentcdodds/babel-plugin-preval to inline the prefix value at build time instead of using process.env

sandulat:
@flybayer I've tried `preval`, however, it seems like the env variables, paths, package.json, will always be scoped to `@blitzjs/core` and not applications since `preval` is being compiled during `@blitzjs/core` build, and basically `_blitzjs_core_` will always be the prefix. Am I missing something here?

flybayer:
@sandulat hmm, I'm sure there's some way to set this up where preval only runs at the actual app build. It might require generating a bit of code with the blitz compiler, like we do with the RPC stuff.

PR #1263, opened 2020-10-06
add @default(now())  to generated updatedAt field

brunocrosier:

I noticed that the current schema setup generates something like:
```updatedAt     DateTime     @updatedAt```

The problem here is that when you first create a database entry using this model, it will have a `createdAt` value of `now()` but an `updatedAt` value of 1st Jan 1970

This PR should make it so that when a database entry is first created, the `createdAt` and `updatedAt` values will be the same.

It should now look like:

```updatedAt     DateTime     default(now()) @updatedAt```

### Checklist

- [ ] Tests added for changes

### DIFF
--- a/examples/auth/db/schema.prisma
+++ b/examples/auth/db/schema.prisma
@@ -22,7 +22,7 @@ generator client {
 model User {
   id             Int       @default(autoincrement()) @id
   createdAt      DateTime  @default(now())
-  updatedAt      DateTime  @updatedAt
+  updatedAt      DateTime  @default(now()) @updatedAt
   name           String?
   email          String    @unique
   hashedPassword String?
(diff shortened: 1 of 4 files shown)


flybayer:
Hey @brunocrosier, according to Prisma this is not needed. `@updatedAt` also sets the default. If it's not working like that, then see if you can make a minimal reproduction with latest prisma, and open [a Prisma issue](https://github.com/prisma/prisma).

aem:
You could potentially reproduce this if you're adding fields via studio or accessing the DB directly, the docs do mention that this is enforced in the Prisma Client code

brunocrosier:
Yes this was happening to me while using studio

aem:
Ok, I think in that case this is a bug in Prisma Studio and should be opened against their project - behavior shouldn't be different when using studio vs. client

brunocrosier:
Agreed!

Issue #1037, opened 2020-09-10
Prompt to choose database engine on app creation


yhoiseth:
Extracted from blitz-js/legacy-framework#764. Writing this up as a new issue because blitz-js/legacy-framework#764 was closed.

### What do you want and why?

A way to choose which database engine to use (Postgres or SQLite) on app creation.

### Possible implementation(s)

@flybayer wrote:

> I think we should add a prompt to blitz new that asks if you want sqlite or postgres (same as we do now for form library).

- [ ] For postgres we want to change the role to be an enum instead of a string.

### Additional context

@flybayer also wrote:

> We are going to remove the `provider = ["sqlite", "postgres"]` thing because it's causing too many problems.

(I couldn’t find an issue covering this point.)

yhoiseth:
I don’t think I’ll get around to fixing this any time soon. So I removed myself as an assignee.

flybayer:
Cool no problem! That's perfect @yhoiseth

Dohxis:
I would love to work on this 🙂

PR #1268, opened 2020-10-06
Add a prompt to select database for new project

Dohxis:
Closes: #1037 

Added two new prompts to `blitz new` command: one for asking which database provider to use and another to specify the database URL. The consequences of specifying the database URL is that now `.env.local` and `.env.test.local` are the same. I am not sure if we want to leave it like this.

### Checklist

- [ ] Tests added for changes


packages/generator/src/generators/app-generator.ts
```diff
@@ -35,6 +37,8 @@ export class AppGenerator extends Generator<AppGeneratorOptions> {
     return {
       name: this.options.appName,
       username: await username(),
+      database: this.options.database === "SQLite" ? "sqlite" : "postgresql",
```
flybayer, line 40:
```suggestion
      database: this.options.database.toLowerCase(),
```

packages/cli/src/commands/new.ts
```diff
@@ -66,7 +67,29 @@ export class New extends Command {
...
-      const promptResult: any = await this.enquirer.prompt({
+      const databaseChoices: Array<{name: AppGeneratorOptions["database"]; message?: string}> = [
+        {name: "SQLite", message: "SQLite (default)"},
+        {name: "PostgreSQL"},
+      ]
+
+      const databasePromptResult: any = await this.enquirer.prompt({
+        type: "select",
+        name: "database",
+        message: "Pick a database engine (you can switch to something else later if you want)",
```
flybayer, line 78:
```suggestion
        message: "Pick a database engine (you can change later)",
```

packages/cli/src/commands/new.ts
```diff
@@ -66,7 +67,29 @@ export class New extends Command {
...
+
+      const databaseUrlPromptResult: any = await this.enquirer.prompt({
+        type: "input",
+        name: "databaseUrl",
+        message: `Specify url for ${databasePromptResult.database} (you can change to something else later if you want)`,
+        initial:
+          databasePromptResult.database === "SQLite"
+            ? "file:./db.sqlite"
+            : `postgresql://${await username()}@localhost:5432/${appName}`,
+      })
```
flybayer, line 90: Let's skip this prompt, and instead just set the URL automatically like we have been. The reason is the vast majority of folks won't want to change this.

packages/generator/templates/app/.env.test.local
```diff
@@ -1,7 +1,3 @@
 # THIS FILE SHOULD NOT BE CHECKED INTO YOUR VERSION CONTROL SYSTEM
 
-# SQLite is ready to go out of the box, but you can switch to Postgres
-# by first changing the provider from "sqlite" to "postgres" in the Prisma
-# schema file and by second swapping the DATABASE_URL below.
-DATABASE_URL="file:./db_test.sqlite"
-# DATABASE_URL=postgresql://__username__@localhost:5432/__name___test
+DATABASE_URL="__databaseUrl__"
```
flybayer, line 3: We definitely need to fix this to have the test db name with `_test` suffix.

flybayer (changes requested):
@Dohxis thank you!! Sorry for the review delay. Overall this looks great, just a few things to address.

packages/cli/src/commands/new.ts
Dohxis: The reason I added this is when you select `PostgreSQL` the default URL is almost always wrong, for example on my dev setup I always have `postgres@localhost` this is the default user used by PostgreSQL. I think allowing to change this is quite a nice touch and if you want to leave the default one you can just press <Enter> as it suggests you the right one.

packages/generator/templates/app/.env.test.local
Dohxis: I was also thinking the same but left it like this per your comment on Slack. Going to change it 🙂

flybayer: Ah oops, I totally misunderstood your question there 😅

packages/cli/src/commands/new.ts
flybayer: I just realized yesterday that the username thing is unnecessary and yes the default username should be `postgres`, so let's make that change. And then do you think it's fine to leave out this prompt?

Dohxis: Yes I think it would be enough, for now, I will update the code 🙂

packages/cli/src/commands/new.ts
```diff
@@ -66,7 +66,19 @@ export class New extends Command {
...
+        type: "select",
+        name: "database",
+        message: "Pick a database engine (you can change later)",
+        choices: databaseChoices,
+      })
+
+      const formPromptResult: any = await this.enquirer.prompt({
         type: "select",
         name: "form",
         message: "Pick a form library (you can switch to something else later if you want)",
```
flybayer, line 84:
```suggestion
        message: "Pick a form library",
```

packages/generator/src/generators/app-generator.ts
flybayer: Is this change ok to make?

Dohxis: Yeah, I think that's a good change.

flybayer:
Awesome! In final testing I realized there's one more thing.

It's super nice for people who first choose sqlite to have the postgres connection string in the env file but commented out. Can we add that back? (to both `.env.local` and `.env.local.test`)

I think you should be able to do something like this in the template:
```
if (process.env.database === 'sqlite') {
DATABASE_URL=__databaseUrl__
# DATABASE_URL=postgresql://postgres@localhost:5432/__name__
} else {
DATABASE_URL=__databaseUrl__
}
```

Also, leave out the `"`s from around the url.

aem:
I think this looks good as is (once @flybayer confirms that his changes have been addressed), but I noticed that the prompt code doesn't really flow nicely in the generator and is pretty repetitive. I made https://github.com/blitz-js/blitz/issues/1298 to track a nicer way to prompt the user during the generation process

packages/generator/templates/app/.env.local
```diff
@@ -1,8 +1,9 @@
...
-# by first changing the provider from "sqlite" to "postgres" in the Prisma
-# schema file and by second swapping the DATABASE_URL below.
-DATABASE_URL="file:./db.sqlite"
-# DATABASE_URL=postgresql://__username__@localhost:5432/__name__
+if (process.env.database === 'sqlite') {
+DATABASE_URL=__databaseUrl__
+# DATABASE_URL=postgresql://postgres@localhost:5432/__name__
+} else {
+DATABASE_URL=__databaseUrl__
+}
```
flybayer, line 9: @aem this if/else template syntax isn't working here. Do you know why?

aem: we can only use the if/else syntax in `.(t|j)sx?` files right now because it requires parsing the AST. I can try to rig something up with basic string matching for non-JS/TS files but it'll take a bit

flybayer: Could this be parsed as a shell script? 🤔

Flavyoo:
@Dohxis How are things going with this PR?

kevinlangleyjr:
Is there a reason for not adding MySQL as a database option? 

It's supported by Prisma, just like SQLite and PostgreSQL.

flybayer:
> Is there a reason for not adding MySQL as a database option?
> 
> It's supported by Prisma, just like SQLite and PostgreSQL.

We can definitely add mysql too

flybayer:
Fyi, the blocker on this PR is making our template parsing work for conditional statements in `.env` files

Issue #950, opened 2020-08-27
Blitz recipes don't work with layout.


oiojin831:
Using chakra-ui recipes turns only  `<Component {...pageProps} />` in to Chakra aware Code.
```ts
{getLayout(
  <ThemeProvider>
    <CSSReset />
    <Component {...pageProps}
  </ThemeProvider>
)}
```

It Chakra's `<ThemeProvider>` should wrap layout as well
```ts
<ThemeProvider>
  <CSSReset />
  {getLayout(
    <Component {...pageProps}
  )}
</ThemeProvider>
````

LoriKarikari:
Thanks for reporting this issue @oiojin831.

flybayer:
This is also a problem for all the other recipes that add context, like emotion etc.

konradkalemba:
@flybayer I can work on this.

PR #1279, opened 2020-10-08
Fix all recipes to add their context providers above getLayout()

konradkalemba:
Closes: #950 

This PR makes recipes wrap not only `Component` but also `getLayout` with context providers.

### Checklist

- [ ] Tests added for changes


flybayer:
Sweet, thank you!!!

PR #1258, opened 2020-10-06
Adds tests for useSession in core super tokens

clgeoio:
### What are the changes and their implications?
Adds some testing to help close out: #757 

exports a type for `useSession` instead of casting with `as`

### Checklist

- [x] Tests added for changes
- [ ] PR submitted to [blitzjs.com](https://github.com/blitz-js/blitzjs.com) for any user facing changes

PR #1283, opened 2020-10-08
Add initialPublicData param to useSession hook to allow server-fetched sessions

mattfwood:

Currently, if someone is using `getSessionContext` in `getServerSideProps`, the session still has to be re-fetched from `PublicDataStore` when using the `useSession` hook.

When working on my own app, i wanted to avoid the flickering of a login screen while the session was being fetched:
```ts
export const getServerSideProps: GetServerSideProps = async ({
  req,
  res,
}) => {
  const session = await getSessionContext(req, res);
  const publicData = session?.publicData;

  return { props: { publicData } }
}

// in render
const Home: BlitzPage = () => {
  const session = useSession(); // => useSession is undefined while loading, despite being fetched server-side
}
```

This change would allow public to be optionally passed to the session and avoid any brief flickering / loading states while the publicData is being fetched from `PublicDataStore`.

```ts
const Home: BlitzPage = ({ publicData }) => {
  const session = useSession(publicData); // => useSession is undefined while loading, despite being fetched server-side
}
```

### Checklist

- [ ] Tests added for changes

I plan to add tests to this dependent on https://github.com/blitz-js/blitz/pull/1258 being merged.

### DIFF
--- a/packages/core/src/supertokens.ts
+++ b/packages/core/src/supertokens.ts
@@ -100,8 +100,8 @@ export const parsePublicDataToken = (token: string) => {
   }
 }
 
-export const useSession = () => {
-  const [publicData, setPublicData] = useState(publicDataStore.emptyPublicData)
+export const useSession = (initialPublicData?: PublicData) => {
+  const [publicData, setPublicData] = useState(initialPublicData ?? publicDataStore.emptyPublicData)
   const [isLoading, setIsLoading] = useState(true)
 
   useIsomorphicLayoutEffect(() => {
--- a/packages/core/src/supertokens.ts
+++ b/packages/core/src/supertokens.ts
@@ -102,12 +102,15 @@ export const parsePublicDataToken = (token: string) => {
 
 export const useSession = (initialPublicData?: PublicData) => {
   const [publicData, setPublicData] = useState(initialPublicData ?? publicDataStore.emptyPublicData)
-  const [isLoading, setIsLoading] = useState(true)
+  // skip loading if initial public data is provided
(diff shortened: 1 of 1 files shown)


packages/core/src/supertokens.ts
```diff
@@ -100,8 +100,8 @@ export const parsePublicDataToken = (token: string) => {
   }
 }
 
-export const useSession = () => {
-  const [publicData, setPublicData] = useState(publicDataStore.emptyPublicData)
+export const useSession = (initialPublicData?: PublicData) => {
+  const [publicData, setPublicData] = useState(initialPublicData ?? publicDataStore.emptyPublicData)
   const [isLoading, setIsLoading] = useState(true)
```
engelkes-finstreet, line 105: I think if we pass the publicData as props we should set the isLoadingState to false and not run the `useIsomorphicLayoutEffect` at all.

mattfwood: Good call, just made these changes

flybayer:
Sweet, thanks @mattfwood!

1. Let's change the useSession() prop to be be an object so that we can add other things in the future.

```tsx
const Home: BlitzPage = ({ publicData }) => {
  const session = useSession({initialPublicData: publicData}); 
}
```

2. I think we should still run the effect, even if initial data is provided, because the useSession hook should contain exactly what the public data is on the client, even if that's different that what was initially provided by the server. But definitely keep isLoading as false, like you have now.

3. And then also open a PR to the docs repo to document this feature.

flybayer:
@mattfwood are you able to wrap this up?

Issue #1003, opened 2020-09-03
EACCES: permission denied even with sudo npm install


sw-yx:
### What is the problem?

i am failing to install blitz even with `sudo` enabled.

```bash
sudo npm install -g blitz
Password:
npm WARN deprecated resolve-url@0.2.1: https://github.com/lydell/resolve-url#deprecated
npm WARN deprecated urix@0.1.0: Please see https://github.com/lydell/urix#deprecated
npm WARN deprecated chokidar@2.1.8: Chokidar 2 will break on node v14+. Upgrade to chokidar 3 with 15x less dependencies.
npm WARN deprecated mkdirp@0.5.3: Legacy versions of mkdirp are no longer supported. Please update to mkdirp 1.x. (Note that the API surface has changed to use Promises in 1.x.)
npm WARN deprecated fsevents@1.2.13: fsevents 1 will break on node v14+ and could be using insecure binaries. Upgrade to fsevents 2.
/usr/local/bin/blitz -> /usr/local/lib/node_modules/blitz/bin/blitz

> fsevents@1.2.13 install /usr/local/lib/node_modules/blitz/node_modules/next/node_modules/fsevents
> node install.js

gyp WARN EACCES current user ("nobody") does not have permission to access the dev dir "/Users/wanshawn/Library/Caches/node-gyp/12.18.0"
gyp WARN EACCES attempting to reinstall using temporary dev dir "/usr/local/lib/node_modules/blitz/node_modules/next/node_modules/fsevents/.node-gyp"
gyp WARN install got an error, rolling back install
gyp WARN install got an error, rolling back install
gyp ERR! configure error 
gyp ERR! stack Error: EACCES: permission denied, mkdir '/usr/local/lib/node_modules/blitz/node_modules/next/node_modules/fsevents/.node-gyp'
gyp ERR! System Darwin 18.7.0
gyp ERR! command "/usr/local/bin/node" "/usr/local/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "rebuild"
gyp ERR! cwd /usr/local/lib/node_modules/blitz/node_modules/next/node_modules/fsevents
gyp ERR! node -v v12.18.0
gyp ERR! node-gyp -v v5.1.0
gyp ERR! not ok 

> fsevents@1.2.13 install /usr/local/lib/node_modules/blitz/node_modules/watchpack-chokidar2/node_modules/fsevents
> node install.js

[...]

PR #1285, opened 2020-10-09
(newapp) Add yarn/npx to global install instructions in readme

cajotafer:
Helps with: #1003 

There is only a small change in the main `README.md` file. I was trying to install and run blitzjs for very first time and I had this same issue #1003 . Even I don't know how to fix real the problem, I found the answers in that issue helpful to be able to continue so I think is a good idea to let the other people know that they also have `yarn` and `npx` as options.

### Checklist

- [ ] Tests added for changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,9 @@ You need Node.js 12 or newer
 
 #### Install Blitz
 
-Run `npm install -g blitz`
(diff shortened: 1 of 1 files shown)


flybayer:
Thanks @cajotafer!

@all-contributors add @cajotafer for docs

Issue #984, opened 2020-09-01
Add `setQueryData` utility for updating query cache


jan-wilhelm:
### What do you want and why?
Some components don't need the actual data of a `useQuery` hook but just need access to the `mutate` and `refetch` functions. Having to wrap those in a `<Suspense  />` component just for that purpose is unintuitive and confusing.

Especially for optimistic UIs it is not very handy that I have to call useQuery  just to be able to update the value of that query.

Example: Think of a `ChatMessageInput` component that renders a textarea for a chat application. There’s a `getChat` query. Now when sending a message through the textarea, I want the ChatMessageInput component to be able to optimistically update the value of that query.

Currently, it seems like I need to do this:

```typescript
const [chat, { mutate, refetch }] = useQuery(getChat, {chatId});
const [message, setMessage] = useState("");

const submit = async () => {
    mutate(
      {/*...*/},
      { refetch: false }
    );
    await sendMessage({ chatId, message });
  }
}
```

Instead, being able to use something like
```typescript
import { queryCache } from 'react-query' 
queryCache.invalidateQueries(/*...*/)
```
without actually having to load the hook would be preferable here.

This is especially true if you take into account that you would want such a `ChatMessageInput` to render the textarea even while the chat query is loading. Therefore you would have to either pass the same UI component as a fallback component to a wrapping `<Suspense>` or disable suspense mode for this query.

### Possible implementation(s)
Not sure, not too familiar with how Blitz internally uses `react-query`

flybayer:
Hey @jan-wilhelm that definitely makes sense. I need to investigate this and see if we can do something to make this nice, but for sure right now you can access `react-query` directly and manually modify the `queryCache`.

flybayer:
Here's [our internal usage of useQuery](https://github.com/blitz-js/blitz/blob/canary/packages/core/src/use-query.ts#L63).

And [here's where we define the `mutate` function](https://github.com/blitz-js/blitz/blob/canary/packages/core/src/utils/query-cache.ts#L12). As I'm looking at this right now, I wonder if we don't have a bug because the queryKeys don't match between those 🤔

flybayer:
@jan-wilhelm ok what if we added a `getQueryKey` function that you could use anywhere in your app like this:

```ts
import getChat from 'app/queries/getChat'
import {getQueryKey} from 'blitz'
import {queryCache} from 'react-query'

// ...

const queryKey = getQueryKey(getChat, {/* exact same params I use with useQuery */})
[...]

PR #1291, opened 2020-10-10
Add setQueryData to react-query-utils AND rename mutate to setQueryData

aericson:
Closes: #984 

1. Add `setQueryData` to `react-query-utils`.
     - Receives (resolver, params, newData, opts) as argument.
2. Some refactoring on `react-query-utils`.
3. Changed arguments to `getQueryCacheFunctions` to keep it more consistent with the other functions from this module and to reduce code duplication. This doesn't seem to be a documented function so I figured it would be ok to change it.
4. Renamed and `mutate` returned by `useQuery` to `setQueryData` as sugested [here](https://github.com/blitz-js/blitz/issues/984#issuecomment-706345381).

### Checklist
- [x] Tests added for changes
    - https://github.com/blitz-js/blitzjs.com/pull/227


flybayer:
Sweet, looks good to me!!

packages/core/src/utils/react-query-utils.ts
```diff
@@ -114,6 +103,34 @@ export function invalidateQuery<TInput, TResult, T extends QueryFn>(
...
+
+  return new Promise((res) => {
+    queryCache.setQueryData(queryKey, newData)
+    let result: void | ReturnType<typeof queryCache.invalidateQueries>
+    if (opts.refetch) {
+      result = invalidateQuery(resolver, params)
+    }
+    if (isClient) {
+      // Fix for https://github.com/blitz-js/blitz/issues/1174
+      requestIdleCallback(() => {
```
jan-wilhelm, line 125: as shown in #1214 , requestIdleCallback is not supported in safari. Could you change this to use setTimeout maybe?

aericson: Thanks for taking a look at this!

I believe this was addressed by https://github.com/blitz-js/blitz/pull/1247 it's using the shim, not `window.requestIdleCallback`.

jan-wilhelm: yeah, exactly. That shim should work here too, I believe

aericson: Sorry, I'm a bit confused. It's already using that shim.

jan-wilhelm: oh, sorry. I didn't see the import and thought that this was implicitly using the function from "window". Thanks for clearing this up :)

aericson: No worries!

aericson:
Added some new tests, will try to put the time into some docs tomorrow or the day after.

aericson:
Docs added! From my side, this is ready to go. 😁

flybayer:
Sweeet, thanks!!

Issue #1286, opened 2020-10-09
Export `next/config` in `@blitz/core`


LoriKarikari:
### What do you want and why?

There currently isn't a Blitz equivalent to `next/config`. If you want to use it you have to do `import getConfig from 'next/config'`. This should become `import { getConfig } from 'blitz'`.

### Possible implementation(s)
This export needs to be added to this file: 
https://github.com/blitz-js/blitz/blob/canary/packages/core/src/index.ts

davidlutta:
Hello, open-source noob here 😅 with a question. 
Would the following implementation be satisfactory ? 
```js
export {getConfig as getConfig} from "next/config"
```

flybayer:
@davidlutta this should even work!

```ts
export {getConfig} from "next/config"
```

PR #1300, opened 2020-10-11
Export next/config in blitz/core

davidlutta:
Closes: #1286 

Exporting next/config in blitz/core


packages/core/src/index.ts
```diff
@@ -59,6 +59,8 @@ export {
 
 export {default as dynamic} from "next/dynamic"
 
+export {getConfig} from "next/config"
```
flybayer, line 62: I think this actually needs to be like this:

```suggestion
export {default as getConfig} from "next/config"
```

Sorry for leading you astray in slack!

davidlutta: Thanks a lot for the correction. Let me fix it

Issue #1303, opened 2020-10-12
blitz build does not return correct exit codes


markhaehnel:
### What is the problem?
Command `blitz build` doesn't return correct exit codes.
It returns `0` even if the build fails.

### Steps to Reproduce

1. Have a syntactical error in your code
2. Run `blitz build`

### Versions

```
macOS Big Sur | darwin-x64 | Node: v14.12.0

blitz: 0.24.1 (global)
blitz: 0.24.1 (local)

  Package manager: npm 
  System:
    OS: macOS 11.0
    CPU: (4) x64 Intel(R) Core(TM) i3-8100B CPU @ 3.60GHz
    Memory: 10.66 GB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.12.0 - /usr/local/bin/node
    Yarn: Not Found
    npm: 6.14.8 - /usr/local/bin/npm
    Watchman: Not Found
  npmPackages:
    @prisma/cli: 2.8.1 => 2.8.1 
    @prisma/client: 2.8.1 => 2.8.1 
    blitz: 0.24.1 => 0.24.1 
    react: 0.0.0-experimental-7f28234f8 => 0.0.0-experimental-7f28234f8 
    react-dom: 0.0.0-experimental-7f28234f8 => 0.0.0-experimental-7f28234f8 
    typescript: 4.0.3 => 4.0.3 
```

### Other
Return code does not equal the `next build` return code
```
Failed to compile.

./pages/api/runtime-estimate.ts:24:65
Type error: Cannot find name 'a'.

  22 |     try {
  23 |       const data = await ssrGetRuntimeEstimate(params)
> 24 |       const parsedData = await RuntimeEstimateSchema.parseAsync(a)
     |                                                                 ^
  25 |       res.status(200).json(parsedData)
  26 |     } catch (ex: any) {
  27 |       res.status(500).json({ error: "Internal server error" })
info  - Creating an optimized production build .'next build' failed with status code: 1

# mhaehnel @ Mark-MacMini:
$ echo $?    
0
```

jamiedavenport:
Thanks for raising this 😃. I believe this issue is [here](https://github.com/blitz-js/blitz/blob/canary/packages/blitz/bin/blitz#L13) where the result of `spawn.sync` is being ignored.

markhaehnel:
Thanks for the pointer. 
I will take a look at it and fix this in a PR.

markhaehnel:
[...]

PR #1316, opened 2020-10-13
Fix blitz CLI not using proper exit codes

markhaehnel:
Closes: #1303

blitz/cli now returns exit codes of the spawned child process.
This allows for better use of blitz command in CI/CD environments.

### Checklist

- [X] Tests added for changes


jamiedavenport:
Thanks for opening this. The fix looks good. Could you add some tests for this so that we can avoid regression in future?

markhaehnel:
> Could you add some tests so that we can avoid regression in future?
Sure, i will try. 🙂

markhaehnel:
I've added a basic test that checks the return type but encountered a problem. I've asked on slack for input on some ideas. (https://blitzjs.slack.com/archives/CUM2C68T1/p1602605269128000)

packages/blitz/test/bin.test.ts
```diff
@@ -0,0 +1,15 @@
...
+  it("should return exit code 0 if command is successfull", () => {
+    const result = spawn.sync("blitz", ["--version"])
+    expect(result.status).toBe(0)
+  })
+  it.skip("should return exit code 1 if command is unsuccessfull", () => {
+    // TODO: Commands not available dont return exit code 1.
+    // Need do setup a environment to get it failing.
+
+    const result = spawn.sync("blitz", ["notAvailableCommand"])
+    expect(result.status).toBe(1)
```
jamiedavenport, line 13: I think something like the change below might work 
```suggestion
    const result = spawn.sync("./bin/blitz", ["build"], {stdio: "inherit"})
    expect(result.status).toBe(1)
```

packages/blitz/test/bin.test.ts
```diff
@@ -0,0 +1,15 @@
+import spawn from "cross-spawn"
+
+describe("Binary blitz", () => {
+  it("should return exit code 0 if command is successfull", () => {
+    const result = spawn.sync("blitz", ["--version"])
```
jamiedavenport, line 5: I think this also needs to change to `./bin/blitz`. I think just spawning `blitz` might run the globally installed CLI on your machine.

```suggestion
    const result = spawn.sync("./bin/blitz", ["--version"], {stdio: "inherit"})
```

markhaehnel:
Now the test should work, but it doesnt run in CI because the lerna scopa "blitz" doesnt get tested

flybayer:
@markhaehnel thanks! I fixed the blitz pkg tests not running. Looks like your new test is failing with exit code of `2` instead of `1`

markhaehnel:
My tests are running now, but can't explain to me why the example builds fail. 🤷‍♂️
Looks like a timeout.

markhaehnel:
Should be good now. Have a nice weekend!

flybayer:
Sweet, thank you!!

@all-contributors add @markhaehnel for code

Issue #1297, opened 2020-10-11
Mutations incorrectly warns about deprecation even when using invoke()


aericson:
### What is the problem?
Even when using `invoke()` with mutations one will still get the warning `Directly calling queries/mutations is deprecated in favor of invoke(queryFn, params)`. I believe that the cause is that `fromInvoke` is being passed down to the `mutate` function that was returned by react-query which doesn't pass that on to the resolver.

I investigated it and tried to fix it but I got stuck. I can take this issue myself but would need some pointers.

### Steps to Reproduce

Add a mutation call using invoke such as:
```
             const [createProductMutation] = useMutation(createProduct)
-            const product = await createProductMutation({data})
+            const product = await invoke(createProductMutation, {data})
```

You will still see the warning.
### Versions
Using canary branch and examples/store app.
```
You are using alpha software - if you have any problems, please open an issue here:
      https://github.com/blitz-js/blitz/issues/new/choose

debug: blitzPkgPath: /Users/aericson/projects/blitz/packages/blitz/dist/index.js
debug: cliPkgPath: /Users/aericson/projects/blitz/packages/cli/lib/src/index.js 

macOS Big Sur | darwin-x64 | Node: v14.12.0

blitz: 0.24.1 (global)
blitz: 0.24.1 (local)

  Package manager: yarn 
  System:
    OS: macOS 11.0
    CPU: (4) x64 Intel(R) Core(TM) i7-6660U CPU @ 2.40GHz
    Memory: 85.56 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.12.0 - /usr/local/bin/node
    Yarn: 1.22.5 - /usr/local/bin/yarn
    npm: 6.14.8 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  npmPackages:
    @prisma/cli: Not Found
    @prisma/client: Not Found
    blitz: Not Found
    react: Not Found
    react-dom: Not Found
    typescript: 4.0.3 => 4.0.3 
```

AxDSan:
This is weird, I'm on this and I'm not getting the error anymore (I used to get it before whereas I was doing a direct mutation/query call, but I switched to `invoke()` and it's not happening anymore.)

![](https://i.imgur.com/bgB2fcP.jpg)

flybayer:
@aericson in you example, invoke is not needed. This is correct:

```ts
[...]

PR #1320, opened 2020-10-13
Remove incorrect deprecation warning when using useMutation

aericson:
Closes: #1297 

See #1297. There is a warning, of deprecation, telling to use `invoke()` when one is using `useMutation`.
I've also added a simple test for the useMutation since there wasn't any.

### Checklist

- [x] Tests added for changes


packages/core/src/use-mutation.ts
```diff
@@ -15,10 +15,13 @@ export function useMutation<TResult, TError = unknown, TVariables = undefined, T
...
 ) {
-  validateQueryFn(mutationResolver)
+  const enhancedResolverRpcClient = sanitize(mutationResolver)
 
-  return useReactQueryMutation(mutationResolver, {
-    throwOnError: true,
-    ...config,
-  }) as MutationResultPair<TResult, TError, TVariables, TSnapshot>
+  return useReactQueryMutation(
+    (variables?: TVariables) => enhancedResolverRpcClient(variables, {fromQueryHook: true}),
```
flybayer, line 21: Unless I'm mistaking, the previous type for this was `MutationFunction` which is typed `(variables: TVariables)` and so the new code should be like this to match.
```suggestion
    (variables: TVariables) => enhancedResolverRpcClient(variables, {fromQueryHook: true}),
```

flybayer (commented):
Sweet, looks great! Just one comment.

aericson:
> Sweet, looks great! Just one comment.

good eye! thanks, changed it.

flybayer (approved):
Awesome!

PR #1323, opened 2020-10-14
(newapp) Fix tests not working out of box

peter50216:
Currently `blitz new test; cd test; blitz db migrate; blitz test` would fail with
```
TypeError: Cannot read property 'createContext' of undefined
````

This is caused by react-query has `require("react")` inside a folder named react, and adding `"."` to moduleDirectories seems to make jest add current directory to module path for ALL dependencies.

Fix this by using `"<rootDir>"` instead of `"."`.

### DIFF
--- a/packages/generator/templates/app/jest.config.js
+++ b/packages/generator/templates/app/jest.config.js
@@ -15,7 +15,7 @@ module.exports = {
     "^.+\\.(ts|tsx)$": "babel-jest",
   },
   // This makes absolute imports work
-  moduleDirectories: ["node_modules", "."],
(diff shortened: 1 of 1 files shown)


flybayer (approved):
Awesome, thanks @peter50216!!

flybayer:
Awesome, thanks @peter50216!!

@all-contributors add @peter50216 for code

iamtekeste:
I am not sure how you figured this out but this was very helpful and thank you for sharing your solution @peter50216 <3

iamtekeste:
I am not sure how you figured this out but this was very helpful and thank you for sharing your solution @peter50216 <3

Issue #1305, opened 2020-10-12
Change Recipes to add providers outside of ErrorBoundary


flybayer:
### What do you want and why?

Currently our recipes add providers inside the ErrorBoundary like this.

```ts
export default function App({ Component, pageProps }: AppProps) {
  const getLayout = Component.getLayout || ((page) => page)

  return (
    <ErrorBoundary
      FallbackComponent={RootErrorFallback}
    >
      <ChakraProvider resetCSS>
        {getLayout(<Component {...pageProps} />)}
      </ChakraProvider>
    </ErrorBoundary>
  )
}
```

But I think we should change them to instead add providers at the root of the return statement like this.

```ts
export default function App({ Component, pageProps }: AppProps) {
  const getLayout = Component.getLayout || ((page) => page)

  return (
    <ChakraProvider resetCSS>
      <ErrorBoundary
        FallbackComponent={RootErrorFallback}
      >
        {getLayout(<Component {...pageProps} />)}
      </ErrorBoundary>
    </ChakraProvider>
  )
}
```

The reason is that often you want to use the providers inside the error boundary fallback component (like using Chakra components there, for example).

A nice side benefit of this is that our logic for adding providers becomes less brittle.

konradkalemba:
I can do this.

PR #1324, opened 2020-10-14
Fix recipes to add their providers at the root of App's return function

konradkalemba:
Closes: #1305 

This PR makes recipes to add their providers at the root of JSX element returned by custom App component.

In my approach, firstly, all JSX elements are found and then those that are first level children of return statement of App function declaration are chosen to get wrapped with providers. This approach won't work as desired if someone changes the name of App component however I think it's rarely the case.

Another approach could be finding `<Component />` JSX element and then recursively checking if its parent is at the root of the return statement -- if so wrap it with providers. So if you prefer it over the original one, let me know. 

### Checklist

- [ ] Tests added for changes


flybayer (approved):
Sweet, thanks!!

Issue #913, opened 2020-08-18
Add support for `blitz.config.ts`


flybayer:
### What do you want and why?

I want to transpile our config file so that we can use Typescript with `blitz.config.ts`. One big reason for this is currently our built in session management is configured in this file. And for people not using prisma, they need to define their database access hooks to get/set the `session` in their DB. The straightforward way to do this would be import `db` from `./db/index.ts`. But currently that doesn't work since `blitz.config.js` isn't transpiled.

### Possible implementation(s)

I think this might be as simple as [adding config to with-blitz.ts](https://github.com/blitz-js/blitz/blob/canary/packages/server/src/with-blitz.ts#L18). Something like this:

```
"../../_blitz.config.js": "./blitz.config",
```

And then [change our `getConfig` function](https://github.com/blitz-js/blitz/blob/canary/packages/config/src/index.ts) to use `_blitz.config.js`.

And also [change this line](https://github.com/blitz-js/blitz/blob/canary/packages/server/src/stages/rpc/index.ts#L85) to use `_blitz.config.js`

jamiedavenport:
Could I work on this? 🥺

jamiedavenport:
> I think this might be as simple as adding config to with-blitz.ts. Something like this:

If only things were that simple 😢 

Looks to me as though the `blitz.config.ts` file would already need to be transpiled before Webpack runs, since the generated `next.config.js` file depends on it and Next doesn't support `next.config.ts` (https://github.com/vercel/next.js/issues/5318)

https://github.com/blitz-js/blitz/blob/canary/packages/server/src/stages/config/index.ts#L45

So my current plan of action is to add a transpilation step using either Typescript or Babel into the config stage. Their main reason for not supporting `next.config.ts` is performance, is that something that we need to be worried about @flybayer @ryardley

flybayer:
@jamiedavenport the perf argument is pretty lame since it's one single file compared to the actual app build that's many files.

Yeah, adding this to the server config stage sounds perfect.

PR #1325, opened 2020-10-14
Add support for `blitz.config.ts`

jamiedavenport:
Closes: #913 

### Checklist

- [x] Tests added for changes


packages/installer/src/executors/executor.tsx
```diff
@@ -38,9 +38,15 @@ export function Frontmatter({executor}: {executor: ExecutorConfig}) {
...
-        <Text color="#8a3df0" bold>{verticalBorder}</Text>
+        <Text color="#8a3df0" bold>
+          {verticalBorder}
+        </Text>
+        <Text color="#8a3df0" bold>
+          ⎪&nbsp;&nbsp;&nbsp;{executor.stepName}&nbsp;&nbsp;&nbsp;⎪
+        </Text>
+        <Text color="#8a3df0" bold>
+          {verticalBorder}
+        </Text>
```
jamiedavenport, line 49: These changes are due to linting when I've merged `canary` into my branch

packages/installer/src/executors/file-transform-executor.tsx
```diff
@@ -65,7 +65,9 @@ export const Propose: Executor["Propose"] = ({cliArgs, onProposalAccepted, step}
           }
 
           return (
-            <Text {...styleProps} key={idx}>{line}</Text>
+            <Text {...styleProps} key={idx}>
+              {line}
+            </Text>
```
jamiedavenport, line 70: Same here, linting

jamiedavenport:
A couple of things to note:
- This solution copies the config over into the root as `_blitz.config.js` because I was experiencing this problem (https://github.com/blitz-js/blitz/pull/726) but I'm sure there must be a more elegant way.
- Wrt testing, I'm not sure what a good unit test would look like for this yet. So I've relied on the cypress tests in the `auth` example. Do we think that's sufficient enough?

packages/server/package.json
```diff
@@ -29,6 +29,7 @@
   "module": "dist/server.esm.js",
   "types": "dist/packages/server/src/index.d.ts",
   "dependencies": {
+    "@babel/core": "7.11.6",
```
flybayer, line 32: Let's use a `^` range here so that we use the same version Next.js. Or, even we could maybe remove it from here since next includes it.

ryardley:
I would recommend having a look at the pages test https://github.com/blitz-js/blitz/blob/5572c9c3239b7beb001cec1914e4b59fe5ee10c2/packages/server/src/stages/pages/pages.test.ts and writing one for the config stage.

ryardley:
> This solution copies the config over into the root as _blitz.config.js because I was experiencing this problem (#726) but I'm sure there must be a more elegant way.

Not sure I understand what is going on here? Sounds to me like we probably shouldn't have to do this.

flybayer (commented):
Sweet!!

First thing: we must also be able to import other `.ts` files into `blitz.config.ts` and have it work correctly. I don't think this will work right now with this pr, right? Because for this to work, we basically need to run webpack on the config file to bundle it, right? I should have thought about and mentioned this in slack when we were talking about just using babel.

As for where to put the built file... My first thought is to put it at `.next/__blitz.config.js` because that's where we are bundling the `db/index.ts` file to as `.next/__db.js`. We can ensure that the .js blitz config always lives there, and then we can change `@blitzjs/config` to read from there. And then also the [config stage](https://github.com/blitz-js/blitz/blob/canary/packages/server/src/stages/config/index.ts) would need changed to read from there too.

flybayer:
[ncc](https://github.com/vercel/ncc) would be the easy way to bundle `blitz.config.ts`

packages/server/src/stages/config/index.ts
```diff
@@ -43,14 +57,26 @@ export const createStageConfig: Stage = ({config, processNewFile, processNewChil
...
     )
   }
 
   // No need to filter yet
-  const stream = transform.file((file) => {
+  const stream = transform.file(async (file) => {
+    if (isBlitzTsConfigPath(file.path)) {
+      const res = await bundle(file.path, {
+        quiet: true,
+        externals: ["@blitzjs/server", "@next/bundle-analyzer"], // FIXME: Figure out why this wouldn't work? Do we need to bundle everything?
```
jamiedavenport, line 71: Not sure what to do here. `ncc` failed to bundle these 3rd party modules, although it shouldn't have to because they're already built. `ncc` doesn't seem to provide a nice way to do this (https://github.com/vercel/ncc/issues/409) so I'm considering:
* Maintain this list manually, thinking that there shouldn't be a lot of 3rd party modules that the config will be expected to import (mostly just auth stuff?)
* Programmatically get all installed modules with `yarn list --json` or similar

What are your thoughts? @flybayer

ryardley: Maybe we should do this first - outside the file-pipeline?

flybayer:
@jamiedavenport sorry I haven't had time to dig into this yet. Had some other life stuff interfere this week and some high priority bugs I'm working through. I'll get back to this as soon as I can.

jamiedavenport:
@ryardley @flybayer would either of you guys be able to review this soon? 😃

flybayer:
@jamiedavenport yeah, so sorry. I've been focused 100% on fixing a bunch of bugs for official beta status, and new features are a bit lower priority than that.

This week I'm very busy and have vacation, so hopefully I can give this full attention in first part of December.

flybayer:
I need to really think through all potential edge cases and also test in multiple production deployments (Vercel and Render).

flybayer:
NOTE: also need to update recipes to work with `blitz.config.ts`

flybayer:
@jamiedavenport I'm really sorry I let this drop. I think subconsciously I knew ncc wasn't the right approach here (but at the time I didn't understand some of these tools well enough to know what else to do).

I've now implemented this using esbuild in a much cleaner way: https://github.com/blitz-js/blitz/pull/2283

I really, really appreciate the effort you put into this. I know it's a bummer to put a lot of effort into something that doesn't even get used. So for that I'm sorry 🙏

Issue #1333, opened 2020-10-15
Can't install any recipe


alexturpin:
### Steps to Reproduce

1. Run `blitz install tailwind`
2. Go through steps until it says "Dependencies installed! Press ENTER to continue". Press ENTER.
3. This results in `ERROR  Text string "" must be rendered inside <Text> component`

### Versions

```
You are using alpha software - if you have any problems, please open an issue here:
      https://github.com/blitz-js/blitz/issues/new/choose

debug: blitzPkgPath: C:\Users\alext\src\cocktails\node_modules\blitz\dist\index.js
debug: cliPkgPath: C:\Users\alext\src\cocktails\node_modules\blitz\node_modules\@blitzjs\cli\lib\src\index.js 

Windows 10 | win32-x64 | Node: v12.16.1

blitz: 0.24.1 (global)
blitz: 0.24.1 (local)

  Package manager: npm
  System:
    OS: Windows 10 10.0.19041
    CPU: (12) x64 AMD Ryzen 5 2600 Six-Core Processor
    Memory: 8.86 GB / 15.93 GB
  Binaries:
    Node: 12.16.1 - C:\Program Files\nodejs\node.EXE
    Yarn: Not Found
    npm: 6.13.4 - C:\Program Files\nodejs\npm.CMD
    Watchman: Not Found
  npmPackages:
    @prisma/cli: 2.8.1 => 2.8.1
    @prisma/client: 2.8.1 => 2.8.1
    blitz: 0.24.1 => 0.24.1
    react: 0.0.0-experimental-7f28234f8 => 0.0.0-experimental-7f28234f8
    react-dom: 0.0.0-experimental-7f28234f8 => 0.0.0-experimental-7f28234f8
    typescript: 4.0.3 => 4.0.3
```

### Other

Full stack trace

```
$ blitz install tailwind
You are using alpha software - if you have any problems, please open an issue here:
      https://github.com/blitz-js/blitz/issues/new/choose


  ERROR  Text string "" must be rendered inside <Text> component

 C:/Users/alext/AppData/Local/Temp/blitz-recipe-blitz-js-blitz/node_modules/ink/src/reconciler.ts:122:10

 - createTextInstance (C:/Users/alext/AppData/Local/Temp/blitz-recipe-blitz-js-blitz/node_modules/ink/src/reconciler.ts:122:10)
[...]

PR #1339, opened 2020-10-15
Fix can't install any recipe

nahue:
Closes: #1333

* After reading through https://stackoverflow.com/a/59108109 i realized that we use conditional rendering (meaning that sometimes it renders <Text>undefined</Text>) so i fixed them.
* Also there was a couple strings without any <Text> wrapper.
* Also i fixed a <Spinner> that was outside of <Text>

### Checklist

- [ ] Tests added for changes


engelkes-finstreet (approved):
Awesome 🚀  Install works again :-)

Issue #1331, opened 2020-10-14
Error: db.connect is not a function, when using without Prisma


flybayer:
### What is the problem?

Can't use queries without prisma

### Steps to Reproduce

1. Remove prisma and session management middleware
2. Run a regular query that doesn't access a database
3. See the error shown below.

### Other

```
17:31:33.252.000	 ERROR 	Error while processing the request

 Error 	db.connect is not a function
error stack:
• rpc.ts:24 <anonymous>
    packages/server/src/rpc.ts:24:9

• middleware.ts:121 dispatch
    packages/core/src/middleware.ts:121:32

• supertokens.ts:127 <anonymous>
    packages/server/src/supertokens.ts:127:12

• middleware.ts:121 dispatch
    packages/core/src/middleware.ts:121:32

• middleware.ts:129 <anonymous>
    packages/core/src/middleware.ts:129:12

• middleware.ts:59 <anonymous>
    packages/core/src/middleware.ts:59:11

• helpers.js:565 _catch
    node_modules/babel-plugin-transform-async-to-promises/helpers.js:565:16

• core.cjs.development.js:746 handleRequestWithMiddleware
    packages/core/dist/core.cjs.development.js:746:28

code frame:
  019 |     const log = baseLogger.getChildLogger({prefix: [resolver._meta.name + "()"]})
  020 |
  021 |     if (req.method === "HEAD") {
  022 |       // Warm the lamda and connect to DB
  023 |       if (typeof connectDb === "function") {
> 024 |         connectDb()
                ^
  025 |       }
  026 |       res.status(200).end()
  027 |       return next()
  028 |     } else if (req.method === "POST") {
  029 |       // Handle RPC call
```

### How To Fix

Here's the line that needs changed to work without a connect function defined: https://github.com/blitz-js/blitz/blob/canary/packages/server/src/stages/rpc/index.ts#L45

nksaraf:
Hi, I am going to take a stab at this, seems simple enough! Excited to contribute to this project.

twhite96:
Fixed this by creating a new Blitz site. My version was so outdated I couldn't run anything.

flybayer:
v0.17 😱 happy halloween lol!

PR #1342, opened 2020-10-17
Fix db.connect error when using without prisma

nksaraf:
Closes: #1331 

Where the `db` and `connect` functions are imported in the generated code for the API file, there wasn't a check for if the `db` doesn't have a connect function (or is `undefined` itself). Added this so that an error is not thrown if somebody uses this without Prisma or another `db` setup

### Checklist

- [ ] Tests added for changes


flybayer:
Sweet, thank you!

@all-contributors add @nksaraf for code

PR #1351, opened 2020-10-18
Adds test to assert useMutation validation

peaonunes:
Closes: ??

Relates to the discussion I raised on: https://blitzjs.slack.com/archives/C011EQ0J75M/p1603015853243500

I was looking into the reasons why we cannot test components using `useMutation` and `useQuery`, then I looked at blitz core code I noticed that the validation of the `useMutation` query is not being tested. Therefore we can accidentally introduce a regression so I took the time to add the test case for that.

It increases the confidence while making changes on `useMutation` 🙌 

### Checklist
- [x] Tests added for changes


flybayer:
Thanks!

PR #1362, opened 2020-10-19
(newapp) Move `tsc` from pre-commit to pre-push hook

### LINKED ISSUE #1353: Move 'tsc' command to pre-push hook
### What is the problem?
I want to suggest to move the `tsc` command from the `pre-commit` hook to the `pre-push` hook in the Blitz boilerplate project. The reason for this change is that I find myself quite often in the middle of changes where I only want to commit a chunk of it (e.g. a style change). 
Because there is no official way of running `tsc` on specific files you are not able to only check for errors in your staged files. Therefore, when `tsc` is running on the `pre-commit` hook I am not able to commit my changes in chunks when there is a TS error in some other file that is not related at all (unless I use `--no-verify` of course).
If you move `tsc` to the `pre-push` hook you are able to split your commits in chunks while still not being able to push code that breaks your application.

### Steps to Reproduce
1. Make a change in some random file that will trigger a TS error. E.g.
```
// Type 'number' is not assignable to type 'string'
const foo: string = 1;
```
2. Make a change in some other random file that doesn't trigger a TS error.
3. Stage the file that has no TS error
4. When committing this file it will trigger a TS error and you can not commit the change.

machadolucasvp:
Closes #1353

### DIFF
--- a/packages/generator/templates/app/package.json
+++ b/packages/generator/templates/app/package.json
@@ -21,8 +21,8 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "tsc && lint-staged && pretty-quick --staged",
-      "pre-push": "npm run lint && npm run test"
+      "pre-commit": "lint-staged && pretty-quick --staged",
+      "pre-push": "tsc && npm run lint && npm run test"
     }
   },
   "lint-staged": {


machadolucasvp:
Hello @flybayer, could you review this PR? :smile:

flybayer:
Thank you @machadolucasvp!! So sorry for the long delay

@all-contributors add @machadolucasvp for code

PR #1370, opened 2020-10-20
fix types of auth mutations in examples and generator

utkarshkukreti:
By typing them as `LoginInputType` and `SignupInputType`, the user could access fields in the object before validation. Now they can only do after validation.

Screenshot 2020-10-21 at 1 33 30 AM

This change fixes a possible source of runtime type errors.

### Checklist

- [ ] Tests added for changes

I don't think this kind of change requires adding a test?

### DIFF
--- a/examples/auth/app/auth/mutations/login.ts
+++ b/examples/auth/app/auth/mutations/login.ts
@@ -1,8 +1,8 @@
 import {Ctx} from "blitz"
 import {authenticateUser} from "app/auth/auth-utils"
-import {LoginInput, LoginInputType} from "../validations"
+import {LoginInput} from "../validations"
 
-export default async function login(input: LoginInputType, {session}: Ctx) {
+export default async function login(input: unknown, {session}: Ctx) {
   // This throws an error if input is invalid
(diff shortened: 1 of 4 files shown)


flybayer:
Hey @utkarshkukreti! I appreciate the effort. That said, I don't think this is a change we actually want to make. 

In theory your change makes sense. It improves the typing inside the resolvers, however it also removes typing when actually using queries/mutations from the frontend or elsewhere. Before this change, you get an TS error if passing wrong params from the frontend. After this change, you don't get a TS error. Instead you only get an error at runtime.

So I think the net benefit of your change is negative. I think it's better to have proper typing when using resolvers than fixing the minor case of types before validation.

What do you think?

utkarshkukreti:
@flybayer oh, didn't think of that. Yeah, the current approach is definitely more useful for users than my proposed change. It would be nice to solve both points at some point as it's very easy to forget to validate form input server side. I don't know enough about how Blitz works to suggest anything, but maybe one day!

flybayer:
@utkarshkukreti Ok :) Appreciate the effort anyways!

Issue #225, opened 2020-04-24
Blitz needs to restart after schema change and CRUD generated


osirvent:
**What is the problem?**

`TypeError: Cannot read property 'findMany' of undefined` after schema changed and CRUD
generated

**Steps to Reproduce:**

1. Follow initial example (User Model and generate CRUD for user)
2. With `blitz` started, add a new model on the schema (i.e. `Comment`) .
3. `blitz db migrate`
4. `blitz generate all comment`
5. go to http://localhost:3000/comments
6. appears `TypeError: Cannot read property 'findMany' of undefined` error on the browser
7. appears `> Running getComments(null)
✕ getComments failed: TypeError: Cannot read property 'findMany' of undefined` on the console
8. Restart Blitz
9. Everything OK

**Versions:**

```
debug: local
debug: pkgPath: /Users/osirvent/dev/blitzjs/tests/test2/node_modules/@blitzjs/cli

macOS Catalina darwin-x64 node-v12.16.2

blitz: 0.6.2 (global)
blitz: 0.6.2 (local)
```

**Supporting Documentation**
`×TypeError: Cannot read property 'findMany' of undefined
getComments
./app/_rpc/comments/queries/getComments.ts:4

  1 | import db, {FindManyCommentArgs} from 'db'
  2 | 
  3 | export default async function getComments(args: FindManyCommentArgs) {
> 4 |   const comments = await db.comment.findMany(args)
  5 | 
  6 |   return comments
  7 | }

    at /Users/osirvent/dev/blitzjs/tests/test2/node_modules/
blitzjs/server/dist/server.cjs.development.js:1610:34
    at _catch (/Users/osirvent/dev/blitzjs/tests/test2/node_modules/
blitzjs/server/dist/server.cjs.development.js:1566:18
    at /Users/osirvent/dev/blitzjs/tests/test2/node_modules/
blitzjs/server/dist/server.cjs.development.js:1609:32
apiResolver
/Users/osirvent/dev/blitzjs/tests/test2/node_modules/next/dist/next-server/server/api-utils.js:46:15
processTicksAndRejections
internal/process/task_queues.js:97:5
async DevServer.handleApiRequest
/Users/osirvent/dev/blitzjs/tests/test2/node_modules/next/dist/next-server/server/next-server.js:464:9
async Object.fn
/Users/osirvent/dev/blitzjs/tests/test2/node_modules/next/dist/next-server/server/next-server.js:386:37
async Router.execute
[...]

PR #1374, opened 2020-10-22
Feat: Add next-prisma-plugin

JoseRFelix:
Closes: #225

#### Changes

- Add [next-prisma-plugin](https://github.com/prisma-labs/next-prisma-plugin)
- Update examples and generator to export one Prisma instance.

#### Implications

Blitz server won't need to restart after schema change and CRUD generated.

### Checklist

- [ ] Tests added for changes

### DIFF
--- a/examples/auth/db/index.ts
+++ b/examples/auth/db/index.ts
@@ -1,15 +1,6 @@
 import {PrismaClient} from "@prisma/client"
 export * from "@prisma/client"
 
-let prisma: PrismaClient
-
-if (process.env.NODE_ENV === "production") {
-  prisma = new PrismaClient()
-} else {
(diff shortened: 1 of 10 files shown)


flybayer:
Thanks @JoseRFelix! Initially I was hoping prisma could somehow fix the multi connection error that necessitated the globalThis.prisma code. However, it seems quite unlikely and we need to keep that code in.

However, currently the prisma plugin won't work with that code. So I've opened https://github.com/prisma-labs/next-prisma-plugin/issues/20 to fix this and make them compatible.

So need to let this set for a bit to see what happens with that.

flybayer:
I'm going to close this for now. See https://github.com/blitz-js/blitz/issues/225#issuecomment-783696018

Issue #1298, opened 2020-10-11
Create static framework for generation prompts


aem:
### What do you want and why?

In https://github.com/blitz-js/blitz/pull/1268 it's becoming apparent that we'll want to continue to add prompts to the app generation process (and potentially other generators), we should add a way to statically declare prompts on the generator class that automatically get executed (and the results passed into the templates). 

### Possible implementation(s)

Should be able to define an array of enquirer configs as a class property and the base `Generator<T>` can just execute them and then add the results to the object that gets passed to the subclass's template function

engelkes-finstreet:
Hey @aem 
I am thinking about the implementation for this right now. This is what I came up with

Extend the constructor with all the prompts you want to run in the generator
`constructor(protected readonly options: T, protected readonly prompts?: Array<Prompt<S>>)`

and use it like this

     const formChoices: Array<{name: AppGeneratorOptions["form"]; message?: string}> = [
        {name: "React Final Form", message: "React Final Form (recommended)"},
        {name: "React Hook Form"},
        {name: "Formik"},
      ]

      const formPrompt: Prompt<AppGeneratorOptions["form"]> = {
        name: "form",
        message: "Pick a form library (you can switch to something else later if you want)",
        choices: formChoices
      }

     const generator = new (require("@blitzjs/generator").AppGenerator)({
        destinationRoot,
        appName,
        dryRun,
        useTs: !flags.js,
        yarn: !npm,
        version: this.config.version,
        skipInstall,
        skipGit: flags["no-git"],
      }, [
        formPrompt
      ])

I added another property `formResults` to the Generator class so that you can later access it this way:

    switch (this.promptResults.form) {
      case "React Final Form":
        type = "finalform"
        pkg.dependencies["final-form"] = "4.x"
        pkg.dependencies["react-final-form"] = "6.x"
        break
      case "React Hook Form":
        type = "hookform"
        pkg.dependencies["react-hook-form"] = "6.x"
        break
      case "Formik":
        type = "formik"
        pkg.dependencies["formik"] = "2.x"
        break
    }

What do you think about this?

engelkes-finstreet:
I thought about this in the evening and I guess we should move everything inside the generator. Let me give this another try and I will create a draft PR.

beerose:
I'm closing this issue because of inactivity. We can open a new one or reopen this one if we have more demand for the framework for generation prompts.

PR #1375, opened 2020-10-23
Add PromptRunner and refactor new app command to show usage.

engelkes-finstreet:
Closes: [1298](https://github.com/blitz-js/blitz/issues/1298)

Adds a PromptRunner to generator which runs all prompts before the write action occurs and adds them to the `promptResult` object for later usage.

### DIFF
--- a/packages/generator/src/generator.ts
+++ b/packages/generator/src/generator.ts
@@ -111,12 +111,45 @@ function replaceJsxConditionals(program: Collection<j.Program>, templateValues: 
   return program
 }
 
+export type Prompt = {
+  name: string
+  message: string
+  choices: any
+}
+
+export class PromptRunner {
+  protected readonly enquirer: Enquirer
+
+  constructor(protected readonly prompts: Array<Prompt>) {
+    this.enquirer = new Enquirer()
+  }
+
+  async run(): Promise<any> {
+    let result = {}
+
+    for (let prompt of this.prompts) {
+      const promptResult = await this.enquirer.prompt({
+        ...prompt,
+        type: "select",
+      })
+
+      result = {
+        ...result,
+        ...promptResult,
+      }
+    }
+
+    return result
+  }
+}
+
 /**
  * The base generator class.
  * Every generator must extend this class.
  */
 export abstract class Generator<
-  T extends GeneratorOptions = GeneratorOptions
+  T extends GeneratorOptions = GeneratorOptions,
+  S extends object = {}
 > extends EventEmitter {
   private readonly store: Store
 
@@ -127,6 +160,7 @@ export abstract class Generator<
   private useTs: boolean
   private prettier: typeof import("prettier") | undefined
 
+  promptResults?: S
   prettierDisabled: boolean = false
   unsafe_disableConflictChecker = false
   returnResults: boolean = false
@@ -149,6 +183,10 @@ export abstract class Generator<
   abstract async getTemplateValues(): Promise<any>
 
   abstract getTargetDirectory(): string
+
+  getPromptRunner(): PromptRunner | undefined {
+    return undefined
+  }
 
   filesToIgnore(): string[] {
     // allow subclasses to conditionally ignore certain template files
@@ -280,6 +318,7 @@ export abstract class Generator<
       process.chdir(this.options.destinationRoot!)
     }
 
+    this.promptResults = await this.getPromptRunner()?.run()
     await this.write()
     await this.preCommit()
 
--- a/packages/generator/src/generators/app-generator.ts
+++ b/packages/generator/src/generators/app-generator.ts
@@ -1,4 +1,4 @@
-import {Generator, GeneratorOptions} from "../generator"
+import {Generator, GeneratorOptions, Prompt, PromptRunner} from "../generator"
 import spawn from "cross-spawn"
 import chalk from "chalk"
 import username from "username"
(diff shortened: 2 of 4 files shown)


packages/generator/src/generator.ts
```diff
@@ -150,6 +184,10 @@ export abstract class Generator<
 
   abstract getTargetDirectory(): string
 
+  getPromptRunner(): PromptRunner | undefined {
+    return undefined
```
clgeoio, line 188: suggestion: return an empty prompt runner here `new PromptRunner([])` to avoid needing to return undefined. It makes the interface a bit cleaner and removes the need to have optional chaining everywhere.

packages/generator/src/generator.ts
```diff
@@ -111,12 +111,45 @@ function replaceJsxConditionals(program: Collection<j.Program>, templateValues:
...
+}
+
+export class PromptRunner {
+  protected readonly enquirer: Enquirer
+
+  constructor(protected readonly prompts: Array<Prompt>) {
+    this.enquirer = new Enquirer()
+  }
+
+  async run(): Promise<any> {
```
clgeoio, line 127: Suggestioin, Instead of returning `Promise<any>` here, could we pass a generic parameter to the `PromptRunner` class, which sets the return type of this method?
e.g. `PrompRunner<T>`  and `run<T>()`

Returning `any` reduces the

packages/generator/src/generators/app-generator.ts
```diff
@@ -15,10 +15,13 @@ export interface AppGeneratorOptions extends GeneratorOptions {
   version: string
   skipInstall: boolean
   skipGit: boolean
+}
+
+export interface AppGeneratorPromptOptions {
   form: "React Final Form" | "React Hook Form" | "Formik"
```
clgeoio, line 21: suggestion: These can be extracted to an enum
``` 
enum FormOptions {
 FINAL_FORM = 'React Final Form',
...
}
```
Then can be referred to in other places by their enum:
```
    const formChoices: Array<{name: FormOptions; message?: string}> = [
      {name: FormOptions.FINAL_FORM, message: `${FormOptions.FINAL_FORM} (recommended)`},
    ]

    const formPrompt: Prompt = {
      name: "form",
      message: "Pick a form library (you can switch to something else later if you want)",
      choices: formChoices,
    }
```
```

File: packages/generator/src/generator.ts [hunk 4]
```diff
@@ -111,12 +111,45 @@ function replaceJsxConditionals(program: Collection<j.Program>, templateValues:
   return program
 }
 
+export type Prompt = {
```
clgeoio, line 114: suggestion: Pass a generic in this type to set the type of choices
```
export interface Prompt<T> {
  name: string
  message: string
  choices: T[];
}

Comment (clgeoio) [reply packages/generator/src/generator.ts:114]: also nit: Prefer the use of `interface` over `type` because interfaces [can be extended](https://www.typescriptlang.org/docs/handbook/advanced-types.html#interfaces-vs-type-aliases) whereas types cannot

File: packages/generator/src/generators/app-generator.ts (hunk 3 above)
Comment (aem) [reply packages/generator/src/generators/app-generator.ts:21]: I'd rather just leave it as the discriminated union. It makes the error messages a bit more helpful (`expected one of "a" "b" or "c"` rather than `expected a FormOption`)

File: packages/generator/src/generators/app-generator.ts [hunk 5]
```diff
@@ -38,6 +41,22 @@ export class AppGenerator extends Generator<AppGeneratorOptions> {
...
+    ]
+
+    const formPrompt: Prompt = {
+      name: "form",
+      message: "Pick a form library (you can switch to something else later if you want)",
+      choices: formChoices,
+    }
+
+    return new PromptRunner([formPrompt])
+  }
```
aem, line 58: So I'm curious why we'd take this approach of having to return a `PromptRunner` - it seems like we could just make this a class property, right? 

```
// prompts: Record<string, PromptConfig>
this.prompts = {
  formChoices: formPrompt
} as const;
```

That should allow us to safely type access to `this.promptResults` without having to introduce a new method to the generator interface.

engelkes-finstreet: I do not know how I can make the typing work for promptResult here to be honest. If I do it like you suggest I can run the prompts like this

    for (let prompt of this.prompts) {
      const promptResult = await this.enquirer.prompt({
        ...prompt,
        type: "select",
      })

      this.promptResults = {
        ...this.promptResults,
        [prompt.name]: promptResult
      }
    }

But this gives me TypeError for this.promptResults

_Error:(298, 7) TS2322: Type '(S & { [x: string]: object; }) | {}' is not assignable to type 'S | undefined'.
  Type '{}' is not assignable to type 'S'.
    '{}' is assignable to the constraint of type 'S', but 'S' could be instantiated with a different subtype of constraint 'object'._

This is the reason why I went with the Runner.

flybayer:
Hey @engelkes-finstreet are you able to finish this?

engelkes-finstreet:
Hey @flybayer,
could you take a look at my last comment. I am not sure how to get @aem s recommendations working. Maybe you have an idea?

packages/generator/src/generators/app-generator.ts
flybayer: @engelkes-finstreet I'm not totally following here, but you can definitely make it work with `as any`:
```ts
  this.promptResults = {
    ...this.promptResults,
    [prompt.name]: promptResult
  } as any
```

Do that to get past this. Then I can check the code and try to get a proper fix or it.

Issue #1376, opened 2020-10-23
`blitz db seed` requires typescript as production dependency


mweibel:
### What is the problem?
`blitz new` adds `typescript` as a devDependency, [`blitz db seed` requires typescript](https://github.com/blitz-js/blitz/blob/canary/packages/cli/src/utils/setup-ts-node.ts). When using `npm install --production` and trying to execute `blitz db seed`, it silently fails:

```
# ./node_modules/.bin/blitz db seed
You are using alpha software - if you have any problems, please open an issue here:
      https://github.com/blitz-js/blitz/issues/new/choose
```

Using `DEBUG=* ./node_modules/.bin/blitz db seed` showed me that it's somehow working until it executes the `db.js` file:
```
[...]
 @oclif/config:@blitzjs/cli require /app/node_modules/blitz/node_modules/@blitzjs/cli/lib/src/commands/db.js +98ms
  @oclif/config start prerun hook +1ms
  @oclif/config prerun hook done +1ms
  blitz:db init version: @oclif/command@1.5.20 argv: [ 'seed' ] +0ms
```

### Steps to Reproduce

1. `blitz new foo`
1. `cd foo`
1. `rm -Rf node_modules`
1. `npm install --production`
1. put stuff for a db seed in `db/seeds.ts`
1. `blitz db seed` fails without actual output

### Versions

```
» npx blitz version --verbose
You are using alpha software - if you have any problems, please open an issue here:
      https://github.com/blitz-js/blitz/issues/new/choose

@blitzjs/cli/0.24.3 darwin-x64 node-v14.12.0
```

PR #1363, opened 2020-10-19
Fix `blitz db seed` not working

nemesv:
Closes: #1221 

The `seed` method tries to load the `db/seeds.ts` TypeScript module which fails unless ts-node is imported.
It was working in development mode because oclif automatically sets up ts-node.
The current unite tests were passing because loading of the `db/seeds.ts` is mocked.

### Checklist

- [ ] Tests added for changes
- [ ] PR submitted to [blitzjs.com](https://github.com/blitz-js/blitzjs.com) for any user facing changes

PR #1377, opened 2020-10-23
(newapp) Move typescript to production dependencies (so `blitz db seed` works on prod deploy)

mweibel:
Closes: #1376

Adds typescript as a production dependency by default.


patosullivan:
#1363 doesn't fix it?

mweibel:
No, because ts-node requires typescript. Typescript is only a dev dependency. See linked issue :-)

Fix would be either to build seeds.ts into js on a prod build or require typescript as a prod dependency.

flybayer:
Thanks @mweibel! Makes sense to me.

@all-contributors add @mweibel for code

Issue #1356, opened 2020-10-18
RootErrorFallback has incorrect "Error" interface for some errors and is therefore casted to "any"


egonm12:
### What is the problem?
This problem consists of two issues:
1. The type of `error` in the interface `FallbackProps` from `react-error-boundary` is inconsistent with the actual error the `RootErrorFallback` contains in some cases. Because of this, the errors that are handled inside the `RootErrorFallback` component are therefore casted to `any`. E.g.:
```
 return (
      <ErrorComponent
        statusCode={(error as any).statusCode}
        title="Sorry, you are not authorized to access this"
      />
    )
```

2. A fallback for an `AuthenticationError` is implemented inside the `RootErrorFallback` component but it will never get there because inside `LoginForm` itself it returns another error when an Authentication Error arises.

### Steps to Reproduce
1.  Let the ErrorBoundary component handle the error of login form:
In `LoginForm.tsx` add the following inside the component
```
import { useErrorHandler } from "react-error-boundary"

export const LoginForm = (props: LoginFormProps) => {
  const handleError = useErrorHandler();
  
  ...
}
```
2. Replace the code in the `catch` block of the `onSubmit` with:
```
catch (error) {
  handleError(error)
}
```
3. Go to `_app.tsx` and add a log inside the `RootErrorFallback` method.
```
console.log(JSON.stringify(error))
```
4. Go to the login page and enter an incorrect email address and submit it
5. See in the console that the `error` object has a different interface than defined by `FallbackProps` : 
```
{"name":"AuthenticationError","statusCode":401,"message":"You must be logged in to access this"}
```
As already explained, the `error` inside `RootErrorFallback` has a different interface defined.

### Versions

```
debug: blitzPkgPath: /Users/egonmeijers/projects/todo-app/node_modules/blitz/dist/index.js
debug: cliPkgPath: /Users/egonmeijers/projects/todo-app/node_modules/@blitzjs/cli/lib/src/index.js

macOS Catalina | darwin-x64 | Node: v12.16.1

blitz: 0.24.2 (global)
blitz: 0.24.2 (local)

  Package manager: yarn
  System:
    OS: macOS 10.15.7
    CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
    Memory: 226.69 MB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 12.16.1 - ~/.nvm/versions/node/v12.16.1/bin/node
[...]

PR #1388, opened 2020-10-27
Add custom type guards for error checking

egonm12:
Closes: https://github.com/blitz-js/blitz/issues/1356

With instanceof type guards TypeScript is able to infer the type of the error. This is only possible because the custom errors extend the default 'Error' type. Therefore it is not needed to cast an error to 'any'. As an extra addition I've added a union type for the different custom errors that are build in.

### Checklist

- [ ] Tests added for changes

### DIFF
--- a/examples/auth/app/pages/_app.tsx
+++ b/examples/auth/app/pages/_app.tsx
@@ -1,5 +1,12 @@
-import {AppProps, ErrorComponent, useRouter} from "blitz"
-import {ErrorBoundary} from "react-error-boundary"
+import {
+  AppProps,
+  BlitzError,
+  ErrorComponent,
+  isAuthenticationError,
+  isAuthorizationError,
+  useRouter,
+} from "blitz"
+import {ErrorBoundary, FallbackProps} from "react-error-boundary"
 import {queryCache} from "react-query"
 import LoginForm from "app/auth/components/LoginForm"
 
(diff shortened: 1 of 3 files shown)


flybayer:
@egonm12 did you confirm this actually works when running the code? I want to make sure our current error serialization code doesn't break `instanceof`. Previously I was under the impression that `instanceof` didn't work right now, but I forget why :)

egonm12:
@flybayer I was not aware of this so thanks for the heads up! Looking at [this Github thread ](https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#extending-built-ins-like-error-array-and-map-may-no-longer-work) it seems to not work indeed but when trying something similar in the [TypeScript playground](https://www.typescriptlang.org/play?#code/FDDGBsEMGdoAgIIFcAuALApgOxQS1JHgPZYCiATuUeXBgB4rYAm8FVNA3sHHFpALYY4AXjgAiZOmx4CxMpWpjucaCkJJoAYSJMhogCwAGAIzLQJVeSSgU1ABSDYkAOZ7xATSJI4-DSjgARkLgRM6uTHC4WHC2cJCgoBiwMWi40GIAlHBcPDzQSAAOGOQOSdAuGBkgPAC+yq7+APoQGJDkAMpqoADWdlk5ueQYKEjk0ShWGMp1dWAW-sXsIrwYAO6IqJg4+IS4JGz2GQDcc1jQROAYAHQhznaL1JFnaliJRABmG1Lbsnvy7BkgA) it seems to work fine.

I will create an alternative with a custom type guard for now.

flybayer:
@egonm12 ok thanks! Last thing: can you open a PR to the docs repo to document to two new utilities?

egonm12:
@flybayer I will. 

Would it make sense to also create type guards for the other custom errors? They will not be used at the moment but might come in handy for users later.

flybayer:
@egonm12 I should have thought about this a bit more before having you document those type guards. Because I just went ahead and fixed the core issue of `instanceof` not working. As of https://github.com/blitz-js/blitz/pull/1414 `instanceof` works, so we no longer need the custom type guards.

egonm12:
@flybayer your fix was the one we eventually wanted anyway so nice that it is fixed 👍 . I will close this PR.

Issue #1017, opened 2020-09-07
Make `blitz new` use the latest published template instead of what's installed globally


flybayer:
### What do you want and why?

Currently `blitz new` will generate a new app with the template version of what's installed. So if global blitz version is `0.17.0`, then the new app template will be the template as of `0.17.0`, but the actual blitz version inside the new app will be the latest blitz version.

That results in a bad and confusing user experience.

### Possible implementation(s)

Instead of using the template packaged with npm, `blitz new` would download the new app template from the latest npm version and use that for new apps.

So if latest version is `0.21.0`, a global blitz version of `0.17.0` should create a new app with the template in `0.21.0`. The end result is the same as if the user upgraded the global cli version and then created a new app.

This may not be the best way. I'm open to suggestions for a better way! 

### Additional Details

The entry point for `blitz new` is [this command file](https://github.com/blitz-js/blitz/blob/canary/packages/cli/src/commands/new.ts). And here is the [actual app generator](https://github.com/blitz-js/blitz/blob/canary/packages/generator/src/generators/app-generator.ts). The new app template files are here: https://github.com/blitz-js/blitz/tree/canary/packages/generator/templates/app

cc @aem

aem:
I think what we really probably should do is prompt the user to globally upgrade Blitz if the version is out of date? If there were breaking changes in the generator/CLI packages between versions we could be trying to install a template that's incompatible with the globally installed CLI.

flybayer:
@aem hmm yeah maybe so. If we do that, then we might as well change all our docs to use `npx blitz new myapp` so that npx takes care of always having the latest version.

Without doing that, we'd somehow have to download the latest template and install the latest generator package.

aem:
could we add a pre-run step to prompt the user to upgrade if it's not the most current version? ask if they want to upgrade first, and if so do an `(npm i -g|yarn global add) blitz` first?

flybayer:
Yeah that actually sounds pretty good!

hardfire:
I like the ideas to 
 1. change the docs to use `npx blitz` instead of `blitz`. I personally have switched to doing this instead of using `yarn` or the global `blitz`
 2. add a pre run step to upgrade (if they want to)

engelkes-finstreet:
What do you think about letting the user choose which version he wants to install? I guess for now you always want to have the latest version but maybe you want to use some older versions in the future.

flybayer:
[...]

PR #1397, opened 2020-10-30
Upon `blitz new`, prompt user for upgrade if blitz version is outdated

abuuzayr:
Closes: #1017 

- If blitz global package is outdated, it will ask the user if they wish to upgrade

image

*Note:* While this is working and ready for review, I have some questions:

- Do we need to check if the global package was installed by `yarn` or `npm`? i.e. if it was installed by `npm` and `yarn` is installed, upgrade via `npm` instead of `yarn`. 
- Is testing the output required and possible? I saw in `new.test.ts` that the tests do not cover everything (e.g. that all the flags are working as intended) and I had some difficulty on writing tests against the code that I am adding, so I am wondering if this is something that is possible and that the tests should be more comprehensive.

### Checklist

- [ ] Tests added for changes

### DIFF
--- a/packages/cli/src/commands/new.ts
+++ b/packages/cli/src/commands/new.ts
@@ -65,7 +65,7 @@ export class New extends Command {
     debug("flags: ", flags)
 
     if (!flags["skip-upgrade"]) {
-      const latestVersion = (await getLatestVersion("blitz")).value
+      const latestVersion = (await getLatestVersion("blitz")).value || this.config.version
       if (lt(this.config.version, latestVersion)) {
         const upgradeChoices: Array<{name: string; message?: string}> = [
           {name: "yes", message: `Yes - Upgrade to ${latestVersion}`},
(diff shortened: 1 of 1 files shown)


flybayer:
Thanks!

> * Do we need to check if the global package was installed by `yarn` or `npm`? i.e. if it was installed by `npm` and `yarn` is installed, upgrade via `npm` instead of `yarn`.

Yes, this is important because if installed globally by both, the behavior is indeterminate.

> * Is testing the output required and possible? I saw in `new.test.ts` that the tests do not cover everything (e.g. that all the flags are working as intended) and I had some difficulty on writing tests against the code that I am adding, so I am wondering if this is something that is possible and that the tests should be more comprehensive.

Yeah not sure exactly how to write a test for this either. You can skip that, will just have to test manually.

abuuzayr:
Added check for whether the global package is installed by `yarn` or `npm` first before running upgrade. I opted to check `yarn` first because `npm list -g` takes 22s to return a response, with or without `--depth=0 --silent`, at least on my machine. `yarn global list` takes 2s. 

If we should check npm first, do let me know!

packages/cli/src/commands/new.ts
```diff
@@ -50,13 +53,74 @@ export class New extends Command {
...
+          {
+            name: "no",
+            message: `No - Continue with old version (${this.config.version}) - NOT recommended`,
+          },
+        ]
+
+        const promptUpgrade: any = await this.enquirer.prompt({
+          type: "select",
+          name: "upgrade",
+          message: "Your blitz CLI version is outdated. Upgrade?",
```
flybayer, line 81:
```suggestion
          message: "Your global blitz version is outdated. Upgrade?",
```

flybayer:
Sweet, looks good. Thank you!!

PR #1402, opened 2020-11-01
remove typescript bits from package.json generated for js projects

### LINKED ISSUE #1309: `package.json` generated for JS projects contains bits for TS projects
### What is the problem?
When generating a new blitz JS app with `blitz new --js`, the generated `package.json` contains fragments of stuff that is very much not JS. 

Some of these are: 
* `lint` script runs not only for `.js` but for `.ts`/`.tsx` extensions
* `pre-commit` husky hook runs `tsc` 
* `lint-staged` runs eslint against `.ts`/`.tsx` files
* multiple TS-related devdependencies such as `@types/..` `typescript` `ts-jest` and the TS eslint plugin and parser 

### Steps to Reproduce

1. Run `blitz new test --js --no-git`
2. Look at `package.json` inside the project folder

### Versions

```
debug: blitzPkgPath: C:\Users\Weil\AppData\Roaming\npm\node_modules\blitz\dist\index.js
debug: cliPkgPath: C:\Users\Weil\AppData\Roaming\npm\node_modules\blitz\node_modules\@blitzjs\cli\lib\src\index.js 

Windows 10 | win32-x64 | Node: v14.12.0

blitz: 0.24.1 (global)

  Package manager: yarn
  System:
    OS: Windows 10 10.0.19041
    CPU: (4) x64 AMD Ryzen 3 1200 Quad-Core Processor
    Memory: 5.64 GB / 15.94 GB
  Binaries:
    Node: 14.12.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.5 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 6.14.8 - C:\Program Files\nodejs\npm.CMD
    Watchman: Not Found
  npmPackages:
    @prisma/cli: Not Found
    @prisma/client: Not Found
    blitz: Not Found
    react: Not Found
    react-dom: Not Found
    typescript: Not Found
```

bcye:
Closes: #1309 

- create two seperate files for JS and TS package.json (package.js.json and package.ts.json)
- rename the selected one to package.json during the generation process, discard the other

### Checklist

- [x] Tests added for changes **(does this matter here? seems to be covered by existing ones)**


flybayer:
@bcye thanks so much! This solution would definitely work, but it would be better to only have a single package.json that has a few conditional changes based on whether it is JS or TS. Because with two package.json's, it will be easy for them to get out of sync.


@aem does our custom template engine support conditional checks in json files?

flybayer:
Bump @aem

flybayer (commented):
@beerose looks good, but we should remove `@types` packages from js package.json. Once that's changed, you are good to merge without another check from me.

Also I remembered that we run `tsc` inside the husky pre-push hook. We need to remove that for JS version. Can do that as a separate PR though.

beerose:
> Also I remembered that we run tsc inside the husky pre-push hook. We need to remove that for JS version. Can do that as a separate PR though.

Yep! Will work on that in a separate pr

flybayer:
@bcye so sorry for the long delay on this!!! I was thinking we needed to make changes in our codegen process, but ultimately decided to just go with your original approach for now, even though there is some duplication we have to worry about.

Issue #1407, opened 2020-11-03
TypeError: Cannot read property '_clientVersion' of undefined (0.25.0-canary.1 )


isoppp:
### What is the problem?

Causes following error when open localhost:3000.
( But the result seems to be no problem. )

```
(node:31313) UnhandledPromiseRejectionWarning: TypeError: Cannot read property '_clientVersion' of undefined
    at $connect (/Users/isoppp/workspace/oss/_testing/node_modules/@prisma/client/runtime/index.js:77960:32)
    at eval (webpack-internal:///./pages/api/auth/mutations/logout.ts:33:81)
    at /Users/isoppp/workspace/oss/_testing/node_modules/@blitzjs/server/dist/server.cjs.development.js:1585:11
    at dispatch (/Users/isoppp/workspace/oss/_testing/node_modules/@blitzjs/core/dist/core.cjs.development.js:858:32)
    at /Users/isoppp/workspace/oss/_testing/node_modules/@blitzjs/server/dist/server.cjs.development.js:2397:12
    at dispatch (/Users/isoppp/workspace/oss/_testing/node_modules/@blitzjs/core/dist/core.cjs.development.js:858:32)
    at /Users/isoppp/workspace/oss/_testing/node_modules/@blitzjs/core/dist/core.cjs.development.js:866:12
    at /Users/isoppp/workspace/oss/_testing/node_modules/@blitzjs/core/dist/core.cjs.development.js:769:30
    at _catch (/Users/isoppp/workspace/oss/_testing/node_modules/@blitzjs/core/dist/core.cjs.development.js:732:18)
    at Object.handleRequestWithMiddleware (/Users/isoppp/workspace/oss/_testing/node_modules/@blitzjs/core/dist/core.cjs.development.js:768:28)
```

### Steps to Reproduce

1. `npx blitz@canary new .`
2. `yarn run blitz run migrate`
3. `yarn start`

### Versions

```
blitz: 0.25.0-canary.1 (local)

  Package manager: yarn 
  System:
    OS: macOS 10.15.6
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 1.49 GB / 16.00 GB
    Shell: 3.1.0 - /usr/local/bin/fish
  Binaries:
    Node: 12.16.0 - /var/folders/xt/4fshmzms3s13prm5t364p_xm0000gp/T/yarn--1604420868303-0.857300376614728/node
[...]

PR #1419, opened 2020-11-05
Fix cannot read property '_clientVersion' of undefined

isoppp:
Closes: #1407

If `require(db).default.connect` (or `$connect`) is a function. It didn not have instance, so it had been lost `this` scope.
Support this case by keeping an instance when `require(db).default.connect` is provided.

This PR has to care about both #1407 and #1331.
Manually tested brand new project for 1407 and another project which has `db` import in a query but never used it.

### Checklist

- [ ] Tests added for changes


packages/server/src/stages/rpc/index.ts
```diff
@@ -42,13 +42,21 @@ let db${useTypes ? ": any" : ""}
 let connect${useTypes ? ": any" : ""}
 try {
   db = require('db').default
-  connect = require('db').connect ?? db?.$connect ?? db?.connect ?? (() => {})
-} catch(err) {}
+  if (require('db').connect) {
+    connect = require('db').connect
+  } else if (db?.$connect || db?.connect) {
+    connect = () => db.$connect ? db.$connect() : db.connect()
```
isoppp, line 48: this line is for #1407

packages/server/src/stages/rpc/index.ts
```diff
@@ -42,13 +42,21 @@ let db${useTypes ? ": any" : ""}
...
+  } else {
+    connect = () => {}
```
isoppp, line 50: this line is for #1331

flybayer:
Sweet, thanks!!

Issue #1443, opened 2020-11-10
Static pre-rendering with useQuery broken in `0.25.0-canary.5`


flybayer:
### What is the problem?

Because of https://github.com/blitz-js/blitz/pull/1422 `useQuery` is disabled on first render which means that useQuery doesn't throw and therefore the loading fallback is not rendered.

So we need to redo #1422 with a different approach

### Versions

```
0.25.0-canary.5
```

PR #1444, opened 2020-11-10
Fix broken static pre-rendering with useQuery

flybayer:
Closes: #1443 

Fix broken static pre-rendering with useQuery. Stop using the `enabled` flag since that messes with suspense loading. Instead use an empty query function during pre-rendering so the actual query resolver doesn't get called.

### Checklist

- [x] Tests added for changes


flybayer:
I have opened https://github.com/tannerlinsley/react-query/issues/1271 to hopefully have a first class solution for this at some point.

Issue #1421, opened 2020-11-06
The blitz dev cache must be busted if blitz version changes.


flybayer:
### What do you want and why?

Currently the `.blitz` dev cache is not cleared between `blitz start` invocations for performance. That's great, but we've found that when the blitz version changes we need to bust that cache. 

https://github.com/blitz-js/blitz/issues/1407 is an example error caused by the lack of this.

### Possible implementation(s)

Basically we want to clean the cache anytime the installed blitz version changes.

Setting [this `clean` config item](https://github.com/blitz-js/blitz/blob/canary/packages/server/src/dev.ts#L26) to true will clean the cache. That spot is a possible place to do this.

rayandrews:
Let me elaborate.

If blitz version in `package.json` is changed, we need to set `clean` in config to `true` right?

flybayer:
@rayandrews yes. So we need to cache the blitz version somehow inside `.blitz` so that each time `blitz start` runs, we can check the current blitz version with the cached version.

Also, we need to check the actual installed version of blitz, not just the version range in the app package.json. I think we can get the actual installed version with something like this: 

```ts
const pkgJson = require(path.join(require.resolve('blitz'), 'package.json'))
```

rayandrews:
I see, it makes sense now for the installed version one @flybayer 

However, how about the blitz version inside `.blitz`? Do we just use the `package.json` inside `.blitz` folder?

flybayer:
No, we don't currently story the blitz version. We could add a new file or maybe add a new field to `.blitz/caches/dev/_manifest.json`.

@ryardley any opinion on this?

rayandrews:
I see, we can cached the version via this [function](https://github.com/blitz-js/blitz/blob/canary/packages/server/src/stages/manifest/index.ts#L71)

and re-read it while running in development to make sure the version is same or not @flybayer

ryardley:
Maybe we can store the blitz version in the cache hash?

rayandrews:
could you give me the pointer of the cache hash?

I want to work on this.

is it similar like `file-hash` [here](https://github.com/blitz-js/blitz/blob/canary/packages/file-pipeline/src/pipeline.ts#L80) or different? 

@flybayer @ryardley

ryardley:
@rayandrews I was talking about the file tree hashing happening here: https://github.com/blitz-js/blitz/blob/canary/packages/server/src/build-hash.ts#L26

rayandrews:
[...]

PR #1446, opened 2020-11-10
Automatically clear `.blitz` cache when you change blitz versions

rayandrews:
Closes: #1421 

Add blitz version checking in `dev` mode to bust cache if the installed version is mismatched with the cache one.

Add new file in `devFolder` named `blitz-version`

### Checklist

- [x] Tests added for changes


rayandrews:
got to fix some failed tests

rayandrews:
I got problems with the `require.resolve` call in the `jest` test.

It turns out `server` package does not have `blitz` as the dependency (which is make sense).

Any idea how to do it? @ryardley

packages/server/src/blitz-version.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export async function isVersionMatched(buildFolder: string = ".blitz/caches") {
+  const versionStore = resolve(buildFolder, "blitz-version")
+  if (!(await pathExists(versionStore))) return false
+
+  try {
+    const buffer = await readFile(versionStore)
+    const version = getBlitzVersion()
+    const read = buffer.toString().replace("\n", "")
```
flybayer, line 21: You probably also want to trim any leading/lagging whitespace.

packages/server/src/blitz-version.ts
```diff
@@ -0,0 +1,32 @@
...
+    const path = require.resolve("blitz/package.json")
+    const pkgJson = require(path)
+    return pkgJson.version
+  } catch {
+    return null
+  }
+}
+
+export async function isVersionMatched(buildFolder: string = ".blitz/caches") {
+  const versionStore = resolve(buildFolder, "blitz-version")
```
flybayer, line 15: 1. I think maybe `blitz-version.txt` would be better?
2. There's a couple places in this file where you use the file name. Can you extract that out into a constant?

flybayer (commented):
Sweet, thanks!

You can use the `resolve-from` package to resolve blitz from the process CWD.

```ts
import resolveFrom from "resolve-from"
const blitzPath = resolveFrom.silent(process.cwd(), "blitz")
```

Add resolve-from as a dependency to the `server` package, but use the same version that's currently used in the `blitz` package.

packages/server/src/blitz-version.ts
```diff
@@ -0,0 +1,32 @@
+import {pathExists, readFile, writeFile} from "fs-extra"
+import {resolve} from "path"
+
+export function getBlitzVersion(): string | null {
+  try {
+    const path = require.resolve("blitz/package.json")
```
ryardley, line 6: Not sure if this is warranted but I am wondering if we should set an env var for this as the process launches? 🤔

packages/server/src/blitz-version.ts
rayandrews: got it trimming is important here

packages/server/src/blitz-version.ts
rayandrews: 1. I agree with that
2. do you need me to define it on the same file or different file?

packages/server/src/blitz-version.ts
rayandrews: not sure what you mean, is it putting the `blitz version` in `process.env`?

packages/server/src/blitz-version.ts
flybayer: same file is fine

rayandrews:
> Sweet, thanks!
> 
> You can use the `resolve-from` package to resolve blitz from the process CWD.
> 
> ```ts
> import resolveFrom from "resolve-from"
> const blitzPath = resolveFrom.silent(process.cwd(), "blitz")
> ```
> 
> Add resolve-from as a dependency to the `server` package, but use the same version that's currently used in the `blitz` package.

both `require.resolve` and `resolve-from` are not working on test environment but working nicely on `dev` command

So the issues are :
- this test case is always failed https://github.com/blitz-js/blitz/blob/canary/packages/server/test/dev.test.ts#L49
- Maybe we just should mock the file?

packages/server/src/blitz-version.ts
ryardley: Just an idea for solving the issues with `require.resolve`. It might not work if env isn't passed down between contexts - although it might be fine. I would investigate if anything already exists first and if not I would test if setting an env var in `blitz` package becomes available to `@blitzjs/server`. If so then one option might be to set the version there as a global and use that. This may not be the best option. Tradeoff is that we then would be relying on remote part of the system. The other alternative I see here would be to manually look for the package.json and parse it with fs. There are possibly a bunch of other ways to do this too so feel free to think of something else.

flybayer:
@rayandrews yeah I think fine to mock it.

packages/server/src/blitz-version.ts
rayandrews: I think the `process.env` will be passed throughout the entire process (unless we do some multi-node processes which I am afraid `blitz` does not do it)

The parsing of package.json is the one that I have done (however still quite problematic using `require.resolve` because the module path needs to be resolved well). However, in my code, if `require.resolve` fails, I will return `null` value. This will ensure everything works well, however, this will lead to mismatch version between the cache and installed one and in extend will result to clean cache every `dev` command is called

Brute-force looking for `blitz` package.json might be able to fix this even if `require.resolve` fails.

Want me to try the `process.env` approach? @ryardley

flybayer: @ryardley require.resolve is working fine, the only issue is in our test environment.

ryardley: Then it makes sense to just fix that

rayandrews:
Voila the tests have been fixed!

@ryardley @flybayer

flybayer:
Awesome!! Works good!

Let's make one last change to rename `blitz-version.txt` to `_blitz-version.txt` so that it matches `_manifest.json` which we also create in `.blitz/caches/dev/`.

@ryardley ready for final review.

rayandrews:
Got it @flybayer!

I have updated the code and will wait until the CI tasks are completed

flybayer:
Awesome, thank you!!

@all-contributors add @rayandrews for code

rayandrews:
Your welcome @flybayer!

Really happy and looking forward to contribute again!

Issue #734, opened 2020-07-09
Need ability to import isomorphic code from query/mutation files into React components


flybayer:
### What do you want and why?

I am defining input validation schemas in my queries and mutations like this:

```ts
//app/users/mutations/signup.ts
import db from "db"
import {SessionContext} from "blitz"
import {hashPassword} from "app/auth"
import * as z from "zod"

export const SignUpInput = z.object({
  email: z.string().email(),
  password: z.string().min(10).max(100),
})

export default async function signup(
  input: z.infer<typeof SignUpInput>,
  ctx: {session?: SessionContext} = {},
) {
  // This throws an error if input is invalid
  const {email, password} = SignUpInput.parse(input)
  const hashedPassword = await hashPassword(password)
  const user = await db.user.create({data: {email, hashedPassword, role: "user"}})

  await ctx.session.create({userId: user.id, roles: [user.role]})

  return user
}
```

And then I want to import that `SignUpInput` object in my component and use it for client-side validation like the below.

```ts
//app/components/SignUpForm.tsx
import signUp, {SignUpInput} from 'app/users/mutations/signup.ts'

// ...

const validInput = SignUpInput.parse(unknownInput)
```

But that currently doesn't work. Which is especially bad the Typescript server thinks the above code is valid (that you can import SignUpInput), but then it fails at compile time.

### Implementation

I don't know exactly, but currently we are moving the entire file at compile time to insert the API layer. But instead of that, we need to only swap out the `default` export and leave everything else alone.

PR #1456, opened 2020-11-12
Fix ability to import from pages, API routes, and queries/mutations

Skn0tt:
Closes: #1013
Closes: #734

Blitz's file pipeline moves files from *A* to *B*, but doesn't update any imports to *A*.
This PR makes sure that imports are rewritten, such that importing from any of the moved files doesn't break compilation.

### Checklist

- [X] Tests added for changes


flybayer:
Still in progress?

Skn0tt:
Yup

Skn0tt:
So I was able to make this kind of work - you can now import stuff from `api/` or `pages/`, which at least fixes Quirrel's API for now.

There's one problem though: When importing from the frontend (see `examples/store/app/admin/pages/admin/products/index.tsx`), you'll see
```
TypeError: undefined is not a constructor (evaluating 'new _prisma_client__WEBPACK_IMPORTED_MODULE_0__["PrismaClient"]()')
```
And that makes sense! Since our `db/index.ts` has side-effects, webpack can't eliminate that from the frontend-bundle, so the browser (rightfully) fails at importing Prisma.

I see two options that we can do:

1. Forbid people from importing `api/` routes in the frontend. Seems a bit harsh.
2. Tell webpack that it can eliminate `db/index.ts` for the frontend bundle. Seems doable, but what about non-isomorphic imports with side-effects in userland?

Any other ideas @ryardley @flybayer @aem?

flybayer:
I think it's fine to ensure anything in the `db` folder doesn't included in the frontend.

> but what about non-isomorphic imports with side-effects in userland?

What would be an example case for this?


Also, related issue to consider: https://github.com/blitz-js/blitz/issues/734

packages/server/src/with-blitz.ts
```diff
@@ -24,6 +24,7 @@ export function withBlitz(nextConfig: any) {
         } else {
           config.module = config.module || {}
           config.module.rules = config.module.rules || []
+          config.module.rules.push({test: /db/, use: {loader: "null-loader"}})
```
Skn0tt, line 27: This makes sure that anything in the `db` folder is never included in the frontend bundle.

Skn0tt:
> > but what about non-isomorphic imports with side-effects in userland?
> 
> What would be an example case for this?

I'm not sure tbh. That's a problem that existed before and now that I think about it, it's out of scope for this PR.

Skn0tt:
https://github.com/blitz-js/blitz/pull/1456/commits/565752e2c9e7c9115123d9f3440dd0ca9e1fc111 closes https://github.com/blitz-js/blitz/issues/734.

packages/server/src/stages/rewrite-imports/index.ts
```diff
@@ -0,0 +1,55 @@
...
+import path from "path"
+
+const isJavaScriptFile = (filepath: string) => filepath.match(/\.(ts|tsx|js|jsx)$/)
+
+const isInAppFolder = (s: string, cwd: string) => s.replace(cwd + path.sep, "").indexOf("app") === 0
+
+/**
+ * Returns a Stage that converts relative files paths to absolute
+ */
+export const createStageRewriteImports: Stage = ({config: {cwd}}) => {
```
Skn0tt, line 11: This stage is very similar to https://github.com/blitz-js/blitz/blob/07b7a8a14809097d119c7b080a31abea10ca3d58/packages/server/src/stages/relative/index.ts, maybe we can refactor them to share code.

Skn0tt:
> > but what about non-isomorphic imports with side-effects in userland?
> 
> What would be an example case for this?

I found one! The [current CI fail](https://github.com/blitz-js/blitz/pull/1456/checks?check_run_id=1402885086) seems to be caused by webpack trying to bundle `secure-password` in the frontend, which depends on Node APIs.

Skn0tt:
It fails because of this line in `auth-utils.ts`:

```ts
const SP = new SecurePassword()
```

Because this is side-effectful, webpack tries to bundle it in the frontend.

We could easily fix this by changing the line like so:

```ts
const SP = typeof window === "undefined" ? new SecurePassword() : undefined;
```

But isn't that kind of ugly? How does Next.js handle this?

Skn0tt:
Alright, so Next.js fails when importing Node APIs in the frontend: https://codesandbox.io/s/node-imports-test-z9wio
I guess that's the sensible way 😅

packages/server/src/stages/relative/pattern-relative-import.test.ts
```diff
@@ -17,7 +17,7 @@ describe("patternRelativeImport", () => {
   })
 
   it("matches an import statement with no specifier", () => {
-    const input = `import from "./ding"`
+    const input = `import "./ding"`
```
Skn0tt, line 20: `import from` isn't actual module syntax, so I changed this :D

packages/server/src/stages/rewrite-imports/index.ts
```diff
@@ -0,0 +1,85 @@
...
+      ,
+      origin,
+      ,
+    ] = args as string[]
+
+    const parts = origin.split("/")
+
+    if (parts.includes("pages")) {
+      if (parts[0] === "app") {
+        parts.splice(0, 1)
```
Skn0tt, line 51: If it's an import from a page, say from `app/pages/mypage.tsx`, we'll rewrite that import to `pages/mypage.tsx`.

packages/server/src/stages/rewrite-imports/index.ts
```diff
@@ -0,0 +1,85 @@
...
+      }
+    }
+
+    if (parts.includes("api")) {
+      parts.splice(0, parts.indexOf("api"), "pages")
```
Skn0tt, line 56: If it's an import from an API Route, say from `app/users/api/myRoute`, we'll rewrite it to `pages/api/myRoute`.

packages/server/src/stages/rewrite-imports/index.ts
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    if (starImport) {
```
Skn0tt, line 59: If it's a "star import" (what are they called irl?) like `import * as query from "app/queries/myQuery"`, I don't really know what to do. We'll just keep it as is for the moment.

In a perfect world, we'd somehow rewrite it to `app/_resolvers/queries/myQuery` but keep `query.default` as it is - I'm not so sure about how that's supposed to work, though.

packages/server/src/stages/rewrite-imports/index.ts
```diff
@@ -0,0 +1,85 @@
...
+      // I don't know yet what to do
+      return args[0]
+    }
+
+    if (parts.includes("queries") || parts.includes("mutations")) {
```
Skn0tt, line 64: Now the juicy part, which deals with #734. Whenever we see smth like `import { myFunc } from "app/queries/myQuery"`, we rewrite that to `app/_resolvers/queries/myQuery`.

When we see smth like `import myQuery, { myFunc } from "app/queries/myQuery"`, we rewrite it to the following two:
```ts
import myQuery from "app/queries/myQuery"
import { myFunc } from "app/_resolvers/queries/myQuery"
```

packages/server/src/stages/rewrite-imports/index.ts
```diff
@@ -0,0 +1,85 @@
...
+      ,
+      ,
+      starImport,
+      defaultImportName,
+      combinedComma,
+      namedImportNames,
+      ,
+      origin,
+      ,
+    ] = args as string[]
```
ryardley, line 45: Nice :) Now that's a few groups!

packages/server/src/stages/rewrite-imports/rewrite-imports.test.ts
```diff
@@ -0,0 +1,213 @@
...
+}
+            `,
+          },
+        ],
+        expectedOutput: [
+          {
+            path: normalize("/projects/blitz/blitz/app/anyFile.ts"),
+            contents: `
+import query from "app/users/queries/getUser"
+import { someFunction } from "app/_resolvers/users/queries/getUser";
```
ryardley, line 203: Should we be testing for dynamic imports too? `const bar = import('foo/queries/bar')`

packages/server/src/stages/relative/pattern-relative-import.test.ts
ryardley: Nice work

packages/server/src/stages/rewrite-imports/index.ts
```diff
@@ -0,0 +1,85 @@
...
+    const newContents = replaceImports(contents)
+    file.contents = Buffer.from(newContents)
+
+    return file
+  })
+
+  return {stream}
+}
+
+export const patternImport = /(import\s+)(?:(?:(\*\s+as\s+\w+)|(\w+\s*)?(,\s*)?(?:(\{\s*\w+\s*(?:,\s*\w+\s*)*\}))?))(\s+from\s+["'])?([\w\.\\\/]+)(["']\)?)/gs
```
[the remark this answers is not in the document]
Skn0tt: Yes it is: https://regex101.com/r/DO9kaT/4

Kinda proud that I got that working 😅

packages/server/src/stages/rewrite-imports/index.ts
Skn0tt: It is!

packages/server/src/stages/rewrite-imports/rewrite-imports.test.ts
Skn0tt: How do these work? Does `import("foo/queries/bar")` import the default export (the query)? Or is it like an `import *`?

Because if they work like `import *`, I'm not so sure as to how we can rewrite those in a sensible way...

packages/server/src/stages/rewrite-imports/rewrite-imports.test.ts
```diff
@@ -0,0 +1,213 @@
...
+        "react",
+        '"',
+      ],
+      [`import db, {ProductCreateArgs} from "db"
+      type CreateProductInput = {
+        data: ProductCreateArgs["data"]
+      }`]: ["import ", , "db", ", ", "{ProductCreateArgs}", ' from "', "db", '"'],
+    }
+
+    Object.entries(successes).forEach(([input, expectedOutput]) => {
```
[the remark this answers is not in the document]
Skn0tt: Makes sense ☺️ Done in https://github.com/blitz-js/blitz/pull/1456/commits/9c683cc4cf44499b8daab7f724d2bcd6ecedba19

packages/server/src/stages/rewrite-imports/rewrite-imports.test.ts
ryardley: > How do these work? Does import("foo/queries/bar") import the default export (the query)? Or is it like an import *?

I have no idea. Would have to look it up.

Skn0tt: Tried it, it imports the whole module.
Screenshot 2020-11-16 at 11 10 08

If you have an idea about how to rewrite these, I'm all ears. Otherwise, I'd say we let it slide for the moment (I don't think that a lot of people will import these dynamically, anyway)

ryardley: You can match the regex for it like this. 

```ts
/import\s?\(["']([^"']+)["']\);?/g
```

ryardley: Might have some corner cases.

Skn0tt: The problem is not detecting the import, but what to rewrite it to. See https://github.com/blitz-js/blitz/pull/1456#discussion_r523812784 for reference.

For imports from `api` and `pages` though, that's not a problem and I'll implement it.

packages/server/src/stages/rewrite-imports/index.ts
Skn0tt: I've also tried having the generated `queries/myQuery` file export everything from the resolvers: https://github.com/blitz-js/blitz/pull/1456/commits/aa67b5a3679dec03f510d2cf14be5b24b7b4c42a

This has the downside that it also bundles all transitive dependencies of the query, a lot of which cannot be tree-shaken because they have side-effects on the module level (even our default `auth-utils.ts` since it initialises `secure-password` on the top level).

One possible solution would be to make the additional export in https://github.com/blitz-js/blitz/pull/1456/commits/aa67b5a3679dec03f510d2cf14be5b24b7b4c42a only appear when there's actually some other file importing from there, but that would required coordination with the RPC stage.

packages/server/src/stages/rewrite-imports/rewrite-imports.test.ts
ryardley: Hmm sounds like we're almost getting into AST territory...

ryardley: Still it can be done with regex.

ryardley: Actually then there is the issue when you take a var and use it later 

```ts
const foo = import('thing/blat/boo');
//
foo.default()
```

ryardley: Perhaps we should do a double whammy and use an ast to solve this as well as https://github.com/blitz-js/blitz/issues/734 ?

Skn0tt: Could you help me understand what improvements an AST would bring in this case? Currently, I don't see a way to rewrite * imports and dynamic imports so that they work _at all_. What would you replace the code example above with?

I'll take a shot at the conditional exports mentioned in https://github.com/blitz-js/blitz/pull/1456#discussion_r524200987, I think that could be successfull.

Skn0tt: Heureka ☺️ I guess I solved it with 15bb9cc. The `rpc` stage now generates both a `myQuery.ts` and a `myQuery.named.ts` that includes named exports.

In the rewrite-imports stage, imports that include any named imports are rewritten to use `myQuery.named.ts`.

Skn0tt:
Anybody have an idea why 15bb9cc broke the Windows build?

Skn0tt:
> Anybody have an idea why 15bb9cc broke the Windows build?

Hmm, seems like it's been something else.

Skn0tt:
Anyway, seems like we've solved the main problem. Yay ☺️

flybayer:
Hallelujah!! Beautiful work mate!

I did run into a snag, I refactored the auth example to import zod schemas from the mutation files but the build failed with this:

```
error - ../../../../../node_modules/node-gyp-build/index.js:1:0
Module not found: Can't resolve 'fs'
null
```

I pushed the commit with this change.

Skn0tt:
That's because `mutations/login` imports `auth-utils`, which in turn imports `fs` via `secure-password`. Sadly, there's nothing we can do about that. Next.js would fail at that, too.

flybayer:
Oh, I thought you fixed that somehow. So probably we want to add null-loaders for secure-password and maybe also for auth-utils.

flybayer:
Another idea to consider separately would be to add ability to exclude a file from the client side bundle with a magic comment.

```
// blitz: server-only
```

Skn0tt:
> we want to add null-loaders for secure-password

Another option would be to make `auth-utils` side-effect free by replacing `const SP = new SecurePassword()` with `const SP = () => newSecurePassword()`. That'd allow it do be tree-shaken correctly.

flybayer:
Just making it side-effect free didn't work. Had to also add null-loader for secure-password.

Issue #388, opened 2020-05-01
`blitz start` broken in 0.8.0-canary.4


flybayer:
### Steps to Reproduce: 

1. `yarn global add blitz@0.8.0-canary.4`
2. `blitz new myapp`
3. `cd myapp`
4. `blitz start`
5. Fails with
```
/t/a1> blitz s
You are using alpha software - if you have any problems, please open an issue here:
  https://github.com/blitz-js/blitz/issues/new/choose

◢ Preparing for launchError: File not found with singular glob: /private/tmp/a1/node_modules/npm-normalize-package-bin//bin/ (if this was purposeful, use `allowEmpty` option)
    at Glob.<anonymous> (/private/tmp/a1/node_modules/glob-stream/readable.js:84:17)
    at Object.onceWrapper (events.js:417:26)
    at Glob.emit (events.js:310:20)
    at Glob.EventEmitter.emit (domain.js:482:12)
    at Glob._finish (/private/tmp/a1/node_modules/glob/glob.js:197:8)
    at done (/private/tmp/a1/node_modules/glob/glob.js:182:14)
    at Glob._processSimple2 (/private/tmp/a1/node_modules/glob/glob.js:688:12)
    at /private/tmp/a1/node_modules/glob/glob.js:676:10
    at Glob._stat2 (/private/tmp/a1/node_modules/glob/glob.js:772:12)
    at lstatcb_ (/private/tmp/a1/node_modules/glob/glob.js:764:12)
```

Maybe this is related to changes in #365?

#### Weird workaround

1. `rm -rf node_modules`
2. `yarn`
3. `blitz start` -> works

cc @ryardley @Skn0tt

flybayer:
@Skn0tt it seems the gitignore check is crawling the file system to and compiling a list of all ignored files. Is that what it's doing?

Skn0tt:
Yup. Didn't know about the `allowEmpty` option 🙈 I'm gonna provide a PR real quick

Skn0tt:
As can be seen from the stack trace, `allowEmpty` is an option on `glob-stream`, which is a dependency two layers below `fast-glob`.  Let's see how we can interface with that 🧐

Skn0tt:
After some investigation, it seems like `glob-stream` is just a devDependency to `fast-glob`. The real reason we've got `glob-stream` is `vinyl-fs`:

```sh
$ yarn why glob-stream
=> Found "glob-stream@6.1.0"
info Reasons this module exists
[...]

PR #1457, opened 2020-11-13
Omit inclusion rules found in the global gitignore

### LINKED ISSUE #1413: error: File not found with singular glob
### What is the problem?

When trying to start an brand new project I get the following error 

```
Preparing for launch[Error: File not found with singular glob: /Users/talbet/Experiments/btest/TAGS/ (if this was purposeful, use `allowEmpty` option)]
```

Could be related to #388 but TAGS makes me think it is unrelated.

### Steps to Reproduce

1. run `blitz new btest` (have also tried with --npm and --no-git)
2. run initial migration inside btest directory
3. run `blitz start`

### Versions

```
macOS Mojave | darwin-x64 | Node: v14.15.0

blitz: 0.24.3 (global)
blitz: 0.24.3 (local)

  Package manager: npm
  System:
    OS: macOS Mojave 10.14.6
    CPU: (4) x64 Intel(R) Core(TM) i7-6567U CPU @ 3.30GHz
    Memory: 673.28 MB / 16.00 GB
    Shell: 5.7.1 - /usr/local/bin/zsh
  Binaries:
    Node: 14.15.0 - /usr/local/bin/node
    Yarn: 1.22.5 - ~/.yarn/bin/yarn
    npm: 6.14.8 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  npmPackages:
    @prisma/cli: 2.10.2 => 2.10.2
    @prisma/client: 2.10.2 => 2.10.2
    blitz: 0.24.3 => 0.24.3
    react: 0.0.0-experimental-7f28234f8 => 0.0.0-experimental-7f28234f8
    react-dom: 0.0.0-experimental-7f28234f8 => 0.0.0-experimental-7f28234f8
    typescript: 4.0.5 => 4.0.5
```

### Other

I have tried to create an empty TAGS directory to see if that helps, but I still get the error.

dmzza:
Closes #1413 

It would be better to use glob's `allowEmpty` option, but since we cannot, and some inclusion rules which may not apply to a Blitz project cause `blitz start` to fail with this error:
`Error: File not found with singular glob`
I just omit the inclusion rules that we find in the global gitignore file. 

The downside is that if you meant for those inclusion rules in your global gitignore file to apply to your Blitz projects, you may now encounter a mysterious situation where those files are ignored by an exclusion rule in your global gitignore file.

My preference would be to avoid reading the global gitignore file and just accept the performance hit from monitoring extra files that might be ignored in the global gitignore. 

But out of respect for the author's intent, I submit this PR. I'm happy to discuss it. And I can add tests for my solution here if you think it makes sense. 

### Checklist

- [ ] Tests added for changes
- ~~PR submitted to [blitzjs.com](https://github.com/blitz-js/blitzjs.com) for any user facing changes~~

### DIFF
--- a/packages/server/src/parse-chokidar-rules-from-gitignore.ts
+++ b/packages/server/src/parse-chokidar-rules-from-gitignore.ts
@@ -47,14 +47,18 @@ export function getAllGitIgnores(rootFolder: string) {
 
   return files.filter(isControlledByUser).map((file) => {
     let prefix = ""
+    let isGlobal = false
 
     if (file.match(localRepoIgnore)) prefix = file.split(localRepoIgnore)[0]
-    else if (file.match(globalIgnore)) prefix = ""
-    else prefix = file.split(".gitignore")[0]
+    else if (file.match(globalIgnore)) {
+      prefix = ""
+      isGlobal = true
+    } else prefix = file.split(".gitignore")[0]
 
     return {
       gitIgnore: fs.readFileSync(file, {encoding: "utf8"}),
       prefix,
+      isGlobal,
     }
   })
 }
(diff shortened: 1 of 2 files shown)


flybayer:
Thanks @dmzza! Question: why can't we use the `allowEmpty` option?

@ryardley what are your thoughts on this?

dmzza:
@flybayer I haven't looked closely, but when I read this https://github.com/blitz-js/blitz/issues/388#issuecomment-622330807 I figured @Skn0tt looked into `allowEmpty` already and ruled it out.

flybayer:
@dmzza ok I think your suggestion to just ignore the global gitignore is probably the best option

dmzza:
@flybayer After a closer look, we can use the `allowEmpty` through vinyl-fs. I tried it out and it worked for me. I submitted #1493 to replace my first attempt here.

Issue #1459, opened 2020-11-14
Next image component external domains not working in production


gielcobben:
### What is the problem?

I set some external domains for the next image component but the images are not working in production. If you go to one of the image urls directly in the browser it returns `"url" parameter not allowed`. The url looks like `example.com/_next/image?url=...`

I put the following in `blitz.config.js`:

```
module.exports = {
  images: {
    domains: ["example.s3.amazonaws.com"]
  }
};
```

### Steps to Reproduce

1. Set external domains in `blitz.config.js`.
1. `blitz start --production` or Deploy your app.
1. Go to one of the image urls in the browser and see an error: `"url" parameter not allowed`.

### Versions

```
macOS | darwin-x64 | Node: v12.18.0

blitz: 0.25.0 (global)
blitz: 0.25.0 (local)

  Package manager: yarn
  System:
    OS: macOS 11.0.1
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 176.34 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 12.18.0 - ~/.nvm/versions/node/v12.18.0/bin/node
    Yarn: 1.21.1 - /usr/local/bin/yarn
    npm: 6.14.4 - ~/.nvm/versions/node/v12.18.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  npmPackages:
    @prisma/cli: 2.11.0 => 2.11.0
    @prisma/client: 2.11.0 => 2.11.0
    blitz: 0.25.0 => 0.25.0
    react: 0.0.0-experimental-4ead6b530 => 0.0.0-experimental-4ead6b530
    react-dom: 0.0.0-experimental-4ead6b530 => 0.0.0-experimental-4ead6b530
    typescript: 3.8.3 => 3.8.3
```

flybayer:
@gielcobben does it work locally with `blitz start --production`?

gielcobben:
Nope

flybayer:
I'll try to look at this soon unless someone else does first

mhmdAljefri:
@flybayer I'm getting the same issue when I upgraded blitz from `0.29.7` to `0.31.0`!

flybayer:
@mhmdAljefri can you please try with `0.33.1`? I think it should be working there

mhmdAljefri:
@flybayer Yes it works, thank you

PR #1463, opened 2020-11-14
Fix image component external domains not working in production

flybayer:
Closes: #1459

With the new image component, Next.js is now reading the `next.config.js` file at _runtime_. Our compiled `next.config.js` file is in `.blitz/caches/build` but we execute `next start` with a cwd of the original project which doesn't have `next.config.js`.

So this fix leaves `.next` inside `.blitz/caches/build/` **and** it runs `next start` in the context of `.blitz/caches/build/`.

During a serverless build on Vercel, we already add `next.config.js` so that should keep working. However for other serverless environments like serverless framework, I think this problem may still exist.

The ultimate solution, imo, is to make a PR to Next.js that copies the next.config.js file into `.next` and then change `next start` to read from there. We can investigate this later.

### DIFF
--- a/packages/server/src/build.ts
+++ b/packages/server/src/build.ts
@@ -1,4 +1,4 @@
-import {move, pathExists, remove} from "fs-extra"
+import {copy, pathExists, remove} from "fs-extra"
 import {resolve} from "path"
 import {saveBuild} from "./build-hash"
 import {normalize, ServerConfig} from "./config"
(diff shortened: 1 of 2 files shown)


Flavyoo (approved):

flybayer:
@ryardley can you review this?

ryardley (approved):
Makes sense to me.

Issue #1427, opened 2020-11-07
useParam is better to return undefined when the param is missing


isoppp:
### What do you want and why?

```
const query = {}
cosnt params = useParams() // returns {}
cosnt params = useParams('number') // returns {}
const param = useParam('key', "number") // returns NaN
```

1. I want to check param like `if(param)` but `NaN` will be `true`. It might not be useful.
2. Between `useParams` and `param` is not consistent, I'm a bit confusing.
3. Type implementation is `number | undefined`, but never returned undefined I guess

https://github.com/blitz-js/blitz/blob/canary/packages/core/src/use-params.ts#L92
```
export function useParam(key: string, returnType: "number"): number | undefined`
```

### Possible implementation(s)

Want to change these test results to be `undefined`.

`useParam("doesnt-exist", TYPE)` in https://github.com/blitz-js/blitz/blob/canary/packages/core/test/use-params.test.ts

or

change type definition such as `number | undefined` to be `number` if it truly never returned `undefined`

### Additional context

When first render, `useRouter().query` is always return empty object {}  as mentioned in #1344 ~~#1334~~, this modification might be helpful.

flybayer:
@isoppp yeah, you have a great point.

We should change `useParam` to always return `undefined` if the requested param doesn't exist. This will be a change to all three types, string, number, and array.

HendrikCromboom:
To pull from some of the more back end minded techs:

I always loved the: mixedType (number OR undefined)

That way you could inject the typed return value rather than hardcoding it.

MrLeebo:
We've [talked about the implications of type checking URL parameters before](https://github.com/blitz-js/blitz/issues/530), but now that nextjs supports [rewrites](https://nextjs.org/docs/api-reference/next.config.js/rewrites), we could revisit the idea of defining a type mapping for parameters and lettings blitz generate the corresponding rewrites to enforce it. That would make it impossible for the param to have an invalid value in the first place.

flybayer:
@MrLeebo regardless if we add that, there will still be the case of `undefined` for static pre-rendering and first render, which this issue is mainly about.

KATT:
I did a PR above, but still remaining issue in regards to `NaN` - I personally really dislike `NaN` and always prefer to throw, or return `undefined`, or return `null` whenever I parse numbers

KATT:
[...]

PR #1474, opened 2020-11-17
Change `useParam()` to always return `undefined` if param doesn’t exist

KATT:
I just had a look at "good first issue" and found this one and fixed it as I interpreted the discussion.

I'm still not convinced that `NaN` is _ever_ is good to return as specified in #1427. Maybe we want to do an `isNaN`-check as well, treat it as "unset", and return undefined?

---

Closes: #1427

### Checklist

- [x] Tests added for changes


packages/core/src/use-params.ts
```diff
@@ -97,6 +97,11 @@ export function useParam(
 ): undefined | number | string | Array<string | undefined> {
   const params = useParams()
   const rawValue = params[key]
+  const exists = params.hasOwnProperty(key)
```
KATT, line 100: possibly

```suggestion
  const exists = params[key] !== undefined
```
or

```suggestion
  const exists = params[key] != null
```

Flavyoo, line 100: Nit, what do you think about rename exists to something like `keyExists`. This is not a blocking change, but it helps with readability.

Flavyoo: You can also try the following since it will handle case for rawValue being undefined and null.

``
const exists = !!rawValue
``

I also think it's cleaner to set `exists` before accessing the rawValue. The way it is now, we don't need to call hasOwnProperty like you suggested above if we know rawValue is falsey.

Flavyoo (changes requested):

flybayer:
Sweet! Flavyoo has some good comments, and then let's go ahead and return `undefined` instead of `NaN`.

packages/core/src/use-params.ts
KATT: No, falsey is not correct - it will break empty strings.

https://github.com/blitz-js/blitz/blob/9819543f47cc03c52eaceb829bbdeff441aeb9e7/packages/core/test/use-params.test.ts#L116-L117

It think it has to be either hasOwnProperty (above), strict undefined check (above) or non-strict null-check (above), just pick one of those :) 

> You can also try the following since it will handle case for rawValue being undefined and null.

`!= null` does just that - but the `null`-check is redundant which is why I also proposed `!== undefined`

KATT: I could make this change if you want, but I'll wait for the resolution of the above discussion first

KATT (commented):
@Flavyoo @flybayer the requested changes wasn't valid, so please let me know what to do

KATT:
Made another suggestion with `typeof rawValue === undefined` in the top instead.

packages/core/src/use-params.ts
Flavyoo: Thanks for catching that

Flavyoo (approved):

packages/core/src/use-params.ts
Flavyoo: Going to mark this as resolved :)

packages/core/src/use-params.ts
```diff
@@ -98,6 +98,10 @@ export function useParam(
   const params = useParams()
   const rawValue = params[key]
 
+  if (typeof rawValue === "undefined") {
+    return undefined
+  }
+
   if (returnType === "number") {
     // Special case because Number("") === 0
     if (rawValue === "") {
```
Flavyoo, line 107: Right after this line, it is returning `NAN`. should that be returning `undefined` instead?

Flavyoo: I'm not sure how to view the entire file and still provide comments on unchanged code :(

Flavyoo (changes requested):

packages/core/src/use-params.ts
KATT: I wish it did returned `NaN` too -- I have a WIP where I refactor the whole thing and suggest to never return `NaN` but it would be a separate change imo

flybayer: @katt I think we should go ahead and do the NaN change with this

KATT: Done this now - also see #1491

flybayer:
Sweeet, thanks!

Issue #1484, opened 2020-11-18
Update Tailwind Recipe for 2.0


flybayer:
### What do you want and why?

Tailwind 2.0 [was just released](https://blog.tailwindcss.com/tailwindcss-v2), so we need to update our tailwind recipe.

### Possible implementation(s)

Make any needed changes in https://github.com/blitz-js/blitz/blob/canary/recipes/tailwind/index.ts#L19. For sure the tailwind version number needs changed from 1 to 2. That could be all, need to check the release notes and see if anything else has changed.

Here is new installation instructions: https://tailwindcss.com/docs/installation

After changing the recipe locally on your computer, you can run `blitz install [PATH_TO_MONOREPO]/recipes/tailwind` to test your changes in any app your on your computer.

PR #1486, opened 2020-11-19
(recipe) Update tailwind recipe to v2

Mzaien:
Followed the instructions but it needs testing

Closes: #1484

changed tailwind version number to 2
### Checklist

- [ ] Tests added for changes


Skn0tt:
Tailwind v2 comes with the new `darkMode` option which I think we should also add to our template:
```js
darkMode: false, // or 'media' or 'class'
```

Also, when I tried out Tailwind v2 a couple of minutes ago, it failed because of a version of PostCSS that's too old - maybe also take a look at that.

flybayer (changes requested):
Thanks!

There's some more changes that are needed as shown here: https://github.com/vercel/next.js/pull/19278/files

- Remove `postcss-present-env` and add the new required deps
- Update the postcss config template
- Remove the `future` key from tailwind config

pbteja1998:
> Thanks!
> 
> There's some more changes that are needed as shown here: https://github.com/vercel/next.js/pull/19278/files
> 
> * Remove `postcss-present-env` and add the new required deps
> * Update the postcss config template
> * Remove the `future` key from tailwind config

@flybayer I don't think postcss-preset-env needs to be removed. Checkout the default postcss config that comes with Next.js [here](https://nextjs.org/docs/advanced-features/customizing-postcss-config#customizing-plugins). It already has `autoprefixer` in it. We just have to add `tailwindcss` to [this config](https://nextjs.org/docs/advanced-features/customizing-postcss-config#customizing-plugins) and also add `postcss 8` as a dev dependency.

We also no longer need to have a separate stylesheet. We can just import from `tailwindcss/tailwind.css`. I added all these in a separate PR(#1490) a while back. I did not see this PR then. I already tested the recipe I added in that PR too. So that config works. 

And another thing is that all of these can be devDependencies.

Mzaien:
Thanks to @pbteja1998  you helped me a lot i'm a beginner in this field  And also thanks for @flybayer and @Skn0tt

flybayer:
@Mzaien sorry for the review delay here!

The [Next.js Tailwind v2 update](https://github.com/vercel/next.js/pull/19278/files) is merged, so let's follow that.

- [ ] Remove `postcss-flexbugs-fixes`
- [ ] Remove `postcss-preset-env`
- [ ] Add `autoprefixer` dependency
- [ ] Change all dependencies to use a major version number like `"2"` instead of `"latest"` to prevent breaking change issues when major version changes
- [ ] Keep `app/styles/index.css` but remove `button.css`. 
  - @pbteja1998 you are right it's not needed, but it's very likely any app using tailwind will have some global styles to add. Having it there makes this easier. In general with Blitz we default to include the files you're likely to use vs the minimum possible.
- [ ] Change `postcss.config.js` to have this content as prescribed by the tailwind docs:
```js
// postcss.config.js
module.exports = {
  plugins: {
    tailwindcss: {},
    autoprefixer: {},
  }
}
```

flybayer (approved):

flybayer:
Awesome, thank you!!

@all-contributors add @Mzaien for code

PR #1491, opened 2020-11-20
Change `useParam` / `useParams` to always return `undefined` for empty/invalid values instead of `""` and `NaN` (patch)

KATT:
Related: #1427 #1474 

- Refactor useParam / useParams so that useParams does all the typecasting magic and useParam simply gets `params[key]`
- `useParam` with number will **never** return `NaN`

### Checklist.

- [x] Tests added for changes


packages/core/src/use-params.ts
```diff
@@ -90,38 +94,13 @@ export function useParams(returnType?: "string" | "number" | "array") {
...
-      return []
-    }
-    if (!Array.isArray(rawValue)) {
-      return [rawValue]
-    }
-    return rawValue
-  }
+  returnType?: ReturnTypes,
+): undefined | number | string | string[] {
+  const params = useParams(returnType as any) // unsure why i need to typecast here
```
KATT, line 102: ❓ ❗ 

Unsure why I need to typecast as any here and why typescript doesn't get it. Haven't used method overloading in TS much before so there is prob some quirk.

KATT: When changing to

```suggestion
  const params = useParams(returnType)
```

I get the following error:

```
No overload matches this call.
  The last overload gave the following error.
    Argument of type '"string" | "number" | "array" | undefined' is not assignable to parameter of type 'undefined'.
      Type '"string"' is not assignable to type 'undefined'.ts(2769)
use-params.ts(42, 17): The last overload is declared here.
use-params.ts(47, 17): The call would have succeeded against this implementation, but implementation signatures of overloads are not externally visible.
```

dmzza: It looks like what you want is a generic function. You could probably replace lines 42-45 with just this:

```
export function useParams<T>(returnType: T): Dict<T>
```

Also, I think your definition on line 41 already covers the definition on line 42.

KATT: Thanks @dmzza. I've played around a bit with generics but couldn't get it to work either. `Dict<T>` won't work as it's string literals and not actually the type that's passed in.


If you - or anyone else watching - wanna clone it and play around, here's a little shortcut:

```sh
git clone git@github.com:KATT/blitz.git katt-blitz
cd katt-blitz/packages/core
git checkout feature/return-strict-or-undefined
yarn
yarn test:watch
```

rayandrews: just add like this in `useParams` and all will be good!

```typescript
export function useParams(): Dict<string | string[]>
export function useParams(returnType?: ReturnTypes): Dict<string | string[]>
export function useParams(returnType: "string"): Dict<string>
export function useParams(returnType: "number"): Dict<number>
export function useParams(returnType: "array"): Dict<string[]>
```

KATT: Amazing! Thank you @ryandrews 🙏

packages/core/src/types.ts
```diff
@@ -64,6 +64,10 @@ export interface CancellablePromise<T> extends Promise<T> {
 
 export type QueryFn = (...args: any) => Promise<any>
 
+export type Dict<T> = Record<string, T | undefined>
```
KATT, line 67: Shortcut to writing `Record<string, string|undefined>` etc

Iguanya (approved):
well that's helpful

flybayer:
Awesome, thanks @KATT!!

flybayer (approved):

Issue #303, opened 2020-04-27
[Feature] Add custom server support for websockets, etc


dajinchu:
### What do you want and why?

Apologies if this is already supported, but it doesn't seem possible to run websockets since there's no way to hook into the http server? 

### Possible implementation(s)

**There could be dedicated support for websockets.**

1. In the queries and mutations you could import 'ws' similar to 'db', which gets a singleton of a SocketIO.Server which you can use to emit things.
2. Using the blitz config, you could specify a function that binds listeners to Socket.IO server `(io: SocketIO.Server) => void`.

**There could be a more generic way to hook into the blitz server**

Basically just do [what next.js does](https://nextjs.org/docs/advanced-features/custom-server)

This has the advantage of making it obvious that websockets don't work in a serverless deployment.

Zeko369:
If I'm not mistaken @flybayer suggested having an ability to expose the server in the future for full-stack deploys, because then we have a problem of the app not being able to be hosted on now/zeit/vercel/howeverItIsCalledNow

ryardley:
This is a good suggestion however it is not easily compatible with our primary deployment target which right now is serverless api functions. This sort of usecase could be abstracted away by an eventual underlying architecture that enables serverside resolvers can call clientside resolvers via long polling or something like that but for now this is likely a backlog item.

flybayer:
I would love to support websockets for `blitz start --production`, which runs as a server.

I don't have much experience here, so I'd love ideas and/or prototypes for how this could be implemented.

wKovacs64:
> And maybe it is just enough to expose a custom server. Probably should do that regardless of websocket support.

Yep. I've written a websocket app in vanilla Next.js and the custom server feature was all I needed. Explicit, baked-in support (as DJ mentioned) could be supported down the road but I doubt the juice would be worth the squeeze (at least initially).

dajinchu:
100% I think the custom server would be the best option. Websockets are a niche use case, with this project being geared towards serverless, so it probably doesn't make sense for baked in support. 
However, IMO blitz has a lot of great features other than serverless, but needing to use websockets or other long-running server features shouldn't preclude you from using the rest of blitz. 

Seems like exposing the server Next-style and adding an example/installer/recipe would be quite nice.

flybayer:
Ok let's do that. Support a custom `server.js` file. 

If `server.js` is present, `blitz start` will use that instead of `next dev/start`

dajinchu:
[...]

PR #1492, opened 2020-11-20
Add custom server support!

KATT:
Closes: #303 

Custom server support is nice to have in order to add stuff like websockets and such.

- New function `createBlitzApp` which starts blitzjs programmatically without starting an actual web server and returns a `requestHandler` - similar and built on how [nextjs does it](https://nextjs.org/docs/advanced-features/custom-server).
- Change the `last-build`-file to contain the whole tree of diff. Background: needed to debug why CI was telling me the build wasn't clean which led me to exclude `dist/`, `build/` and `.tsbuildinfo` in the `build-hash.ts:getInputArtifacts`
- Some more `debug()`s added whilst debugging, might as well keep them
- There is a working example in [`examples/custom-server`](https://github.com/KATT/blitz/tree/issues/303-custom-server/examples/custom-server) with added cypress tests that run in CI

### Further help wanted below:

- How could one access/generate blitz context/db from their own server easily? I would like to make a full example with WebSockets, but need some guidance on how to create a blitz context that could be used in the WebSocket-server - the whole flow of a request with `handleRequestWithMiddleware`/`getAllMiddlewareForModule`/etc is a bit hard to grasp. Maybe `createBlitzApp` could return some useful methods where one can authorize the user and access the blitz ctx.
- It would be nice if it didn't need a separate tsconfig and that blitz would transpile it for us with it's tsconfig as part of build, but blitz seems to delegate that part to nextjs so not sure if it's feasible. But right now it basically is on pair with how nextjs does it, which is good enough for me at least.

### Checklist

- [x] Tests added for example
- [x] PR submitted to [blitzjs.com](https://github.com/blitz-js/blitzjs.com) for any user facing changes: https://github.com/blitz-js/blitzjs.com/pull/307


examples/plain-js/package.json
```diff
@@ -2,6 +2,7 @@
   "name": "@examples/plain-js",
   "version": "0.25.1-canary.1",
   "scripts": {
+    "dev": "node server",
```
KATT, line 5: will make a separate example for custom server, just keeping it here whilst developing

packages/server/src/create-blitz-app.ts
```diff
@@ -0,0 +1,60 @@
...
+  if (dev) {
+    // dev
+    const stages = configureStages({writeManifestFile, isTypescript})
+
+    await transformFiles(rootFolder, stages, devFolder, {
+      ignore,
+      include,
+      watch,
+      clean,
+    })
```
KATT, line 41: In `server/dev.ts` we have this:

https://github.com/blitz-js/blitz/blob/75b63ed087265db268b2c0f9025c2377ae270da4/packages/server/src/dev.ts#L21-L30

**🙋‍♂️  Two questions regarding this**

- What is the purpose of `readyForNextDev`?
- What is `manifest` used for and is it something I need to consider here?

packages/server/src/next-utils.ts
```diff
@@ -59,6 +59,7 @@ export async function nextStartDev(
   const transform = createOutputTransformer(manifest, devFolder).stream
   const {spawnCommand, spawnEnv, availablePort} = await createCommandAndPort(config, "dev")
 
+  console.log({spawnCommand, cwd, spawnEnv})
```
KATT, line 62:
```suggestion

```

KATT, line 62:
```suggestion
```

packages/server/src/create-blitz-app.ts
```diff
@@ -0,0 +1,63 @@
...
+  } else {
+    // prod
+    if (!(await alreadyBuilt(buildFolder))) {
+      debug("not built - building..")
+      await build(serverConfig)
+      debug('build done.')
+    } else {
+      debug("already built")
+    }
```
KATT, line 52: I kinda think this would be better in this case - less magic

```suggestion
    if (!(await alreadyBuilt(buildFolder))) {
      throw new Error('Blitz is not built - make sure to building it before running')
    }
```

packages/server/src/build-hash.ts
```diff
@@ -1,34 +1,52 @@
-import {hashElement} from "folder-hash"
-import {pathExists, readFile, writeFile} from "fs-extra"
+import {hashElement,HashElementNode} from "folder-hash"
+import {pathExists,readFile,writeFile} from "fs-extra"
 import {resolve} from "path"
 
-export async function getInputArtefactsHash() {
+const debug = require("debug")("blitz:build-hash")
+
+export async function getInputArtifacts() {
```
KATT, line 7: Artefact is UK-english, artifiact is US.. generally I guess we should default to US

examples/custom-server/server/index.ts
```diff
@@ -0,0 +1,27 @@
...
+const dev = NODE_ENV === "development"
+
+async function main() {
+  const { requestHandler } = await createBlitzApp({ dev })
+
+  createServer((req, res) => {
+    const parsedUrl = parse(req.url!, true)
+    const { pathname } = parsedUrl
+
+    if (pathname === "/hello") {
```
KATT, line 16: 🙋‍♂️ Question here

Would there be a way to generating/accessing blitz context for requests here? Would not be useful for http-methods like here as one would just write them as API-functions, but if one instead did a WebSocket-server it would be useful to authenticate and access blitz context within this.

KATT:
Changing this from a draft to proper PR - as the code is working and tests are added - in hope that it'll trigger some discussion around it.

flybayer:
@KATT thank you for this!! Sorry I haven't gotten to review it yet. I'm on vacation rest of this week, so could be next week before I review, but will try to as soon as I can.

KATT:
Don't apologise & please do take your time, there are no fires burning 😊

packages/server/src/create-blitz-app.ts
```diff
@@ -0,0 +1,62 @@
...
+  }
+  const dir = dev ? devFolder : buildFolder
+  const app = next({dev, dir})
+  const requestHandler = app.getRequestHandler()
+
+  await app.prepare()
+
+  return {
+    requestHandler,
```
Skn0tt, line 60: Should we return the whole ˋappˋ here? That would make the API somewhat more similar to Next.

KATT:
I'm on the verge of closing this PR, it's too much work to keep it up to date, there's breaking changes / version updates weekly that requires manual work and a long CI pipeline to wait for.

packages/server/src/create-blitz-app.ts
KATT: Easy too do. Is it useful?

flybayer:
Hey @KATT sorry, I feel bad about this. You can leave it sit, and I'll take it over the finish line asap.

examples/custom-server/server/index.ts
flybayer: Yes, one can do `const session = await getSessionContext(req, res)` here to get the session.

flybayer:
### Update

I just pushed a big refactor that changes `server.js` to work exactly like next.js but by importing `@blitzjs/server` instead of `next`. And then this also adds support so that `blitz start` will automatically use the custom server if it is present so you don't have to run the script yourself. And `blitz start --production` works as expected.

This removes ability for hot reloading of the custom server for now. We can add this, but since it's a bit complex let's roll with this for now.

This doesn't support `server.ts`, but we should be able to add this later with the work done in #1325.

Thank you so much @KATT for this, and again I'm so sorry I let this drag so long. Not one of my finer moments :(

KATT (commented):
Thanks for taking over. I don't mind the wait. Not sure if I agree that automagically detecting the custom server is the best way to go, but you're the boss. :)

> This removes ability for hot reloading of the custom server for now. We can add this, but since it's a bit complex let's roll with this for now.

This, unfortunately, makes it very hard to work with, the way I approached it was (the same I've done with next.js-projects before,) to use nodemon and invoke it. Not ideal as it's a bit slow, but it works as one would expect. I think the reference example should at least contain a `yarn dev` command that makes it auto restart on changes until there's a hot reload functionality even for custom server.

```js
"dev": "cross-env NODE_ENV=development nodemon --watch server --ext ts,js,tsx --exec \"blitz start\""
```

☝️  that would probably do?

---

Secondly, did you get rid of the TypeScript-support in the custom server or am I missing something? I don't want to write JavaScript 😄

flybayer:
Yeah good point that I can keep nodemon for running `blitz start`. I'll add that.

> Secondly, did you get rid of the TypeScript-support in the custom server or am I missing something? I don't want to write JavaScript 😄

Yeah, I said this in my last comment:

> This doesn't support server.ts, but we should be able to add this later with the work done in #1325.

One could still do the manual build for server.ts like you had, but for official TS support we need to also bundle it in case users import TS code into server.ts. Because in that case, a simple tsc build of just server.ts won't work.

Issue #1455, opened 2020-11-12
Error "Failed to execute 'btoa'" occur when setting multibyte string to session.create


kyken:
### What is the problem?

setting string containing multi-byte characters (e.g. Japanese) as a session.create argument will result in the following message: "Error Failed to execute 'btoa': The string to be encoded contains characters outside of the Latin1 range." error occurs.

### Steps to Reproduce

1. run `blitz new xxxxx`
2. modify the `session.create` in the auth module to include a multi-byte string.
3. following error occurs when signup or login

```
 Error  Failed to execute 'btoa': The string to be encoded contains characters outside of the Latin1 range.
```

### Versions

```
macOS Catalina | darwin-x64 | Node: v12.18.0

blitz: 0.25.0 (local)

  Package manager: yarn 
  System:
    OS: macOS 10.15.7
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 401.34 MB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 12.18.0 - ~/.nodebrew/current/bin/node
    Yarn: 1.22.10 - ~/.nodebrew/current/bin/yarn
    npm: 6.14.4 - ~/.nodebrew/current/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  npmPackages:
    @prisma/cli: 2.11.0 => 2.11.0 
    @prisma/client: 2.11.0 => 2.11.0 
    blitz: 0.25.0 => 0.25.0 
    react: 0.0.0-experimental-4ead6b530 => 0.0.0-experimental-4ead6b530 
    react-dom: 0.0.0-experimental-4ead6b530 => 0.0.0-experimental-4ead6b530 
    typescript: 4.0.5 => 4.0.5 
```

jamiedavenport:
Thanks for raising this issue. I believe the issue is at the `btoa` call sites in https://github.com/blitz-js/blitz/blob/canary/packages/server/src/supertokens.ts

We're using the https://www.npmjs.com/package/b64-lite library so the docs should be useful.

konradkalemba:
I think I should add `b64-lite` to `package.json` of core package since it hadn't been used there until this PR, shouldn't I?

PR #1527, opened 2020-11-24
Fix to allow for unicode chars in session data

konradkalemba:
Closes: #1455 

Originally used `atob` and `btoa` methods from [`b64-lite`](https://github.com/kevlened/b64-lite) package don't support unicode characters. Fortunately, within the same package there are alternative methods that allow for unicode (`fromBase64` and `toBase64`). This PR basically replaces the former methods with the latter ones.

As for decoding public data on browser-end, [native `atob` method doesn't support unicode chars either](https://stackoverflow.com/a/30106551/3902198). To fix this we could use code snippet provided in the linked SO answer, but since the package we use for base64 operations is very small I just used its `fromBase64` method. 

### Checklist

- [x] Tests added for changes
- [ ] <s>PR submitted to [blitzjs.com](https://github.com/blitz-js/blitzjs.com) for any user facing changes</s>

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -43,6 +43,7 @@
     "@blitzjs/config": "0.26.0",
     "@blitzjs/display": "0.26.0",
     "bad-behavior": "1.0.1",
+    "b64-lite": "1.4.0",
     "cookie-session": "1.4.0",
     "deepmerge": "4.2.2",
     "lodash": "^4.17.19",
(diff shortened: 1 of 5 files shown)


konradkalemba:
I think I should add `b64-lite` to `package.json` of core package since it hadn't been used there until this PR, shouldn't I?

flybayer:
@konradkalemba awesome, thanks! And yes, technically it should be in core package.json too

flybayer (approved):

konradkalemba:
@flybayer Done!

PR #1532, opened 2020-11-26
Fix tree shaking for imports from queries/mutations

Skn0tt:
Before, the RPC generator produced files like this:

```ts
// getReferer.ts

import {getIsomorphicEnhancedResolver} from '@blitzjs/core'
import * as resolverModule from 'app/_resolvers/queries/getReferer'
export * from 'app/_resolvers/queries/getReferer'
export default getIsomorphicEnhancedResolver(
  resolverModule,
  'app/_resolvers/queries/getReferer',
  'getReferer',
  'query',
  undefined,
  {
    warmApiEndpoints: false
  }
)
```

People would import that (generated) file to access methods exported in line 3 of it - but they wouldn't want the rest of it to become part of the bundle.
Yet, this is precisely what happened: Because the whole `resolverModule` is passed into `getIsomorphicEnhancedResolver` ([which actually _may_ use it](https://github.com/blitz-js/blitz/blob/4d54e8962185b0475c513b82ec8c6b9dce86fd72/packages/core/src/rpc.ts#L217)), it couldn't be shaken away.

This commit fixes this. It introduces a new variant of the generated resolver file:

```ts
// getReferer.client.ts
import {getIsomorphicEnhancedResolver} from '@blitzjs/core'
export * from 'app/_resolvers/queries/getReferer'
export default getIsomorphicEnhancedResolver(
  undefined,
  'app/_resolvers/queries/getReferer',
  'getReferer',
  'query',
  undefined,
  {
    warmApiEndpoints: false
  }
)
```

The `NormalModuleReplacementPlugin` makes sure that the frontend build is rewritten to the `.client.ts` one, while the server build gets access too the full resolver.

I could locally verify that this allows `getProducts` from the store example to be successfully eliminated from the bundle.
I will add an integration test for that in the next commit.

Closes: ??

### Checklist

- [ ] Tests added for changes

### DIFF
--- a/examples/store/app/products/queries/getProducts.ts
+++ b/examples/store/app/products/queries/getProducts.ts
@@ -25,6 +25,8 @@ export default async function getProducts(
     console.log("HTTP referer:", ctx.referer)
   }
 
+  console.log("this line should not be included in the frontend bundle")
+
   const products = await db.product.findMany({
     where,
     orderBy,
--- a/examples/store/assert-tree-shaking-works.js
+++ b/examples/store/assert-tree-shaking-works.js
@@ -1,0 +1,25 @@
+#!/usr/bin/env node
+
+const glob = require("glob")
+const fs = require("fs")
+
+glob("./.next/static/chunks/**/*", {nodir: true}, (err, matches) => {
+  if (err) {
+    throw err
+  }
+
+  const offenders = []
+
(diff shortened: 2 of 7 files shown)


flybayer:
Sweeeet, thanks Simon!!

Issue #1508, opened 2020-11-21
`.DS_Store` should be ignore by duplicate page detection


flybayer:
### What is the problem?

```
◤ /app/_resolvers/billing/mutations/startMapTestingSubscription.ts✕ Warning: You have created conflicting page routes:
- /app/account/pages/.DS_Store
- /app/testing/proctor/pages/.DS_Store
✖ Uh oh something broke
DuplicatePathError: Warning: You have created conflicting page routes:
    at transformFn (/Users/b/c/blitz/packages/server/src/stages/pages/index.ts:88:14)
    at processInput (/Users/b/c/blitz/packages/file-pipeline/src/transform.ts:92:5)
    at DestroyableTransform._transform (/Users/b/c/blitz/packages/file-pipeline/src/transform.ts:69:11)
    at DestroyableTransform.Transform._read (/Users/b/c/blitz/packages/file-pipeline/node_modules/readable-stream/lib/_stream_transform.js:177:10)
    at DestroyableTransform.Transform._write (/Users/b/c/blitz/packages/file-pipeline/node_modules/readable-stream/lib/_stream_transform.js:164:83)
    at doWrite (/Users/b/c/blitz/packages/file-pipeline/node_modules/readable-stream/lib/_stream_writable.js:409:139)
    at writeOrBuffer (/Users/b/c/blitz/packages/file-pipeline/node_modules/readable-stream/lib/_stream_writable.js:398:5)
    at DestroyableTransform.Writable.write (/Users/b/c/blitz/packages/file-pipeline/node_modules/readable-stream/lib/_stream_writable.js:307:11)
    at DestroyableTransform.ondata (/Users/b/c/blitz/packages/file-pipeline/node_modules/readable-stream/lib/_stream_readable.js:681:20)
    at DestroyableTransform.emit (events.js:315:20) {
  pathType: 'pages',
  paths: [
    [
      '/Users/b/c/ahts/homeschool-boss/app/account/pages/.DS_Store',
      '/Users/b/c/ahts/homeschool-boss/app/testing/proctor/pages/.DS_Store'
    ]
  ]
}
```

### Versions

```
    "blitz": "0.25.1-canary.1",
```

### Other

Here's the code for this: https://github.com/blitz-js/blitz/blob/canary/packages/server/src/stages/pages/index.ts#L87

piotrski:
[...]

PR #1534, opened 2020-11-27
Fix blitz compiler to ignore gitignored files during processing

tpatel:
Closes: #1508

The change makes the package `file-pipeline` ignore the `.gitignore` patterns passed in the `config.ignore` configuration.

This changes requires another dependency to be added: `micromatch`. This library is already included by `fast-glob` so it shouldn't increase the bundle size.

I've added tests to make sure the ignored patterns are respected.

The "weak point" in this PR is that I've used an easy way to convert `.gitignore` patterns into glob patterns (supported by fast-glob/micromatch) by:

- prepending `**/` to gitignore patterns
- both prepending `**/` and appending `/**/*` to gitignore patterns

According to the tests it works on several patterns from `.gitignore` so we should be good.

### Checklist

- [x] Tests added for changes

### DIFF
--- a/packages/file-pipeline/package.json
+++ b/packages/file-pipeline/package.json
@@ -37,6 +37,7 @@
     "fs-extra": "9.0.0",
     "gulp-if": "3.0.0",
     "merge-stream": "2.0.0",
+    "micromatch": "4.0.2",
     "ora": "4.0.4",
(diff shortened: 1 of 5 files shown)


flybayer:
Saweeet, thank you!!

@all-contributors add @tpatel for code

PR #1547, opened 2020-11-27
Fix regression in 0.26.1 if you have a `test` folder inside `app/` (patch)

flybayer:

Fix regression in 0.26.1 that didn't work if you have a `test` folder inside `app/`


flybayer:
@tpatel fyi, adding the `**` before and after the ignore patterns broke some stuff. So removed that and instead added the `**` glob as necessary in the server config.

tpatel:
@flybayer damn good catch!
There might be an issue if the patterns are loaded directly from the `.gitignore` file then 🤔
I'll give it a second look tomorrow 👀

PR #1552, opened 2020-11-28
Fix message in the `blitz generate pages` edit page

jonstuebe:

Modified the alert text that is triggered when a resource edit throws an error. Before it was displaying the same value as the "Creation" of the resource.

### Checklist

- [x] Tests added for changes *n/a*
- [x] PR submitted to [blitzjs.com](https://github.com/blitz-js/blitzjs.com) for any user facing changes *n/a*


flybayer:
Sweet, thank you!!

@all-contributors add @jonstuebe for code

Issue #1556, opened 2020-11-29
Blitz install recipe not working


johnslemmer:
### What is the problem?

I'm getting an error when I try to install a recipe.  I've tried material and tailwind.

```
    Error: Cannot find module '@blitzjs/config'
    Require stack:
    - /private/var/folders/8g/3pzgl0_j4fd6ks3tb7cn9bf51wfvxy/T/blitz-recipe-blitz-js-blitz/node_modules/@blitzjs/display/dist/display.cjs.development.js
    - /private/var/folders/8g/3pzgl0_j4fd6ks3tb7cn9bf51wfvxy/T/blitz-recipe-blitz-js-blitz/node_modules/@blitzjs/display/dist/index.js
    - /private/var/folders/8g/3pzgl0_j4fd6ks3tb7cn9bf51wfvxy/T/blitz-recipe-blitz-js-blitz/node_modules/@blitzjs/installer/dist/installer.cjs.development.js
    - /private/var/folders/8g/3pzgl0_j4fd6ks3tb7cn9bf51wfvxy/T/blitz-recipe-blitz-js-blitz/node_modules/@blitzjs/installer/dist/index.js
    - /private/var/folders/8g/3pzgl0_j4fd6ks3tb7cn9bf51wfvxy/T/blitz-recipe-blitz-js-blitz/index.ts
    - /Users/jslemmer/Code/blitz/tail/node_modules/@blitzjs/cli/lib/src/commands/install.js
    - /Users/jslemmer/Code/blitz/tail/node_modules/@blitzjs/cli/node_modules/@oclif/command/node_modules/@oclif/config/lib/plugin.js
    - /Users/jslemmer/Code/blitz/tail/node_modules/@blitzjs/cli/node_modules/@oclif/command/node_modules/@oclif/config/lib/config.js
    - /Users/jslemmer/Code/blitz/tail/node_modules/@blitzjs/cli/node_modules/@oclif/command/node_modules/@oclif/config/lib/index.js
    - /Users/jslemmer/Code/blitz/tail/node_modules/@blitzjs/cli/node_modules/@oclif/command/lib/command.js
    - /Users/jslemmer/Code/blitz/tail/node_modules/@blitzjs/cli/node_modules/@oclif/command/lib/index.js
    - /Users/jslemmer/Code/blitz/tail/node_modules/@blitzjs/cli/lib/src/index.js
    - /Users/jslemmer/.config/yarn/global/node_modules/blitz/dist/cli.js
[...]

PR #1561, opened 2020-11-30
Add missing dependencies to `config` and `display` packages

konradkalemba:
Closes: #1556 

This PR adds missing required dependencies for `@blitzjs/config` and `@blitzjs/display` packages. Therefore fixes the problem with modules not found when installing recipes.

### Checklist

- [ ] <s>Tests added for changes</s>
- [ ] <s>PR submitted to [blitzjs.com](https://github.com/blitz-js/blitzjs.com) for any user facing changes</s>

### DIFF
--- a/packages/config/package.json
+++ b/packages/config/package.json
@@ -31,5 +31,9 @@
     "type": "git",
     "url": "https://github.com/blitz-js/blitz"
   },
+  "dependencies": {
+    "next": "10.0.2",
+    "pkg-dir": "4.2.0"
+  },
   "gitHead": "d3b9fce0bdd251c2b1890793b0aa1cd77c1c0922"
 }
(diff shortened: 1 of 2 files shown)


aaronshaf:
I assume this will fix `Error: Cannot find module '@blitzjs/config'`

konradkalemba:
@aaronshaf Yes, it does fix that. 

However I'm not really sure why building packages fail during CI processes. :/ They're building correctly on my local machine.

flybayer:
Hey @konradkalemba thank you for attempting this! In an effort to get this fixed asap, I made the PR which you must have seen. Issue was specifically with the installer package, but I see there's another issue with the next dependency. I'm going to make a PR shortly to fix that in config package with dynamic require.

konradkalemba:
@flybayer Yeah, I saw both of your PRs. I didn't know it could be fixed like that!

Issue #1511, opened 2020-11-22
Change passport adapter to accept array of strategy objects with "authenticateOptions" for each strategy


markylaing:
### What do you want and why?
`passportAuth` accepts an object with `authenticateOptions` and an array of strategies. The same authenticate options are applied regardless of strategy. This is obviously fine if you only have one strategy, but will not work if there are multiple strategies that require different options.

I came across this issue while writing an app that needs some permissions from either google or facebook (user choice). Code with comments below.

### Possible implementation(s)
In terms of types you might have something like:
```typescript
export declare type BlitzPassportConfig = {
    successRedirectUrl?: string;
    errorRedirectUrl?: string;
    strategies: Required<BlitzPassportStrategy>[];
    secureProxy?: boolean;
};
export declare type BlitzPassportStrategy = {
    authenticateOptions?: AuthenticateOptions;
    strategy: Strategy
}
```
A simpler option might be to have a map of authentication options who's keys are the strategy names.
```typescript
export declare type BlitzPassportConfig = {
    successRedirectUrl?: string;
    errorRedirectUrl?: string;
    authenticateOptions?: {[key:string]:AuthenticateOptions};
    strategies: Required<Strategy>[];
    secureProxy?: boolean;
};
```
and change [this line](https://github.com/blitz-js/blitz/blob/canary/packages/core/src/passport-adapter.ts#L70) to read:
```typescript
connectMiddleware(passport.authenticate(strategy.name, {...config.authenticateOptions[strategy.name]})),
```
### Additional context
In the below example, even though the google strategy options specify scopes as `["google","scopes"]`, `passport.authenticate` will be called with the facebook scopes.
```typescript
import { passportAuth } from "blitz"
import db from "db"
import { Strategy as GoogleStrategy } from "passport-google-oauth20"
import { Strategy as FacebookStrategy } from "passport-facebook"

export default passportAuth({
  successRedirectUrl: "/",
  errorRedirectUrl: "/",
  authenticateOptions: {scope: ["facebook","scopes"]},
  strategies: [
    new GoogleStrategy(
      {
        // Client info
        scope: ["google", "scopes"],
      },
      async function (accessToken, refreshToken, params, profile, done) {
        // Create user and call done
        done(undefined, { })
      }
    ),
    new FacebookStrategy(
      {
        // Client info
        // Note that scopes can't be passed in to the facebook strategy and must be
        // defined in authenticateOptions
        profileFields: ["name","email"], 
[...]

PR #1564, opened 2020-12-01
Change passport adapter to accept an array of objects with authenticate options for each strategy

markylaing:
Closes: https://github.com/blitz-js/blitz/issues/1511

Changes the authentication configuration which is passed into the passport adapter so that each strategy can have independent authentication options.

This means that each strategy can request different scopes.
### Checklist

- [x] Tests added for changes
- [x] PR submitted to [blitzjs.com](https://github.com/blitz-js/blitzjs.com) for any user facing changes (https://github.com/blitz-js/blitzjs.com/pull/276).


packages/core/src/passport-adapter.ts
```diff
@@ -51,13 +51,18 @@ export function passportAuth(config: BlitzPassportConfig) {
       "No Passport strategies found! Please add at least one strategy.",
     )
 
-    const strategy = config.strategies.find((strategy) => strategy.name === req.query.auth[0])
-    assert(strategy, `A passport strategy was not found for: ${req.query.auth[0]}`)
+    const blitzStrategy = config.strategies.find(
+      (strategy) => strategy.strategy.name === req.query.auth[0],
```
jamiedavenport, line 55: `strategy.strategy.name` looks a bit strange to me.

Think something like this would be a bit cleaner?
```suggestion
      ({strategy}) => strategy.name === req.query.auth[0],
```

packages/core/src/passport-adapter.ts
```diff
@@ -51,13 +51,18 @@ export function passportAuth(config: BlitzPassportConfig) {
...
+    )
+    assert(blitzStrategy, `A passport strategy was not found for: ${req.query.auth[0]}`)
+
+    const {strategy, authenticateOptions} = blitzStrategy
 
     passport.use(strategy)
+    const strategyName = strategy.name as string
```
jamiedavenport, line 62: Do you need the type assertion here?

markylaing: L75 & L84 `passport.authenticate(...)` can't be called with the `strategy.name` as it is a `string | undefined` (optional).

Happy to pass in `strategy.name as string` to those functions instead if you prefer?

packages/core/src/passport-adapter.ts
markylaing: I've made the suggested change.

flybayer:
Awesome, thank you @markylaing!

@all-contributors add @markylaing for code, doc

PR #1589, opened 2020-12-10
fix: typo? Prepped -> Prepared

beeplin:
This maybe trivial. I googled and found 'prepared' can also be written as 'prepped', but since the previous prompt says 'preparing', it might be better to use 'prepared' here.


flybayer:
Hey @beeplin, thanks for your contribution! I actually just merged a PR today that totally changes this, so this is no longer applicable :)

@all-contributors add @beeplin for code

beeplin:
@flybayer sure, waiting for the new release~!

Issue #1586, opened 2020-12-08
blitz db migrate error: "save" doesn't exist in the new version of Prisma Migrate v2.13.0


robdrosenberg:
# Workaround

**Use Prisma 2.13 and call `prisma migrate` directly instead of `blitz db migrate`**

or

Switch back to Prisma 2.12
```
dependencies": {
    "@prisma/cli": "2.12.0",
    "@prisma/client": "2.12.0",
```

---

### What is the problem?
I upgraded to version 0.27 and after generating a model and running `blitz db migrate` I get the following error.
![image](https://user-images.githubusercontent.com/20813991/101527760-46e5f600-3943-11eb-90a1-c9c943a1aba3.png)
Here's the link mentioned in the error: https://www.prisma.io/docs/guides/prisma-guides/prisma-migrate-guides/add-prisma-migrate-to-a-project
### Steps to Reproduce

1. make sure blitz is on version 0.27 and prisma is version 2.13.0
2. run `blitz db migrate`

### Versions

```
macOS Catalina | darwin-x64 | Node: v14.7.0

blitz: 0.27.0 (global)
blitz: 0.27.0 (local)

  Package manager: yarn 
  System:
    OS: macOS 10.15.7
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 769.20 MB / 32.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 14.7.0 - ~/.nvm/versions/node/v14.7.0/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.8 - ~/.nvm/versions/node/v14.7.0/bin/npm
    Watchman: Not Found
  npmPackages:
    @prisma/cli: 2.13.0 => 2.13.0 
    @prisma/client: 2.13.0 => 2.13.0 
    blitz: 0.27.0 => 0.27.0 
    react: 0.0.0-experimental-4ead6b530 => 0.0.0-experimental-4ead6b530 
    react-dom: 0.0.0-experimental-4ead6b530 => 0.0.0-experimental-4ead6b530 
    typescript: 4.1.2 => 4.1.2 
```

### Other

Please include applicable logs and screenshots that show your problem.
package.json:
```
{
  "name": "mysite",
  "version": "1.0.0",
  "scripts": {
    "start": "blitz start",
    "studio": "blitz db studio",
    "build": "blitz build",
[...]

PR #1605, opened 2020-12-14
use new prisma migrate dev command

jackbravo:
Closes: #1586 

This just changes the migrate command to use the new prisma migrate `dev` command instead of `save` and `up`. But for production we should be using the new `deploy` command.

The current implementation checks the environment and if running in production it changes the actions. My opinion is that we should have different commands for dev and deploy. Maybe `migrate` and `migrate-deploy`. This facilitates testing deploy commands on your local environment.

Comments?

### Checklist

- [ ] Tests added for changes


flybayer:
Hey @jackbravo! Thanks for taking a stab at this. However, I'm thinking we should remove the blitz db command altogether:

I've opened an RFC for this: https://github.com/blitz-js/blitz/discussions/1611

jackbravo:
I agree completely with what you say there :-p. While working on this I was thinking the same, that it would be much better to just use prisma commands directly:

- they have clear documentation
- we avoid oversimplifying their commands leaving their options out, specially because their are also on a preview state so they are prone to changing at any given moment
- we have less code to maintain

flybayer:
@all-contributors add @jackbravo for code

Issue #904, opened 2020-08-18
Cancelling installer still outputs success message


dwightwatson:
### Steps to Reproduce

1. Start an installer, for example `blitz install chakra`.
2. When it outputs dependencies and asks you to press `ENTER` to install, press `control+C` (or otherwise cancel).
3. See the successful output:

```
🎉 The recipe for Chakra UI completed successfully! Its functionality is now fully configured in your Blitz app.
```

Cancelling a recipe before completion shouldn't display the success message - probably doesn't need any message at all.

### Versions

```
debug: blitzPkgPath: /Users/dwight/Sites/myapp/node_modules/blitz/dist/index.js
debug: cliPkgPath: /Users/dwight/Sites/myapp/node_modules/@blitzjs/cli/lib/src/index.js 

macOS Catalina | darwin-x64 | Node: v14.8.0

blitz: 0.17.0 (global)
blitz: 0.17.1-canary.5 (local)
```

### Other

```
+––––––––––––––––––––––––––+
⎪   Add npm dependencies   ⎪
+––––––––––––––––––––––––––+
Chakra requires some other dependencies like emotion to work

Adding some shiny new dependencies.
If this list looks good, press ENTER to install.

Dependencies to be installed:
   📦 @chakra-ui/core@latest
   📦 @emotion/core@latest
   📦 @emotion/styled@latest
   📦 emotion-theming@latest

Dev Dependencies to be installed:

🎉 The recipe for Chakra UI completed successfully! Its functionality is now fully configured in your Blitz app.
```

aem:
Good catch. I'll address this in the upcoming revamp of the UX for the recipe runner

flybayer:
@aem sounds good. You can assign yourself if you decide to take an issue :)

flybayer:
The problem is [this line of code](https://github.com/blitz-js/blitz/blob/canary/packages/installer/src/recipe-executor.tsx#L37) runs for both success and failure.

arjundubey-cr:
@flybayer I would like to work on this issue. Can you please assign it to me?

flybayer:
@arjundubey-cr yes! Sorry for the long delay!

PR #1609, opened 2020-12-15
solves success message error on cancelling installer

arjundubey-cr:
Closes: #904 

- created a boolean variable hasExecuted to check whether waitUntilExit() gets executed or not
- log the output message only if the variable hasExecuted is true

### Checklist

- [ ] Tests added for changes

### DIFF
--- a/packages/installer/src/recipe-executor.tsx
+++ b/packages/installer/src/recipe-executor.tsx
@@ -7,6 +7,8 @@ import * as NewFileExecutor from "./executors/new-file-executor"
 import {RecipeRenderer} from "./recipe-renderer"
 import {RecipeMeta} from "./types"
 
+let hasExecuted:boolean = true;
+
 type ExecutorConfig =
   | AddDependencyExecutor.Config
   | FileTransformExecutor.Config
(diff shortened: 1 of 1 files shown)


packages/installer/src/recipe-executor.tsx
```diff
@@ -31,11 +33,12 @@ export class RecipeExecutor<Options extends RecipeMeta> {
...
       return
     }
-
-    log.info(
-      `\n🎉 The recipe for ${this.options.name} completed successfully! Its functionality is now fully configured in your Blitz app.\n`,
-    )
+    if(hasExecuted){
+      log.info(
+        `\n🎉 The recipe for ${this.options.name} completed successfully! Its functionality is now fully configured in your Blitz app.\n`,
+    )}
```
flybayer, line 42: Hey @arjundubey-cr thank you!

This definitely solves the problem, but I think it'd be a bit cleaner to move the success log message up right after `await waitUntilExit()`. And then you don't need the new `hasExecuted` variable

arjundubey-cr: yes, it is a great idea indeed

PR #1613, opened 2020-12-16
Add missing exports for `LinkProps`, `ImageProps`, and `ErrorProps`

Kosai106:

In most of my projects I find myself needing to use `LinkProps` from `next/link`, but doing two imports for this feels weird to me.

I've also added `ImageProps` and `ErrorProps`.

```tsx
import { Link, Image } from "blitz"
import { LinkProps, ImageProps } from "next/link"
```

as compared to:
```tsx
import { Link, LinkProps, Image, ImageProps } from "blitz"
```

### Checklist
Not really sure if either are applicable here. Let me know and I'll update the PR accordingly.

- [ ] Tests added for changes

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -63,8 +63,7 @@ export {
 
 export {default as dynamic} from "next/dynamic"
 
-export {default as ErrorComponent} from "next/error"
-export {ErrorProps} from "next/error"
+export {default as ErrorComponent, ErrorProps} from "next/error"
 
 export {default as getConfig} from "next/config"
 
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -44,8 +44,7 @@ export {
 
 export {default as Head} from "next/head"
 
-export {default as Link} from "next/link"
-export {LinkProps} from "next/link"
+export {default as Link, LinkProps} from "next/link"
 
 export {default as Router} from "next/router"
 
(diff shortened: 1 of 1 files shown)


packages/core/src/index.ts
```diff
@@ -45,10 +45,12 @@ export {
 export {default as Head} from "next/head"
 
 export {default as Link} from "next/link"
+export {LinkProps} from "next/link"
```
flybayer, line 48:
```suggestion
export {default as Link, LinkProps} from "next/link"
```

packages/core/src/index.ts
```diff
@@ -45,10 +45,12 @@ export {
...
 
 export {default as Router} from "next/router"
 
 export {default as Image} from "next/image"
+export {ImageProps} from "next/image"
```
flybayer, line 53:
```suggestion
export {default as Image, ImageProps} from "next/image"
```

packages/core/src/index.ts
```diff
@@ -63,6 +65,7 @@ export {
 export {default as dynamic} from "next/dynamic"
 
 export {default as ErrorComponent} from "next/error"
+export {ErrorProps} from "next/error"
```
flybayer, line 68:
```suggestion
export {default as ErrorComponent, ErrorProps} from "next/error"
```

flybayer:
Sweeet, thank you @Kosai106!!

@all-contributors add @Kosai106 for code

packages/core/src/index.ts
Kosai106: Huh, I was going to do it like this but TS complained locally so I ended up not doing it.
It's much cleaner like this though, thanks for taking care of that. 👍

PR #1623, opened 2020-12-18
Remove implicit auto-build with `blitz start --production`. `blitz build` now a pre-requisite (major)

KATT:
### What do you want and why?

Hey! I think the auto-build when doing `yarn start` in prod is an anti-pattern.

I don't know the history of this feature but I think it just adds quite a lot of complexity which will increase over time. I already saw this myself when doing the PR #1492 and got unexpected side-effects when something is/isn't in the exclude-array. 

A start command should not implicitly also build the project, that seems wrong to me. 

Simply remove/simplify `alreadyBuilt` and throw an error instead of `build()` when project isn't built. It's the responsibility of the dev to run build when they deploy (this is defacto standard across most software afaik)

No automatic building of blitz in prod.

### Checklist

- [x] Tests added for changes


KATT:
<details>
<summary>Windows test seem flaky and I can't restart them</summary>


```
@blitzjs/file-pipeline: FAIL src/helpers/agnostic-source/agnostic-source.test.ts
@blitzjs/file-pipeline:   ● Console
@blitzjs/file-pipeline:     console.log src/helpers/agnostic-source/agnostic-source.test.ts:51
@blitzjs/file-pipeline:       Starting test: include a folder doesn't exist
@blitzjs/file-pipeline:     console.log src/helpers/agnostic-source/agnostic-source.test.ts:53
@blitzjs/file-pipeline:       expected [
@blitzjs/file-pipeline:         'D:\\a\\blitz\\blitz\\packages\\file-pipeline\\src\\helpers\\agnostic-source\\fixtures\\one',
@blitzjs/file-pipeline:         'D:\\a\\blitz\\blitz\\packages\\file-pipeline\\src\\helpers\\agnostic-source\\fixtures\\two'
@blitzjs/file-pipeline:       ]
@blitzjs/file-pipeline:     console.log src/helpers/agnostic-source/agnostic-source.test.ts:60
@blitzjs/file-pipeline:       started stream
@blitzjs/file-pipeline:   ● agnosticSource › include a folder that doesn't exist
@blitzjs/file-pipeline:     ENOENT: no such file or directory, lstat 'D:\a\blitz\blitz\packages\file-pipeline\src\helpers\agnostic-source\fixtures\three'
```

</details>

flybayer:
Thanks @KATT, I think you are right. This will certainly remove flakiness with production deployments. We originally did that because that's how Rails works, but in today's deployment environments most have separate build and runtime stages, so the DX improvement isn't even noticed.

flybayer (changes requested):
@KATT ok last thing needed for this is to open a PR to update our docs for this change

KATT:
Will do after xmas 🎄🎅

KATT:
Wouldn't it now be cleaner to have two separate commands?

Let's say

```bash
blitz dev # starts in dev mode
blitz start # starts in prod
```

Not to do in this PR, but might do a follow-up if you'd agree.

KATT:
I can't see the auto-build-feature being mention in the docs but I did a PR linked above

flybayer (approved):

flybayer:
> Wouldn't it now be cleaner to have two separate commands?
> 
> Let's say
> 
> ```shell
> blitz dev # starts in dev mode
> blitz start # starts in prod
> ```
> 
> Not to do in this PR, but might do a follow-up if you'd agree.

I don't personally like this, but feel free to create a GH discussion for this to see what others think

Issue #1567, opened 2020-12-01
react-query-utils/getQueryKey doesn't work when called from inside getStaticProps


jxe:
### What is the problem?

I'm trying to prefetch queries on the server, then ship the react-query cache to the client in a page prop, to reuse the cache on the client, as described [here](https://react-query.tanstack.com/docs/guides/ssr#integrating-with-nextjs). 

Except, this doesn't work:

```typescript
import queryFunction from 'api/queries/queryFunction'
export function getStaticProps(){
  console.log(getQueryKey(queryFunction, {id:1})) 
  // outputs an array with an empty string: [ '', { json: { id: 1 }, meta: undefined } ]
}
```

I'll need the key to be the same on the client and server, in order to reuse the cache. You can check out [this tweet](https://twitter.com/edelwax/status/1333747091525525504?s=20) and [this RFC](https://github.com/blitz-js/blitz/discussions/1530) for more on what I'm up to.

### Versions

```
macOS Big Sur | darwin-x64 | Node: v14.12.0

blitz: 0.18.0 (global)
blitz: 0.26.0 (local)

  Package manager: yarn 
  System:
    OS: macOS 11.1
    CPU: (4) x64 Intel(R) Core(TM) i5-8210Y CPU @ 1.60GHz
    Memory: 58.09 MB / 8.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.12.0 - /usr/local/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.8 - /usr/local/bin/npm
    Watchman: Not Found
  npmPackages:
    @prisma/cli: 2.12.1 => 2.12.1 
    @prisma/client: 2.12.1 => 2.12.1 
    blitz: 0.26.0 => 0.26.0 
    react: 0.0.0-experimental-4ead6b530 => 0.0.0-experimental-4ead6b530 
    react-dom: 0.0.0-experimental-4ead6b530 => 0.0.0-experimental-4ead6b530 
    typescript: 4.2.0-dev.20201112 => 4.2.0-dev.20201112 
```

MrLeebo:
[...]

PR #1624, opened 2020-12-18
fix `getQueryKey()` to work on  the server (patch)

jxe:
Closes: #1567 

This fix 
* (a) makes getQueryKey work on the server; 
* (b) makes useQuery work on the server if the query is already pre-cached, but the query won't run otherwise. 

(b) is relevant to #1530.


jxe:
Hmmm, looks like I broke things. I guess the problem is that react-query isn't suspending on the server side, and it should. I'm not sure if there's an API to make rq suspend on enabled:false or when idle. But I guess this could be done in the useQuery wrapper before it returns. But this means, I think, using the top-secret suspense data fetching api inside the useQuery wrapper? (That is, it should [throw a suspender](https://github.com/gaearon/suspense-experimental-github-demo/blob/master/src/api.js)?)

jxe:
I can't tell why that windows CI build failed.

My rather hacky second commit makes this work for my own purposes. There are some problems with it, that I don't intend to fix:

1. react-query's 'isLoading' should perhaps be set to true on the server. But with my approach the query isIdle on the server, but not isLoading.
2. I didn't mod blitz's useInfiniteQuery and usePagedQuery, just useQuery
3. It's kind of ugly
4. My dirty little implementation of placeholderData doesn't work the same as the implementation in react query 3

However, it lets me do what I wanted: write isomorphic useQueries on the server and client and pass the queryCache from server to client. I'm happy to live on my own little branch for a while until this gets implemented in a better way in blitz proper.

flybayer:
Hey thanks! The windows example CI is currently busted. Not sure what's the deal with it.

Could you do the queryKey fix separately? Then we can go ahead and ship that. Have to think about the placeholder thing. Might be good to wait till RQ v3 upgrade, maybe some things will work better after that.

Btw, in the meantime you can use https://github.com/ds300/patch-package to patch the blitz source code in your app to get what you want.

flybayer:
@jxe are you interested in reworking this PR to jest fix the getQueryKey issue?

> Ok yeah we can rework that a bit so that `emptyQueryFn` is only returned for use by useQuery hooks but not by `getQueryKey`

jxe:
OK, I redid it, without the placeholder stuff, as you asked. I still didn't patch usePaginatedQuery and useInfiniteQuery, so this is incomplete. But perhaps you can merge it, and someday follow my lead with the other queries, and also rewrite your routerIsReady stuff to work like what I did—using enable:false rather than a fake queryFn. I think fake queryFns are a bad idea.

flybayer:
Hey @jxe, this still has changes to the query hooks which is what I was asking you to remove.

We can't use `enabled:false` because that also disable suspense. The router is ready fake query fns are the only known solution to that.

jxe:
@flybayer oh I guess I still don't understand. 

- If I just change getQueryKey, server side useQuery will execute during server rendering. So something inside useQuery must disable them. 
- I disabled them this way (rather than with a fake queryFn) because I do prefetching in my SSG. If the queries have been prefetched, the SSG useQuery should return data. By setting enabled:false, I get it to return data in this case, but not run otherwise.

Also, re your second point, it seems to me that using a fake queryFn is equivalent to what I've done (`throw new Promise(() => {})`). Is that not true?

flybayer:
@jxe fyi you need to merge latest canary into this branch which has some changes here because `useSession` now uses suspense by default.

I agree that this change adds the functionality removed from `sanitize`:
```
const routerIsReady = useRouter().isReady || isServer
```

But then you don't need the `enabled` override or the manual `throw promise` because react-query will throw the promise on the server.

flybayer:
> I disabled them this way (rather than with a fake queryFn) because I do prefetching in my SSG. If the queries have been prefetched, the SSG useQuery should return data. By setting enabled:false, I get it to return data in this case, but not run otherwise.

Fixing useQuery to work with SSG is a separate issue. I'm definitely interesting in fixing this, but I'd rather take that up in a separate PR because trying to fix both at once is making it confusing :D

jxe:
@flybayer I think to disable on the server that line should be `const routerIsReady = useRouter().isReady && isClient`

Sure, I can make this PR just do that and send another one that lets useQuery work (if it has precached data) during SSG.

flybayer:
Awesome, thank you!

jxe:
OK, @flybayer here's PR #1, with just the getQueryKey fix

flybayer:
Sweet, thank you!!

@all-contributors add @jxe for code

PR #1628, opened 2020-12-19
Remove `pretty-ms` dependency (move code into core) (patch)

KATT:
Discussion: #1626 

Remove `pretty-ms` dep. Most handlers should hopefully take less than a second so it shouldn't have much impact.


packages/core/src/invoke.ts
```diff
@@ -60,8 +59,8 @@ export async function invokeWithMiddleware<TInput, TResult>(
 
       const result = await enhancedResolver(params, res.blitzCtx)
 
-      const duration = prettyMs(new Date().getTime() - startTime)
-      log.info(chalk.dim("Finished", "in", duration))
+      const duration = new Date().getTime() - startTime
+      log.info(chalk.dim(`Finished in ${duration}ms`))
```
flybayer, line 63: To remove pretty-ms, we need to replicate the full behavior. For example pretty-ms will convert to seconds if greater than 1000 ms.

flybayer (changes requested):

packages/core/src/invoke.ts
KATT: I'd say if you have handlers taking more than 1000 ms I'd say you have bigger problems

KATT: Seems like an irrelevant optimization for me, but I can do it

KATT: I reluctantly did it

flybayer (approved):
Thanks :)

PR #1629, opened 2020-12-19
Remove unused deps

KATT:

Remove some unused dependencies


KATT:
Did this a bit to hastly - will do package by package instead which is less error-prone

Issue #1573, opened 2020-12-03
Windows CI apps test never finishes


flybayer:
### What is the problem?

Our CI `Build & Test Examples (windows-latest)` test stopped working last week I think. It just never finishes and eventually times out. I have no idea what might have changed or what the issue is.

### Steps to Reproduce

1. Open a PR with some change
2. Watch CI task `Build & Test Examples (windows-latest)` never finish

PR #1493, opened 2020-11-20
Fix File not found with singular glob

dmzza:
When reading `.gitignore` files, especially the global gitignore file, there may be
inclusion rules that refer to paths which don't actually exist in your project.

Closes: #1413 

### What are the changes and their implications?
This PR supersedes #1457 and fixes #1413 which crashes the file pipeline and prevents `blitz start` from working if your global gitignore contains an inclusion rule like this:

```
!TAGS/
```
and your Blitz project doesn't contain a `TAGS/` folder.

Allowing some included paths to not exist (using the `allowEmpty: true` option in vinyl-fs) fixes this problem. Although if there really are missing paths that should exist, vinyl-fs will no longer produce an error to warn you. 

### Checklist

- [x] Tests added for changes
~~PR submitted to [blitzjs.com](https://github.com/blitz-js/blitzjs.com) for any user facing changes~~

PR #1635, opened 2020-12-20
Skip flaky windows packages test

KATT:
Tests that _sometimes_ fail are a pain in the butt.

- Attempted to fix the flakiness of the windows packages test. Really just updated related deps to the faulty code `chokidar` & `vinyl` (didn't help)
- Ended up skipping the test

### Notes

I have **not** fixed the broken windows examples test #1573 . No idea what's going on there, realised that was broken whilst doing this PR.

<details><summary>The culprit seems to this test that _sometimes_ fails</summary>

https://github.com/blitz-js/blitz/blob/d9c529943630de96b394d86b2ae5dcaedce787db/packages/file-pipeline/src/helpers/agnostic-source/agnostic-source.test.ts#L50-L72
</details>

It was added in #1493 although it just showcased something underlying that was there before. 

A real hell to test as I'm not on Windows, so it's very slow trial-and-error. 😬 

Anyway, updating packages _seems_ to work, but I wouldn't bet on it _staying_ fixed. Possibly we should remove, rewrite, or `skip()` the test.


flybayer:
Thank you @KATT, this was definitely needed.

flybayer:
Thank you @KATT, this was definitely needed.

PR #1641, opened 2020-12-22
Disable windows example testing

KATT:

Disable windows example testing until #1573/#1639 is fixed


flybayer:
fixed!

Issue #1579, opened 2020-12-06
blitz/cli should throw more specific errors


Zeko369:
# Example

![image](https://user-images.githubusercontent.com/3064377/101269873-852cbb00-3773-11eb-87b2-38b95d96964f.png)

When running any `db` command `runPrisma` is called and if `getPrismaBin` errors out it just throws "Prisma not found". This functions calls `require(@blitzjs/server)` meaning that if it's broken it's still going to throw "Prisma not broken".

# Fix implementation

throw specific error in getPrismBin function and if it's not that, log the thrown error

flybayer:
Here's the code where this needs changed: https://github.com/blitz-js/blitz/blob/canary/packages/cli/src/commands/db.ts#L13

Probably this should log the actual error instead of this custom error.

arjundubey-cr:
@flybayer I would like to work on this issue.

PR #1648, opened 2020-12-25
Fix `blitz db` commands showing incorrect error (patch)

arjundubey-cr:
Closes: #1579 

updated db.ts to provide more specific error.
### Checklist

- [ ] Tests added for changes


packages/cli/src/commands/db.ts
```diff
@@ -9,10 +9,8 @@ const runPrisma = async (args: string[], silent = false) => {
   if (!prismaBin) {
     try {
       prismaBin = await getPrismaBin()
-    } catch {
-      throw new Error(
-        "Oops, we can't find Prisma Client. Please make sure it's installed in your project",
-      )
+    } catch (err) {
+      log.error(err)
```
KATT, line 13: you still probably wanna throw an error

arjundubey-cr: Ok I'll do that

arjundubey-cr: I've updated the code

packages/cli/src/commands/db.ts
```diff
@@ -9,7 +9,8 @@ const runPrisma = async (args: string[], silent = false) => {
   if (!prismaBin) {
     try {
       prismaBin = await getPrismaBin()
-    } catch {
+    } catch (err) {
+      log.error(err)
       throw new Error(
         "Oops, we can't find Prisma Client. Please make sure it's installed in your project",
       )
```
flybayer, line 16: Thanks! An even better solution is just to throw the original error which fixes the original issue. And this code is going to be removed soon anyways, so not too big of a deal.

```suggestion
      throw err
```

flybayer (approved):

PR #1661, opened 2020-12-31
Prisma 2.13+ Support: Remove all `blitz db` commands except `blitz db seed` and use prisma CLI directly (major)

jackbravo:
Closes:  #1586

- This deletes most migrate commands
- Renames `blitz migrate seed` to `blitz seed`
- Changes the `new` command to use the new prisma migrate dev command instead of our blitzjs migrate command

It is not modifying documentation for most of this changes. For example, for production we should be using the new prisma migrate deploy command.

This change also removes the use of two files for environment variables: `.env` and `.env.local`, where `.env` was encourged to be added to source control systems because it didn't contain the database URL. Instead it now uses a single `.env` file that is ignored from source control systems (it is added to the generated .gitignore file).

This changes are being discussed in RFC #1611.

### Checklist

- [x] Tests added for changes
- [x] PR submitted to [blitzjs.com](https://github.com/blitz-js/blitzjs.com) for any user facing changes: https://github.com/blitz-js/blitzjs.com/pull/321


flybayer:
Thank you @jackbravo!!

- [ ] Let's not make any changes to `.env` and `.env.local` in this PR and discuss that separately if you'd like that to be changed
- [ ] Let's keep `blitz db seed` working as `blitz db seed` instead of renaming to `blitz seed`

flybayer:
Also need to remove `/@prisma[\\/]client/,` from `with-blitz.ts` which should allow importing prisma enums into the client bundle.

packages/cli/src/commands/new.ts
```diff
@@ -156,9 +156,22 @@ export class New extends Command {
...
+                ".env.local",
+                "prisma",
+                "migrate",
+                "dev",
+                "--preview-feature",
+                "--name",
+                "Initial Migration",
+              ],
+              {stdio: "inherit"},
+            )
```
flybayer, line 174: This replacement code for the env loading doesn't have feature parity (dotenv is limited compared to dotenv-flow).

So keep the `require("dotenv-expand")(require("dotenv-flow").config({silent: true}))` line and then use spawn to just call prisma.
```
spawn.sync('npx', ["prisma", ...], {stdio: "inherit"})
```

flybayer:
Update on prisma 2.13 plan: Current plan is to add a `blitz prisma` command that will proxy everything to the prisma CLI as a workaround until they add proper support for env var loading.

flybayer:
Hey @jackbravo, I'd like to get this merged asap, so do you know when you might can finish? If it'll be awhile maybe I can take over for you :)

flybayer:
I have added the `blitz prisma` command in https://github.com/blitz-js/blitz/pull/1685

flybayer:
If I don't hear back from you by Monday, then I'll go ahead and finish it. Hopefully that's ok :)

jackbravo:
> If I don't hear back from you by Monday, then I'll go ahead and finish it. Hopefully that's ok :)

that's ok 👍 . I've been really busy. Sorry for not being able to advance on this!

flybayer:
Ok np, I'm working on this now

flybayer:
@all-contributors add @jackbravo for code

PR #1666, opened 2021-01-01
[RFC] Change `db/index.ts` to be side-effect free and all uses of `db.something` to `db().something`

flybayer:

**This is a proposal for a semi-breaking change to the `db/index.ts` file.**

Until now, the `db/index.ts` file has instantiated PrismaClient at the top level as a side effect. We started with this because it's quite elegant to just call `db.user.findMany()` instead of `db().user.findMany()`.

The primary motivation to change this to be side-effect free is so that we can import Prisma enums from `db`. Once we finish [Prisma 2.13](https://github.com/blitz-js/blitz/pull/1661) support, you will be able to do `import {RoleEnum} from '@prisma/client'` but you can't do `import {RoleEnum} from 'db'` because it's not side effect free.

The other motivation is that last year I learned that side effects in your JS files are very bad. You almost always run into something at some point that breaks because of it.

**What the change would entail:**

- New apps would have `db/index.ts` as side effect free
- `blitz generate` would generate code that has `db()`
- Existing apps would continue to work because in Blitz we can support both `db` being PrismaClient and it being a function that returns PrismaClient.
- We could create a little codemod that would update usage of `db` to `db()` throughout your app

What do you all think?

### Checklist

- [ ] Tests added for changes

### DIFF
--- a/examples/auth/db/index.ts
+++ b/examples/auth/db/index.ts
@@ -1,15 +1,7 @@
 import {PrismaClient} from "@prisma/client"
 export * from "@prisma/client"
 
-let prisma: PrismaClient
-
-if (process.env.NODE_ENV === "production") {
-  prisma = new PrismaClient()
-} else {
-  // Ensure the prisma instance is re-used during hot-reloading
-  // Otherwise, a new client will be created on every reload
-  globalThis["prisma"] = globalThis["prisma"] || new PrismaClient()
-  prisma = globalThis["prisma"]
+export default function db(): PrismaClient {
+  globalThis.prisma = globalThis.prisma || new PrismaClient()
+  return globalThis.prisma
 }
-
-export default prisma
--- a/examples/auth/app/auth/auth-utils.ts
+++ b/examples/auth/app/auth/auth-utils.ts
@@ -18,7 +18,7 @@ export const verifyPassword = async (hashedPassword: string, password: string) =
 }
 
 export const authenticateUser = async (email: string, password: string) => {
-  const user = await db.user.findFirst({where: {email}})
+  const user = await db().user.findFirst({where: {email}})
 
   if (!user || !user.hashedPassword) throw new AuthenticationError()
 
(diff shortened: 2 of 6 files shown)


examples/auth/db/index.ts
```diff
@@ -1,15 +1,7 @@
...
-let prisma: PrismaClient
-
-if (process.env.NODE_ENV === "production") {
-  prisma = new PrismaClient()
-} else {
-  // Ensure the prisma instance is re-used during hot-reloading
-  // Otherwise, a new client will be created on every reload
-  globalThis["prisma"] = globalThis["prisma"] || new PrismaClient()
-  prisma = globalThis["prisma"]
+export default function db(): PrismaClient {
```
Flavyoo, line 4: In order to allow for using the existing method to query the db, does it make sense to export the result of calling db?

flybayer: That's what we are doing currently and is a side-effect :D I.e. what we are trying to fix.

**But**, I just had an idea of using `Proxy` to make this side effect free but still use `db.model.etc`. Will experiment and report back.

Flavyoo: Thanks for clarifying, didn't know that was a side-effect.

flybayer:
### Update

So a Proxy would work for this, like so:

```ts
import {PrismaClient} from "@prisma/client"

const ProxyPrisma = new Proxy(PrismaClient, {
  construct(target, args) {
    if (typeof window !== "undefined") return {}
    globalThis["db"] = globalThis["db"] || new target(...args)
    return globalThis["db"]
  },
})

export * from "@prisma/client"
export default new ProxyPrisma()
```

Which means we could add an API to blitz similar to this:

```ts
import {ServerOnlyPrisma} from "blitz"
import {PrismaClient} from "@prisma/client"

export * from "@prisma/client"
export default new ServerOnlyPrisma(PrismaClient)
```

flybayer:
Closing in favor of https://github.com/blitz-js/blitz/pull/1720

Issue #461, opened 2020-05-06
Blitz console does not work in plain JS projects


tmns:
### What is the problem?

In a freshly generated JS project, attempting to access the `db` object within `blitz console` causes a `ReferenceError` to be thrown.  

This in turn blocks you from being able to interact with the db via the console, such as performing queries like `db.question.findMany().then(console.log)`, which is given as an example in the tutorial here: https://github.com/blitz-js/blitz/blob/canary/TUTORIAL.md#playing-with-the-api.

### Steps to Reproduce:

1. `blitz new myapp --js`
2. `cd myapp`
3. Follow the steps in either the default `index.js` page or the [relevant tutorial section](https://github.com/blitz-js/blitz/blob/canary/TUTORIAL.md#creating-models) for editing your model and migrating your DB
4. `blitz console`
5. Attempt to access the db object, like so:
```bash
$ blitz console
[..snip..]
◢ Loading your codeCouldn't find tsconfig.json. tsconfig-paths will be skipped
✔ Loading your code
⚡️ > db.question.findMany().then(console.log)
Thrown:
ReferenceError: db is not defined
```

Note there is also an error about `tsconfig.json` that is printed out next to the first `Loading your code`. I'm not sure if that is actually related to the reference error though.

### Versions:

```
Linux 4.13 | linux-x64 | Node: v12.3.1

blitz: 0.8.1 (global)
```

---

# Update from @flybayer 

The core issue is the node repl is unable to load es modules.

You can add an error log statement to [this line](https://github.com/blitz-js/blitz/blob/canary/packages/cli/src/utils/load-blitz.ts#L50) to see the following error for every file it's loading.

```
SyntaxError: Cannot use import statement outside a module
    at wrapSafe (internal/modules/cjs/loader.js:1070:16)
    at Module._compile (internal/modules/cjs/loader.js:1120:27)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1176:10)
    at Module.load (internal/modules/cjs/loader.js:1000:32)
    at Function.Module._load (internal/modules/cjs/loader.js:899:14)
    at Module.require (internal/modules/cjs/loader.js:1042:19)
    at require (internal/modules/cjs/helpers.js:77:18)
    at Object.exports.forceRequire (/Users/b/c/blitz/packages/cli/src/utils/module.ts:7:10)
    at /Users/b/c/blitz/packages/cli/src/utils/load-blitz.ts:45:24
    at Array.map (<anonymous>)
[...]

PR #557, opened 2020-05-26
Fix Blitz Console in plain JS projects

enzoferey:
Closes: #461 

### What are the changes and their implications?

- Added `esm` dependency.
- Changed `require` to required function provided by `esm` dependency. I don't know the internals of `esm` but of course it may be slower to load the code into the console as it needs an extra step (but no other solution I would say 🤷‍♂️). 

### Breaking change: no

Issue #1233, opened 2020-10-02
Blitz console not loading queries/mutations


simonedelmann:
### What is the problem?

Blitz console does not find any queries/mutations. (Furthermore it does not find any files except those which are in `integrations` or `utils` folder.)

### Steps to Reproduce

1. New blitz app
2. Rename `auth-utils.ts` to `auth.ts` (to avoid side-effects because of kebab-case)
3. Move `auth.ts` to `/integrations` folder
4. Open blitz console and enter `auth`. => It works. 
5. Move `auth.ts` to another location and check if it can be found: 
```
// These works
integrations/auth.ts
utils/auth.ts

// These does not work
integration/auth.ts
app/auth.ts
app/auth/auth.ts
db/auth.ts
/auth.ts
```

The same is for all queries/mutations. There are not found until their files are in `integrations` or `utils`. So seems like the Module resolving is broken somehow. 

### Versions

```
0.23.2-canary.3
```

Alucard17:
Hey @flybayer can I work on this.

flybayer:
@Alucard17 yes!

Alucard17:
Hey @flybayer @simonedelmann as per the code mentioned below,
It expects the queries and mutation should be in app directory, is this behavior intentional or should I change it ?
https://github.com/blitz-js/blitz/blob/bd1063a965eeccfd091a199cdc5c3ab57d4832ba/packages/repl/src/utils/load-blitz.ts#L11

flybayer:
@Alucard17 yes, queries and mutations must be inside app.

flybayer:
@Alucard17 did you make any progress on this? Or should I mark it back for anyone else to work on?

flybayer:
fyi for anyone, related code is here: https://github.com/blitz-js/blitz/blob/canary/packages/repl/src/utils/load-blitz.ts

denvaar:
Is there a workaround for this while it's being worked on?

flybayer:
@denvaar should be able to manually require it inside the repl, something like this:

```
const getUsers = require('app/users/queries/getUsers').default
```

PR #1686, opened 2021-01-09
Fix `blitz console` not working in vanilla JS projects

Flavyoo:
Closes: #461 

This continues work from https://github.com/blitz-js/blitz/pull/557

- Added `esm` dependency.
- Changed `require` to required function provided by `esm` dependency. I don't know the internals of `esm` but of course it may be slower to load the code into the console as it needs an extra step (but no other solution I would say 🤷‍♂️). 

### Breaking change: no


Flavyoo:
@flybayer I have done the following validation.

1. Created a new app with a fresh installation of blitz and verified that I am able to access the `db` object and load data from the database.
2. Updated this branch and ran the plain-js example and was also able to load the `db` object.

For both cases, the running app errors out when going to `/projects` in the URL with the error `Cannot read property 'findMany' of undefined.` I thought this issue was only in the blitz console but it looks like it is in the running apps also. I believe this is the issue you were mentioning previously.

flybayer:
@Flavyoo ok very good. Did you run `blitz generate all projects name` and `blitz db migrate` before going to `/projects`?

Also, did you check if you can use queries/mutations in the console both ways? (This was the issue I remember)

Flavyoo:
Yes I did. I followed the steps in the index page.

However, I ran the steps again and it is working for the blitz app I created.

I validated these ways:

New App with Fresh Blitz Installation:
- db.project.findMany().then(console.log) # works
- await db.project.findMany() # works

Plain Js App from this PR:
- db.project.findMany().then(console.log) # works
- await db.project.findMany() #works

flybayer:
Ok good, so what I meant by testing queries/mutations is does this work?

```
// have to provide a mock ctx object as second param
await getProjects({}, {authorize: ()=>{}})
```

Flavyoo:
I see, I will try that

Flavyoo:
@flybayer Looks like I can't run `await getProjects({}, {authorize: ()=>{}})`. 
I get this error:

```
Uncaught ReferenceError: getProjects is not defined
    at REPL1:1:16
```

flybayer:
@Flavyoo ok, so query/mutation loading has been broken in published blitz for awhile: https://github.com/blitz-js/blitz/issues/1233

It was working in our repo until this PR though. So this PR actually makes it break in our mono repo same as production which is probably a good thing :)

PR #1709, opened 2021-01-13
(newapp) Fix studio package.json script to use new blitz prisma command

malkomalko:
Closes: n/a

I noticed the prisma studio package.json script was wrong after the latest updates.


flybayer:
Yes, thank you!!

Issue #1711, opened 2021-01-13
Logout mutation not updating currentUser across the app


gabrilator:
Hey! Having some of trouble with logout mutations since 0.28. 
### What is the problem?
The logout mutation doesn't update currentUser across all the app components
### Steps to Reproduce

1. blitz new name 
2. Extract the boilerplate logout button into a component, i.e LogoutButton
```
const LogoutButton = () => {
    const [logoutMutation] = useMutation(logout)
    return (
        <button   onClick={async () => {await logoutMutation()}}>
          Logout
        </button>
    )
}
```
3. Replace the default logout button with LogoutButton.
4. After logging in, try to logout with the new LogoutButton. The logout function is called, but the user information will still show (currentUser is not invalidated). 
When refreshing the page, currentUser will be null throughout the app and the user will be logged out (as expected).

I'm not sure if I'm missing something here.  Any idea why I'm getting this behaviour? Thanks!

 Thanks in advance!

flybayer:
Hmm, something's odd because that doesn't make sense. All query cache is deleted in the blitz framework when it receives the api response.

When it [gets the header noting session revoked](https://github.com/blitz-js/blitz/blob/canary/packages/core/src/rpc.ts#L86), it [clears all query cache](https://github.com/blitz-js/blitz/blob/canary/packages/core/src/public-data-store.ts#L38). 

Did you reproduce in a brand new app or only in your current app?

gabrilator:
Hey @flybayer, I reproduced this  in a brand new app.

gabrilator:
Hey! I realised the same thing happens with login and signup mutations! Do I open a new issue?

PR #1717, opened 2021-01-13
Fix logout not triggering re-renders for all queries (patch)

flybayer:
Closes: #1711 

Fix logout not triggering re-renders for all queries.

Root cause is a bug in react-query version that we are on. Added a temporary fix to remove once we upgrade to react-query v3.

### Checklist

- [ ] Tests added for changes


nksaraf:
Hi @flybayer, I saw that this fix is there till react query v3 is used. Are there any plans to do that before beta?

flybayer:
@nksaraf no, react-query v3 after beta (unless someone besides me does it!)

Issue #639, opened 2020-06-06
[Rollup help needed] Add new`@blitzjs/framework` pkg with separate client & server bundles


flybayer:
### What do you want and why?

Currently we have framework code scattered around various packages, mostly `@blitzjs/core` and `@blitzjs/server`. We want to consolidate this code into `@blitzjs/framework` and then remove `@blitzjs/core`.

We are currently using tsdx for these packages, but I don't think we can use it for `framework`. Because in addition to isomorphic code, `framework` will have server only code that should be bundled separately from the client code to ensure server code never ends up in a client bundle.

We also want deep imports so we can do something like this:

```ts
import something from '@blitzjs/framework/client'
import something from '@blitzjs/framework/server'
```

I'd love to have this sooner rather than later :)

Additional context: #306

PR #1727, opened 2021-01-16
Fix `blitz generate` bug when model name ends with number

koolii:
Closes: #1715

- Use util functions in generator instead of prolix functions in `packages/cli/src/commands/generate.ts`.
- Add unit tests
- Rename utils/plurals.ts to utils/inflector.ts
- Reduce utils/kebab-case.ts to utils/inflector.ts

### WHY

```
Steps to Reproduce
1.Run blitz generate all test002

CREATE    app/test002-s/pages/test002-s/[test002Id]/edit.tsx
CREATE    app/test002-s/pages/test002-s/[test002Id].tsx
CREATE    app/test002-s/pages/test002-s/index.tsx
CREATE    app/test002-s/pages/test002-s/new.tsx
CREATE    app/test002S/components/Test002Form.tsx
CREATE    app/test002S/queries/getTest002.ts
CREATE    app/test002S/queries/getTest002S.ts
CREATE    app/test002S/mutations/createTest002.ts
CREATE    app/test002S/mutations/deleteTest002.ts
CREATE    app/test002S/mutations/updateTest002.ts
```

#### test002 => test002S

I believe the problem is to transform call order.
it should be called from `pascalCase()`.

refs: https://github.com/blitz-js/blitz/pull/1727/files#diff-e7569fc2058f33e8e3349b21677c545ca06beec413fd06985d153079bc69b332L53-L64

#### test002 (=> test002S) => test002-s

it also has the reason above.
In addition, the PageGenerator transforms modelName(e.g. test002S) to kebab-case.
refs: https://github.com/blitz-js/blitz/blob/canary/packages/generator/src/generators/page-generator.ts#L58

### Checklist

- add plurals unit tests at the moment.
- use plural helpers imported from @blitzjs/generator during `blitz generate`

- [x] Tests added for changes


malkomalko (approved):
👍🏻 I'd maybe argue that we should consider renaming this file, as it's doing more than pluralization.  A common term for this is string inflection https://api.rubyonrails.org/classes/ActiveSupport/Inflector.html

koolii:
SGTM :)

malkomalko (approved):
❤️ Love

packages/cli/src/commands/generate.ts
```diff
@@ -7,14 +7,17 @@ import {
   FormGenerator,
   ModelGenerator,
   QueryGenerator,
+  singleCamel,
+  singlePascal,
+  pluralCamel,
+  pluralPascal,
```
malkomalko, line 13: Food for thought for later.  Maybe it would be nice to create a new package for some of these util based files that can then be shared across other packages.  Right now we're exporting from the generator, but there's no reason these should be tied to that project.  🤷🏻

flybayer: The goal is to reduce the number of packages with a better build setup: https://github.com/blitz-js/blitz/issues/639

flybayer:
Thank you!!

Issue #1732, opened 2021-01-17
Disable `react-query` suspense when using legacy React


anteprimorac:
### What do you want and why?
Currently switching `reactMode` experimental option to `legacy` requires going to every `useQuery` call and setting `suspense` to `false`.

### Possible implementation(s)
Make `defaultQueryConfig.suspense` dependable on `reactMode` experimental option.

malkomalko:
👋🏻 Ante.  An amazing community member posted this gist for helping with the above https://gist.github.com/Zeko369/a0bd1acbb6feeeee512294f9ab6d44e7.

❓ Do we want to think about swapping this in during the compilation phase as a default if we know that we're using legacy mode.

PR #1733, opened 2021-01-17
Disable react query suspense in legacy react mode

anteprimorac:
Closes: #1732 

Disable react query suspense when project opts-in legacy react mode.

### Checklist

- [x] Tests added for changes


packages/core/src/utils/react-query-utils.ts
```diff
@@ -146,7 +147,27 @@ export const retryFunction = (failureCount: number, error: any) => {
...
+  } else {
+    // On the client side load reactMode from the env variable
+    if (process.env.NEXT_PUBLIC_REACT_MODE) {
+      reactMode = process.env.NEXT_PUBLIC_REACT_MODE
+    }
+  }
+
+  return {
+    // Set default value for suspense option to true only when concurrent react mode is active
+    suspense: reactMode === "concurrent" ? true : undefined,
```
bacongravy, line 170: Is this correct? I think Suspense is also supported in "blocking" mode.  This check might be better written as `reactMode !== "legacy"`.

packages/core/src/utils/react-query-utils.ts
```diff
@@ -146,7 +147,27 @@ export const retryFunction = (failureCount: number, error: any) => {
   return false
 }
 
-export const defaultQueryConfig = {
-  suspense: true,
-  retry: retryFunction,
+export function getDefaultQueryConfig() {
```
bacongravy, line 150: This is probably a premature optimization, but I wonder if this function should be memo-ized.  I don't think its return value can change after it has been run once.

packages/server/src/with-blitz.ts
```diff
@@ -77,6 +79,10 @@ export function withBlitz(nextConfig: any) {
 
         return config
       },
+      env: {
+        ...(normalizedConfig.env || {}),
+        NEXT_PUBLIC_REACT_MODE: experimental.reactMode,
```
bacongravy, line 84: I don't think this needs to be prefixed with `NEXT_PUBLIC_`.  It could instead be `BLITZ_REACT_MODE`.  Environment variables set via the `env` key in config don't require the prefix; the prefix is only required for environment variables picked up from `.env.local`.

See https://nextjs.org/docs/api-reference/next.config.js/environment-variables and https://nextjs.org/docs/basic-features/environment-variables#exposing-environment-variables-to-the-browser

packages/core/src/utils/react-query-utils.ts
anteprimorac: `getConfig` is already memoized and reading from `process.env` should be quick

flybayer:
Thank you very much for the effort on this @anteprimorac! But as discussed in Slack it's better to go with the `__BLITZ_DATA__` data approach in https://github.com/blitz-js/blitz/pull/1735.

So going to go ahead and close this one.

PR #1735, opened 2021-01-17
Automatically disable `useQuery` suspense when not using concurrent mode

bacongravy:
Closes https://github.com/blitz-js/blitz/issues/1732

Currently the `defaultQueryConfig` for `useQuery` sets `{ suspense: true }`.  This is a good default for `suspense` because Blitz enables React `concurrent` mode by default, and the default app template uses `Suspense`.

However, if a user of Blitz changes the `reactMode` to `legacy` and removes `Suspense` from their app, the default no longer makes sense, and the user must manually override the config in each call to useQuery.

This change updates the `defaultQueryConfig` to depend on the `reactMode`.

Making this change required plumbing through the blitzConfig from the server through to the client by way of a new `__BLITZ_DATA__` mechanism modeled after `__NEXT_DATA__`.

Some implementation notes:

* Exporting `BlitzScript` from `core/src/index.ts` has the side-effect of setting the `window.__BLITZ_DATA__` attribute when running in a browser environment and must happen after the value has appeared in the HTML document but before using anything else depends on that value being set.
* The `BlitzScript` element embeds the serialized `__BLITZ_DATA__`  in the HTML document and must be rendered before the `NextScript` element. (When the `NextScript` element is rendered it loads the scripts for the app, including the code in `core/src/index.ts`, causing the side-effect described above.)

### Checklist

- [ ] Tests added for changes

### DIFF
--- a/packages/core/src/blitz-data.tsx
+++ b/packages/core/src/blitz-data.tsx
@@ -1,16 +1,28 @@
 import {getConfig} from "@blitzjs/config"
 import htmlescape from "htmlescape"
 import React from "react"
+import {isClient} from "./utils"
 
-// Automatically deserialize __BLITZ_DATA__ in a browser environment
-if (typeof window !== "undefined") {
-  if (document.getElementById("__BLITZ_DATA__")) {
-    deserializeAndSetBlitzDataOnWindow()
+export function _getBlitzRuntimeData() {
+  return {suspenseEnabled: getConfig().experimental?.reactMode !== "legacy"}
+}
+
+export function getBlitzRuntimeData() {
+  if (isClient) {
+    return window.__BLITZ_DATA__
+  } else {
+    if (!global.__BLITZ_DATA__) {
+      global.__BLITZ_DATA__ = _getBlitzRuntimeData()
+    }
+    return global.__BLITZ_DATA__
   }
 }
 
-export function getBlitzData() {
-  return {config: getConfig()}
+// Automatically deserialize __BLITZ_DATA__ in a browser environment
+if (isClient) {
+  if (document.getElementById("__BLITZ_DATA__")) {
+    deserializeAndSetBlitzDataOnWindow()
+  }
 }
 
 export function deserializeAndSetBlitzDataOnWindow() {
@@ -20,7 +32,7 @@ export function deserializeAndSetBlitzDataOnWindow() {
     )
     window.__BLITZ_DATA__ = data
   } catch (e) {
-    console.error(e)
+    console.error("Error deserializing __BLITZ__DATA__", e)
   }
 }
 
@@ -30,7 +42,7 @@ export function BlitzData() {
       id="__BLITZ_DATA__"
       type="application/json"
       dangerouslySetInnerHTML={{
-        __html: htmlescape(getBlitzData()),
+        __html: htmlescape(_getBlitzRuntimeData()),
       }}
     />
   )
--- a/packages/core/src/blitz-data.tsx
+++ b/packages/core/src/blitz-data.tsx
@@ -1,0 +1,37 @@
+import {getConfig} from "@blitzjs/config"
+import htmlescape from "htmlescape"
+import React from "react"
+
+// Automatically deserialize __BLITZ_DATA__ in a browser environment
+if (typeof window !== "undefined") {
+  if (document.getElementById("__BLITZ_DATA__")) {
+    deserializeAndSetBlitzDataOnWindow()
+  }
+}
+
+export function getBlitzData() {
+  return {config: getConfig()}
+}
+
+export function deserializeAndSetBlitzDataOnWindow() {
+  try {
+    const data: typeof window["__BLITZ_DATA__"] = JSON.parse(
+      document.getElementById("__BLITZ_DATA__")!.textContent!,
+    )
+    window.__BLITZ_DATA__ = data
+  } catch (e) {
+    console.error(e)
+  }
+}
+
+export function BlitzData() {
+  return (
+    <script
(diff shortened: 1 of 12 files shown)


bacongravy:
This can be manually tested by modifying the `auth` example.

1. In `examples/auth/blitz.config.js`: Add`reactMode: "legacy"` to the `experimental` record
1. In `examples/auth/app/pages/index.tsx`: Remove the use of `<Suspense>`
1. Run the app and confirm that it renders without errors

bacongravy:
I see that @anteprimorac beat me to it and created another, similar PR: https://github.com/blitz-js/blitz/pull/1733

Instead of plumbing the blitz.config.js contents through a `__BLITZ_DATA__` script element, that PR uses `process.env.__NEXT_REACT_MODE` to determine the current React mode.  That value is set in [Next.js's webpack config](https://github.com/vercel/next.js/blob/b02df3f487ba6fbd9ba9bcd823c77a7b5391fc73/packages/next/build/webpack-config.ts).

A similar approach could be taken for the rest of the contents of blitz.config.js, in Blitz's webpack config.

bacongravy:
I think the other PR is a better implementation of disabling suspense in legacy mode and recommend going with it instead of this one.

@flybayer Let me know if you're interested in this PR from the perspective of support for a `__BLITZ_DATA__` mechanism.  If it would be useful, I could remove the bits from this PR that impact suspense.  But, what data would you want sent inside of `__BLITZ_DATA__` in that case?

packages/core/src/blitz-data.tsx
```diff
@@ -0,0 +1,37 @@
...
+}
+
+export function deserializeAndSetBlitzDataOnWindow() {
+  try {
+    const data: typeof window["__BLITZ_DATA__"] = JSON.parse(
+      document.getElementById("__BLITZ_DATA__")!.textContent!,
+    )
+    window.__BLITZ_DATA__ = data
+  } catch (e) {
+    console.error(e)
```
flybayer, line 23: More specific error message
```suggestion
    console.error("Error deserializing __BLITZ__DATA__", e)
```

packages/core/src/config.ts
```diff
@@ -0,0 +1,7 @@
+export function getBlitzConfig() {
+  if (typeof window !== "undefined") {
+    return window.__BLITZ_DATA__?.config
+  } else {
+    return undefined
+  }
+}
```
flybayer, line 7: And remove this file because of above changes

flybayer (commented):
Awesome, nice work!!

bacongravy:
I pushed a commit with the suggested changes. I also changed `defaultQueryConfig` to be a function, `getDefaultQueryConfig`.  (Attempting to use `getConfig()` by way of `getBlitzRuntimeData()` from the top-level of `react-query-utils.ts` results in a failure at server-startup time.)

bacongravy:
I see there are some failing tests, I'll look into those.

bacongravy:
The tests are failing because the test environment looks like the client environment (window is defined), but since the BlitzData element hasn't been rendered by the test, there is no script element to retrieve the runtime data from.

I see a couple of options: either update each test that relies on the runtime data (so far, just anything that uses useQuery) to include a <BlitzData /> element, or, modify `getBlitzRuntimeData()` to detect that it is running in a test and behave like it does on the server.

packages/core/src/utils/react-query-utils.ts
```diff
@@ -146,7 +147,10 @@ export const retryFunction = (failureCount: number, error: any) => {
...
 
-export const defaultQueryConfig = {
-  suspense: true,
-  retry: retryFunction,
+export function getDefaultQueryConfig() {
+  const {suspenseEnabled} = getBlitzRuntimeData()
+  return {
+    suspense: !!suspenseEnabled,
+    retry: retryFunction,
+  }
```
flybayer, line 155: We should turn this into a hook and use `useMemo`. Dependency array will be `[]` because it will never change

flybayer:
You can add a check for `process.env.JEST_WORKER_ID === undefined` in `isClient` to detect if in a test environment or not.

packages/core/src/blitz-data.tsx
```diff
@@ -0,0 +1,49 @@
...
+import React from "react"
+import {isClient} from "./utils"
+
+export function _getBlitzRuntimeData() {
+  return {suspenseEnabled: getConfig().experimental?.reactMode !== "legacy"}
+}
+
+export function getBlitzRuntimeData() {
+  if (isClient) {
+    return window.__BLITZ_DATA__
```
bacongravy, line 12: I can get the tests passing again by changing this line to:

```
    if (!window.__BLITZ_DATA__) {
      window.__BLITZ_DATA__ = _getBlitzRuntimeData()
    }
    return window.__BLITZ_DATA__
```

But that would blow up if it got executed on the client.

bacongravy: (It _shouldn't_ get executed on the client, because __BLITZ_DATA__ should already be set on the client...)

bacongravy: Maybe it would be better to move this into the tests:

```
window.__BLITZ_DATA__ = _getBlitzRuntimeData()
```

bacongravy: I pushed a commit that updates the tests.

packages/core/src/utils/react-query-utils.ts
bacongravy: Done.

flybayer:
Sweeeeet!!!

@all-contributors add @bacongravy for code

PR #1741, opened 2021-01-18
(newapp) Add `changePassword` mutation

flybayer:
**Stacked on https://github.com/blitz-js/blitz/pull/1740**

This adds the `changePassword` mutation to new apps that everyone will need. It doesn't add a form or page for it since those will usually be custom.


packages/generator/templates/app/app/auth/mutations/changePassword.ts
```diff
@@ -0,0 +1,21 @@
+import { Ctx, NotFoundError, SecurePassword } from "blitz"
+import db from "db"
+import { authenticateUser } from "./login"
+import { ChangePasswordInput, ChangePasswordInputType } from "../validations"
+
+export default async function updatePassword(input: ChangePasswordInputType, ctx: Ctx) {
```
flybayer, line 6:
```suggestion
export default async function changePassword(input: ChangePasswordInputType, ctx: Ctx) {
```

PR #1758, opened 2021-01-20
fix: auth form error message

bravo-kernel:
### Currently

The signup and login pages show error message `[object Object]` when there is e.g. no database connection present, as can seen in this before-screenshot:

blitz-before

### After this PR

The expected error message is shown on screen, as can be seen in this after-screenshot:

blitz-after

### Disclaimer

I'm no blitz/ts guru so feel free to close if this is unsuitable. I just felt the time to describe the issue would take more time than creating the PR.

### Checklist

- [ ] Tests added for changes

### DIFF
--- a/packages/generator/templates/app/app/auth/components/LoginForm.tsx
+++ b/packages/generator/templates/app/app/auth/components/LoginForm.tsx
@@ -30,7 +30,7 @@ export const LoginForm = (props: LoginFormProps) => {
             } else {
               return {
                 [FORM_ERROR]:
-                  "Sorry, we had an unexpected error. Please try again. - " + error.toString(),
+                  "Sorry, we had an unexpected error. Please try again. - " + error.message.toString(),
               }
             }
           }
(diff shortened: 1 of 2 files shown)


flybayer:
Hey thanks! I'm unable to reproduce the `[object Object]` issue 🤔  What browser is that?

[According to MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/toString), `error.toString()` should work in call cases.

Flavyoo:
@bravo-kernel Are you still able to reproduce this issue? If not, then maybe we can close it.

bravo-kernel:
I will try to reproduce this weekend and report back.

bravo-kernel:
I can no longer reproduce this behavior using @blitzjs/cli/0.29.7 win32-x64 node-v14.15.4.

Flavyoo:
Thanks for confirming @bravo-kernel

PR #1763, opened 2021-01-21
doc: Describe precise version of node.js

tarunama:
describe precise version of node.js in [Getting started](https://blitzjs.com/docs/getting-started)

### Checklist

- [ ] Tests added for changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Blitz brings back the **simplicity and conventions** of server-rendered framewor
 
 ### Quick Start
 
-You need Node.js 12 or newer
+You need Node.js 12.20.0 or newer
 
 #### Install Blitz
 


flybayer:
@tarunama as far as I know any v12 version and up works. The `.node-version` setting is just a convenience, not a requirement.

tarunama:
@flybayer 
I have misunderstood. Thank you for your review !

nitaking:
@flybayer 
There is the following in `package.json/engines`, which causes an error with yarn in node v12.20 and below. (or npm install --engine-strict)

This means that if you want to use node v12.20 or above as a development environment, you need to make sure that it works as well.
Ideally, package.json/engines should be set to `"node":">=12"` if you want to use 12 or more. What do you think? (The
 .node-version is fine as it is)

```json
  "engines": {
    "node": ">=12.20.0"
  },
```

https://github.com/blitz-js/blitz/blob/9536caa07383847e400a686a3bb2e2b216476bb5/package.json#L16

```bash
 $ yarn
yarn install v1.22.4
[1/5] 🔍  Validating package.json...
error root@0.0.0: The engine "node" is incompatible with this module. Expected version ">=12.20.0". Got "12.17.0"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

flybayer:
Oh good point, then we should change that to just be `>=12`

PR #1779, opened 2021-01-22
[RFC] Prefix all `ctx.session` methods with `$` and move `publicData` fields to top level of `ctx.session`

flybayer:

Our `useSession` hook returns an object with all of the `publicData` properties on it like `session.customProp`. But on the server to access that you have to do `ctx.session.publicData.customProp` which is quite cumbersome and not consistent with the frontend.

So why not change this to be like Prisma client where custom things like `authorize()` are prefixed with `$` and then all publicData properties can be accessed at the top level.

```diff
-await ctx.session.authorize()
+await ctx.session.$authorize()

const userId = ctx.session.userId

-const orgId = ctx.session.publicData.orgId
+const orgId = ctx.session.orgId
```

What do you all think?

### Checklist

- [x] Tests added for changes

### DIFF
--- a/examples/auth/app/auth/mutations/login.ts
+++ b/examples/auth/app/auth/mutations/login.ts
@@ -31,7 +31,7 @@ export default async function login(input: LoginInputType, {session}: Ctx) {
   // This throws an error if credentials are invalid
   const user = await authenticateUser(email, password)
 
-  await session.create({userId: user.id, roles: [user.role]})
+  await session.$create({userId: user.id, roles: [user.role]})
 
   return user
 }
--- a/examples/auth/app/auth/mutations/logout.ts
+++ b/examples/auth/app/auth/mutations/logout.ts
@@ -1,5 +1,5 @@
 import {Ctx} from "blitz"
 
 export default async function logout(_: any, {session}: Ctx) {
-  return await session.revoke()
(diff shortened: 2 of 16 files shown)


matthieuh:
The fact that we can access the publicData properties at directly on ctx.session seems a good idea but I'm not seeing the relation ans benefit of prefixing authorize method with a `$`.

flybayer:
> The fact that we can access the publicData properties directly on ctx.session seems a good idea but I'm not seeing the relation ans benefit of prefixing authorize method with a `$`.

The reason to prefix our methods with `$` is to prevent naming clashes with custom user defined public data fields.

MrLeebo:
Do you worry that this might set up an expectation that a user should be able to write 

```ts
ctx.session.orgId = 12;
``` 

to update their session? Or give the proxy a setter so that it works...

flybayer:
> Do you worry that this might set up an expectation that a user should be able to write
> 
> ```ts
> ctx.session.orgId = 12;
> ```
> 
> to update their session? Or give the proxy a setter so that it works...

I don't think this PR makes a different whether it is `ctx.session.orgId` or `ctx.session.publicData.orgId` 🤷  And in the React community we are used to not being able to set properties directly.

PR #1788, opened 2021-01-24
Remove requirement of `publicData.roles`, change new app to use `publicData.role`, and fix public data types

Kosai106:
Closes: [Slack thread](https://blitzjs.slack.com/archives/C011EQ0J75M/p1611434916079100?thread_ts=1611434048.078800&cid=C011EQ0J75M)

By default, Blitz assumes the use of a `roles` array which isn't always the case and can cause type issues if not included.

This PR aims to remove the notion of roles in order to leave that decision up to the user, although I've kept `role` in the auth example as a reference for new users.

Feel free to let me know if I missed something or misunderstood what you had in mind @flybayer.

### Checklist

- [x] Tests ~added~ updated for changes


packages/core/src/types.ts
```diff
@@ -32,8 +32,7 @@ export interface BlitzRouter extends NextRouter {
 }
 
 export interface DefaultPublicData {
-  userId: any
-  roles: string[]
+  userId: string | number | null
```
Kosai106, line 35: Let me know if you prefer having this as its own PR.

packages/server/src/supertokens.test.ts
```diff
@@ -141,7 +140,6 @@ describe("supertokens", () => {
 
       const publicData = JSON.parse(publicDataStr)
       expect(publicData.userId).toBe(1)
-      expect(publicData.roles[0]).toBe("admin")
```
flybayer, line 144: Leave this test in here. We still want this to work, even though it won't be required.

packages/generator/templates/app/app/auth/mutations/login.ts
```diff
@@ -25,7 +25,7 @@ export default async function login(input: LoginInputType, ctx: Ctx) {
   // This throws an error if credentials are invalid
   const user = await authenticateUser(email, password)
 
-  await ctx.session.create({ userId: user.id, roles: [user.role] })
+  await ctx.session.create({ userId: user.id })
```
flybayer, line 28: The new app template should now use `role`. So the mutations need changed for this and also the `types.ts` file.

packages/server/src/supertokens.ts
```diff
@@ -224,7 +221,7 @@ export class SessionContextClass implements SessionContext {
   isAuthorized(input?: any) {
     if (!this.userId) return false
 
-    return config.isAuthorized(this.roles, input)
+    return config.isAuthorized(input)
```
flybayer, line 224: `isAuthorized` should now take `this.publicData` as first argument

packages/server/src/supertokens.ts
```diff
@@ -246,8 +243,8 @@ export class SessionContextClass implements SessionContext {
   }
 
   async setPublicData(data: Record<any, any>) {
-    if (this.userId && data.roles) {
-      await updateAllPublicDataRolesForUser(this.userId, data.roles)
+    if (this.userId) {
+      await updateAllPublicDataRolesForUser(this.userId)
```
flybayer, line 247: TODO: We need to decide best approach here. Until now `publicData` is also session specific. The only thing that could change across sessions was `roles`.

Another option is add a config item for which `publicData` fields should be synced across sessions. Actually I kind of like this idea. `roles` and `role` could be set in this list by default.

packages/server/src/supertokens.ts
```diff
@@ -723,7 +719,7 @@ export async function createNewSession(
 }
 
 export async function createAnonymousSession(req: IncomingMessage, res: ServerResponse) {
-  return await createNewSession(req, res, {userId: null, roles: []}, undefined, {anonymous: true})
+  return await createNewSession(req, res, {userId: null}, undefined, {anonymous: true})
```
flybayer, line 722: I think we need to add a "default publicData" config that would be used here for anonymous sessions. Because with as setup where `roles` is used, you always want `roles` to be an array type.

flybayer (commented):
Sweet, thank you!!

In addition to comments below, `simpleRolesIsAuthorized()` needs updated:
- First argument is now `publicData`
- Should continue to work as it has until now with either `publicData.roles` **or** `publicData.role`
- Error if both `publicData.roles` and `publicData.role` is present
- Error if neither `publicData.roles` or `publicData.role` is present

Lastly I just merged https://github.com/blitz-js/blitz/pull/1779 which caused a number of merge conflicts. Sorry about that!

packages/server/src/supertokens.ts
flybayer: Another great suggestion by @Zeko369: sync all publicData across session by default but add a second config param to `setPublicData` to disable this: `{thisSessionOnly: true}` 

- [ ] Need to check and make sure that `publicData` set during `session.create()` is also synced across sessions.

packages/server/src/supertokens.ts
flybayer: @Kosai106 **UPDATE:** canary branch already changed `config.isAuthorized` implementation so only minimal changes should be needed in this PR for that

packages/server/src/supertokens.ts
flybayer: After some discussion I think we should go with this solution:

> Another option is add a config item for which publicData fields should be synced across sessions. Actually I kind of like this idea. roles and role could be set in this list by default.

packages/server/src/supertokens.ts
Kosai106: Excellent - I haven't had time to work on this but I'll be sure to rebase. 👍

flybayer:
@Kosai106 are you able to finish this up? If not I can take it over

Kosai106:
@flybayer I don't think I'll have time in the next few weeks I'm afraid so feel free to make any changes yourself or if another maintainer wants to take over. 👍

Kosai106:
Thanks for taking care of finishing this @flybayer 💜

PR #1796, opened 2021-01-26
Improvements to `isAuthorized`: (1) Ability to strongly type, (2) Change first arg to have `ctx`, (3) Allow multiple params, (4) add conditional `$authorize()`

flybayer:

This

1. Adds ability to strongly type the input of `ctx.session.$authorize()` and `ctx.session.$isAuthorized()`
2. Changes `isAuthorized` signature to take a single object with `{ctx, args: []}`
3. Allows `isAuthorized` to accept multiple params like `ctx.session.$authorize('read', 'project')`
4. Add optional second `{if: boolean}` argument to the default `simpleRolesIsAuthorized` implementation.
   5. Enables this: `ctx.session.$authorize('admin', {if: id !== ctx.session.userId})`

### Checklist

- [ ] Tests added for changes

### DIFF
--- a/examples/auth/app/users/queries/getUser.ts
+++ b/examples/auth/app/users/queries/getUser.ts
@@ -7,6 +7,7 @@ type GetUserInput = {
 
 export default async function getUser({where}: GetUserInput, ctx: Ctx) {
   ctx.session.$authorize()
+  ctx.session.$isAuthorized()
 
   const user = await db.user.findFirst({where})
 
--- a/examples/auth/app/users/queries/getUser.ts
+++ b/examples/auth/app/users/queries/getUser.ts
@@ -7,7 +7,6 @@ type GetUserInput = {
 
 export default async function getUser({where}: GetUserInput, ctx: Ctx) {
   ctx.session.$authorize()
-  ctx.session.$isAuthorized()
 
   const user = await db.user.findFirst({where})
 
--- a/examples/auth/types.ts
+++ b/examples/auth/types.ts
@@ -11,7 +11,7 @@ declare module "blitz" {
     userId: User["id"]
     views?: number
   }
-  export interface Authorization {
+  export interface Session {
     isAuthorized: typeof simpleRolesIsAuthorized
   }
 }
--- a/examples/auth/types.ts
+++ b/examples/auth/types.ts
@@ -1,5 +1,6 @@
-import {DefaultCtx, SessionContext, DefaultPublicData, DefaultAuthorize} from "blitz"
-// import {simpleRolesIsAuthorized} from "@blitzjs/server"
+import {DefaultCtx, SessionContext, DefaultPublicData} from "blitz"
+
+import {simpleRolesIsAuthorized} from "@blitzjs/server"
 import {User} from "db"
 
 declare module "blitz" {
@@ -10,8 +11,7 @@ declare module "blitz" {
     userId: User["id"]
     views?: number
   }
-  // export type IsAuthorized = typeof simpleRolesIsAuthorized
-  export interface Authorize extends DefaultAuthorize {
-    (roleOrRoles?: string | string[], options?: {if?: boolean}): boolean
+  export interface Authorization {
+    isAuthorized: typeof simpleRolesIsAuthorized
   }
 }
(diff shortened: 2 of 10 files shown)


examples/auth/app/users/queries/getUser.ts
```diff
@@ -7,6 +7,7 @@ type GetUserInput = {
 
 export default async function getUser({where}: GetUserInput, ctx: Ctx) {
   ctx.session.$authorize()
+  ctx.session.$isAuthorized()
```
flybayer, line 10: These aren't working correctly. `$authorize()` is typed `...args: any[]`. 

`$isAuthorized()` picks up the correct type but has two overloads for `...args: any[]` which allows incorrect usage 🤔

examples/auth/types.ts
```diff
@@ -9,4 +10,8 @@ declare module "blitz" {
     userId: User["id"]
     views?: number
   }
+  // export type IsAuthorized = typeof simpleRolesIsAuthorized
+  export interface Authorize extends DefaultAuthorize {
+    (roleOrRoles?: string | string[], options?: {if?: boolean}): boolean
+  }
 }
```
flybayer, line 17: User provided type override (same approach as `PublicData` which works fantastically)

packages/core/src/supertokens.ts
```diff
@@ -25,16 +26,14 @@ export type SessionConfig = {
...
 
 export interface SessionContextBase extends PublicData {
   $handle: string | null
   $publicData: unknown
-  $authorize(input?: any): asserts this is AuthenticatedSessionContext
-  $isAuthorized(input?: any): boolean
-  // authorize: (roleOrRoles?: string | string[]) => void
-  // isAuthorized: (roleOrRoles?: string | string[]) => boolean
+  $authorize(...args: Parameters<Authorize>): asserts this is AuthenticatedSessionContext
+  $isAuthorized: Authorize
```
flybayer, line 36: Here's where the `Authorize` type (which user can modify) is used.

packages/server/src/supertokens.ts
```diff
@@ -222,22 +243,22 @@ export class SessionContextClass implements SessionContext {
     return this._kernel.publicData
   }
 
-  $authorize(input?: any) {
+  $authorize(...args: Parameters<Authorize>) {
```
flybayer, line 246: This usage of `Authorize` shouldn't be exposed to the end user app. The type definition for this class comes from `@blitzjs/core` which is where users import it from. This here in `@blitzjs/server` is only the implementation.

examples/auth/types.ts
flybayer: The only thing the user really needs to modify is the arguments to the function. The return type is always fixed. So maybe there's another way to structure this that is better.

flybayer:
I made more progress by doing this, where the types are modified at a higher level.

```ts
export interface SessionContextBase extends Authorization, PublicData {
  $handle: string | null
  $publicData: unknown
  // $authorize(...args: AuthorizeArgs): asserts this is AuthenticatedSessionContext
  // $isAuthorized: Authorize
  ...
}
```

```ts
declare module "blitz" {
  export interface Ctx extends DefaultCtx {
    session: SessionContext
  }
  export interface PublicData extends DefaultPublicData {
    userId: User["id"]
    views?: number
  }

  export interface Authorization extends DefaultAuthorization {
    $authorize(
      roleOrRoles?: string | string[],
      options?: {if?: boolean},
    ): asserts this is AuthenticatedSessionContext
    $isAuthorized(roleOrRoles?: string | string[], options?: {if?: boolean}): boolean
  }
}
```

This results in proper types in the user's app, but that's too ugly/tedious types in user code. Ideally the type could be defined like this somehow:

```ts
import {simpleRolesIsAuthorized} from '@blitzjs/core'

export interface Authorize extends typeof simpleRolesIsAuthorized {}
// or
export interface Authorize extends typeof makeAuthorize(simpleRolesIsAuthorized) {}
```

Basically the param types for `$authorize()` and `$isAuthorized()` are all the params of the `isAuthorized` implementation, i.e. `simpleRolesIsAuthorized` **except** the first one.

Type signature for `simpleRolesIsAuthorized`:
```ts
export function simpleRolesIsAuthorized(
  {ctx}: any,
  roleOrRoles?: string | string[],
  {if: condition = true}: {if?: boolean} = {},
) {}
```

flybayer:
@beerose I fixed the loss of argument names!! 👉 https://github.com/blitz-js/blitz/pull/1796/commits/f0ae53495c62a3be71ad7a7cb82c3940571a91c1

PR #1799, opened 2021-01-26
inject isAuthorized args type

beerose:
Closes: a twitter thread

A user can inject a type of `isAuthorized`:

```ts
declare module "blitz" {
  export interface Authorization {
    isAuthorized: typeof simpleRolesIsAuthorized
  }
}
```
which then will be used for parameters of `$authorize` and `$isAuthorized` omitting the first argument (`ctx`).

I had to set TypeScript 4.1 in resolutions to make `Tail` (variadic tuple) work (though it's possible to make it work in older TS versions using some hacks).

Limitations: it seems we lost parameter names on module boundaries (they display nicely as a named tuple when the code is in the same file).


flybayer:
Awesome!!! Thank you!! Loosing the argument names isn't ideal, but better than no types at all!

@all-contributors add @beerose for code

PR #1801, opened 2021-01-26
Add `pipe.resolver()`, `pipe.zod()`, `pipe.authorize()`, and `pipe.authorizeIf()` for queries/mutations

flybayer:
Closes: https://github.com/blitz-js/blitz/discussions/1787

This adds some awesome pipe utilities for queries/mutations.

**Benefits:**

- Easily validate input data with a zod schema
- Easily require user to be logged in
- Less having to manually specify TS types
- Get the resolver input properly typed without having to do `export type SignupInputType = z.infer<typeof SignupInput>`
- For advanced cases you can easily compose functions

```ts
import {pipe} from "blitz"
import db from "db"
import * as z from "zod"

export const CreateProject = z.object({
  name: z.string(),
  dueDate: z.date().optional(),
  orgId: z.number().optional(),
})

export default pipe.resolver(
  pipe.zod(CreateProject),
  pipe.authorize(),
  // Set default orgId
  (input, {session}) => ({orgId: session.orgId, ...input}),
  async (input, ctx) => {

    console.log("Creating project...", ctx.session.orgId)
    const project = await db.project.create({
      data: input,
    })
    console.log("Created project")

    return project
  },
)
```

### Checklist

- [ ] Tests added for changes


packages/core/src/pipes.ts
```diff
@@ -0,0 +1,195 @@
...
+  hi: PipeFn<H, I>,
+  ij: PipeFn<I, J>,
+  jk: PipeFn<J, K>,
+  kl: PipeFn<K, L>,
+  lm: PipeFn<L, M>,
+  mn: PipeFn<M, N>,
+  no: PipeFn<N, O>,
+  op: PipeFn<O, P>,
+): (input: A, ctx: Ctx) => P
+function resolver<A, B, C, D, E, F, G, H, I, J, K, L, M, N, O, P, Q>(
```
Flavyoo, line 152: Just curious, why do you stop at Q?

packages/core/src/pipes.ts
```diff
@@ -0,0 +1,195 @@
...
+  kl: PipeFn<K, L>,
+  lm: PipeFn<L, M>,
+  mn: PipeFn<M, N>,
+  no: PipeFn<N, O>,
+  op: PipeFn<O, P>,
+  pq: PipeFn<P, Q>,
+): (input: A, ctx: Ctx) => Q
+function resolver(..._args: unknown[]): unknown {
+  // Convert to real array
+  const functions: PipeFn<unknown, unknown>[] = [].slice.call(arguments)
```
Flavyoo, line 172: Where is `arguments` defined?

packages/core/src/pipes.ts
flybayer: Because you have to stop somewhere and that's quite a lot! :)

packages/core/src/pipes.ts
flybayer: It's part of Javascript.
👉 https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments

flybayer: But good point, I should just use `args` here. don't need `arguments` any more, so refactored that out.

Flavyoo (approved):

Issue #1826, opened 2021-01-29
`enhancePrisma` does not work with TS `strict: true`


flybayer:
### What is the problem?

Set `strict: true` in tsconfig.json and you get this error:

```ts
db/index.ts:4:38 - error TS2345: Argument of type 'typeof PrismaClient' is not assignable to parameter of type 'Constructor<unknown>'.
  Types of parameters 'optionsArg' and 'args' are incompatible.
    Type 'unknown' is not assignable to type 'Subset<PrismaClientOptions, PrismaClientOptions> | undefined'.
      Type 'unknown' is not assignable to type 'Subset<PrismaClientOptions, PrismaClientOptions>'.

4 const EnhancedPrisma = enhancePrisma(PrismaClient)

app/users/queries/getUsers.ts:16:26 - error TS2339: Property 'user' does not exist on type 'EnhancedPrismaClientAddedMethods'.

16   const count = await db.user.count()
                            ~~~~
```

### Versions

```
0.30.0-canary.2
```

PR #1827, opened 2021-01-29
Fix `enhancePrisma` with TS `strict:true`

flybayer:
Closes: https://github.com/blitz-js/blitz/issues/1826

Fix `enhancePrisma` with TS `strict:true`

### Checklist

### DIFF
--- a/packages/core/src/prisma-utils.ts
+++ b/packages/core/src/prisma-utils.ts
@@ -1,7 +1,7 @@
 import {exec} from "npm-run"
 
 interface Constructor<T = unknown> {
-  new (...args: any[]): T
+  new (...args: never[]): T
 }
 
 interface EnhancedPrismaClientAddedMethods {
--- a/packages/core/src/prisma-utils.ts
+++ b/packages/core/src/prisma-utils.ts
@@ -1,7 +1,7 @@
 import {exec} from "npm-run"
 
 interface Constructor<T = unknown> {
-  new (...args: unknown[]): T
+  new (...args: any[]): T
 }
 
 interface EnhancedPrismaClientAddedMethods {
--- a/examples/auth/app/api/auth/[...auth].ts
+++ b/examples/auth/app/api/auth/[...auth].ts
@@ -1,4 +1,4 @@
-import {passportAuth} from "blitz"
+import {passportAuth, PublicData} from "blitz"
 import db from "db"
 import {Strategy as TwitterStrategy} from "passport-twitter"
 import {Strategy as GitHubStrategy} from "passport-github2"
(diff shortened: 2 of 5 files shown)


packages/core/src/prisma-utils.ts
```diff
@@ -1,7 +1,7 @@
 import {exec} from "npm-run"
 
 interface Constructor<T = unknown> {
-  new (...args: unknown[]): T
+  new (...args: any[]): T
```
hasparus, line 4: Ouch, I botched it. I wanted to write "any possible constructor", and instead I wrote "any constructor that doesn't care about its arguments". 

Could we use `never` here? `any` is an escape hatch, and it tends to "leak". In "fully strict" TypeScript you'd have `never` and `unknown` in all places where `any` is commonly used. (unknown is top type, never is bottom type, any is "shut up TypeScript, I ain't got time"). It's a bit tricky — I had to pull the PR and tested it to make sure. 

[TS Playground with an example](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgMIHsQGcxQK4JjpQBiU6AtgHLpgAWoA5gDwAqyAvMniANYjoA7iAB8yAN4AoZMhARByABQA6VXCiMsALlkQAbtADaAXQCUO1pIC+kyaEixEKDNlwEipchQDqdOJAMoNk5uPgFhMSkZOQUVNQ1tUP4hEBNzZEsbO3BoeCQ0TBx8QmJvYHoAUSwEOAAHCAAJfwQ6YK4eZIiJaV1Y1WV1TR04EABPNItrWwQAGzgsLGQGvAoR7pkZBEK3EqhFWrwAIxngBFk4CggdIqZTCRsssFH6pZWRlyL3YhCn+vQYV6rEAAblsAHowcg3BBJL8UAB9ACMIWWQI+Ow8yAgAA9ICAACaLdHFDxkSg0ehMZAAfih+BQOngMywEFBkghyCZLNhzwRACYUW8QMSvlAsbiIASidsScQyT4-AFoDS6XgGZy4MzWeDIdCeS94QBmQVomWi8V4wkFVyyqBlSrVOqNZp0FXQ5CMzUs4FAA)

```suggestion
  new (...args: never[]): T
```


---

BTW I assumed that `strict: false` is here in examples tsconfigs because of Prisma codegen.
What do you think about adding strict typecheck to catch problems like this?

flybayer: ok sweet, thanks!

flybayer: I'll see about making auth example work with strict mode

Issue #1820, opened 2021-01-29
pipe.authorize() doesn't work as type guard


merodiro:
### What is the problem?
The new `pipe.authorize()` doesn't work as a type guard like `ctx.session.$authorize()`

### Steps to Reproduce

create a new mutation with this code
```js
export default pipe.resolver(pipe.zod(CreatePost), pipe.authorize(), async ({ body }, ctx) => {
  const post = await db.post.create({
    data: {
      body,
      userId: ctx.session.userId,
    },
  })

  return post
})
```

### Expected
`ctx.session.userId` to always be `number`

### Current
`ctx.session.userId` is `number | null`

### Versions

```
blitz: 0.30.0-canary.2 (local)

  Package manager: yarn 
  System:
    OS: Linux 4.19 Ubuntu 20.04.2 LTS (Focal Fossa)
    CPU: (6) x64 Intel(R) Core(TM) i5-9400F CPU @ 2.90GHz
    Memory: 11.53 GB / 12.44 GB
    Shell: 3.1.0 - /usr/bin/fish
  Binaries:
    Node: 15.6.0 - ~/.config/nvm/15.6.0/bin/node
    Yarn: 1.22.5 - /usr/bin/yarn
    npm: 7.4.0 - ~/.config/nvm/15.6.0/bin/npm
    Watchman: Not Found
  npmPackages:
    @prisma/cli: ~2.15 => 2.15.0 
    @prisma/client: ~2.15 => 2.15.0 
    blitz: 0.30.0-canary.2 => 0.30.0-canary.2 
    react: 0.0.0-experimental-3310209d0 => 0.0.0-experimental-3310209d0 
    react-dom: 0.0.0-experimental-3310209d0 => 0.0.0-experimental-3310209d0 
    typescript: 4.1.3 => 4.1.3
```

flybayer:
NOTE: in canary branch, this is now `resolver.authorize` in `packages/core/src/resolver.ts`

hasparus:
**how I thought about it**

So... there's no way to use `asserts` or `is` on a `ctx` parameter, because there is no "ctx" variable in common scope, right? 
`Ctx` is set in `PipeFn` type, and that's it, no function earlier in the chain can modify it...

**end result**

![image](https://user-images.githubusercontent.com/15332326/106328956-f1fa9800-6280-11eb-854e-a6dda6c5b9c1.png)

**new authorize function**

[...]

PR #1831, opened 2021-01-29
Fix `resolver.authorize(` doesn't work as type guard (with great effort)

hasparus:
Closes: https://github.com/blitz-js/blitz/issues/1820

Functions in `resolver.pipe` can now modify the type of `Ctx`.

### Checklist

- [ ] ~~Changes covered by tests (tests added if needed)~~ — There's no way to test this automatically (yet?)
- [ ] ~~PR submitted to [blitzjs.com](https://github.com/blitz-js/blitzjs.com) for any user facing changes~~ — not needed

---

I left overloads higher than `n` commented, because I want to discuss it.

I highly doubt anybody is going to use more than 12 function inside `pipe`, and you can always compose functions inside.
Each overload is slightly affecting user's TypeScript performance (as I'm too lazy to write more overloads, I'm also too lazy to back this claim with benchmarks, so you'll have to trust me :P)


packages/core/src/resolver.ts
```diff
@@ -1,195 +1,366 @@
...
+      if (isResultWithContext(lastResult)) {
+        ctx = lastResult.ctx as Ctx
+        lastResult = lastResult.value
+      }
     }
     return lastResult
   }
 }
 
+interface ResolverAuthorize {
```
hasparus, line 340: I declared the type of `resolver.authorize`, because it was getting pretty complex, and while "types first, implementation later, functions as interfaces" is probably the most tiring way to write code, it results in the shortest error messages and tooltips

image

packages/core/src/resolver.ts
```diff
@@ -1,195 +1,366 @@
...
+//   hi: PipeFn<H, I>,
+//   ij: PipeFn<I, J>,
+//   jk: PipeFn<J, K>,
+//   kl: PipeFn<K, L>,
+//   lm: PipeFn<L, M>,
+//   mn: PipeFn<M, N>,
+//   no: PipeFn<N, O>,
+//   op: PipeFn<O, P>,
+//   pq: PipeFn<P, Q>,
+// ): (input: A, ctx: Ctx) => Q
```
flybayer, line 323:
```suggestion
```

packages/core/src/resolver.ts
andrekovac: @hasparus Very interesting comment. 

Where can I read more about the "types first, implementation later, functions as interfaces" philosophy?

I understand "types first, implementation later". But what exactly does "functions as interfaces" mean in this context? How does this fit together with the fact that `PipeFn` is also a function, but you defined it as type alias and not as `interface`?

PR #1834, opened 2021-01-30
Fix bug in `resolver.authorize()` type

flybayer:

Fix bug in `resolver.authorize()` type from #1831 

```
app/projects/mutations/createProject.ts:12:3 - error TS2345: Argument of type '<T>(input: T, ctx: Ctx) => ResultWithContext<T, AuthenticatedMiddlewareCtx>' is not assignable to parameter of type 'PipeFn<{ dueDate?: Date | undefined; name: string; }, { dueDate?: Date | undefined; name: string; }, unknown, AuthenticatedMiddlewareCtx>'.
  Types of parameters 'ctx' and 'c' are incompatible.
    Type 'unknown' is not assignable to type 'Ctx'.

12   resolver.authorize(),
```

cc @hasparus. Apparently you had branched from a bit earlier in canary? not sure how this passed CI on your branch. Oh well, this works.

### DIFF
--- a/examples/auth/app/projects/mutations/createProject.ts
+++ b/examples/auth/app/projects/mutations/createProject.ts
@@ -9,6 +9,7 @@ export const CreateProject = z.object({
 
 export default resolver.pipe(
   resolver.zod(CreateProject),
+  (input, _ctx) => ({extraFieldForIntegrationTesting: _ctx.session.userId, ...input}),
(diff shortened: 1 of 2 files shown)


hasparus:
Sorry! :D That `= Ctx` makes so much sense. How could I forget about it?

PR #1865, opened 2021-02-04
fix: enable resolver.authorize to accept Ctx

camsloanftc:
resolver.authorize should accept a standard `Ctx` in addition to `AuthenticatedMiddlewareCtx`. This is useful for testing purposes.

Closes: N/A

### Checklist

### DIFF
--- a/packages/core/src/resolver.ts
+++ b/packages/core/src/resolver.ts
@@ -275,7 +275,7 @@ interface ResolverAuthorize {
   (...args: Parameters<SessionContextBase["$authorize"]>): <T, C>(
     input: T,
     ctx: C,
-  ) => ResultWithContext<T, AuthenticatedMiddlewareCtx>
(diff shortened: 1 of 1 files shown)


packages/core/src/resolver.ts
```diff
@@ -275,7 +275,7 @@ interface ResolverAuthorize {
   (...args: Parameters<SessionContextBase["$authorize"]>): <T, C>(
     input: T,
     ctx: C,
-  ) => ResultWithContext<T, AuthenticatedMiddlewareCtx>
+  ) => ResultWithContext<T, AuthenticatedMiddlewareCtx | Ctx>
```
flybayer, line 278: Thanks Cam, but this breaks the "typeguard" feature of authorize(). This return type must be AuthenticatedMiddlewareCtx.

Without knowing the code and error you're having, I'm not sure what the correct solution would be.

camsloanftc:
I am going to open an issue with more details since there are some other downstream effects here.

Issue #1840, opened 2021-02-01
resolver pipes without resolver.zod as the first argument shows TypeScript error when in strict mode


jan-wilhelm:
### What is the problem?
When in TS strict mode, the code
```typescript
export default resolver.pipe(resolver.authorize(), (input, ctx)) => {

}
```
shows a type error for the `resolver.authorize()` part.
The error is `Argument of type '<T, C>(input: T, ctx: C) => ResultWithContext<T, AuthenticatedMiddlewareCtx>' is not assignable to parameter of type 'PipeFn<unknown, T, Ctx, AuthenticatedMiddlewareCtx>'`.

The error does not occur when you put a `resolver.zod()` call as the first argument to `resolver.pipe`, before the authorizer.
### Steps to Reproduce

1. Enabled TS strict mode (`strict: true` in tsconfig)
2. Make a new file and type in the above code
3. You will see the TypeScript error

### Versions

```
blitz: 0.30.0-canary.2 (global)
blitz: 0.30.0-canary.4 (local)

  Package manager: yarn 
  System:
    OS: macOS 11.1
    Shell: 5.8 - /usr/local/bin/zsh
  Binaries:
    Node: 15.6.0 - /usr/local/bin/node
    Yarn: 1.22.10 - /usr/local/bin/yarn
    npm: 7.4.3 - /usr/local/bin/npm
    Watchman: Not Found
  npmPackages:
    @prisma/cli: ~2.15.0 => 2.15.0 
    @prisma/client: ~2.15.0 => 2.15.0 
    blitz: 0.30.0-canary.4 => 0.30.0-canary.4 
    react: 0.0.0-experimental-3310209d0 => 0.0.0-experimental-3310209d0 
    react-dom: 0.0.0-experimental-3310209d0 => 0.0.0-experimental-3310209d0 
    typescript: 4.1.3 => 4.1.3 
```

reiwa:
I have the same problem and I need to use unknown for this case.

```ts
export default resolver.pipe(resolver.authorize(), (_: unknown, ctx)) => {

}
```

flybayer:
I think the ideal solution would look like this:

```ts
import {Ctx, NotFoundError, resolver} from "blitz"
import db, {Prisma} from "db"

type GetProjectInput = Pick<Prisma.ProjectFindFirstArgs, "where">

export default resolver.pipe<GetProjectInput>(
  resolver.authorize(), 
  async (input, ctx) => {
[...]

PR #1878, opened 2021-02-05
fix typechecking of resolver pipe without input validation

beerose:
Closes: https://github.com/blitz-js/blitz/issues/1840

### Checklist


flybayer:
Hey, thanks for looking into this!

But it actually didn't fix anything 😅 . Tests didn't catch it because the auth example didn't have the latest template code (with resolver.pipe). I fixed that on canary (thought it was already done, but no). Hence why the test failed now.

But somehow you inspired me with the correct fix which I made here: https://github.com/blitz-js/blitz/pull/1881

Issue #1867, opened 2021-02-04
Enable mocking Ctx for queries/mutations that use resolver.authorize


camsloanftc:
### What do you want and why?
I would like to be able to test my protected mutation routes that use resolver.authorize. The problem is that this generates a type error when I try to mock out the Ctx.

```ts
// createTour.ts
export default resolver.pipe(
  resolver.zod(CreateTour),
  resolver.authorize(["admin", "user"]),
  async function createTour({ name, steps }: z.infer<typeof CreateTour>) {
    // etc...
  }
)

// createTour.test.ts
import { Ctx } from "blitz"
import createTour from "./createTour"

describe("createTour mutation", () => {
  it("does a thing", async () => {
    await expect(
      createTour({ name: "My fave Tour", steps: [] }, {
        session: { $authorize: () => {} },
      } as Ctx)
    ).resolves.not.toThrow()
  })
```

This actually runs no problem but it results in a type error in the test file when I try to cast as `Ctx`, because it does not overlap with `AuthenticatedMiddlewareCtx`
```Argument of type 'Ctx' is not assignable to parameter of type 'AuthenticatedMiddlewareCtx'.ts(2345)```

### Possible implementation(s)
We could export the `AuthenticatedMiddlewareCtx` type, but maybe that is an internal so @flybayer suggested we could potentially just enable `resolver.authorize` to accept `Ctx` as well.

I gave a shot at this in #1865 but it broke things 😅 so I am back here to file an issue like I should have done in the first place.

### Additional context
![image](https://user-images.githubusercontent.com/16295659/106948926-55118200-66fa-11eb-9aa3-29164e286db6.png)

PR #1881, opened 2021-02-06
Fix all known `resolver.pipe` type issues AND update `getModelNames` template to use resolver.pipe

flybayer:
Closes: https://github.com/blitz-js/blitz/issues/1840
Closes: https://github.com/blitz-js/blitz/issues/1867

Fix those bugs!!

And update `getModelNames` template to use resolver.pipe (couldn't until this fix)

### Checklist

### DIFF
--- a/packages/core/src/resolver.ts
+++ b/packages/core/src/resolver.ts
@@ -38,14 +38,14 @@ function pipe<A, B, C, D, E, CA = Ctx, CB = CA, CC = CB, CD = CC, CE = CD>(
   bc: PipeFn<B, C, CB, CC>,
   cd: PipeFn<C, D, CC, CD>,
   de: PipeFn<D, E, CD, CE>,
-): (input: A, ctx: CA) => E
+): (input: A, ctx: CA) => EnsurePromise<E>
 function pipe<A, B, C, D, E, F, CA = Ctx, CB = CA, CC = CB, CD = CC, CE = CD, CF = CE>(
   ab: PipeFn<A, B, CA, CB>,
   bc: PipeFn<B, C, CB, CC>,
   cd: PipeFn<C, D, CC, CD>,
   de: PipeFn<D, E, CD, CE>,
   ef: PipeFn<E, F, CE, CF>,
-): (input: A, ctx: CA) => F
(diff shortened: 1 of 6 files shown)


packages/core/src/resolver.ts
```diff
@@ -23,37 +23,37 @@ type PipeFn<Prev, Next, PrevCtx extends Ctx, NextCtx = PrevCtx> = (
   c: PrevCtx,
 ) => Next extends ResultWithContext ? never : Next | ResultWithContext<Next, NextCtx>
 
-function pipe<A, Z>(ab: (i: A, c: Ctx) => Z): (input: A, ctx: Ctx) => Z
+function pipe<A, Z>(ab: (i: A, c: Ctx) => Z): (input: A, ctx: Ctx) => EnsurePromise<Z>
 function pipe<A, B, C, CA = Ctx, CB = CA, CC = CB>(
   ab: PipeFn<A, B, CA, CB>,
   bc: PipeFn<B, C, CB, CC>,
-): (input: A, ctx: CC) => C
+): (input: A, ctx: CA) => EnsurePromise<C>
```
flybayer, line 30: The core issue here was the type of `ctx` in this return function. Before we used the ctx type of the **last** function in the pipe. But this fixes it to instead be the ctx type of the **first** function in the pipe.

And then `EnsurePromise` fixes the return type in cases where the resolver pipe doesn't have an async function (resolver.pipe always returns a promise regardless of inputs)

hasparus: Does `EnsurePromise` truly match the runtime here? Doesn't that async function inside of `pipe` actually create a new `Promise` while EnsurePromise would preserve your "custom promises" (e.g. Bluebird).

```
interface MyPromise<T> extends PromiseLike<T> {
    abort(): void;
}

// MyPromise<number>
type result = EnsurePromise<MyPromise<number>>
```

flybayer: Currently we don't support custom promises. 

Without this conditional when you have an async function in the pipe, then the return type is `Promise<Promise<Result>>` which breaks types in the frontend with useQuery

Issue #1752, opened 2021-01-19
Queries run before session data is fully loaded


flybayer:
### What is the problem?

Queries run while `session.isLoading === true`. This causes the following code to first run with wrong input data:

```ts
const session = useSession()
const [data] = useQuery(getData, {where: {id: session.userId}})
```

### Steps to Reproduce

1. Use above query and watch server logs. You will see `{where: {id: undefined}}`

### Versions

```
0.29.2
```

### Other

I attempted a fix in https://github.com/blitz-js/blitz/pull/1751 but that didn't work because `useEffect` doesn't run when a component is suspended. However the use query hooks do re-render if `router` changes (which we depend on by waiting for `router.isReady === true` before running the query).

### How To Fix

We need to change `useSession` and publicdata implementation to use React Context instead of the proxy thing we are using currently. 

- Add our own `SessionContext` provider to keep session data
- Change `useSession` to use from that context. 
- Add a new `withBlitzAppRoot` HOC which will wrap the component with `SessionContext.Provider` (you can test by manually adding to `_app.tsx`
- Add babel plugin to automatically wrap default export in `_app` with `withBlitzAppRoot`
  - Our babel config is here: https://github.com/blitz-js/blitz/blob/canary/packages/babel-preset/src/index.ts
  - You can add `packages/babel-preset/src/add-blitz-app-root.ts` and then import and include that file in the above preset file.

jan-wilhelm:
could you explain briefly what's causing this? From my understanding, I would've assumed that useSession simply suspends while loading, causing React to stop rendering and not calling any consecutive hooks.

flybayer:
`useSession` doesn't use suspense (maybe it should)

mtford90:
Seem to be experiencing something like this myself. If I log into my app, some of the queries run before the session is configured - this causes an AuthenticationError which boots the user back to the login screen immediately.

If I then refresh the page all is well & the user is logged in.

mtford90:
As expected the following workaround seems to fix this issue:

```ts
import { useSession } from "@blitzjs/core"
import { FirstParam, PromiseReturnType, QueryFn } from "@blitzjs/core/dist/packages/core/src/types"
import { QueryOptions } from "react-query"
import { useQuery } from "blitz";

// TODO: Remove this once Blitz prevents queries being fired when the session is loading
[...]

PR #1885, opened 2021-02-08
Disable queries until session is loaded.

engelkes-finstreet:
Closes: [1752](https://github.com/blitz-js/blitz/issues/1752)


mtford90:
Careful with this - because `useSession` does not suspend I found that this fix broke suspense for all usage of `useQuery` - see my further comments in #1752

engelkes-finstreet:
Thank you for the clarification. I misunderstood your comment in the other thread.

mtford90:
no problem @engelkes-finstreet - i had a solution working locally for this within my own project - I just drafted a PR to show you how I handled this:

https://github.com/blitz-js/blitz/pull/1888

flybayer:
Hey @engelkes-finstreet thanks for the attempt, but as I described in https://github.com/blitz-js/blitz/issues/1752, I already tried this and it doesn't work.

The proper fix is to change useSession to use context (which I describe in the issue).

So I'll go and close this :)

PR #1888, opened 2021-02-08
Change `useSession` to use suspense by default

mtford90:
Closes https://github.com/blitz-js/blitz/issues/1752

@engelkes-finstreet @flybayer this is what I had in mind for fixing #1752 - seems to be working well within my own project but like I said in the issue comments - not sure if my approach is correct as I'm new to both suspense & this codebase. 

What made this less straightforward to fix is that `useSession` did not suspend and therefore `PublicData.isLoading` could not be used to determine whether or not the query is enabled without breaking `suspense` for all `useQuery` invocations...

Let me know what you think about the solution and perhaps we can solidify this.


flybayer:
Thank you @mtford90, this sort of puts a band-aid on the problem instead of fixing the real issue. To fix the real issue we need to add a React Context for the session state as described in https://github.com/blitz-js/blitz/issues/1752

Regarding useSession suspense, I have another idea to explore to eliminate the need first render problem with isLoading/suspense.

flybayer:
Well, I've added a session context provider and that didn't fix it. So you might be right that we need to implement suspense here. Still investigating

flybayer:
**Update!** So it turns out we don't actually have to throw a promise on the client. If we throw a promise on the server, then we can immediately set public data on the client in the first render. I've made this change. It greatly simplifies the implementation.

flybayer:
@all-contributors add @mtford90 for code

mtford90:
much simpler - good work!

Issue #1874, opened 2021-02-05
When using useQuery with suspense disabled, the typings should suggest that the data can be undefined whilst loading


mtford90:
### What is the problem?

```
const [data] = useQuery(foo, null, {suspense: false});

const x = data.x; // This should be a type error (accessing a property on possible undefined)
```

### Versions

```
You are using beta software - if you have any problems, please open an issue here:
      https://github.com/blitz-js/blitz/issues/new/choose

debug: blitzPkgPath: /Users/mtford/Playground/bo/app/node_modules/blitz/dist/index.js
debug: cliPkgPath: /Users/mtford/Playground/bo/app/node_modules/blitz/node_modules/@blitzjs/cli/lib/src/index.js 

macOS Catalina | darwin-x64 | Node: v12.20.0

blitz: 0.29.7 (global)
blitz: 0.29.7 (local)

  Package manager: yarn 
  System:
    OS: macOS 10.15.6
    CPU: (16) x64 Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
    Memory: 32.82 GB / 64.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 12.20.0 - ~/.nvm/versions/node/v12.20.0/bin/node
    Yarn: 1.22.4 - ~/.nvm/versions/node/v12.20.0/bin/yarn
    npm: 6.14.8 - ~/.nvm/versions/node/v12.20.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  npmPackages:
    @prisma/cli: ~2.16.0 => 2.16.0 
    @prisma/client: ~2.16.0 => 2.16.0 
    blitz: 0.29.7 => 0.29.7 
    react: 0.0.0-experimental-3310209d0 => 0.0.0-experimental-3310209d0 
    react-dom: 0.0.0-experimental-3310209d0 => 0.0.0-experimental-3310209d0 
    typescript: 4.1.3 => 4.1.3 

```

flybayer:
Fixed in #1893

PR #1893, opened 2021-02-09
Fix useQuery hooks return type when `enabled` or `suspense` could be false (patch)

LBrian:
Closes: ??

As the issue #1874, currently `data` of `useQuery` type is always casted to `TResult` which is not implied the actual union type of `data`, also there is no `resolvedData` in `InfiniteQueryResult` but `data`.

### Checklist


flybayer:
Hey @LBrian thank you for the effort on this. Currently we have the types optimized for the case when you are using suspense, and we want to keep that. The type change you have here is a regression for our "happy path" (with suspense).

So the correct fix is to adjust the return type such that result is only possibly undefined if `suspense: false` or `enabled: false`

LBrian:
> regression



> Hey @LBrian thank you for the effort on this. Currently we have the types optimized for the case when you are using suspense, and we want to keep that. The type change you have here is a regression for our "happy path" (with suspense).
> 
> So the correct fix is to adjust the return type such that result is only possibly undefined if `suspense: false` or `enabled: false`

Hi @flybayer, thanks for the feedback. I aware the "happy path" is intentional and was thinking to just fix for the cases, but I had put myself into conundrums in the past by casting type intentionally for "happy cases" and that's why I suggested a regression solution. Anyway, I can totally understand and will revert my solution.

LBrian:
I believe we don't need to handle `suspenseEnabled` in the `blitz.config.js` since we only expect TS raise error when explicit `suspense: false` and `enabled: false`

flybayer:
Sweet, very nice!!! I just pushed a commit that makes it even stricter, so that result includes undefined unless suspense and enabled are explicitly `true`. If they are set to a variable, then result will include undefined since we don't know what the variable will be at runtime.

flybayer:
@all-contributors add @LBrian for code

flybayer (approved):

Issue #1868, opened 2021-02-04
Prompt user to run prisma migration during `blitz generate`


flybayer:
### What do you want and why?

Currently `blitz generate` simply prints a message about running `prisma migrate`.

We should instead prompt the user with yes/no, do you want to run `prisma migrate`. If yes, then we will run prisma migrate for them, saving them an extra step of typing.

Also, we need to only do this if prisma is installed.

```
~/c/b/e/auth> blitz g all task --parent organization
You are using beta software - if you have any problems, please open an issue here:
      https://github.com/blitz-js/blitz/issues/new/choose

✔ Model for 'task' created successfully:

> model Task {
>   id        Int      @default(autoincrement()) @id
>   createdAt DateTime @default(now())
>   updatedAt DateTime @updatedAt
>   room      String
> }

Now run `blitz prisma migrate dev --preview-feature` to add this model to your database

CREATE    app/pages/organizations/[organizationId]/tasks/[taskId].tsx
CREATE    app/pages/organizations/[organizationId]/tasks/[taskId]/edit.tsx
CREATE    app/pages/organizations/[organizationId]/tasks/index.tsx
CREATE    app/pages/organizations/[organizationId]/tasks/new.tsx
CREATE    app/tasks/components/TaskForm.tsx
CREATE    app/tasks/queries/getTask.ts
CREATE    app/tasks/queries/getTasks.ts
CREATE    app/tasks/mutations/createTask.ts
CREATE    app/tasks/mutations/deleteTask.ts
CREATE    app/tasks/mutations/updateTask.ts
```

camilo86:
Is anyone working on this? If not, I can get a pr soon

PR #1894, opened 2021-02-09
Automatically prompt to run `prisma migrate` after `blitz generate` (minor)

camilo86:
Closes: #1868

Adds a prompt to automatically run Prisma migrations when running `blitz generate`.

Screen Shot 2021-02-08 at 10 36 35 PM

### Checklist

### DIFF
--- a/packages/cli/src/commands/generate.ts
+++ b/packages/cli/src/commands/generate.ts
@@ -64,11 +64,11 @@ function ModelNames(input: string = "") {
 
 const generatorMap = {
   [ResourceType.All]: [
-    ModelGenerator,
     PageGenerator,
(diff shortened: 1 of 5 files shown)


flybayer:
Sweet, thank you!!!

I took liberty to fix some issues.

- Migrate must be in the `postWrite()` hook because schema file hasn't been written to disk yet in `write()`
- Switched prompt to use `confirm` type instead of `select
- Use `npm-run` instead of `cross-spawn` which ensures it uses prisma binary from node_modules
- Re-organized the generator order so that model is last

flybayer (dismissed):

flybayer (approved):

Issue #1884, opened 2021-02-07
0.30.0-canary.6 - Reset password doesn't submit form and throws an error


Slein2012:
### What is the problem?
If a password reset link is requested via forget-password everything works fine, but when opening the link to reset-password?token=... it is not possible to submit the form. There is no error message in the browser's console or in the executing terminal. When commenting the form's schema with //schema={ResetPassword} , the form submits, but throws the error: 
ERROR resetPassword() 
 Error  1 validation issue(s)

  Issue #0: invalid_type at token
  Required

### Steps to Reproduce

1. blitz new app
2. run blitz, create an account and logout
3. go to forget-password and insert email, then click on the generated reset-password link
4. insert new password and confirm password, submit -> nothing happens
5. comment schema={ResetPassword} in app\auth\pages\reset-password.tsx -> submitting works, but throws error invalid_type at token

### Versions

debug: blitzPkgPath: *\node_modules\blitz\dist\index.js
debug: cliPkgPath: *\node_modules\blitz\node_modules\@blitzjs\cli\lib\src\index.js 

Windows 10 | win32-x64 | Node: v14.15.0

blitz: 0.30.0-canary.6 (local)

  Package manager: npm
  System:
    OS: Windows 10 10.0.19042
    CPU: (4) x64 Intel(R) Core(TM) i5-6400 CPU @ 2.70GHz
    Memory: 470.78 MB / 7.95 GB
  Binaries:
    Node: 14.15.0 - C:\Program Files\nodejs\node.EXE
    Yarn: Not Found
    npm: 6.14.8 - C:\Program Files\nodejs\npm.CMD
    Watchman: Not Found
  npmPackages:
    @prisma/client: ~2.16 => 2.16.0
    blitz: 0.30.0-canary.6 => 0.30.0-canary.6
    prisma: ~2.16 => 2.16.0
    react: 0.0.0-experimental-3310209d0 => 0.0.0-experimental-3310209d0
    react-dom: 0.0.0-experimental-3310209d0 => 0.0.0-experimental-3310209d0
    typescript: 4.1.3 => 4.1.3

### Other

Error message from step five:
```
INFO resetPassword() Starting with input: 
{
  password: '[***]',
  passwordConfirmation: '[***]'
}
20:45:47.879 ERROR resetPassword() 
 Error  1 validation issue(s)

  Issue #0: invalid_type at token
  Required

details:
{
  errors: [
    {
      code: 'invalid_type',
      expected: 'string',
[...]

PR #1902, opened 2021-02-10
(newapp) fix react-hook-form reset password page

flybayer:
Closes: https://github.com/blitz-js/blitz/issues/1884

fix react-hook-form reset password page


Slein2012 (changes requested):
In the new line 30 it should be `token: query.token as string` in my opinion.

flybayer:
@Slein2012 yes you are right. Fixed in https://github.com/blitz-js/blitz/commit/190bc65c53f47ebd3aa8ba0738a581c4f3cf36fb

PR #1909, opened 2021-02-11
Automatically clear console on `blitz dev` (can disable with `cli.clearConsoleOnBlitzDev = false` in blitz.config.js

JuanM04:

Like Snowpack, the console will be cleaned when `blitz dev` ~~or `blitz start` are~~ is run. This should be a QoL feature for the developers.

### Checklist

### DIFF
--- a/packages/cli/src/commands/dev.ts
+++ b/packages/cli/src/commands/dev.ts
@@ -1,6 +1,4 @@
-import {log} from "@blitzjs/display"
 import {ServerConfig} from "@blitzjs/server"
-import {getConfig} from "@blitzjs/config"
 import {Command, flags} from "@oclif/command"
 
 export class Dev extends Command {
@@ -39,9 +37,11 @@ export class Dev extends Command {
 
     try {
       const dev = (await import("@blitzjs/server")).dev
+      const {getConfig} = await import("@blitzjs/config")
 
       const blitzConfig = getConfig()
       if (blitzConfig.cli?.clearConsoleOnBlitzDev !== false) {
+        const {log} = await import("@blitzjs/display")
         log.clearConsole()
       }
 
--- a/packages/cli/src/commands/dev.ts
+++ b/packages/cli/src/commands/dev.ts
@@ -22,6 +22,9 @@ export class Dev extends Command {
     ["no-incremental-build"]: flags.boolean({
       description: "Disable incremental build and start from a fresh cache",
     }),
+    ["no-clear-console"]: flags.boolean({
+      description: "Prevent console clearing",
+    }),
   }
 
   async run() {
@@ -38,7 +41,9 @@ export class Dev extends Command {
 
     try {
       const dev = (await import("@blitzjs/server")).dev
-      log.clearConsole()
+
+      if (!flags["no-clear-console"]) log.clearConsole()
+
       await dev(config)
     } catch (err) {
       console.error(err)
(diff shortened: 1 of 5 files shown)


Skn0tt:
Hi @JuanM04! Thanks for taking the time to contribute.

Do you think we should skip the log-clearing step while running in CI? Also, how would this work in the context of e.g. `concurrently`?

JuanM04:
Hmm... @Skn0tt maybe we could add a flag. Like `--no-clear-console`

JuanM04:
Also, but I don't know if this would work, we could make it clear the console only if `NODE_ENV === 'production'`

Skn0tt:
I see how it's desirable to have your console cleared upon running `blitz start` - but for me personally, I'd be totally fine just running `clear` before. How does everybody else think about this? Is this really a QoL improvement?

JuanM04:
@Skn0tt I think you are right, that it may not be in `blitz start`, but yes in `blitz dev`. I changed the code and added `--no-clear-console`

paulm17:
> I see how it's desirable to have your console cleared upon running `blitz start` - but for me personally, I'd be totally fine just running `clear` before. How does everybody else think about this? Is this really a QoL improvement?

I actually find myself doing clear before blitz dev lol.  But either way, I don't have a strong opinion either way.

flybayer:
Thanks @JuanM04! I think we should make this configurable from `blitz.config.js`. (Then use `getConfig` from `@blitzjs/config` to read this config item in the cli command file)

```js
module.exports = {
  cli: {
    clearConsoleOnBlitzDev: boolean
  },
}
```

I'm not sure what the default should be. I personally would disable this on all my projects.

I've started a poll here: https://twitter.com/blitz_js/status/1359897748393312259

JuanM04:
@flybayer Perfect! I'll remove `--no-clear-console` and add `clearConsoleOnBlitzDev`. Also, I'll put `true` as default, but I'll change it if the poll says so

flybayer:
Cool sounds good. Also looks like you have a failing test

packages/cli/src/commands/dev.ts
```diff
@@ -1,4 +1,6 @@
+import {log} from "@blitzjs/display"
 import {ServerConfig} from "@blitzjs/server"
+import {getConfig} from "@blitzjs/config"
```
flybayer, line 3: Change these imports to be dynamic (like `@blitzjs/server` is below). This is so that the code is only parsed and loaded when using this command.

JuanM04: Something like this?

```js
try {
  const dev = (await import("@blitzjs/server")).dev
  const {log} = (await import("@blitzjs/display"))
  const {getConfig} = (await import("@blitzjs/config"))
```

flybayer: The first is fine. It's not a network request, so it's fast and doesn't need Promise.all

flybayer:
In addition to fixing the test, can you also open a PR to the docs repo to document this?

JuanM04:
Sure. Also, I'm trying to figure out where that error comes from, because it's from `examples/auth/app/auth/mutations/forgotPassword.test.ts` and it's a timeout 🤔

flybayer:
Looks like test failure is not related to your change. I pushed a commit that increases default timeout for server env

flybayer:
Thanks again!

@all-contributors add @JuanM04 for code

flybayer:
@all-contributors add @JuanM04 for docs

PR #1940, opened 2021-02-15
allow useQuery to run during SSG

jxe:
This allows useQuery to run on the server, iff there's precached data. We do this by calling useReactQuery with enabled:false and seeing if there's data returned. If there's no data, we throw a suspender to emulate the previous behavior.

To prefetch data, you can run something like this in getStaticProps

```js
import { QueryCache } from "react-query"
const cache = new QueryCache()
const queryKey = getQueryKey(queryFn, queryArgs)
await cache.prefetchQuery(queryKey, () => queryFn(queryArgs))
```

and then you can pass the resulting cache to the client with 

```js
import { dehydrate } from "react-query/hydration"
return { props: { queryCache: dehydrate(cache) } }
```

So you get identical results on first render.


MrLeebo:
@jxe @flybayer I resolved the merge conflicts and added an exclusion so useQuery doesn't run during SSG when in legacyMode. The failing test passed on my local environment though...

packages/core/src/use-query-hooks.ts
```diff
@@ -49,14 +50,15 @@ export function useQuery<T extends QueryFn, TResult = PromiseReturnType<T>>(
     throw new Error("useQuery is missing the first argument - it must be a query function")
   }
 
-  const suspense =
-    options?.enabled === false || options?.enabled === null ? false : options?.suspense
+  const legacyMode = getConfig?.().experimental?.reactMode === "legacy"
```
flybayer, line 53: @MrLeebo as of nextjs 10.2 (already on our canary), it's now `experimental.reactRoot = true/false`

MrLeebo:
Switched to reactRoot and would you look at that, a green checkmark! It's what I was hoping for, but you can never be sure what you're going to get 😁

flybayer:
Ok last thing @MrLeebo, do we need to update or add any docs?

MrLeebo:
@flybayer Definitely. At minimum, the content at https://blitzjs.com/docs/pages#automatic-static-optimization will need to be updated, and there may be other broader repercussions I'm not even aware of yet.

Issue #1972, opened 2021-02-19
Change session cookie prefix to have an explict config instead of using package.json `name` field


flybayer:
### What do you want and why?

Currently we are using the name field of your `package.json` for setting the session cookie prefix. [As @lcswillems highlighted](https://github.com/blitz-js/blitz/issues/1957#issuecomment-781992049), this isn't ideal.

So we should change this to be an explicit config:

```diff
module.exports = {
  middleware: [
    sessionMiddleware({
+      cookiePrefix: "myApp",
      isAuthorized: simpleRolesIsAuthorized,
    }),
  ],
}
```

And we'll add that to templates for new apps.

### Possible implementation(s)

1. Update config type: https://github.com/blitz-js/blitz/blob/canary/packages/core/src/supertokens.ts#L21
2. Change `sessionMiddleware` implementation to return `name` and `config` static properties on the function, something like this:

```ts
export const sessionMiddleware = (sessionConfig: Partial<SessionConfig> = {}): Middleware => {
  assert(
    sessionConfig.isAuthorized,
    "You must provide an authorization implementation to sessionMiddleware as isAuthorized(userRoles, input)",
  )
  config = {
    ...defaultConfig,
    ...sessionConfig,
  } as Required<SessionConfig>

  const middleware = async (req, res, next) => {
    if (req.method !== "HEAD" && !res.blitzCtx.session) {
      // This function also saves session to res.blitzCtx
      await getSessionContext(req, res)
    }
    return next()
  }
  middleware.name = 'blitzSessionMiddleware'
  middleware.config = {
    cookiePrefix:
  }
  return middleware
}
```
3. Change the blitz runtime data function to read `config.middleware`, find middleware with `name = 'blitzSessionMiddleware'` and then get the cookie prefix config from that:
   - Here: https://github.com/blitz-js/blitz/blob/canary/packages/core/src/blitz-data.tsx#L11-L17
4. Add `cookiePrefix` to new app template as `cookiePrefix: "__name__"`:
   - Here: https://github.com/blitz-js/blitz/blob/canary/packages/generator/templates/app/blitz.config.js#L4-L8

piotrski:
Can I grab this one?

flybayer:
Yes!!

flybayer:
Let me know if you have any questions as you go

piotrski:
[...]

Issue #1957, opened 2021-02-18
Weird error if no name field in package.json


lcswillems:
### Steps to Reproduce

Create a new Blitzjs app: `npx blitz new blitz-app`

Remove the `name` field from the `package.json` file.

Run `yarn build`.

This gives this error:

```
TypeError: Cannot read property 'replace' of undefined
    at ee (/mnt/Data/Documents/Travail/Personnel/Projets/Ouverts/Miple/Code/lab/blitz-app/node_modules/@blitzjs/core/dist/core.cjs.production.min.js:1:5987)
    at re (/mnt/Data/Documents/Travail/Personnel/Projets/Ouverts/Miple/Code/lab/blitz-app/node_modules/@blitzjs/core/dist/core.cjs.production.min.js:1:6380)
    at d (/mnt/Data/Documents/Travail/Personnel/Projets/Ouverts/Miple/Code/lab/blitz-app/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:34:498)
    at bb (/mnt/Data/Documents/Travail/Personnel/Projets/Ouverts/Miple/Code/lab/blitz-app/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:37:16)
    at a.b.render (/mnt/Data/Documents/Travail/Personnel/Projets/Ouverts/Miple/Code/lab/blitz-app/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:43:48)
    at a.b.read (/mnt/Data/Documents/Travail/Personnel/Projets/Ouverts/Miple/Code/lab/blitz-app/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:42:83)
    at exports.renderToStaticMarkup (/mnt/Data/Documents/Travail/Personnel/Projets/Ouverts/Miple/Code/lab/blitz-app/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:53:446)
    at renderDocument (/mnt/Data/Documents/Travail/Personnel/Projets/Ouverts/Miple/Code/lab/blitz-app/node_modules/next/dist/next-server/server/render.js:3:715)
    at renderToHTML (/mnt/Data/Documents/Travail/Personnel/Projets/Ouverts/Miple/Code/lab/blitz-app/node_modules/next/dist/next-server/server/render.js:56:103)
[...]

PR #1976, opened 2021-02-19
Use cookie prefix from `config.blitz.js` instead of package.json `name` field

piotrski:
Closes: #1972, #1957

Adds new `cookiePrefix` property to `sessionMiddleware` config.
```diff
module.exports = {
  middleware: [
    sessionMiddleware({
+     cookiePrefix: 'cookie',
      isAuthorized: simpleRolesIsAuthorized,
    }),
  ],
}
```

Instead of using the name from your `package.json` as a session cookie prefix, it uses the `cookiePrefix` from `blitz.config.js`

Should we add information on this somewhere in the docs?

### Checklist


packages/config/src/index.ts
```diff
@@ -33,8 +33,14 @@ export interface BlitzConfig extends Record<string, unknown> {
   _meta: {
     packageName: string
   }
+  middleware?: {
+    (req: any, res: any, next: any): Promise<void> | void
+    type?: string
+    config?: {
+      cookiePrefix?: string
+    }
+  }[]
```
flybayer, line 42: 1. Change `type` to `name` here
2. And allow any attributes on middleware. So probably extract this to an interface that `extends Record<string, any>`

packages/core/src/blitz-data.tsx
```diff
@@ -10,8 +10,11 @@ export type BlitzRuntimeData = {
 
 export function _getBlitzRuntimeData(): BlitzRuntimeData {
   const config = getConfig()
+  const cookiePrefix = config.middleware?.filter(
+    (middleware) => middleware.type === "blitzSessionMiddleware",
```
flybayer, line 14:
```suggestion
    (middleware) => middleware.name === "blitzSessionMiddleware",
```

packages/generator/templates/app/blitz.config.js
```diff
@@ -4,6 +4,7 @@ module.exports = {
   middleware: [
     sessionMiddleware({
       isAuthorized: simpleRolesIsAuthorized,
+      cookiePrefix: '__name__',
```
flybayer, line 7: A bit pedantic but I think this reads better :)
```suggestion
      cookiePrefix: '__name__',
      isAuthorized: simpleRolesIsAuthorized,
```

packages/server/src/supertokens.ts
```diff
@@ -151,13 +151,20 @@ export const sessionMiddleware = (sessionConfig: Partial<SessionConfig> = {}): M
...
+  const middleware: Middleware<{
+    cookiePrefix?: string
+  }> = async (req, res, next) => {
     if (req.method !== "HEAD" && !res.blitzCtx.session) {
       // This function also saves session to res.blitzCtx
       await getSessionContext(req, res)
     }
     return next()
   }
+  middleware.type = "blitzSessionMiddleware"
```
flybayer, line 163:
```suggestion
  middleware.name = "blitzSessionMiddleware"
```

flybayer (changes requested):
Awesome, great job!

Main change is use `name` instead of `type` because we might need `type` for something else later that actually indicates type.

Then please open a PR to the docs repo to update the docs for this change

packages/config/src/index.ts
piotrski: I've changed it to type because the name is read-only `Cannot assign to 'name' because it is a read-only property.`

piotrski: I could do it this way
```tsx
  const blitzSessionMiddleware: Middleware<{
    cookiePrefix?: string
  }> = async (req, res, next) => {
    if (req.method !== "HEAD" && !res.blitzCtx.session) {
      // This function also saves session to res.blitzCtx
      await getSessionContext(req, res)
    }
    return next()
  }
  blitzSessionMiddleware.config = {
    cookiePrefix: config.cookiePrefix,
  }
  return blitzSessionMiddleware
```
This way `name === blitzSessionMiddleware` but is it clear enough?

flybayer: Ah yeah I see. And yeah let's roll with that for now

flybayer:
Hey @piotrski, are you able to finish this up?

piotrski:
Sure. Sorry I had a lot of stuff on my mid and was sure it was already done.

packages/config/src/index.ts
piotrski: Should I switch it or leave the current implementation with type?

flybayer:
@piotrski ok no problem :) Should have said something earlier

flybayer:
@mabadir can you resolve the merge conflicts too?

mabadir:
Yes, next on my list.

mabadir:
The file `packages/core/src/supertokens.ts` was recently removed from `canary. I see @piotrski; made a change to this file to include the `cookiePrefix` to the configuration.
I assume it is safe to remove this file from the PR, as it’s not used in the cores package any longer.

@flybayer any thoughts?

flybayer:
@mabadir 
`packages/core/src/supertokens.ts` is now https://github.com/blitz-js/blitz/blob/canary/packages/core/src/auth/auth-types.ts

and `packages/server/src/supertokens.ts` is now `https://github.com/blitz-js/blitz/blob/canary/packages/core/src/server/auth/sessions.ts`

flybayer:
@mabadir and then we also need a PR to the docs repo to update docs for this change

Issue #1953, opened 2021-02-18
Invalid hook call when installing recipes with NPM


dmvinson:
### What is the problem?
When running the `blitz install tailwind` command, the install fails with a "Invalid hook call" React error.

### Steps to Reproduce

1. Create new blitz project
2. Run `blitz install tailwind`

### Versions

```
You are using beta software - if you have any problems, please open an issue here:
      https://github.com/blitz-js/blitz/issues/new/choose

debug: blitzPkgPath: /Users/me/projects/test/node_modules/blitz/dist/index.js
debug: cliPkgPath: /Users/me/projects/test/node_modules/@blitzjs/cli/lib/src/index.js

macOS Catalina | darwin-x64 | Node: v15.5.1

blitz: 0.30.0 (global)
blitz: 0.30.0 (local)

  Package manager: npm
  System:
    OS: macOS 10.15.6
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 1.01 GB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 15.5.1 - /usr/local/bin/node
    Yarn: Not Found
    npm: 7.3.0 - /usr/local/bin/npm
    Watchman: Not Found
  npmPackages:
    @prisma/client: ~2.17 => 2.17.0
    blitz: 0.30.0 => 0.30.0
    prisma: ~2.17 => 2.17.0
    react: 0.0.0-experimental-3310209d0 => 0.0.0-experimental-3310209d0
    react-dom: 0.0.0-experimental-3310209d0 => 0.0.0-experimental-3310209d0
    typescript: 4.1.5 => 4.1.5

```

### Other

Please include applicable logs and screenshots that show your problem.
Command Output:
```
➜  test git:(master) blitz install tailwind
You are using beta software - if you have any problems, please open an issue here:
      https://github.com/blitz-js/blitz/issues/new/choose


  ERROR  Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
        1. You might have mismatching versions of React and the renderer (such as React DOM)
        2. You might be breaking the Rules of Hooks
        3. You might have more than one copy of React in the same app
        See https://reactjs.org/link/invalid-hook-call for tips about how to debug and fix this problem.

 - 1. You might have mismatching versions of React and the renderer (such as React DOM)
 - 2. You might be breaking the Rules of Hooks
[...]

PR #1986, opened 2021-02-22
Fix `blitz install` when using npm (patch)

hirenchauhan2:
Closes: #1953

When installing any recipe via npm as the default package manager, then it will pass `install` instead of `add` which is for yarn.

### Checklist

### DIFF
--- a/packages/installer/src/executors/add-dependency-executor.tsx
+++ b/packages/installer/src/executors/add-dependency-executor.tsx
@@ -81,10 +81,12 @@ export function getPackageManager() {
  */
 export async function installPackages(packages: NpmPackage[], isDev = false) {
   const packageManager = getPackageManager()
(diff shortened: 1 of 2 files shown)


piotrski:
It will still fail for the user who starts a new blitz project and has no yarn installed.
```
blitz new test
cd test
blitz install emotion
```
When there is no `package-lock.json` we assume the user is using yarn, but we should also check if the yarn is installed.

flybayer:
Sweet, thank you @hirenchauhan2!!

I pushed a commit to address @piotrski concern. Now we default to npm if `yarn.lock` isn't present.

@all-contributors add @hirenchauhan2 for code

flybayer (dismissed):

flybayer (dismissed):

androsj:
Still not working with v0.36,0

PR #1995, opened 2021-02-23
(newapp) Update `global.d.ts` to reference types from nextjs

FDiskas:
Just use reference to nextjs typings. It is already includes such things as `module.scss` and more

Closes: ??

Removes redundant repetetive code. As blidz deaply uses nextjs it also should refare typings from such dependency.

### Checklist


FDiskas:
Related to PR https://github.com/blitz-js/blitz/pull/1929

flybayer:
Ah, very nice!! Can you add a comment explaining what those reference types provide?

FDiskas:
I'm bad in English. Just say what to write and I will do. And in normal nextjs project when you run it first time the file `next-env.d.ts` is generated in the root of your project and content is like in this pull request.
 But if we dive in to the reference `next/types/global`. 
```ts
/// <reference types="node" />

// Extend the NodeJS namespace with Next.js-defined properties
declare namespace NodeJS {
  interface Process {
    readonly browser: boolean
  }

  interface ProcessEnv {
    readonly NODE_ENV: 'development' | 'production' | 'test'
  }
}

declare module '*.module.css' {
  const classes: { readonly [key: string]: string }
  export default classes
}

declare module '*.module.sass' {
  const classes: { readonly [key: string]: string }
  export default classes
}

declare module '*.module.scss' {
  const classes: { readonly [key: string]: string }
  export default classes
}
```

FDiskas:
@flybayer I noteiced that there is `.prettierignore` but `.prettierrc` is missing. Prettier should be included?

packages/generator/templates/app/types.d.ts
```diff
@@ -1,8 +1,2 @@
-declare module "*.module.scss" {
-  const content: { [className: string]: string }
-  export default content
-}
-declare module "*.module.css" {
-  const content: { [className: string]: string }
-  export default content
-}
+/// <reference types="next" />
+/// <reference types="next/types/global" />
```
flybayer, line 2:
```suggestion
// These reference imports provide type definitions for things like styled-jsx and css modules
/// <reference types="next" />
/// <reference types="next/types/global" />
```

flybayer:
> @flybayer I noteiced that there is `.prettierignore` but `.prettierrc` is missing. Prettier should be included?

The prettier config is inside `package.json` :)

flybayer:
Alright on the type change: I added a comment for you. And I want to make this change, but I'm going to wait until we fork next.js to merge this to prevent a breaking change. Because the import path will change to `blitz` instead of `next`

@all-contributors add @FDiskas for code

Issue #1947, opened 2021-02-15
TypeScript does not identify invalid CSS/SCSS module paths


remjx:
### What do you want and why?

This PR https://github.com/blitz-js/blitz/pull/1929 declares `*.module.css` files as valid for import. This means it is possible to `import styles from some-file-that-doesnt-exist.module.css` and TypeScript compiles without any error.

Is there any way it can check that the file actually exists?

### Possible implementation(s)

Unknown

### Additional context

N/A

flybayer:
@remjx I'm not aware of any way to do that via TS, but if the file is missing the build will fail.

Issue #1531, opened 2020-11-26
(CSS) module not found


gielcobben:
### What is the problem?
Since I’ve updated to `v0.26.0` I’m having trouble with importing css files. Maybe it has something todo with #1456 ? I `rm -rf` the directory `.blitz` and installed all versions backwards. Everything works perfect with- and before `v0.25.1-canary.0`. I’m using css modules and a file structure shown below.

##### File structure:
```
/app/users/pages/signin/index.tsx
/app/users/pages/signin/index.module.scss

/app/users/components/SignInForm/index.tsx
/app/users/components/SignInForm/index.module.scss
```

##### How I import css modules:
```
import styles from "./index.module.scss"
```

##### Error:
```
Failed to compile
./pages/users/test/index.tsx:3:0
Module not found: Can't resolve 'users/pages/users/test/index.module.scss'

  1 | import * as React from "react"
  2 | 
> 3 | import styles from "users/pages/users/test/index.module.scss"
  4 | 
  5 | const Test: React.FC = () => {
  6 |   return (
```

### Steps to Reproduce

1. Create a new blitz app by running: `blitz new _bug-css-modules`.
1. Create a new page, for example: `app/users/pages/users/test/index.tsx`
1. Add a css file, for exmaple: `app/users/pages/users/test/index.module.scss`
1. Run the project by running: `blitz start`
1. Navigate to the url of the page you just created: `localhost:3000/users/test`

### Versions

```
debug: blitzPkgPath: /Users/giel/Sites/_bug-css-modules/node_modules/blitz/dist/index.js
debug: cliPkgPath: /Users/giel/Sites/_bug-css-modules/node_modules/@blitzjs/cli/lib/src/index.js

macOS | darwin-x64 | Node: v12.18.0

blitz: 0.26.0 (global)
blitz: 0.26.0 (local)

  Package manager: yarn
  System:
    OS: macOS 11.0.1
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 57.70 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 12.18.0 - ~/.nvm/versions/node/v12.18.0/bin/node
    Yarn: 1.21.1 - /usr/local/bin/yarn
    npm: 6.14.4 - ~/.nvm/versions/node/v12.18.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  npmPackages:
[...]

PR #1997, opened 2021-02-23
typescript support for module css

FDiskas:
Closes: https://github.com/blitz-js/blitz/issues/1947 https://github.com/blitz-js/blitz/issues/1531 

The main reason is to have autocomplete of module css classes in components using module css.
deepin-screen-recorder_Select area_20210223225400

### Checklist

If you are using vscode also you should change typescript version to locally installed one
```
{
  "typescript.tsdk": "node_modules/typescript/lib"
}
```

### DIFF
--- a/packages/generator/templates/app/package.json
+++ b/packages/generator/templates/app/package.json
@@ -48,7 +48,7 @@
     "preview-email": "3.x",
     "prettier-plugin-prisma": "0.x",
     "typescript": "4.x",
-    "typescript-plugin-css-modules": "^3.x",
+    "typescript-plugin-css-modules": "^3.x"
   },
   "private": true
 }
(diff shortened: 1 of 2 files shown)


flybayer:
Hey thanks! This is a good idea, but I'm thinking maybe it should be a recipe instead because not everyone uses css modules? 

Anyone else have thoughts on this?

FDiskas:
Actually a good point to have this as recipe. But in other case - not everyone uses SCSS/SASS but it is already possible to use it out of the box and module css is also comes out of the box - so lets just fully support it

flybayer:
We can still make a recipe, but this is now also documented here: https://blitzjs.com/docs/css#typescript-support

PR #2000, opened 2021-02-23
Improve error message for missing `_document` page (patch)

flybayer:

Improve error message for missing `_document` page

### DIFF
--- a/packages/core/src/blitz-data.tsx
+++ b/packages/core/src/blitz-data.tsx
@@ -17,7 +17,7 @@ export function _getBlitzRuntimeData(): BlitzRuntimeData {
 }
 
 export function getBlitzRuntimeData() {
-  if (isClient && process.env.JEST_WORKER_ID === "undefined") {
+  if (isClient && !process.env.JEST_WORKER_ID) {
     return window.__BLITZ_DATA__
   } else {
     if (!global.__BLITZ_DATA__) {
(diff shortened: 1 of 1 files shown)


andyeskridge:
When running the client test[0] included in the default app template, jest now logs out errors[1] during the test. Before this commit, `deserializeAndSetBlitzDataOnWindow()` was guarded by `if (document.getElementById("__BLITZ_DATA__"))` which kept this error from bubbling up in jest.

[0]
```js
test("renders blitz documentation link", () => {
  // This is an example of how to ensure a specific item is in the document
  // But it's disabled by default (by test.skip) so the test doesn't fail
  // when you remove the the default content from the page

  // This is an example on how to mock api hooks when testing
  mockUseCurrentUser.mockReturnValue({
    id: 1,
    name: "User",
    email: "user@email.com",
    role: "user",
  })
  const { getByText } = render(<Home />)
  const linkElement = getByText(/Documentation/i)
  expect(linkElement).toBeInTheDocument()
})
```

[1]
```
console.error
      Error deserializing __BLITZ__DATA__. Make sure you have a custom _document.js/tsx page that uses <BlitzScript/> TypeError: Cannot read property 'textContent' of null
```

flybayer:
Thanks @andyeskridge, fixed in https://github.com/blitz-js/blitz/pull/2004

PR #2005, opened 2021-02-25
fix "role" of PublicData was lost by `$setPublicData`

marina-ki:
Closes: ??

#### 【problem】
My sessions are like this:
```
{
      isLoading: false
      role: "user"
      userId: 1
      userName: "foo"
}
```

When I use `setPublicData` without "role", "role" is gone.

```tsx
ctx.session.$setPublicData({ userName: "bar" })
```

```
{
      isLoading: false
      userId: 1
      userName: "bar"
}
```

#### 【solution】

It was because of l847(`packages/server/src/supertokens.ts`):

```
config.publicDataKeysToSyncAcrossSessions.forEach((key) => (dataToSync[key] = data[key])
```

This made `dataToSync` as follows:

```
{
  role: undefined,
  roles: undefined,
}
```

### Checklist

### DIFF
--- a/packages/server/src/supertokens.ts
+++ b/packages/server/src/supertokens.ts
@@ -844,8 +844,11 @@ export async function syncPubicDataFieldsForUserIfNeeded(
   data: Record<string, unknown>,
 ) {
   const dataToSync: Record<string, unknown> = {}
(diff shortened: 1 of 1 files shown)


flybayer:
Very nice, thank you!!

PR #2019, opened 2021-02-28
Add `blitz install secureheaders` recipe for setting secure headers

MrLeebo:

Creates a new blitz recipe for setting a default list of security headers. There's probably more improvements that we can make with this. I wasn't sure how to compute hashes for styles (especially since the user might use one of any number of CSS frameworks) so I included 'unsafe-inline' by default.

Sets the following headers:
- Content-Security-Policy (as a meta tag)
- Referrer-Policy (as a meta tag)
- Strict-Transport-Security
- X-Frame-Options
- X-Content-Type-Options
- Permissions-Policy
- X-Powered-By (disabled, nextjs enables this by default)

### Checklist

### DIFF
--- a/packages/core/src/nextjs.ts
+++ b/packages/core/src/nextjs.ts
@@ -12,7 +12,7 @@ export {default as Image} from "next/image"
 export type {ImageProps} from "next/image"
 
 export {default as Document, Html, Head as DocumentHead, Main} from "next/document"
-export type {DocumentContext, DocumentInitialProps} from "next/document"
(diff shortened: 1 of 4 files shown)


flybayer:
Thanks!! Question: should we add some of these directly to the new app template?

MrLeebo:
Potentially yes, but I wouldn't want to introduce a burden on users that don't want to pay attention to security headers because they are working on toy projects or something. `X-Powered-By` is safe to disable in the default template, but the rest are potentially breaking changes so I suppose it depends on what your tolerance for that is.

flybayer:
Ok makes sense.

Issue #1998, opened 2021-02-23
Attempting to upgrade the global blitz version from the blitz new prompt fails


Roesh:
**What is the problem?**

EDIT: There is a really simple workaround mentioned by Brandon in the posts below, npm i -g blitz@latest will upgrade the global blitz version to the latest release.

When trying to create a new blitzjs project, if your global blitz version is not the latest released version, a prompt to upgrade the global blitz version appears. Selecting yes to the prompt causes the following error:
```
blitz new linuxJs
You are using beta software - if you have any problems, please open an issue here:
      https://github.com/blitz-js/blitz/issues/new/choose

✔ Your global blitz version is outdated. Upgrade? · yes
    TypeError: Cannot read property 'toString' of null
```

**What are detailed steps to reproduce this?**

1. Create a new blitz app with blitz new <app name> when your global version is not the latest version
2. Accept the prompt to update the global blitz version
3. The error should occur

**Run `blitz -v` and paste the output here:**

```
Linux 5.4 | linux-x64 | Node: v14.15.5

blitz: 0.30.0-canary.13 (global)
blitz: 0.30.0-canary.13 (local)

  Package manager: npm 
  System:
    OS: Linux 5.4 Linux Mint 19.3 (Tricia)
    CPU: (16) x64 AMD Ryzen 7 2700 Eight-Core Processor
    Memory: 3.97 GB / 15.63 GB
    Shell: 4.4.20 - /bin/bash
  Binaries:
    Node: 14.15.5 - /usr/local/bin/node
    Yarn: Not Found
    npm: 7.5.4 - /usr/local/bin/npm
    Watchman: Not Found
  npmPackages:
    @prisma/client: 2.17.0 => 2.17.0 
    blitz: 0.30.0-canary.13 => 0.30.0-canary.13 
    prisma: 2.17.0 => 2.17.0 
    react: 0.0.0-experimental-3310209d0 => 0.0.0-experimental-3310209d0 
    react-dom: 0.0.0-experimental-3310209d0 => 0.0.0-experimental-3310209d0 
    typescript: 4.1.5 => 4.1.5

```

flybayer:
@Roesh can you please update to 0.30.3 and try again?

Roesh:
Sounds good. I'll need to learn a little more about how npx works since I am on 0.30.0-canary.15 and I'm not sure what the best method to create a single project with 0.30.3. I should be able to have that done by sometime tomorrow

flybayer:
[...]

PR #2025, opened 2021-03-01
Fix Cannot read property 'toString' of null for `blitz new`

Roesh:
Closes: #1998 

Adding checks to make sure the checkYarn buffer and its stdout are not null or undefined before calling .toString() on them. This ensures that the useYarn variable which determines whether we use yarn or npm to upgrade blitz is properly set, and the rest of the upgrade happens without an error being thrown if the user doesn't have yarn installed

### Checklist

I was unable to test this because the issue was only occurring when I didn't have yarn installed, but I needed yarn installed and needed to be running the yarn dev command in order to run the commands using the local blitz version I was working on in the example directory


flybayer:
Woohoo, thank you!

@all-contributors add @Roesh for code

Issue #2015, opened 2021-02-27
next export - Not working


BenedictWHD:
**What is the problem?**

Next export which would be required to deploy on some providers does not work out of the box due to use of `next/image` component within Blitz.

When doing next export after a blitz build, we get the following error, even though personally not making use of either the Blitz Image component or the Next Image component with the application.

Error: Image Optimization using Next.js' default loader is not compatible with `next export`.
  Possible solutions:
    - Use `next start`, which starts the Image Optimization API.
    - Use Vercel to deploy, which supports Image Optimization.
    - Configure a third-party loader in `next.config.js`.
  Read more: https://err.sh/next.js/export-image-api

**What are detailed steps to reproduce this?**

1.Create a fresh Blitz application.
2. add `next export` to "build" in package.json ("build": "blitz build && next export").
3. npm run build in console.

**Run `blitz -v` and paste the output here:**

You are using beta software - if you have any problems, please open an issue here:
      https://github.com/blitz-js/blitz/issues/new/choose

Windows 10 | win32-x64 | Node: v14.15.4

blitz: 0.30.5 (global)
blitz: 0.30.5 (local)

  Package manager: npm
  System:
    OS: Windows 10 10.0.19041
    CPU: (8) x64 Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz
    Memory: 18.57 GB / 31.94 GB
  Binaries:
    Node: 14.15.4 - D:\Code\Node\node.EXE
    Yarn: Not Found
    npm: 7.6.0 - D:\Code\Node\npm.CMD
    Watchman: Not Found
  npmPackages:
    @prisma/client: ~2.17 => 2.17.0
    blitz: 0.30.5 => 0.30.5
    prisma: ~2.17 => 2.17.0
    react: 0.0.0-experimental-3310209d0 => 0.0.0-experimental-3310209d0
    react-dom: 0.0.0-experimental-3310209d0 => 0.0.0-experimental-3310209d0
    typescript: 4.2.2 => 4.2.2

flybayer:
Hey @BenedictWHD, I'm curious, why are you using Blitz instead of Next.js for this?

BenedictWHD:
> Hey @BenedictWHD, I'm curious, why are you using Blitz instead of Next.js for this?

Hi @flybayer, 

Thanks for taking a look, is essentially as Blitz has extra features that I would like to make use of which next doesn't, specifically for me, the ability to schedule functions to run is very useful!

[...]

PR #2039, opened 2021-03-02
Major internal refactor + add multiple entry points + change all imports to be from `blitz` + rename `getSessionContext` to `getSession` + other fixes

flybayer:
Closes: https://github.com/blitz-js/blitz/issues/639
Closes: https://github.com/blitz-js/blitz/issues/2015

- 11% reduction in app bundle sizes
- Adds multiple entry points to `@blitzjs/core`, like `@blitzjs/core/image`, `@blitzjs/core/document`, `@blitzjs/core/server`, etc
- All entrypoints are exported from `blitz`, but with the new babel plugin added in this PR, imports are remapped to their dedicated entrypoint. This optimizes DX and bundle sizes at the same time
- Removes `date-fns` as a dependency
- Removes `zod` as a dependency (now a peer dependency)
- Moves rpc and auth code from `@blitzjs/server` into `@blitzjs/core/server`
- `@blitzjs/server` now only deals with build time operations. And it will go away once we fork next.js
- `core/src/supertokens.ts` moved to `core/src/auth/auth-client.ts`
- `server/src/supertokens.ts` moved to `core/src/server/auth/sessions.ts`

#### Breaking changes:

```diff
--const {sessionMiddleware, simpleRolesIsAuthorized} = require("@blitzjs/server")
++const {sessionMiddleware, simpleRolesIsAuthorized} = require("blitz")

--import {SimpleRolesIsAuthorized} from "@blitzjs/server"
++import {SimpleRolesIsAuthorized} from "blitz"

--import {getSessionContext} from "@blitzjs/server"
++import {getSession} from "blitz"

--const blitz = require("@blitzjs/server")
++const blitz = require("blitz/custom-server")
```

### Checklist


MrLeebo (dismissed):
This looks really good! The imports make a lot more sense now.

Based on the automated checks, it looks like some breaking changes still need to be fixed in the examples/ apps, but other then that, overall I'd say this is a step in the right direction! 💪

Issue #2023, opened 2021-02-28
Generate types for /pages routes


remjx:
### What do you want and why?

TypeScript types for routes in /pages

For example, given `app/pages/user/settings.tsx`, generate `type UserSettingsRoute = "/user/settings"`

This way, we can reference the route in a type-safe manner instead of relying on tests to catch the error.

### Possible implementation(s)

Create a `blitz generate types` command that scans the directory structure for /pages routes, and maps the routes to TypeScript types, e.g.

```route-types.ts
declare namespace BlitzRoutes {
  type AppPagesIndexRoute: "/"
  type UserSettingsRoute: "/user/settings"
}
```

to be used like this:

```
const homeRoute: BlitzRoutes.AppPagesIndexRoute = "/"
return <Link href={homeRoute} />
```

#### Static paths
Given `app/pages/index.tsx`, generate `type IndexRoute = "/"`
Given `app/pages/user/settings`, generate `type UserSettingsRoute = '/user/settings'`

#### Dynamic paths
I don't think this is possible since TypeScript doesn't have a way to use wildcards in string types, i.e. `type String = "app/users/*wildcard*/settings`

remjx:
@engelkes-finstreet's current approach:

```
export const routes = {
  index: "/",
  customers: {
    index: "/customers",
    new: "/customers/new",
    view: (id: number) => `/customers/${id}`,
    edit: (id: number) => `/customers/${id}/edit`,
  },
  users: {
    index: "/users",
    edit: (id: number) => `users/${id}/edit`,
  },
}
```

Skn0tt:
I like this idea! We could then use the generated type in our `Link` adaptation.

What file should we write these types into? Maybe into the .blitz directory?

flybayer:
Oh, this is a fabulous idea!! And I think we can take it a step further to provide routes as a JS object so that you can do something like this:

```tsx
import {Routes} from 'blitz'

<Link href={Routes.LoginPage}>
```

Basically do the same thing as Prisma but for routes.

It would be better if we can generate the field/type name based on something other than the filepath because ideally you can reorganize your files and routes without changing your usage of `Routes`. The first thing that comes to mind is use the name of the default export component in the page. I.e. `export default UserShowPage` would be `Routes.UserShowPage`, regardless of where it is at in the file system.

hasparus:
> Oh, this is a fabulous idea!! And I think we can take it a step further to provide routes as a JS object so that you can do something like this:

[...]

PR #2042, opened 2021-03-03
Add autogenerated `Routes` manifest for use with `<Link>` components

Skn0tt:
Closes: #2023

Give users a Route manifest as described in #2023.

ToDo:
- [x] Generate manifest
- [x] Decide on where the manifest should be generated into
- [x] Do TypeScript shenanigans to allow importing from blitz

### Checklist


Skn0tt:
Alright, so now the route manifest is generated into `node_modules/.blitz/route-manifest.ts`. For the auth example, it'd look like this:

```ts
export default {
  Home: "/",
  Test: "/ssr",
  ForgotPasswordPage: "/forgot-password",
  LoginPage: "/login",
  ResetPasswordPage: "/reset-password",
  SignupPage: "/signup",
  ShowProjectPage: ({ projectId }: { projectId: string }) => `/projects/${projectId}`,
  ProjectsPage: "/projects",
  NewProjectPage: "/projects/new",
  EditProjectPage: ({ projectId }: { projectId: string }) => `/projects/${projectId}/edit`
}
```

For our last TODO "Do TypeScript shenanigans to allow importing from blitz", I need some inspiration. What should we do here?

flybayer:
> For our last TODO "Do TypeScript shenanigans to allow importing from blitz", I need some inspiration. What's the best way of making the manifest accessible to devs?

Follow same approach as prisma, which I expect is adding this to `@blitzjs/core` index file:

```ts
export {default as Routes} from '.blitz/route-manifest.ts'
```

Skn0tt:
Adding `export {default as Routes} from ".blitz/route-manifest"` results in `Error: Cannot find module '.blitz/route-manifest'` on the first run b/c webpack tries to bundle the file before it is emitted. For `blitz dev`, this can be circumvented by emitting first - but for any other source code analysis, that error will be a bummer.
Shipping stub files is the way to go, I think. We can give them contents that give meaningful hints about "this is a stub!".

For housing the stubs, we could create a new package called `@blitz/generated`.
But: It would look like any other package, and in case of Prisma's `.prisma` style code gen ever becoming more common, we wouldn't profit from any optimisations in that regard (e.g. bundlers may decide to opt dev-generated deps out of caching optimisations).

Another solution would be to add a post-install script to `blitz` that creates stubs in `.blitz`.

Prisma has a proxy package `@prisma/client` that forwards to `.prisma` I think.

What do y'all think about this? It's not a decision that would require a breaking change in the future, so we don't need to ponder _that_ long about it ^^

Skn0tt:
I've decided to resemble Prisma: We're still generating into `node_modules/.blitz/route-manifest.js/d.ts`, but now generate it during `postinstall`.

There's two problems that need to be solved, though:

- [ ] I couldn't get `export {default as Routes} from ".blitz/route-manifest"` to work. It always complains about not finding the dependency, and I think it's caused by our monorepo structure. @flybayer maybe you could give it a try?
- [ ] I added the command `dx-generate`, which is used for updating the manifest e.g. before unit tests in CI. I originally wanted to name it `generate` (like Prisma), but we already have that command. `dx-generate` is a bad name. How could we call it?

flybayer:
Yeah, I think do exactly what Prisma does unless we have a good reason not to. Also I would copy their code exactly as much as possible, because they have fixed a lot of things like postinstall not working in monorepos etc.

What exactly is the error with `export {default as Routes} from ".blitz/route-manifest"`? You might try with monorepo `yarn build`. There could be some slight issues we need to workaround with monorepo `yarn dev` because of the preconstruct no-build setup.

On the command naming:

My feeling is that we should use `blitz generate` if possible.

Here's some ideas
```
blitz init
blitz codegen
blitz bootstrap
blitz generate sidecar
blitz generate sdk // not good, prolly want for mobile app sdk
blitz generate init
blitz generate foundation
blitz generate lib
```

I think I quite like `blitz generate lib`. It's short and it clearly denotes some type of library files.

flybayer:
I suspect the `.blitz` module not found error may be coming from preconstruct as a warning that it's not a devDep or peerDep. If that's the case, then update our preconstruct patch to include `.blitz`. We already use patch-package to patch it to hide some of those errors for other libraries.

Skn0tt:
> update our preconstruct patch to include `.blitz`

Good one! We can now `import { Routes } from "blitz"` 🎉

New problem with our example apps, though: Since Blitz's bundle does not sit in `node_modules` b/c of preconstruct, it cannot import from ".blitz", because it's in a different `node_modules` location.

flybayer:
Hmm, good question. Probably some way we can workaround that, although nothing immediately comes to mind 🤔 Probably prisma has had this same issue?

Skn0tt:
Maybe importing from `.blitz` would be a good thing, anyhow? I see that SvelteKit has a TypeScript path alias from `$app` to `.svelte/dev/...`, presumably meant to notify people of "beware, this is a user-generated dependency". Do we really need people to import from `blitz`, or would something like `$blitz` be better?

flybayer:
@Skn0tt from a UX perspective, importing from `blitz` is far superior. I think it importing from `.blitz` made sense, Prisma would have switched to that by now.

Skn0tt:
Maybe we could have a special babel plugin that transpiles `import { Routes } from "blitz"` to `import { Routes } from ".blitz/..."` that's just used in the examples. Or we place a comment with every import stating that "import normally works via "blitz" directly"

Skn0tt:
I haven't found a way to make this work with `.blitz` / `import { Routes } from "blitz"`. eae5a8b changes it to generate into `@blitz/generated` instead. If nobody has an idea on how to make this work, I suggest merging this PR.

We can still add support for an import from "blitz" down the road, without breaking `@blitz/generated`.

Skn0tt:
https://github.com/blitz-js/blitz/pull/2042/commits/88ab77c5e82d99a28e5de0f0a9daf303c3e6d832 renamed `blitz dx-generate` to `blitz codegen`. While I also liked `blitz generate lib`, I don't want users to confuse it with the `blitz generate` scaffolding tool.

flybayer:
@Skn0tt sorry for the delay here. Great work! This is going to be so awesome to have!! And I like `blitz codegen`.

I changed the generated location to be `.blitz` and fixed the build (by manually adding stubs at `packages/core/node_modules/.blitz` (see below). **I tested `import {Routes} from 'blitz'` in a danger release and it works in end-user's apps!**

And for now we'll just have to import from `.blitz` in our monorepo.

To fix our internal blitz build we need to generate stubs at `packages/core/node_modules/.blitz`. We should do like prisma and add [`default-index.js`](https://github.com/prisma/prisma/blob/master/src/packages/client/scripts/default-index.js) and `default-index.d.ts`. Then we can add a postinstall hook in our monorepo to add those stubs.

Also I think we need to beef up the `postinstall` script. Prisma has already solved a ton of problems, especially in regards to monorepos, so I think we should copy everything from [their postinstall](https://github.com/prisma/prisma/blob/master/src/packages/client/scripts/postinstall.js) unless it specifically doesn't apply.

And then also we'll want to update the new app template to use this and add docs for this.

🎉

Skn0tt:
- [x] Adapt Prisma's `postinstall`
- [x] Link our default files as stubs for the build (who needs monorepo hooks when they can have symlinks? ;D)
- [x] Update imports in `examples/auth`
- [ ] Update new app template
- [ ] Docs

Skn0tt:
@flybayer any idea why [Compressed Size](https://github.com/blitz-js/blitz/pull/2042/checks?check_run_id=2255870718) fails?

flybayer:
@Skn0tt 

```
**ERROR** Failed to apply patch for package @preconstruct/cli at path
    
      node_modules/@preconstruct/cli
  
    This error was caused because patch-package cannot apply the following patch file:
  
      patches/@preconstruct+cli+2.0.5.patch
```

🤔

packages/blitz/scripts/postinstall.js
```diff
@@ -0,0 +1,346 @@
...
+    })
+    child.on("error", () => {
+      reject()
+    })
+  })
+}
+
+async function ensureEmptyDotBlitz() {
+  try {
+    const dotBlitzDir = path.join(process.cwd(), "node_modules/.blitz")
```
flybayer, line 174: We need to take the same approach as prisma here so it works correctly with monorepos.

```suggestion
    const dotBlitzDir = path.join(__dirname, "../../.blitz")
```

flybayer:
We also need to add docs for this.

And update the `why-blitz` page to include this as a key feature addition to next.js.

packages/blitz/scripts/postinstall.js
Skn0tt: `__dirname` is not compatible with Blitz's preconstruct setup, it'd generate into the wrong folder. What's the best way to detect that?

```ts
const dotBlitzDir =
  process.env.IS_IN_BLITZ_REPO // Do we already have a detection mechanism for smth like this?
    ? path.join(process.cwd(), "node_modules/.blitz")
    : path.join(__dirname, "../../.blitz")
```

Skn0tt:
Some thoughts that occurred to me while writing the docs:

1. What about query parameters like `/?counter=10`? We could allow a passthrough.  `Routes.Home({ counter: 10 })`. That'd require us to make all routes a function, but I don't think that's a problem.
2. At the moment, `Routes.Product({ pid: 123 })` returns `/products/123`. Would it be better to return `{ pathname: "/products/[pid]", query: { pid: 123 } }` instead? That could make the generated code more legible, especially with respect to 1.
3. Is `Pages` a better naming than `Routes`? `Route` could also refer to `API Route`, `Page` is more specific.

packages/blitz/scripts/postinstall.js
flybayer: The postinstall script on latest canary takes care of this (used for nextjs fork). So check how I do it there.

flybayer:
> 1. What about query parameters like `/?counter=10`? We could allow a passthrough.  `Routes.Home({ counter: 10 })`. That'd require us to make all routes a function, but I don't think that's a problem.

Great point. Probably we should do this. And type the function so that if it's a dynamic route, the route param is required but also any key/value can be passed.

> 2. At the moment, `Routes.Product({ pid: 123 })` returns `/products/123`. Would it be better to return [`{ pathname: "/products/[pid]", query: { pid: 123 } }`](https://nextjs.org/docs/routing/introduction#linking-to-dynamic-paths) instead? That could make the generated code more legible, especially with respect to 1.

This also sounds like a good idea.

> 3. Is `Pages` a better naming than `Routes`? `Route` could also refer to `API Route`, `Page` is more specific.

I'm thinking we should stick with Routes so it leaves the door open for other things besides strictly pages.

---

Also, are we generating the manifest on `blitz dev` and `blitz build`? I feel like we should so it's always up to date. Pages change much more frequently than the prisma schema.

Skn0tt:
> Also, are we generating the manifest on blitz dev and blitz build? I feel like we should so it's always up to date. Pages change much more frequently than the prisma schema.

Yes, we already do. The generation mechanism is part of the normal pipeline.

packages/blitz/scripts/postinstall.js
Skn0tt: Resolved in https://github.com/blitz-js/blitz/pull/2042/commits/10d7afc124bf6fdd47cbb1a6695046e00e1a07dc

Skn0tt:
> ```
> **ERROR** Failed to apply patch for package @preconstruct/cli at path
>     
>       node_modules/@preconstruct/cli
>   
>     This error was caused because patch-package cannot apply the following patch file:
>   
>       patches/@preconstruct+cli+2.0.5.patch
> ```
> 🤔

I can't fix this locally (tried running `npx patch-package @preconstruct/cli`, but that didn't do anything to the patch file).
It worked before d5caf87, though. Could you take a look at it?

packages/generator/templates/page/index.tsx
```diff
@@ -30,7 +30,7 @@ export const __ModelNames__List = () => {
           {__modelNames__.map((__modelName__) => (
             <li key={__modelName__.id}>
               <Link
-                href={`/__parentModels__/${__parentModelId__}/__modelNames__/${__modelName__.id}`}
+                href={Routes.__ModelName__()}
```
flybayer, line 33: Should be this I think.
```suggestion
                href={Routes.Show__ModelName__Page()}
```

packages/generator/templates/page/index.tsx
```diff
@@ -93,11 +93,11 @@ const __ModelNames__Page: BlitzPage = () => {
       <div>
         <p>
           <if condition="parentModel">
-            <Link href={`/__parentModels__/${__parentModelId__}/__modelNames__/new`}>
+            <Link href={Routes.New__ModelName__({ __parentModelId__ })}>
```
flybayer, line 96:
```suggestion
            <Link href={Routes.New__ModelName__Page({ __parentModelId__ })}>
```

packages/generator/templates/page/__modelIdParam__/edit.tsx
```diff
@@ -69,11 +69,11 @@ const Edit__ModelName__Page: BlitzPage = () => {
 
       <p>
         <if condition="parentModel">
-          <Link href={`/__parentModels__/${__parentModelId__}/__modelNames__`}>
+          <Link href={Routes.__ModelNames__List({ __parentModelId__ })}>
```
flybayer, line 72:
```suggestion
          <Link href={Routes.__ModelNames__Page({ __parentModelId__ })}>
```

packages/generator/templates/page/__modelIdParam__.tsx
```diff
@@ -25,12 +25,12 @@ export const __ModelName__ = () => {
 
         <if condition="parentModel">
           <Link
-            href={`/__parentModels__/${__parentModelId__}/__modelNames__/${__modelName__.id}/edit`}
+            href={Routes.Edit__ModelName__({ __parentModelId__, __modelIdParam__: __modelName__.id })}
```
flybayer, line 28:
```suggestion
            href={Routes.Edit__ModelName__Page({ __parentModelId__, __modelIdParam__: __modelName__.id })}
```

flybayer (commented):
Sweet!! Almost there.

Compressed size checks code from different places, so don't worry about it. Will resolve once merged.

zanami:
> Would it be better to return { pathname: "/products/[pid]", query: { pid: 123 } } instead? That could make the generated code more legible

It does return that object now and I'm really unsure how to mark link as active based on asPath or pathname from useRouter and this {pathname: 'somepath/[slug]', query {slug:'whatever'}

Skn0tt:
Hi @zanami! Is this a question you have or an issue you'd like to report? For questions, our [Discord](https://discord.blitzjs.com/) works best! For an issue, could you create a new one? I'd love to understand the problems you're facing.

zanami:
I'm trying to make [ActiveLink](https://github.com/vercel/next.js/blob/canary/examples/active-class-name/components/ActiveLink.js) work (not using 'as'). It does work if href is a string but I couldn't find an easy way to make it work with href = URL object like { pathname: "/products/[pid]", query: { pid: 123 } }. Can't find any exposed utils to convert it back to string to compare to asPath.
No big deal, I don't have to use Routes as it just adds a bit of convenience.

Issue #2026, opened 2021-03-01
Add `blitz export` command


lcswillems:
### What do you want and why?

Next.js has a `next export` command to create a static website.

### Possible implementation(s)

I would like to have the same for Blitz.

flybayer:
Thanks @lcswillems, but closing as a duplicate of #2015

flybayer:
Actually nevermind, this is a separate issue :)

PR #2043, opened 2021-03-03
Add `blitz export` command

Skn0tt:
Closes: #2026

This PR adds the new command "blitz export", which acts as a proxy to "next export".

It also adds an example for this use-case, although it is blocked by #2015 

### Checklist

- [x] PR submitted to [blitzjs.com](https://github.com/blitz-js/blitzjs.com) for any user facing changes (https://github.com/blitz-js/blitzjs.com/pull/402)


flybayer (changes requested):
Ok @Skn0tt, sorry for late review here. Looks good! 

Just one issue is there's now a version mismatch (see failed lint job)

Issue #2052, opened 2021-03-03
Ability to access the session data within a passport strategy


james2406:
### What do you want and why?

I would like to be able to access the session data inside the verify callback of the passport strategy.

This session data is set on an anonymous session via the `ctx.session.$setPublicData()` and `ctx.session.$setPrivateData()` functions.

### Possible implementation(s)

Potentially, the `passportAuth` function could accept a callback which would provide the `ctx.session`, like so:

```javascript
export default passportAuth((ctx) => ({
  successRedirectUrl: "/",
  errorRedirectUrl: "/",
  strategies: [
    {
      strategy: new TwitterStrategy(
        {
          consumerKey: process.env.TWITTER_CONSUMER_KEY,
          //...
        },
        async function (_token, _tokenSecret, profile, done) {
          const email = profile.emails && profile.emails[0]?.value

          // WANT: access ctx.session.publicData here
 
          const user = await db.user.upsert({
            where: {email},
            create: {
              email,
              name: profile.displayName,
            },
            update: {email},
          })

          const publicData = {/* ... */}
          done(null, {publicData})
        },
      ),
    },
    },
  ],
}))
```

flybayer:
Sweet, thanks! Great idea. 


This change needs to be made in the `passport-adapter.ts` file. We'll need to run global middleware at start of function instead of the end, we can access ctx.

So something like this:


```diff
+   // First run global middleware so we can access session
+   const globalMiddleware = getAllMiddlewareForModule({} as any)
+   await handleRequestWithMiddleware(req, res, globalMiddleware)

+   // pseudo code: if passportAuth config is callback, call it with `res.blitzCtx` as argument
 
    // rest of passport code

-    const globalMiddleware = getAllMiddlewareForModule({} as any)
-    await handleRequestWithMiddleware(req, res, [...globalMiddleware, ...middleware])
+    // Now only run the specific passport middleware that we constructed
+    await handleRequestWithMiddleware(req, res, middleware)
```

james2406:
Okay. This makes a lot of sense! I'm happy to take a go at this change and submit a PR.

PR #2058, opened 2021-03-04
Allow the `ctx` to be accessed within passport strategies

james2406:
Closes: https://github.com/blitz-js/blitz/issues/2052

This change allows access to the session inside the verify callback of the passport strategy.

### Checklist


james2406:
I tested this with the example auth project locally and it works as expected.

The only problem is, each strategy displays a typescript error whenever the `passportAuth` config is a callback. This is really strange because the config object types are the same for both the config object and config callback.

If you have any ideas why this could be happening, I'm happy to make the changes so this does not error!

Thanks

flybayer:
wow that is weird 🤔 I tried a few things and couldn't get it to work either.

Also, you can go ahead and open a PR to the docs repo to document this

flybayer:
For anyone else looking at this, do this:

1. `yarn`
2. `yarn dev`
3. Open `examples/auth/app/api/auth/[...auth].ts` and see the type error

james2406:
> wow that is weird 🤔 I tried a few things and couldn't get it to work either.

Yeah. I'll give it another go tomorrow, but something tells me this could be a bug with typescript. It just doesn't make sense.

> Also, you can go ahead and open a PR to the docs repo to document this

Done! https://github.com/blitz-js/blitzjs.com/pull/399

james2406:
Interesting. After a bit more investigation, I found the problem is the environment variables (whos types are guaranteed by the `assert` function) are reverting back to the type `string | undefined` once the callback is introduced.

I guess this makes sense as these variables could technically change.

The option would be to typecast:

```typescript
consumerKey: process.env.TWITTER_CONSUMER_KEY as string,
```

Or add something like an `environment.d.ts` file to the project which would contain the following:

```typescript
declare global {
  namespace NodeJS {
    interface ProcessEnv {
      TWITTER_CONSUMER_KEY: string;
      TWITTER_CONSUMER_SECRET: string;
    }
  }
}

export {};
```

flybayer:
Sweet! Let's go with the `as string` approach because it's more copy-pastable and will prevent people from running into this (hopefully).

And let's add that type cast to the docs to, or at least mention it somehow there

james2406:
> the `as string` approach because it's more copy-pastable

Yeah, it's a good point. I also thought the environment.d.ts file approach could confuse people if they are copy and pasting.

> And let's add that type cast to the docs to, or at least mention it somehow there

Will do.

PR #2073, opened 2021-03-05
Add `blitz install graphql-apollo-server` recipe to implement a GraphQL API endpoint

MrLeebo:

- New `blitz install graphql` recipe which sets up a GraphQL endpoint + Playground at `http://localhost:3000/api/graphql`
- New addBabelPlugin() transform to make it easier to install babel plugins in recipes:

```js
/* Supports simple string style and array with options */

// plugins: ["@emotion"]
return addBabelPlugin(program, "@emotion")

// plugins: [["@babel/plugin-proposal-decorators", {legacy: true}]]
return addBabelPlugin(program, ["@babel/plugin-proposal-decorators", {legacy: true}])
```

**Testing Notes:**

1. Create a new blitz app `blitz new graphql-test`
2. Install this recipe `blitz install graphql`
3. Start the dev server `yarn dev`
4. Visit http://localhost:3000/ and register a new User account
5. Visit http://localhost:3000/api/graphql and perform the following query:

```gql
query {
  allUsers {
    id
    email
    role
  }
}
```

  Your user information should appear!

### Checklist


MrLeebo:
👋  @kn0ll let me know if this is what you had in mind!

MrLeebo:
@flybayer I noticed that making changes to `@blitzjs/installer` (adding a new addBabelPlugin transform) and running `yarn dev` wasn't enough to get the blitz recipes to pick up the changes to the transforms. I had to `yarn build` and `yarn link @blitzjs/installer` specifically for the recipes to pick up the changes. Is that expected, or was that something preconstruct should have been able to sort out?

flybayer:
Oh very cool, thanks! Couple questions

1. Should this be named `graphql-apollo-server` or something more specific than `graphql`?
2. What's your thoughts behind using `type-graphql`? I'm thinking maybe it'd be better for this to be simpler and use basic string templates. GraphQL Nexus would be ideal in my opinion for Typescript, but it doesn't work right now. The nextjs fork will fix that though.

> @flybayer I noticed that making changes to `@blitzjs/installer` (adding a new addBabelPlugin transform) and running `yarn dev` wasn't enough to get the blitz recipes to pick up the changes to the transforms. I had to `yarn build` and `yarn link @blitzjs/installer` specifically for the recipes to pick up the changes. Is that expected, or was that something preconstruct should have been able to sort out?

Were you doing `blitz install` inside the blitz monorepo or outside it? If outside, then yes you have to do `yarn link`, otherwise it should just work.

MrLeebo:
> 1. Should this be named `graphql-apollo-server` or something more specific than `graphql`?

I think that the target audience for these recipes is often someone who is looking for a very prescriptive experience. They want to follow our best recommendation without having to independently research all the choices first. That kind of user is going to know they want GraphQL, but they may not know what all the possible implementations are.

It would be great if we could add prompts to recipes, same as the blitz new command. Then they could do `blitz install graphql` initially and if we have multiple implementations we could give them a prompt to choose and provide some guidance on why you'd pick one option or the other.

And maybe recipes could also kick off other recipes, so `blitz install graphql` could be a "decision" recipe that accepts prompts and then runs `blitz install graphql-apollo-server` based on the user's answers.

> 2. What's your thoughts behind using `type-graphql`? I'm thinking maybe it'd be better for this to be simpler and use basic string templates. GraphQL Nexus would be ideal in my opinion for Typescript, but it doesn't work right now. The nextjs fork will fix that though.

type-graphql was the same kind of thinking as above; I thought that you'd benefit from being able to leverage your typed prisma models to quickly build out your graphql API.

> Were you doing `blitz install` inside the blitz monorepo or outside it? If outside, then yes you have to do `yarn link`, otherwise it should just work.

I tried calling blitz install from outside the repo and from an app inside the examples/ directory, with the same results either way.

flybayer:
I like your idea of having `graphql` be a walkthrough/branch recipe. So maybe go with `graphql-apollo-server` for now and then add `graphql` later?

Also what are you thoughts regarding Nexus? I feel like that would be much better because of the prisma integration

MrLeebo:
👍 will rename and look into Nexus, since I'm not very familiar with it currently.

MrLeebo:
I had a beautiful recipe that worked great and installed a working GraphQL API right out of the box, on blitz 0.31... but then I upgraded to 0.33 and now nothing works 😩

flybayer:
Sweet!

PR #2077, opened 2021-03-06
Fix not being able to use `await` in `blitz console` for some users

Skn0tt:

@paulpanther noticed that `await db.project.findMany()` fails with `SyntaxError: await is only valid in async functions and the top level bodies of modules` when he tries it on CMD / Node v14.15 / Blitz 0.31

It seems to be b/c https://github.com/blitz-js/blitz/blob/canary/packages/blitz/bin/blitz is never executed.

Interestingly enough, it works when using MinGW!

Our assumption: Blitz has two packages that register a binary called `blitz`, CMD chooses the wrong one while MinGW / Bash etc use the right one. This is backed up by the output of `where blitz` (the same as `which blitz` on Unix).

This PR removes the superfluous "bin" entry to (hopefully) solve this issue.

We haven't been able to properly test this, though. @flybayer could you publish his PR as a danger release?

### Checklist

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -17,9 +17,6 @@
     "url": "https://twitter.com/flybayer"
   },
   "main": "lib/src/index.js",
-  "bin": {
-    "blitz": "./bin/run"
-  },
   "files": [
     "/bin",
     "/lib"


flybayer:
This PR is published as `blitz@0.31.2-danger.7a88537f.5`

> Our assumption: Blitz has two packages that register a binary called blitz

How is that possible?

flybayer:
Ah yes! Did you test the danger release?

Issue #2048, opened 2021-03-03
RPC doesn't use basePath


bugzpodder:
### What is the problem?

basePath wildly inconsistent during development and deployment

### What are detailed steps to reproduce this?

1. blitz new app
2. Edit blitz.config.js and add an entry { basePath: "/admin" }

Run yarn dev (development):
RPC: served from /admin/api/users/queries/getCurrentUser
requests to RPC: to /api/users/queries/getCurrentUser, which means it won't connect with above

pages and static assets: served and requested from /admin, so this is ok.
This means localhost:3000/admin will die because it cannot make the RPC request.

---- blitz start (production) ---
pages and static assets: 
served from root, eg /_next/static/chunks/webpack-xxxx.js
request from /admin, eg frontend requests /admin/_next/static/chunks/webpack-xxxx.js
So cannot load page bc static assets are missing

RPC: (not 100% sure)
RPC: served from /api/users/queries/getCurrentUser
requests to RPC: to /admin/api/users/queries/getCurrentUser

### Run `blitz -v` and paste the output here:

```
macOS Catalina | darwin-x64 | Node: v12.19.0

blitz: 0.30.6 (global)
blitz: 0.30.6 (local)

  Package manager: yarn
  System:
    OS: macOS 10.15.7
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 38.39 MB / 32.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 12.19.0 - ~/.nvm/versions/node/v12.19.0/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.8 - ~/.nvm/versions/node/v12.19.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  npmPackages:
    @prisma/client: ~2.17 => 2.17.0
    blitz: 0.30.6 => 0.30.6
    prisma: ~2.17 => 2.17.0
    react: 0.0.0-experimental-3310209d0 => 0.0.0-experimental-3310209d0
    react-dom: 0.0.0-experimental-3310209d0 => 0.0.0-experimental-3310209d0
    typescript: ~4.1 => 4.1.5
```

flybayer:
Ah, this is indeed a bug! Thanks for reporting

PR #2082, opened 2021-03-07
(patch) fix queries & mutations when using `basePath`

bugzpodder:
Closes: #2048 

Fix rpc client calls for dev/production with basePath
Fix production assets serving with basePath (reverts #2017)

### Checklist


packages/server/src/next-utils.ts
```diff
@@ -162,7 +162,7 @@ export async function nextStart(nextBin: string, buildFolder: string, config: Se
       rej("")
     } else {
       const nextjs = spawn(nextBin, spawnCommand, {
-        cwd: process.cwd(),
+        cwd: buildFolder,
```
flybayer, line 165: Base path fix looks great!

Was this change needed?

bugzpodder: @flybayer, yes it seem so.  This will make blitz build + blitz start to serve static assets correctly.  Without this change, blitz start will serve next assets from _next/ instead of ${adminPath}/_next/, while API path is served from ${adminPath}/api and root is served from ${adminPath}/.

Screen Shot 2021-03-08 at 3 45 11 PM
Screen Shot 2021-03-08 at 3 47 38 PM

flybayer: ok yes, makes sense! Thanks!

bugzpodder:
@all-contributors add @Immortalin for code

bugzpodder:
@all-contributors add @bugzpodder for code

PR #2100, opened 2021-03-10
(patch) fix another bug with queries/mutations when using `basePath`

bugzpodder:
In the previous fix, the rpcClient.warm call was fixed but the rpcCall itself was not actually modified.  This PR fixes the rpcCall with correct basePath as well.

### Checklist


flybayer:
Beautiful, thank you!!

PR #2104, opened 2021-03-10
Fix `-d` shorthand for `--dry-run` missing in blitz new

pavestru:

According to [documentation](https://github.com/blitz-js/blitzjs.com/blob/main/app/pages/docs/cli-new.mdx), `blitz new` should support also shorthand flag `-d` for `--dry-run`. It didn't as yet. It has been added in this PR.


flybayer:
@pavestru sweet, thank you!!

Issue #2089, opened 2021-03-08
Typescript compiler flags error in generated forgotPassword.test.ts file


Roesh:
### What is the problem?

In line 12 of \app\auth\mutations\forgotPassword.test.ts, `...jest.requireActual("blitz")!,` causes tsc to throw this error: `Spread types may only be created from object types.ts(2698)`

Seems like the issue is that spread types are not supported on generics (and I'm assuming that the jest.requireActual("blitz") call is returning an object, but the signature denotes that it's "unkown", which is what is causing the issue?
`function jest.requireActual<unknown>(moduleName: string): unknown`

There's an open issue at https://github.com/microsoft/TypeScript/issues/10727, so this might not be a blitz related issue, but it may make sense to implement a workaround in the meanwhile

### Paste all your error logs here:

```
tsc
app/auth/mutations/forgotPassword.test.ts:12:3 - error TS2698: Spread types may only be created from object types.

12   ...jest.requireActual("blitz")!,
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


Found 1 error
```

### Paste all relevant code snippets here:

```
jest.mock("blitz", () => ({
  ...jest.requireActual("blitz")!,      <--------- This line
  generateToken: () => generatedToken,
}))
```

### What are detailed steps to reproduce this?

1. Download latest blitz, create a new app and run the tsc command

### Run `blitz -v` and paste the output here:

```
Windows 10 | win32-x64 | Node: v15.8.0

blitz: 0.31.1 (global)
blitz: 0.31.1 (local)

  Package manager: yarn
  System:
    OS: Windows 10 10.0.18363
    CPU: (16) x64 AMD Ryzen 7 2700 Eight-Core Processor
    Memory: 5.49 GB / 15.93 GB
  Binaries:
    Node: 15.8.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.10 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 7.6.0 - C:\Program Files\nodejs\npm.CMD
    Watchman: Not Found
  npmPackages:
    @prisma/client: ~2.17 => 2.17.0
    blitz: 0.31.1 => 0.31.1
    prisma: ~2.17 => 2.17.0
    react: 0.0.0-experimental-3310209d0 => 0.0.0-experimental-3310209d0
    react-dom: 0.0.0-experimental-3310209d0 => 0.0.0-experimental-3310209d0
[...]

PR #2110, opened 2021-03-13
Fix typescript jest issue in `forgotPassword.test.ts`

Roesh:
Closes: #2089

Modified ...jest.requireActual("blitz")!,to ...jest.requireActual<object>("blitz")!, to prevent typescript compiler from throwing error.

Using the <object> for now, I tried to get the type of that object using a console log but it said "object" as well, but if there's a better type to assign here sometime in the future, we can swap object out with that instead.
console.log(typeof jest.requireActual("@blitzjs/core/server")!)
Console
    console.log
      object
      at app/auth/mutations/forgotPassword.test.ts:12:11

In addition, I also updated the auth examples tests files to be the same as the templates test files

### Checklist


examples/auth/app/auth/mutations/forgotPassword.test.ts
```diff
@@ -8,15 +8,15 @@ beforeEach(async () => {
 })
 
 const generatedToken = "plain-token"
-jest.mock("@blitzjs/core/server", () => ({
-  ...jest.requireActual("@blitzjs/core/server")!,
+jest.mock("blitz", () => ({
+  ...jest.requireActual<object>("blitz")!,
```
flybayer, line 12: Thanks @Roesh!

You have to actually revert this change as it breaks example tests here in the monorepo. Maybe a way to fix it, but needs to be `@blitzjs/core/server` for now

Roesh: Ah, got it. Let me revert that

examples/auth/app/pages/index.test.tsx
```diff
@@ -1,27 +1,25 @@
...
+  })
 
-  const {getByText} = render(<Home />)
-  const element = getByText(/powered by blitz/i)
-  // @ts-ignore
-  expect(element).toBeInTheDocument()
+  const { getByText } = render(<Home />)
+  const linkElement = getByText(/Documentation/i)
+  expect(linkElement).toBeInTheDocument()
 })
```
flybayer, line 25: So some change in this file is still causing tests to fail.

Not sure if `jest.mock` needs to be before the import?

Roesh: Ah I see. Should the test syncing be a separate PR then? I can commit just the single change to satisfy typescript and we can update the auth tests later? 
Also I'm actually not sure how to trigger the CI/CD tests, is there a way to retrigger them and check if the changes I made are OK in the monorepo?

flybayer: it's up to you. If we get tests passing, then fine to keep the change here.

CI auto runs on each `git push`, but you can run these failing tests locally by running `yarn jest` inside `examples/auth`

Roesh: Ah, I see. I reverted back to the original tests, with the only change being the <object> before the tests. I tried to troubleshoot the index.test.tsx file but I'm actually ran into issues with the forgotpassword mutation as well, so I reverted that back to its original version too. I made sure to run yarn jest this time and all tests passed.

Roesh: Ok, that was on me, I forgot to push the changes I'd made after resolving a merge conflict, I reran the yarn jest command and eslint --fix and lint, and those passed and waiting for the results of the CI tests

flybayer:
Thank you!!

Issue #2085, opened 2021-03-07
CSRFTokenMismatchError should have some messaging/hint about requiring the anti-csrf header


wescopeland:
### What is the problem?

Right now, if you hit an API route without including the `anti-csrf` header (such as with Postman or Insomnia, but could be from a client-side call in your app as well), you get a console error signifying a `CSRFTokenMismatchError`. However, it is not immediately clear without going through the docs [here](https://blitzjs.com/docs/session-management#manual-api-requests) that you need to include this `anti-csrf` header.

It would probably be a better developer experience if there was some kind of hint in the error message reminding you to either check this docs page or include the header.

### Paste all your error logs here:

```
CSRFTokenMismatchError
    at getSessionKernel (/Users/wescopeland/Documents/GitHub/achievements.app/node_modules/@blitzjs/core/server/dist/blitzjs-core-server.cjs.dev.js:1032:13)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:94:5)
    at async getSession (/Users/wescopeland/Documents/GitHub/achievements.app/node_modules/@blitzjs/core/server/dist/blitzjs-core-server.cjs.dev.js:704:23)
    at async createXblConnection (webpack-internal:///./app/api/setup/xbl/index.ts:37:19)
    at async apiResolver (/Users/wescopeland/Documents/GitHub/achievements.app/node_modules/next/dist/next-server/server/api-utils.js:8:1)
    at async DevServer.handleApiRequest (/Users/wescopeland/Documents/GitHub/achievements.app/node_modules/next/dist/next-server/server/next-server.js:64:462)
    at async Object.fn (/Users/wescopeland/Documents/GitHub/achievements.app/node_modules/next/dist/next-server/server/next-server.js:56:492)
    at async Router.execute (/Users/wescopeland/Documents/GitHub/achievements.app/node_modules/next/dist/next-server/server/router.js:23:67)
    at async DevServer.run (/Users/wescopeland/Documents/GitHub/achievements.app/node_modules/next/dist/next-server/server/next-server.js:66:1042)
    at async DevServer.handleRequest (/Users/wescopeland/Documents/GitHub/achievements.app/node_modules/next/dist/next-server/server/next-server.js:34:504) {
  statusCode: 401
}
```

### Paste all relevant code snippets here:

```typescript
export default async function createXblConnection(
  req: NextApiRequest,
  res: NextApiResponse
) {
[...]

PR #2113, opened 2021-03-13
Add helpful message when request is missing the anti-csrf header (patch)

adblanc:
Closes: https://github.com/blitz-js/blitz/issues/2085

Added a log warning when there is an anti-crsf mismatch and moved the config to the global object as I had a bug where my config would be undefined when hitting an API route.

### Checklist


packages/core/src/server/auth/sessions.ts
```diff
@@ -655,6 +656,9 @@ export async function getSessionKernel(
     if (enableCsrfProtection && payload.antiCSRFToken !== antiCSRFToken) {
       // await revokeSession(req, res, payload.handle, true)
       setHeader(res, HEADER_CSRF_ERROR, "true")
+      log.warning(
+        "Request must be performed using an anti-csrf token, you can learn more here : https://blitzjs.com/docs/session-management#manual-api-requests",
```
flybayer, line 660: Awesome, thank you so much!!

For this actual warning, let's move it to the top and only log it if the header is actually missing.

So something like this:

```tsx
  const antiCSRFToken = req.headers[HEADER_CSRF] as string

  if (enableCsrfProtection && !antiCSRFToken) {
    log.warning(`This request is missing the ${HEADER_CSRF} header. You can learn about adding this here: https://blitzjs.com/docs/session-management#manual-api-requests`)
  }
```

MrLeebo: 👍 I like that this has a link to the docs rather than trying to dump all of that information into your logs

flybayer:
@adblanc are you able to make the update I requested in my previous comment?

mabadir:
Hello,
I am claiming this PR and will be working on @flybayer recommendations [here](https://github.com/blitz-js/blitz/pull/2113#discussion_r594767920).

packages/core/src/server/auth/sessions.ts
```diff
@@ -612,6 +618,9 @@ export async function getSessionKernel(
     if (enableCsrfProtection && persistedSession.antiCSRFToken !== antiCSRFToken) {
       // await revokeSession(req, res, handle)
       setHeader(res, HEADER_CSRF_ERROR, "true")
+      log.warning(
+        "Request must be performed using an anti-csrf token, you can learn more here : https://blitzjs.com/docs/session-management#manual-api-requests",
+      )
```
flybayer, line 623:
```suggestion
```

PR #2114, opened 2021-03-14
use decodeURIComponent for router.query.next during page authentication redirect

remjx:
I ran into an issue where my `Page.authenticate = { redirectTo: "/login" }` was not working as expected. The initial page was `http://localhost:3000/page-i-refreshed` and the redirect URL was `http://localhost:3000/login?next=%2Fpage-i-refreshed` (this is OK so far). But after I logged in, the URL changed to `http://localhost:3000/%2Fpage-i-refreshed` instead of back to `http://localhost:3000/page-i-refreshed` which generated a 404. 

Use `decodeURIComponent` on the `router.query.next` path before `router.push`

Update `blitz generate` template
Update default login example

### Checklist

- [x] PR submitted to [blitzjs.com](https://github.com/blitz-js/blitzjs.com) for any user facing changes (N/A)


remjx:
I'm having trouble getting the tests to run, I made a post in help channel in Discord https://discord.com/channels/802917734999523368/814628067788062750/820540099795091468

examples/auth/app/auth/pages/login.tsx
```diff
@@ -10,7 +10,7 @@ const LoginPage: BlitzPage = () => {
     <div>
       <LoginForm
         onSuccess={() => {
-          const next = (router.query.next as string) ?? "/"
+          const next = decodeURIComponent(router.query.next as string) ?? "/"
```
flybayer, line 13: Hey, thank you! 

The issue is that `decodeURIComponent(undefined) === "undefined"` and so this doesn't fall back to `"/"`.

So probably need to change like this:
```suggestion
          const next = router.query.next ? decodeURIComponent(router.query.next) : "/"
```

remjx: That makes sense. Updated.

flybayer:
The tests are still failing with a type error. Probably you need to do `decodeURIComponent(router.query.next!)` or cast it like it was.

Make sure the tests pass. You can run `yarn jest` from `examples/auth/` to run these tests.

flybayer:
Yay, thank you!!

PR #2101, opened 2021-03-10
(major) Upgrade to React Query v3

tundera:
Closes https://github.com/blitz-js/blitz/issues/2012

### What are the changes and their implications?

This is still a work in progress. Probably some details I'm glossing over, but I wanted to get the ball rolling on this one, hoping someone will have some input for some of the TypeScript errors coming into play. The bulk of changes are happening in `@blitzjs/core/src/use-query-hooks.ts`.  

A couple of the TypeScript errors getting in the way of progressing are mentioned below. Please take a look if you have time! Would love some help on this one. Thanks 😄 

Issues:
- Overload function error on `useInfiniteQuery` (line 155, `packages/core/src/use-query-hooks.ts`)
- Type conversion error on `useReactQueryMutation` in `useMutation` export (line 27, `packages/core/src/use-mutation.ts`

PR #2134, opened 2021-03-22
Subtemplate management

Maastrich:
Closes: ??

Addition of template comment management for the generator

### Checklist

### DIFF
--- a/packages/generator/src/generator.ts
+++ b/packages/generator/src/generator.ts
@@ -243,9 +243,6 @@ export abstract class Generator<
         log.warning(`Failed trying to run prettier:` + error)
       }
     }
-    if (subTemplateString) {
-      console.log(templatedFile)
-    }
     return templatedFile
   }
 
--- a/packages/generator/src/generator.ts
+++ b/packages/generator/src/generator.ts
@@ -128,6 +128,11 @@ export abstract class Generator<
   private useTs: boolean
   private prettier: typeof import("prettier") | undefined
 
+  // relative path to templates folder of Generators implementation folder
+  static sourceRootFromTemplate(template: string): string {
+    return path.join(__dirname, "../templates", template)
+  }
+
   prettierDisabled: boolean = false
   unsafe_disableConflictChecker = false
   returnResults: boolean = false
@@ -136,7 +141,6 @@ export abstract class Generator<
 
   constructor(protected readonly options: T) {
     super()
-
     this.options = options
     this.store = createStore()
     this.fs = createEditor(this.store)
@@ -199,6 +203,21 @@ export abstract class Generator<
     if (codeFileExtensions.test(pathEnding)) {
       templatedFile = this.replaceConditionals(inputStr, templateValues, prettierOptions || {})
     }
+    // templatedFile.match
+    const subTemplateRegExp = new RegExp(/{\/\* template: (.*) \*\/}/)
+    const subTemplateString = templatedFile
+      .match(subTemplateRegExp)?.[0]
+      .replace(subTemplateRegExp, "$1")
+    if (subTemplateString) {
+      const subTemplatePosition = templatedFile.search(subTemplateRegExp)
+      templatedFile = [
+        templatedFile.slice(0, subTemplatePosition),
+        ...(templateValues.subTemplateValues?.map((values: any) =>
+          this.replaceTemplateValues(subTemplateString, values),
+        ) || []),
+        templatedFile.slice(subTemplatePosition),
+      ].join("\n")
+    }
     templatedFile = this.replaceTemplateValues(templatedFile, templateValues)
     if (!this.useTs && tsExtension.test(pathEnding)) {
       return (
@@ -223,6 +242,9 @@ export abstract class Generator<
       } catch (error) {
         log.warning(`Failed trying to run prettier:` + error)
       }
+    }
+    if (subTemplateString) {
+      console.log(templatedFile)
     }
     return templatedFile
   }
--- a/packages/generator/src/generator.ts
+++ b/packages/generator/src/generator.ts
@@ -204,7 +204,7 @@ export abstract class Generator<
       templatedFile = this.replaceConditionals(inputStr, templateValues, prettierOptions || {})
     }
     // templatedFile.match
-    const subTemplateRegExp = new RegExp(/{\/\* template: (.*) \*\/}/)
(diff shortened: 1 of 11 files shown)


flybayer:
Hey @Maastrich just an update here, I got swamped this week with client work and haven't had much time on Blitz. Reviewing https://github.com/blitz-js/blitz/pull/2101 is first in line. Once I finish that I'll give a thorough review and testing of this.

I appreciate your patience!

Maastrich:
No problem it's not a hurry ;)

packages/generator/src/generators/form-generator.ts
```diff
@@ -42,6 +42,13 @@ export class FormGenerator extends Generator<FormGeneratorOptions> {
       modelNames: this.options.modelNames,
       ModelName: this.options.ModelName,
       ModelNames: this.options.ModelNames,
+      subTemplateValues: this.options.extraArgs?.map((arg: string) => {
+        const [valueName] = arg.split(":")
+        return {
+          fieldName: singleCamel(valueName),
+          FieldName: singlePascal(valueName).replace(/(?!^)([A-Z])/g, " $1"),
```
flybayer, line 49: Also add the following which will put spaces between the words.
- `field_name`
- `Field_name`
- `Field_Name` - ex: `myFirstName` -> `My First Name`

packages/generator/templates/form/__ModelName__Form.tsx
```diff
@@ -6,7 +6,7 @@ export {FORM_ERROR} from "app/core/components/Form"
 export function __ModelName__Form<S extends z.ZodType<any, any>>(props: FormProps<S>) {
   return (
     <Form<S> {...props}>
-      <LabeledTextField name="name" label="Name" placeholder="Name" />
+      {/* template: <LabeledTextField name="__fieldName__" label="__FieldName__" placeholder="__FieldName__" /> */}
```
flybayer, line 9:
```suggestion
      {/* template: <LabeledTextField name="__fieldName__" label="__Field_Name__" placeholder="__Field_Name__" /> */}
```

packages/generator/templates/mutations/create__ModelName__.ts
```diff
@@ -4,12 +4,12 @@ import * as z from "zod"
 
 if (process.env.parentModel) {
   const Create__ModelName__ = z.object({
-    name: z.string(),
+    /* template: __attributeName__: z.__zodTypeName__(), */
```
flybayer, line 7: Can we use `fieldName` here instead of `attributeName` to be consistent?

Also I think we can shorten `zodTypeName` to `zodType`

packages/generator/src/generators/mutations-generator.ts
```diff
@@ -42,6 +50,13 @@ export class MutationsGenerator extends Generator<MutationsGeneratorOptions> {
       modelNames: this.options.modelNames,
       ModelName: this.options.ModelName,
       ModelNames: this.options.ModelNames,
+      subTemplateValues: this.options.extraArgs?.map((arg: string) => {
+        const [valueName, typeName] = arg.split(":")
+        return {
+          attributeName: singleCamel(valueName),
+          zodTypeName: this.getZodTypeName(typeName),
+        }
+      }),
```
flybayer, line 59: - Let's abstract out all these `getTemplateValues` into the base Generator class. Currently all our generators provide the same basic values, so we need to centralize it so they stay consistant. We'll want `subTemplateValues` available in all our generators
  - But we should allow subclasses to add or override the default values.

flybayer (commented):
This is awesome, thank you!!

Let's adjust this to support multiple subtemplate types.

- change the `template:` keyword to `fieldTemplate:` which makes also makes it more clear what it does
- change `getTemplateValues.subTemplateValues: {...}` to be `getTemplateValues.fieldTemplate: {...}`

And one other sort of big thing to address is we want to be able to update files with subtemplates without overwriting the whole file. Currently if you generate some code, customize it, then add a new model field, it will overwrite your customizations. So instead of existing files, we need to just insert the new field above the template comment. I think probably we should only do this custom insertion if the file has a subtemplate comment. If it doesn't, then use the current overwrite behavior.

packages/generator/src/generator.ts
```diff
@@ -199,6 +203,21 @@ export abstract class Generator<
     if (codeFileExtensions.test(pathEnding)) {
       templatedFile = this.replaceConditionals(inputStr, templateValues, prettierOptions || {})
     }
+    // templatedFile.match
+    const subTemplateRegExp = new RegExp(/{?\/\* template: (.*) \*\/}?/)
```
flybayer, line 207: I think we should also support `// template: ...` in addition to `/* template: ... */`

flybayer:
I just thought of something kind of important. Currently we have

```js
{/* fieldTemplate: <LabeledTextField name="__fieldName__" label="__Field_Name__" */}
```

But we don't want to always use `LabeledTextField`. We may want `LabeledDateField`, `LabeledCheckboxField`, etc. 

We don't need to add this in this PR, but we should consider how we can add it later. My first thought is something like this:

```ts
{/* fieldTemplate: <__FieldComponent__ name="__fieldName__" label="__Field_Name__" */}

// And then provide a map in blitz.config.js or in templates/config or something
{
  fieldComponents: {
    LabeledTextField: ['string'],
    LabeledCheckboxField: ['boolean'],
    LabeledDateField: ['date'],
  } 
}
```

Maastrich:
I'll change the vars names, for the rest of what you propose I thought quite the same but I was thinking that those are more new brand features but this PR is just an "upgrade" of existing features.

Anyway, what you proposed is exactly what I thought too add.

May be we should call each other on Monday or something to plan what is the first feat to add etc...

flybayer:
Hey @Maastrich any update on this? :)

flybayer:
@Maastrich sweet! Can you also fix the merge conflicts?

packages/generator/src/generators/form-generator.ts
flybayer: Still todo

packages/generator/templates/form/__ModelName__Form.tsx
flybayer: Still todo

packages/generator/src/generators/mutations-generator.ts
flybayer: Still todo

flybayer:
Superseded by https://github.com/blitz-js/blitz/pull/2679

Issue #2050, opened 2021-03-03
Removing the "db" folder will throw warnings from queries 


markhughes:
### What is the problem?

Removing prisma and the db folder should just remove the functionality of prisma.

However, doing so will cause the following error to appear when compiling query files.

```
wait  - compiling...
warn  - .\pages\api\<removed>\myQuery.ts
Module not found: Can't resolve 'db' in '<removed>\queries'
```

### What are detailed steps to reproduce this?

1. Remove prisma references from package.json
2. Delete 'db' folder
3. Have a query compile in blitz dev mode

### Run `blitz -v` and paste the output here:

```
Windows 10 | win32-x64 | Node: v14.15.4

blitz: 0.30.3 (global)
blitz: 0.30.6 (local)

  Package manager: yarn
  System:
    OS: Windows 10 10.0.19042
    CPU: (8) x64 Intel(R) Core(TM) i7-10610U CPU @ 1.80GHz
    Memory: 2.35 GB / 15.58 GB
  Binaries:
    Node: 14.15.4 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.5 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 6.14.10 - C:\Program Files\nodejs\npm.CMD
    Watchman: Not Found
  npmPackages:
    @prisma/client: Not Found
    blitz: 0.30.3 => 0.30.6
    prisma: Not Found
    react: 0.0.0-experimental-3310209d0 => 0.0.0-experimental-3310209d0
    react-dom: 0.0.0-experimental-3310209d0 => 0.0.0-experimental-3310209d0
    typescript: 4.2.2 => 4.2.2

```

My current solution is to create `db/index.ts` with an empty export, like so:

```
export {}
```

Perhaps something similar should be added to blitz, if it can't find any "db" folder? As we always try to import `db` once compiled

flybayer:
You are right it shouldn't error if db is missing.

This could be caused by this: https://github.com/blitz-js/blitz/blob/canary/packages/server/src/stages/rpc/index.ts#L75-L84 or this: https://github.com/blitz-js/blitz/blob/canary/packages/server/src/supertokens.ts#L57-L62

YashKumarVerma:
I'll try to look into this, this is my time contributing to blitz, so might take a bit long to workup on this.

flybayer:
Awesome, sounds good! Feel free to ask any questions here or in discord #new-contributor-help

[...]

PR #2152, opened 2021-03-27
Fix bug when `db` folder is not present

markhughes:
Closes: #2050

We only include the code in the api handler template now if `db/index.ts` (or `db/index.js`) exists. This removes the warning when webpack compiles the query should that folder not exists, and prevents executing unnecessary code as well. 

### Checklist

- [x] Does this PR warrant documentation updates? If yes, open a second PR with doc changes to [blitzjs.com](https://github.com/blitz-js/blitzjs.com)


flybayer:
Thank you!

Issue #2047, opened 2021-03-03
Blitz breaks Shallow Routing of Next.js


medihack:
### What is the problem?

With [Shallow Routing](https://nextjs.org/docs/routing/shallow-routing) it is possible to update a path without losing component state. The components get rerendered, but will not be destroyed and created again. This works fine in Next.js, but Blitz.js breaks this.

### What are detailed steps to reproduce this?

I created two little different repositories that demonstrate the problem:
https://github.com/medihack/blitz-shallow
https://github.com/medihack/next-shallow

When clicking the button and watching the console you'll see that TestComponent is destroyed and recreated in Blitz.js, but not in Next.js.

### Run `blitz -v` and paste the output here:

```
Linux 4.15 | linux-x64 | Node: v12.19.0

blitz: 0.30.4 (global)
blitz: 0.30.6 (local)

  Package manager: yarn 
  System:
    OS: Linux 4.15 Ubuntu 18.04.5 LTS (Bionic Beaver)
    CPU: (4) x64 Intel(R) Xeon(R) Gold 6248 CPU @ 2.50GHz
    Memory: 2.12 GB / 15.66 GB
    Shell: 4.4.20 - /bin/bash
  Binaries:
    Node: 12.19.0 - /usr/bin/node
    Yarn: 1.22.5 - /usr/bin/yarn
    npm: 6.14.8 - /usr/bin/npm
    Watchman: Not Found
  npmPackages:
    @prisma/client: ~2.17 => 2.17.0 
    blitz: latest => 0.30.6 
    prisma: ~2.17 => 2.17.0 
    react: 0.0.0-experimental-3310209d0 => 0.0.0-experimental-3310209d0 
    react-dom: 0.0.0-experimental-3310209d0 => 0.0.0-experimental-3310209d0 
    typescript: 4.x => 4.2.2
```

flybayer:
Hey @medihack, I'm fairly certain this is because of react concurrent mode which we have enabled by default in blitz.

Can you try this in the next app and see if it still works or not?
1. `npm i react@experimental react-dom@experimental`
2. Add following to `next.config.js`
```js
module.exports = {
  experimental: {
    reactMode: "concurrent",
  },
}
```

And you can disable that in blitz with this in blitz.config.js:

```js
module.exports = {
  experimental: {
    reactMode: "legacy",
  },
}
```

You can also try `"blocking"`

medihack:
[...]

PR #2161, opened 2021-03-28
fix: don't recreate the dom on query parameter changes

sakulstra:
Closes: #2047

`BlitzInnerRoot` is no longer recreated on each route change throwing away the dom.

### Checklist

- [x] Changes covered by tests. As I understand it, regression should be covered by auth tests already.
- [x] Does this PR warrant documentation updates? No.

My React fu isn't great, so please double check if this fix makes sense.
I didn't understand what role:
```
for (let [key, value] of Object.entries(Page)) {
    ;(BlitzInnerRoot as any)[key] = value
  }
```
plays here, but removing it didn't break any test for me :man_shrugging:

### DIFF
--- a/packages/core/src/blitz-app-root.tsx
+++ b/packages/core/src/blitz-app-root.tsx
@@ -1,8 +1,8 @@
-import React, {ComponentPropsWithoutRef, useEffect} from "react"
+import React, {useEffect} from "react"
 import {useAuthorizeIf} from "./auth/auth-client"
 import {publicDataStore} from "./auth/public-data-store"
 import {Head} from "./head"
-import {AppProps, BlitzPage} from "./types"
+import {AppProps} from "./types"
 
 const customCSS = `
   body::before {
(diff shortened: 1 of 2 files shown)


packages/core/src/blitz-app-root.tsx
```diff
@@ -86,7 +78,9 @@ export function withBlitzAppRoot(UserAppRoot: React.ComponentType<any>) {
     return (
       <>
         {noPageFlicker && <NoPageFlicker />}
-        <UserAppRoot {...props} Component={withBlitzInnerWrapper(props.Component)} />
+        <UserAppRoot {...props}>
+          <BlitzInnerRoot Component={props.Component} />
+        </UserAppRoot>
```
flybayer, line 83: This change causes `<BlitzInnerRoot />` to never be rendered in the tree. Check your react dev tools and you'll see.

packages/core/src/blitz-app-root.tsx
```diff
@@ -38,19 +38,11 @@ const NoPageFlicker = () => {
...
 
-export function withBlitzInnerWrapper(Page: BlitzPage) {
-  const BlitzInnerRoot = (props: ComponentPropsWithoutRef<BlitzPage>) => {
-    useAuthorizeIf(Page.authenticate === true)
-
-    return <Page {...props} />
-  }
-  for (let [key, value] of Object.entries(Page)) {
-    ;(BlitzInnerRoot as any)[key] = value
-  }
```
flybayer, line 49: This is needed to to copy all values like `Page.authenticate` and `Page.getLayout` so that `Component` inside user's `_app` will have those values.

flybayer (commented):
Ah good catch! This solution is in the right direction, but current is not working.

We don't have very good test coverage right now. Something we want to fix for 1.0.

To ensure this is working, make sure you can see all the components correctly in react dev tools. Also visit a page that has `Page.authenticate = true` (like `examples/auth/app/pages/projects/new.tsx`) while logged out, and make sure you see the login form.

PR #2169, opened 2021-03-30
For building Recipes: New schema.prisma transform utilities

MrLeebo:
This is a work in progress, but I figured I would post it and request a review and help testing.

Per my notes in https://github.com/blitz-js/blitz/discussions/2036 I set out to create the schema.prisma transforms. Partway through, I had the idea to try to parse the schema.prisma document into a JS object that you could mutate with normal JS operations and then write back out to a file when you're finished.

So that's what this PR introduces, in the `produceSchema()` function. **It also makes it possible for transformers to be async.**

Adds the following transform utilities:

- **produceSchema** (all the others are built off of this)
- addPrismaEnum
- addPrismaField
- addPrismaGenerator
- addPrismaIndex
- addPrismaModel
- setPrismaDataSource

Example Usage:

```ts
return produceSchema(source, ({ doc }) => {
  doc.datamodel.enums.push({
    name: "Role",
    values: [
      { name: "USER", dbName: null },
      { name: "ADMIN", dbName: null },
    ],
  });
});
```

Results (added to schema.prisma):

```prisma
enum Role {
  USER
  ADMIN
}
```

Unfortunately, it isn't perfect. The @prisma/sdk parser [leaves some details out](https://github.com/prisma/prisma/issues/3998) and users aren't going to be thrilled with a recipe that drops part of their data model. I've had success using the IntrospectionEngine to pull missing aspects back into the file, but it isn't exactly ideal and there are probably issues I haven't encountered yet.

For example, it will take aspects that were implicitly named and write their names back out. Like:

```prisma
@@unique([hashedToken, type])
```

Becomes:

```prisma
@@unique([hashedToken, type], name: "Token.hashedToken_type_unique")
```

Which is slightly annoying, but it would be better if it could perfectly match the original code. 

### Checklist

- [x] Does this PR warrant documentation updates? If yes, open a second PR with doc changes to [blitzjs.com](https://github.com/blitz-js/blitzjs.com)


packages/installer/src/transforms/prisma/print-schema.ts
```diff
@@ -0,0 +1,140 @@
...
+    .join("\n  ")
+
+  return `datasource ${name} {
+  ${children}
+}`
+}
+
+function printGenerator({name, provider, output, binaryTargets, previewFeatures}: GeneratorConfig) {
+  const children = [
+    provider && `provider        = "${provider}"`,
```
malkomalko, line 51: These extra spaces before the key look a little odd to me.  Any reason for that?

malkomalko: I think the idea is you are trying to line up the generated code on the `=` sign and `previewFeatures` is the largest key, but that might not be there.  I'm wondering if we want to keep things simple and leave the auto formatting up to a different process / keep it manual.  If not, we'd need to make this piece a little smarter.

malkomalko (approved):
This is awesome.  Left some comments for thoughts around the `printSchema` area.  Maybe it'd be good to wrap a few tests around that function as well.

Great job!  I think being able to add to prisma in recipes is going to open up a world of possibilities.  :)

packages/installer/src/transforms/prisma/print-schema.ts
MrLeebo: 👍 Yeah, purpose of that code was to line up the equals signs, but the whole document gets prettified by the IntrospectionEngine so it isn't necessary. I don't think `printSchema()` needs to be tested independently from `produceSchema()` (the callback function can be a no-op) because it's much more convenient to write your test input as a string than it would be to build out DMMF objects manually.

MrLeebo:
Is it acceptable that we lose comments from schema.prisma files during transformations? None of the parsers or Abstract Syntax Trees (ASTs) that I'm aware of for schema.prisma files preserve comments, so if that's a deal breaker than we'll have to icebox this PR until such an AST exists.

flybayer:
> Is it acceptable that we lose comments from schema.prisma files during transformations? None of the parsers or Abstract Syntax Trees (ASTs) that I'm aware of for schema.prisma files preserve comments, so if that's a deal breaker than we'll have to icebox this PR until such an AST exists.

Yikes, that could be an issue. Can we do some type of pre/post processing with string match/replace to add back the comments at the end?

flybayer:
This is looking really cool. Thanks for working on this!!

MrLeebo:
@flybayer Before I give up and roll back to string manipulations, I want to take a swing at building my own syntax tree using https://chevrotain.io/. I'm still learning it, but theoretically I should be able to build an output object similar to what we're getting from @prisma/sdk and swap the transforms over to that.

I've got a working parser for the simple schema blocks like `datasource` and `generator` already, so results are looking promising.

MrLeebo:
Just checking in to say I'm still working on this, just had a lot of other stuff to get out of the way first. So I do have an AST for schema files that includes the comments. It's a little bit different from the one that the prisma SDK generates, so I'm going to spend a little time trying to better match that, and hopefully I'll have something to show off soon

markhughes:
This will be incredible for creating really powerful recipes!

MrLeebo:
I have created [prisma-ast](https://github.com/MrLeebo/prisma-ast) which parses schema.prisma files without dropping comments or model attributes. I still have some kinks to work out, unit tests to write, and to integrate it into this PR. 

This feature is turning into a lot more work than I expected 😩

MrLeebo:
OK, we're getting better. Using my custom Abstract Syntax Tree, the comments are being preserved. However, I am losing the double line-breaks. Doesn't sound like such a big deal, but prisma autoformats everything into vertical columns using whitespace, and it uses double line-breaks to reset the indent level, so dropping the line-breaks can have an ugly effect on the output.

Look at the screenshot below, how the `user` field gets re-indented with the fields above it now that the line breaks are gone.

Screen Shot 2021-04-23 at 12 10 45 AM

I think I can create another symbol to parse double line-breaks, same as I did for comments, and if that works we will be able to parse the file with total fidelity.

Although, with my luck, a completely new version of Prisma will come out 2 days after this gets merged...

MrLeebo:
I ditched IntrospectionEngine entirely and implemented the logic to auto-format the document myself. This has the added benefit of making it so the printed schema looks the same in unit tests as it does in the real world because we had to skip running IntrospectionEngine for unit tests.

flybayer:
Awesome, great work!! Can you go ahead and add docs for this?

Issue #2196, opened 2021-04-10
suspense refresh/invalidation broken on canary-0


sakulstra:
### What is the problem?

When I use `useInfiniteQuery` with suspense, query parameter change will trigger a fetch, but the data parameter is never updated.

When I disable suspense it seems to work again.

I think this might be related to the react-query v3 update, but i'm not 100% sure.
Might be related to #2191, #2172 and #2184, but am not quite sure.

### Paste all your error logs here:

There are non

### Paste all relevant code snippets here:

```js
const [groups, { isFetching, isFetchingNextPage, fetchNextPage, hasNextPage }] = useInfiniteQuery(
    getFlashLoans,
    (page = { take: 30, skip: 0 }) => ({
      where,
      orderBy,
      ...page,
    }),
    {
      getNextPageParam: (lastGroup) => lastGroup.nextPage,
    }
)
console.log(groups, isFetching, isFetchingNextPage)
```
With this code when `where` changes the console will log:
```
undefined false false
```

Setting suspense to `false`, will fix the problem and the log will change to:
```
undefined true false
[{…}] false false
```

### What are detailed steps to reproduce this?

1. use `useInfiniteQuery` with suspense & change the `where` parameter.
2. observe the issue

### Run `blitz -v` and paste the output here:

```
Linux 5.4 | linux-x64 | Node: v12.14.0

blitz: 0.33.0 (global)
blitz: 0.34.0-canary.0 (local)

  Package manager: yarn 
  System:
    OS: Linux 5.4 Manjaro Linux
    CPU: (4) x64 Intel(R) Core(TM) i7-7500U CPU @ 2.70GHz
    Memory: 449.38 MB / 15.52 GB
    Shell: 5.8 - /usr/bin/zsh
  Binaries:
    Node: 12.14.0 - ~/.nvm/versions/node/v12.14.0/bin/node
    Yarn: 1.21.1 - ~/.nvm/versions/node/v12.14.0/bin/yarn
    npm: 6.13.4 - ~/.nvm/versions/node/v12.14.0/bin/npm
    Watchman: Not Found
  npmPackages:
    @prisma/client: 2.20.1 => 2.20.1 
    blitz: 0.34.0-canary.0 => 0.34.0-canary.0 
    prisma: 2.20.1 => 2.20.1 
    react: 0.0.0-experimental-3310209d0 => 0.0.0-experimental-3310209d0 
[...]

PR #2197, opened 2021-04-10
Fix useInfiniteQuery suspense bug in canary release by upgrading react-query to 3.13.9 (patch)

sakulstra:
Till now it's just the test.

targets: #2196

### Checklist

The issue might not be on the blitz side, but caused by https://github.com/tannerlinsley/react-query/issues/2095 instead.


sakulstra:
There seem sto be an issue with running integration tests in the ci?
`CI / Blitz - Test Integration (pull_request) Successful in 15s` - they  should fail & locally do fail. seems like they aren't even executed here.

flybayer:
@sakulstra oops! 😅 That's hilarious, tests not actually running. I can fix that on Monday.

flybayer:
Thanks! This is caused by https://github.com/tannerlinsley/react-query/issues/2095

PR #2219, opened 2021-04-13
Fix crash when using Blitz inside iframes

### LINKED ISSUE #2218: Blitz apps crash when run in an iFrame in Chrome when third-party cookies are disabled
### What is the problem?

This is quite an obscure problem - we discovered it after integrating a form from our Blitz app in an iframe within our webflow landing page.

If third-party cookies are disabled then a Blitz app will crash when run in an iFrame in chrome.

This is Chrome's default setting when running in incognito.

As a side effect of third-party cookies being disabled, localStorage is also disabled. Because Blitz relies on localStorage within `PublicDataStore`:

```
  updateState(value?: PublicData | EmptyPublicData, opts?: {suppressEvent: boolean}) {
    // We use localStorage as a message bus between tabs.
    // Setting the current time in ms will cause other tabs to receive the `storage` event
    if (!opts?.suppressEvent) {
      // Prevent infinite loop
      localStorage.setItem(this.eventKey, Date.now().toString())
    }
    this.observable.next(value ?? this.getData())
  }
```

This problem can be fixed by wrapping the `localStorage.setItem` call in a `try/catch` block

### Paste all your error logs here:

```
DOMException: Failed to read the 'localStorage' property from 'Window': Access is denied for this document.
```

### Paste all relevant code snippets here:

```
PASTE_HERE (leave the ``` marks)
```

### What are detailed steps to reproduce this?

1. Try to load a blitz app in an iframe in Chrome when third-party cookies are disabled.
2. Observe the following error: `DOMException: Failed to read the 'localStorage' property from 'Window': Access is denied for this document.`

### Run `blitz -v` and paste the output here:

```
➜  app git:(master) blitz -v     
You are using beta software - if you have any problems, please open an issue here:

macOS Catalina | darwin-x64 | Node: v12.20.0

blitz: 0.30.1 (global)
blitz: 0.33.1 (local)

  Package manager: npm 
  System:
    OS: macOS 10.15.6
    CPU: (16) x64 Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
    Memory: 24.00 GB / 64.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 12.20.0 - ~/.nvm/versions/node/v12.20.0/bin/node
    Yarn: 1.22.4 - ~/.nvm/versions/node/v12.20.0/bin/yarn
    npm: 6.14.8 - ~/.nvm/versions/node/v12.20.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  npmPackages:
    @prisma/client: ~2.20.1 => 2.20.1 
    blitz: 0.33.1 => 0.33.1 
    prisma: ~2.20.1 => 2.20.1 
    react: 0.0.0-experimental-3310209d0 => 0.0.0-experimental-3310209d0 
    react-dom: 0.0.0-experimental-3310209d0 => 0.0.0-experimental-3310209d0 
    typescript: 4.1.5 => 4.1.5 

```

mtford90:
Closes #2218

### DIFF
--- a/packages/core/src/auth/public-data-store.ts
+++ b/packages/core/src/auth/public-data-store.ts
@@ -27,7 +27,11 @@ class PublicDataStore {
     // Setting the current time in ms will cause other tabs to receive the `storage` event
     if (!opts?.suppressEvent) {
       // Prevent infinite loop
-      localStorage.setItem(this.eventKey, Date.now().toString())
+      try {
+        localStorage.setItem(this.eventKey, Date.now().toString())
+      } catch (err) {
+        console.error("LocalStorage is not available", err)
+      }
     }
     this.observable.next(value ?? this.getData())
   }


markhughes:
Also seems recommended on MDN:  

> setItem() may throw an exception if the storage is full. Particularly, in Mobile Safari (since iOS 5) it always throws when the user enters private mode. (Safari sets the quota to 0 bytes in private mode, unlike other browsers, which allow storage in private mode using separate data containers.) Hence developers should make sure to always catch possible exceptions from setItem().

see: https://developer.mozilla.org/en-US/docs/Web/API/Storage/setItem

flybayer:
Sweet, thanks!!

PR #2223, opened 2021-04-15
Throw an error if using a query with useMutation or a mutation with useQuery

### LINKED ISSUE #2200: Throw an error if using a query with useMutation or a mutation with useQuery
### What do you want and why?

Throw an error if using a query with useMutation or a mutation with useQuery. These have different expectations and behavior, so we need to throw an error in this case.

Split [sanitize](https://github.com/blitz-js/blitz/blob/canary/packages/core/src/utils/react-query-utils.ts#L82) into `sanitizeQuery` and `sanitizeMutation` and add the type check `queryFn._meta.type` should be `query` or `mutation`, so you know which type is actually being passed in.

### Additional context

https://discord.com/channels/802917734999523368/802989186495152149/830524607429476413

Mokshit06:
Closes: #2200 

This Pull Request closes #2200. It also includes test cases for them.

## Bug Checklist

## Feature Checklist


flybayer:
Thanks @Mokshit06! Looks really good except you have a failing jest test in the auth example.

```
[@examples/auth::jest]   ● renders blitz documentation link
[@examples/auth::jest] 
[@examples/auth::jest]     TypeError: Cannot read property 'type' of undefined
[@examples/auth::jest] 
[@examples/auth::jest]       279 |     var queryFnName = type === "mutation" ? "useMutation" : "useQuery";
[@examples/auth::jest]       280 |
[@examples/auth::jest]     > 281 |     if (enhancedQueryFn._meta.type !== type) {
[@examples/auth::jest]           |                               ^
[@examples/auth::jest]       282 |       throw new Error("\"" + queryFnName + "\" was expected to be called with a " + type + " but was called with a \"" + enhancedQueryFn._meta.type + "\"");
[@examples/auth::jest]       283 |     }
[@examples/auth::jest]       284 |
```

Run `yarn jest` in `examples/auth/` to see that locally.

Mokshit06:
Hey @flybayer, I was looking into the error and it is occurring because the `_meta` property is not defined on the `enhancedQueryFn`. Isn't that supposed to be defined after `validateQueryFn` is called?

flybayer:
@Mokshit06 no, it's already added because `validateQueryFn` ensures that. But I think the solution is to not throw the error if `isNotInUserTestEnvironment()`, same as in `validateQueryFn`, because this is a jest only issue.

flybayer:
Yay!

PR #2228, opened 2021-04-17
Fix `Routes` manifest type error. Change to `string` to `string | number`

JuanM04:

Before this, something like this (taken from blitzjs.com) would throw a TypeScript error:

```tsx
<Link href={Routes.ProductsPage({ productId: 123 })} />
```

## Bug Checklist

### DIFF
--- a/examples/auth/app/pages/projects/[projectId].tsx
+++ b/examples/auth/app/pages/projects/[projectId].tsx
@@ -24,7 +24,7 @@ export const Project = () => {
         <h1>Project {project.id}</h1>
         <pre>{JSON.stringify(project, null, 2)}</pre>
 
-        <Link href={`/projects/${project.id}/edit`}>
+        <Link href={Routes.EditProjectPage({projectId: project.id})}>
           <a>Edit</a>
         </Link>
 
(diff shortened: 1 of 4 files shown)


Skn0tt (changes requested):
Good catch! Could you add an integration test?

The CI fails, but I fail to see why ... any idea @flybayer?

JuanM04:
> Could you add an integration test?

I've already changed `route-import-manifest.spec.ts` to match these changes. If you are asking me to create integration tests (the ones inside `test/`), I have no clue on how to create them.

> The CI fails, but I fail to see why

It failed before I created this PR (see #2223, for example)

Mokshit06:
I think the CI is failing because we aren't running `blitz codegen` to [generate the Routes manifest](https://blitzjs.com/docs/route-manifest#generating).

Skn0tt:
@JuanM04 I've extended the existing integration test to include usage of numbers: 16dc11d

Skn0tt:
@Mokshit06

> I think the CI is failing because we aren't running blitz codegen to generate the Routes manifest.

It looks like Windows doesn't know how to deal with [these](https://github.com/blitz-js/blitz/tree/canary/packages/core/node_modules/.blitz) symlinks. Do you know of a way to make Symlinks work on windows? Otherwise, I'll just copy&paste the files they're linking to :)

Skn0tt (approved):

Issue #2231, opened 2021-04-18
Page.redirectAuthenticatedTo does not accept null type from Routes manifest


remjx:
### What is the problem?

When trying to do
```
Page.redirectAuthenticatedTo = Routes.SomePage().pathname
```

I get a type error because pathname returns `string | null | undefined` and `redirectAuthenticatedTo` expects `string | undefined`

Related question: Is there a recommended way to convert the URL Object from routes to a string? 

### Paste all your error logs here:

```
(property) redirectAuthenticatedTo?: string | undefined
Type 'string | null | undefined' is not assignable to type 'string | undefined'.
  Type 'null' is not assignable to type 'string | undefined'.ts(2322)
```

### Paste all relevant code snippets here:

```
LoginPage.redirectAuthenticatedTo = Routes.DashboardPage().pathname
```

### What are detailed steps to reproduce this?

See above

### Run `blitz -v` and paste the output here:

```
blitz: 0.34.0 (global)
blitz: 0.34.0 (local)
  npmPackages:
    typescript: ~4.2 => 4.2.4
```

PR #2232, opened 2021-04-18
redirectAuthenticatedTo can be null from Routes manifest

remjx:
Closes: #2231 

Update type of `redirectAuthenticatedTo` to accept `null`

null or undefined will have the same effect because the `blitz-app-root` component is just checking if the value is truthy:

- https://github.com/blitz-js/blitz/blob/563720bb5798b0c43d6858ffe502d19d78be1c5b/packages/core/src/blitz-app-root.tsx#L61
- https://github.com/blitz-js/blitz/blob/563720bb5798b0c43d6858ffe502d19d78be1c5b/packages/core/src/blitz-app-root.tsx#L82

## Bug Checklist

## Feature Checklist


flybayer:
Hey @remjx, thank you for the bug report and thanks for this PR. I've actually just fixed it in a better way in https://github.com/blitz-js/blitz/pull/2238 by fixing the types of the Route manifest itself.

PR #2235, opened 2021-04-19
Fix CI Failure on Windows

Skn0tt:

This PR aims to fix the CI failure caused by symlinks on windows.
We'll just use plain files instead, they don't pose any problems :)

## Bug Checklist

## Feature Checklist


Skn0tt:
cc #2228

flybayer:
Oh, I didn't realize we had stuff in node_modules checked into git 😬 

Solution was to remove that and bust the CI cache: https://github.com/blitz-js/blitz/pull/2236

Issue #2244, opened 2021-04-20
blitz console hangs on loading modules


malkosan:
### What is the problem?

Generate a new app, then run `blitz console`.  You'll see all the modules don't load and you get warnings about circular dependencies.

You have to hit enter to see the `>` to enter in db queries, and the queries/mutations are not available because it hangs trying to load the modules.

### Paste all your error logs here:

```
> blitz console
You are using beta software - if you have any problems, please open an issue here:
      https://github.com/blitz-js/blitz/issues/new/choose

You have entered the Blitz console
Tips: - Exit by typing .exit or pressing Ctrl-D
      - Use your db like this: await db.project.findMany()
      - Use your queries/mutations like this: await getProjects({})
Loading Modules 2/20
⚡️ > (node:28563) Warning: Accessing non-existent property 'sessionMiddleware' of module exports inside circular dependency
(Use `node --trace-warnings ...` to show where the warning was created)
(node:28563) Warning: Accessing non-existent property 'simpleRolesIsAuthorized' of module exports inside circular dependency
(node:28563) Warning: Accessing non-existent property 'sessionMiddleware' of module exports inside circular dependency
(node:28563) Warning: Accessing non-existent property 'simpleRolesIsAuthorized' of module exports inside circular dependency
```

### Paste all relevant code snippets here:

```
PASTE_HERE (leave the ``` marks)
```

### What are detailed steps to reproduce this?

1. Create a new blitz app
2. Run `blitz console`

### Run `blitz -v` and paste the output here:

```
>  blitz -v
You are using beta software - if you have any problems, please open an issue here:
      https://github.com/blitz-js/blitz/issues/new/choose

macOS Big Sur | darwin-x64 | Node: v15.3.0

blitz: 0.34.1 (global)
blitz: 0.34.1 (local)

  Package manager: yarn
  System:
    OS: macOS 11.2.3
    CPU: (16) x64 Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
    Memory: 365.79 MB / 32.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 15.3.0 - ~/.nvm/versions/node/v15.3.0/bin/node
    Yarn: 1.22.10 - ~/.nvm/versions/node/v15.3.0/bin/yarn
    npm: 7.0.14 - ~/.nvm/versions/node/v15.3.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  npmPackages:
[...]

PR #2251, opened 2021-04-20
fix `blitz console` circular dependency issues

flybayer:
Closes: https://github.com/blitz-js/blitz/issues/2244

fix `blitz console` circular dependency issues


malkomalko:
🙏🏻

PR #2256, opened 2021-04-21
Add experimental.initServer hook for running code on server start

flybayer:

This adds an experimental feature for running code on server start, for both serverful and serverless deployments.

```js
// blitz.config.js
module.exports = {
  experimental: {
    initServer() {
      console.log("Hello world from initServer")
    },
  },
}
```

If you try this feature, please give me feedback on how well it works and how well you like it.

### DIFF
--- a/nextjs/packages/next/next-server/server/next-server.ts
+++ b/nextjs/packages/next/next-server/server/next-server.ts
@@ -190,6 +190,7 @@ export default class Server {
       assetPrefix,
       generateEtags,
       compress,
+      experimental,
     } = this.nextConfig
 
     this.buildId = this.readBuildId()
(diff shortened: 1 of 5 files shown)


kitze:
@flybayer it's not possible to import a ts file and call a fn in blitz.config.js, unless it also supports blitz.config.ts?

flybayer:
@kitze I'm working on that right now

kitze:
@flybayer amazing!

nextjs/packages/next/next-server/server/next-server.ts
```diff
@@ -246,6 +247,10 @@ export default class Server {
     this.router = new Router(this.generateRoutes())
     this.setAssetPrefix(assetPrefix)
 
+    if (experimental.initServer) {
+      experimental.initServer()
```
andreasasprou, line 251: Can we make this `await  experimental.initServer()`?

andreasasprou: Use case: `connectToMongoDB()` for those who don't use custom server but are not running serverless

PR #2277, opened 2021-04-26
(newapp) Add a `<Suspense>` boundary to `_app.tsx`

Roesh:
> It's best practice to have a Suspense boundary at the root.
> From https://github.com/facebook/react/pull/20970 

This PR adds a base Suspense component to the generated _app.tsx file as a global fallback for components that do not have a Suspense boundary. A note is added as a comment indicating that devs should wrap components lower in the tree with suspense to improve user experience

I have not added an integration test, would one be necessary for this? I am not sure if there is a good way to test a suspense feature, but I can look that up if needed


packages/generator/templates/app/app/pages/_app.tsx
```diff
@@ -9,19 +9,24 @@ import {
...
   return (
-    <ErrorBoundary
-      FallbackComponent={RootErrorFallback}
-      resetKeys={[router.asPath]}
-      onReset={useQueryErrorResetBoundary().reset}
-    >
-      {getLayout(<Component {...pageProps} />)}
-    </ErrorBoundary>
+    // NOTE: This suspense tag provides a convenient global fallback
+    // Make sure to add suspense lower in the tree to optimize loading experience where possible
```
flybayer, line 20: This is a nice touch, but it will break some of our current recipe code. So I'll remove it for now.
```suggestion
```

flybayer:
Thanks!!

Issue #2280, opened 2021-04-29
SuperJSON babel plugin causes prefetched queryKeys to miss the cache


MrLeebo:
### What is the problem?

Alright that title is a bit of a mouthful, but here's what's happening.

Blitz is set up to allow you to prefetch queries while server-rendering the component. The `<BlitzProvider>` component accepts a dehydratedState prop, [as seen here](https://github.com/blitz-js/blitz/blob/canary/packages/core/src/blitz-app-root.tsx#L89), which you can populate by returning 
```ts
return { props: { dehydratedState: dehydrate(queryClient) } }
``` 
from your `getServerSideProps` method.

The problem is, that the prop gets processed by [babel-plugin-superjson-next](https://github.com/blitz-js/blitz/blob/bc0c7287ecd35c5e9e5f22b16bfdcf88b7d7145a/packages/babel-preset/src/index.ts#L12) which changes the shape of the queryKey, causing the cache lookup to miss.

As a fix, I think the plugin could be changed to ignore the `dehydratedState` prop when transforming `getServerSideProps` and `getStaticProps`. Or the plugin could accept a user-configurable list of props to ignore, and we put `dehydratedState` in that list in the blitz preset.

### Paste all your error logs here:

No error. The server renders the Suspense fallback and makes an extra POST to fetch the query.

### Paste all relevant code snippets here:

```ts
const user = useCurrentUser()
```

```ts
export async function getServerSideProps(ctx: GetServerSidePropsContext) {
  const queryClient = new QueryClient()
  await queryClient.prefetchQuery(getQueryKey(getCurrentUser), () =>
    invokeWithMiddleware(getCurrentUser, null, ctx)
  )

  return {
    props: {
      dehydratedState: dehydrate(queryClient),
    }
  }
}
```

### What are detailed steps to reproduce this?

1. In a fresh blitz app, create a page that invokes `useCurrentUser()`
2. Prefetch the getCurrentUser query using the getServerSideProps snippet shown
3. Refresh the page and notice in the DevTools networking tab that the server rendered the Suspense fallback (the cache missed)
4. Check the queryCache and notice that there are two entries, with and without SuperJSON metadata

### Run `blitz -v` and paste the output here:

```
macOS Catalina | darwin-x64 | Node: v14.15.4

blitz: 0.35.0-canary.2 (local)

  Package manager: yarn 
  System:
    OS: macOS 10.15.7
[...]

PR #2281, opened 2021-04-29
Deserialize SuperJSON-ed pageProps before passing to dehydratedState

Skn0tt:
Closes: #2280

When `pageProps` is accessed, it's required to deserialize its contents before using them.

## Bug Checklist

## Feature Checklist


Skn0tt:
@MrLeebo do you want to add an integration test to this?

packages/core/src/blitz-app-root.tsx
```diff
@@ -85,8 +86,15 @@ export function withBlitzAppRoot(UserAppRoot: React.ComponentType<any>) {
       document.documentElement.classList.add("blitz-first-render-complete")
     }, [])
 
+    let dehydratedState = props.pageProps.dehydratedState
+    if (dehydratedState && "_superjson" in props.pageProps) {
+      const {_superjson, ...json} = props.pageProps
+      const deserialized = SuperJSON.deserialize({json, meta: _superjson}) as any
+      dehydratedState = deserialized.dehydratedState
+    }
+
```
MrLeebo, line 95: Do you think it would be better to have a way to skip the prop in during serialization, rather than deserializing it back after? If the prefetched payload is large or deeply nested, this could be an expensive operation just to get back to where it started.

MrLeebo (commented):
I can set up an integration test, probably tonight.

packages/core/src/blitz-app-root.tsx
Skn0tt: Good idea! See https://github.com/blitz-js/babel-plugin-superjson-next/pull/68

MrLeebo:
@Skn0tt Added a test for this. If you inspect the server rendered output from the integration test, (via running `blitz dev` in test/integration/auth) you'll notice it renders the suspense fallback server-side but renders the results without re-fetching on the first client-side render. I also tested this with https://github.com/blitz-js/blitz/pull/1940 merged in, and was getting the fully rendered output from the server render, so once these two PRs are merged, prefetching will be a lot more viable in blitz.

packages/core/src/blitz-app-root.tsx
MrLeebo: Looks good! 👍 Once that gets merged, we'll have to add `dehydratedState` to the blitz babel preset.

Skn0tt:
Updated to the new `babel-plugin-superjson-next` and exclued `dehydratedState`.

nextjs/packages/next/build/webpack-config.ts
```diff
@@ -1511,93 +1511,94 @@ export default async function getBaseWebpackConfig(
...
-    ) {
-      if (!(rule.test instanceof RegExp && Array.isArray(rule.use))) {
-        return
-      }
+    ;[].forEach.call(
+      webpackConfig.module.rules,
+      function (rule: webpack.RuleSetRule) {
+        if (!(rule.test instanceof RegExp && Array.isArray(rule.use))) {
+          return
+        }
```
flybayer, line 1519: Can we revert all the changes to the `nextjs/*` files? Not sure how this happened, but it will cause merge conflicts with the fork.

flybayer:
And you have a build error

nimashoghi:
I'm currently playing around with query prefetching, and my query results include date objects. In this case, I receive a serialization error as this object is being handled by next.

```
Error: Error serializing `.dehydratedState.queries[0].state.data.items[0].createdAt` returned from `getServerSideProps` in "/projects".
Reason: `object` ("[object Date]") cannot be serialized as JSON. Please only return JSON serializable data types.
```

flybayer:
@nimashoghi oops — can you open a new issue for that?

PR #2283, opened 2021-04-29
Add support for `blitz.config.ts`

flybayer:
Closes: https://github.com/blitz-js/blitz/issues/913

Add support for `blitz.config.ts`. We use `esbuild` to bundle your blitz config file and output it at `.blitz/blitz.config.js`.

## Feature Checklist

- [x] Update recipe installer
- [x] Test Vercel deployment


andreasasprou:
Can this be disabled by default for non `.ts` files?

esbuild does not support [features that need a type system](https://esbuild.github.io/content-types/#no-type-system), which are required for libraries like [Typegoose](https://typegoose.github.io/typegoose/).

flybayer:
@andreasasprou so `blitz.config.js` is failing for you because `emitDecoratorMetadata` is in `tsconfig` even though `blitz.config.js` isn't importing any typescript? 🤔

sebastianhoitz:
@flybayer after upgrading to the new `blitz.config.ts` it looks like `next/image` has problems with custom registered domains. It complains they are not added to the config. Did some mapping get lost?

flybayer:
@sebastianhoitz hmm, that's not expected. Can you open an issue?

Issue #2286, opened 2021-04-30
Optional catch-all routes not handled correctly by the routes codegen script


kirakik:
### What is the problem?

I've identified an issue in the autogenerated `Routes`. It seems it doesn't [handle optional catch-all routes](https://blitzjs.com/docs/api-routes#dynamic-api-routes) correctly. 

### Paste all your error logs here:

```
Type error: ',' expected.

   7 |
   8 | export const Routes: {
>  9 |   HomePage(query: { [...answersId: string | number } & ParsedUrlQueryInput): RouteUrlObject;
     |                                                    ^
  10 |   ForgotPasswordPage(query?: ParsedUrlQueryInput): RouteUrlObject;
  11 |   LoginPage(query?: ParsedUrlQueryInput): RouteUrlObject;
  12 |   ResetPasswordPage(query?: ParsedUrlQueryInput): RouteUrlObject;

```

### Paste all relevant code snippets here:

N/A

### What are detailed steps to reproduce this?

1. Create a page with an optional catch-all route (e.g. `pages/[[...slug]].tsx`
2. Run `blitz codegen`
3. Run `blitz build`
4. The build fails because the generated Routes typing is missing a square bracket

### Run `blitz -v` and paste the output here:

```
blitz: 0.34.3 (global)
blitz: 0.34.3 (local)

  Package manager: yarn
  System:
    OS: macOS 11.2.3
    CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
    Memory: 54.01 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.15.4 - ~/.nvm/versions/node/v14.15.4/bin/node
    Yarn: 1.21.1 - /usr/local/Cellar/yvm/3.4.0/shim/yarn
    npm: 6.14.10 - ~/.nvm/versions/node/v14.15.4/bin/npm
    Watchman: Not Found
  npmPackages:
    @prisma/client: ~2.20 => 2.20.1
    blitz: 0.34.3 => 0.34.3
    prisma: ~2.20 => 2.20.1
    react: 0.0.0-experimental-6a589ad71 => 0.0.0-experimental-6a589ad71
    react-dom: 0.0.0-experimental-6a589ad71 => 0.0.0-experimental-6a589ad71
    typescript: ~4.2 => 4.2.4
```

### Please include below any other applicable logs and screenshots that show your problem:

[...]

PR #2290, opened 2021-05-01
Fix optional catch-all handling in routes codegen

kirakik:
Closes: #2286

See the issue for more details, this is essentially fixing the build for projects with optional catch all routes.

## Bug Checklist

- [ ] Integration test added (should I add an integration test for this somewhere?)

### DIFF
--- a/packages/server/src/stages/route-import-manifest/route-import-manifest.test.ts
+++ b/packages/server/src/stages/route-import-manifest/route-import-manifest.test.ts
@@ -4,10 +4,24 @@ import {
   parseParametersFromRoute,
 } from "./route-import-manifest"
 
-test("parseParametersFromRoute", () => {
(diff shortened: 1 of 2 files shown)


Skn0tt (changes requested):
Great work on adding all of the tests.
I don't think an integration test is necessary. WDYT @flybayer?

This PR only fixes parsing, we should also adapt the code generation part to actually mark the optional parameters as being optional.

kirakik:
> This PR only fixes parsing, we should also adapt the code generation part to actually mark the optional parameters as being optional.

Where can I find the code generation part in the codebase?

Skn0tt:
It's in the same file 😊 https://github.com/blitz-js/blitz/blob/ada8e2d4255ac6d0cfdd3084827edec6ed37510c/packages/server/src/stages/route-import-manifest/route-import-manifest.ts#L49

Skn0tt:
Awesome PR @kirakik!!!

Issue #2294, opened 2021-05-03
Setting trailingSlash: true in the blitz configuration creates 308 redirects for all api requests.


cashmagnet:
### What is the problem?

API requests for blitz, from either await direct function calls inside useEffects or useQuery calls are not being created with consideration for the trailingSlash config.

### What are detailed steps to reproduce this?

1. create new blitz application
2. update blitz.config.js as below 
````
const { sessionMiddleware, simpleRolesIsAuthorized } = require("blitz")

module.exports = {
  middleware: [
    sessionMiddleware({
      isAuthorized: simpleRolesIsAuthorized,
    }),
  ],
  trailingSlash: true,

  /* Uncomment this to customize the webpack config
  webpack: (config, { buildId, dev, isServer, defaultLoaders, webpack }) => {
    // Note: we provide webpack above so you should not `require` it
    // Perform customizations to webpack config
    // Important: return the modified config
    return config
  },
  */
}

````

3. run your the blitz server (development or production, no difference) 
4. open up network tab in browser debugging tools
5. navigate to the login screen - http://localhost:3000/login/
6. you will see the getCurrentUser request is being requested without a trailing slash. 

### Run `blitz -v` and paste the output here:

```
Windows 10 | win32-x64 | Node: v12.18.2

blitz: 0.34.3 (global)
blitz: 0.34.3 (local)

  Package manager: yarn
  System:
    OS: Windows 10 10.0.19042
    CPU: (16) x64 Intel(R) Core(TM) i7-10700 CPU @ 2.90GHz
    Memory: 16.19 GB / 31.92 GB
  Binaries:
    Node: 12.18.2 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.4 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 6.14.5 - C:\Program Files\nodejs\npm.CMD
    Watchman: Not Found
  npmPackages:
    @prisma/client: ~2.20 => 2.20.1
    blitz: 0.34.3 => 0.34.3
    prisma: ~2.20 => 2.20.1
    react: 0.0.0-experimental-6a589ad71 => 0.0.0-experimental-6a589ad71
    react-dom: 0.0.0-experimental-6a589ad71 => 0.0.0-experimental-6a589ad71
    typescript: ~4.2 => 4.2.4

```

flybayer:
### To Fix

[...]

PR #2300, opened 2021-05-04
Fix wrong API requests when `trailingSlash` configuration is true

peterpalau:
Closes: #2294

This PR adds the implementation to consider the `trailingSlash` configuration in the API requests from either await direct function calls inside `useEffects` or `useQuery`.

## Checklist

- [x] Added `trailingSlash` to the blitz runtime dat
- [x] Updated `getApiUrlFromResolverFilePath` to add trailing slash if `getBlitzRuntimeData().trailingSlash` is true.


packages/core/src/rpc-client.ts
```diff
@@ -157,7 +158,10 @@ executeRpcCall.warm = (apiUrl: string) => {
 }
 
 const getApiUrlFromResolverFilePath = (resolverFilePath: string) =>
-  resolverFilePath.replace(/^app\/_resolvers/, "/api")
+  resolverFilePath.replace(
+    /^app\/_resolvers/,
+    "/api".concat(getBlitzRuntimeData().trailingSlash ? "/" : ""),
+  )
```
flybayer, line 164: You need to contact the slash on end of resolverFilePath, not `api`. Current code will result in url like `/api//queries/getUser`

flybayer (changes requested):
Thanks!!

Please add an integration test. You can do that by copying `test/integration/queries/` to `test/integration/trailing-slash/` and adding the trailing slash config to that.

cashmagnet:
👋 are things happening here - I don't want to bug anyone for this, just want to know if I should pick it up if needed 🥰

peterpalau:
Hi @cashmagnet

>  are things happening here - I don't want to bug anyone for this, just want to know if I should pick it up if needed

I will be finishing this soon. Thank you.

peterpalau:
I'm sorry, things went wrong here. I'm going to discard this PR.

cashmagnet:
I'm a little confused, I'm really sorry - I just don't want to steal your thunder ⚡ - does this mean you're gonna make another, or just want to be rid of this one and someone else should do it 😄

flybayer:
We have another PR opened for this: 2392

PR #2311, opened 2021-05-08
Use cookie prefix from `config.blitz.js` instead of package.json name field

mabadir:
Replaces: #1976
Closes: #1972, #1957

Adds new `cookiePrefix` property to `sessionMiddleware config`.

```
module.exports = {
  middleware: [
    sessionMiddleware({
+     cookiePrefix: 'cookie',
    }),
  ],
}
```

## Feature Checklist


mabadir:
Next step is to PR the docs for that change.

Also, I think we need to have some test coverage.

flybayer (commented):
Thanks! I fixed a bug where it wasn't working — example app test was failing.

A proper integration test can be added to `test/integration/auth/` by setting the session cookie prefix and then adding a middleware that checks and validates the `res` cookies. Probably should only do it on the first request? Or a certain request like login or something.

flybayer:
Sweet, thanks Mina! I looked and the existing tests sufficiently cover this. I also just pushed a commit to docs canary to add this.

PR #2313, opened 2021-05-08
Add Recipe Auth Tail

Anjianto:
Closes: -

- [x] Authentication with style

## Bug Checklist

- [ ] -

## Feature Checklist

- [x] Authentication with beautiful design
- [x] Lightweight code


flybayer:
Wow, impressive! I think some people will definitely like this.

That said, I'm not sure about adding this as an official recipe because this adds extra burden to us maintainers to update this recipe any time we update the "vanilla" template files. 

What do you think about adding this as a recipe in your own github repo and then people can install via `blitz install anjianto/auth-tailwind`? 

We can add this recipe to our docs/list of recipes. It's just the maintence burden of keeping it up to date that I'm worried about.

Also, I thought "Auth Tail" was a new styling library lol. So maybe "auth-tailwind" is a better name that people will instantly recognize.

Anjianto:
> Wow, impressive! I think some people will definitely like this.
> 
> That said, I'm not sure about adding this as an official recipe because this adds extra burden to us maintainers to update this recipe any time we update the "vanilla" template files.
> 
> What do you think about adding this as a recipe in your own github repo and then people can install via `blitz install anjianto/auth-tailwind`?
> 
> We can add this recipe to our docs/list of recipes. It's just the maintence burden of keeping it up to date that I'm worried about.
> 
> Also, I thought "Auth Tail" was a new styling library lol. So maybe "auth-tailwind" is a better name that people will instantly recognize.

I think you're right about maintainance this recipe.  And than about the recipe name auth-tailwind maybe make people easy to find this (the Auth Tail name from Authentication Tailwind 🤣).

Anjianto:
> > Wow, impressive! I think some people will definitely like this.
> > That said, I'm not sure about adding this as an official recipe because this adds extra burden to us maintainers to update this recipe any time we update the "vanilla" template files.
> > What do you think about adding this as a recipe in your own github repo and then people can install via `blitz install anjianto/auth-tailwind`?
> > We can add this recipe to our docs/list of recipes. It's just the maintence burden of keeping it up to date that I'm worried about.
> > Also, I thought "Auth Tail" was a new styling library lol. So maybe "auth-tailwind" is a better name that people will instantly recognize.
> 
> I think you're right about maintainance this recipe. And than about the recipe name auth-tailwind maybe make people easy to find this (the Auth Tail name from Authentication Tailwind ).

auth-tailwind has been release in https://github.com/Anjianto/auth-tailwind

Issue #2341, opened 2021-05-14
Can't resolve modules placed in 'app/(. +)pages/' path


okmr-d:
### What is the problem?

This error has occurred when I tried to import 'app/web-pages/queries/getWebPages' module

<img width="975" alt="screenshot" src="https://user-images.githubusercontent.com/4679138/118236722-64f6ea00-b4d1-11eb-971d-b2b70fc39457.png">

I tried moving "app/web-pages" to "app/aaapages", but that didn't work.
I tried moving "app/web-pages" to "app/web-page", and it worked fine.

It seems to happen in 'app/(. +)pages' path.

### Paste all your error logs here:

```
error - ./app/pages/web-pages/index.tsx:4:0
Module not found: Can't resolve 'app/web-pages/queries/getWebPages'
  2 | import { Head, Link, usePaginatedQuery, useRouter, BlitzPage, Routes } from "blitz"
  3 | import Layout from "app/core/layouts/Layout"
> 4 | import getWebPages from "app/web-pages/queries/getWebPages"
  5 | 
  6 | const ITEMS_PER_PAGE = 100
  7 | 
```

### Paste all relevant code snippets here:

```
N/A
```

### What are detailed steps to reproduce this?

1. Run `blitz generate all WebPage` command and it will generate the following files:
```
app/pages/web-pages/[webPageId]/edit.tsx
app/pages/web-pages/[webPageId].tsx
app/pages/web-pages/index.tsx
app/pages/web-pages/new.tsx
app/web-pages/components/WebPageForm.tsx
app/web-pages/queries/getWebPage.ts
app/web-pages/queries/getWebPages.ts
app/web-pages/mutations/createWebPage.ts
app/web-pages/mutations/deleteWebPage.ts
app/web-pages/mutations/updateWebPage.ts
```

2. Run `blitz dev`

3. Open `http://localhost:3000/web-pages`

### Run `blitz -v` and paste the output here:

```
macOS Big Sur | darwin-arm64 | Node: v15.12.0

blitz: 0.34.3 (global)
blitz: 0.34.3 (local)

  Package manager: yarn 
  System:
    OS: macOS 11.2.3
    CPU: (8) arm64 Apple M1
    Memory: 70.95 MB / 8.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 15.12.0 - ~/.anyenv/envs/nodenv/versions/15.12.0/bin/node
[...]

PR #2369, opened 2021-05-20
Fix For Regex in PagesPathTransformer Function

Patil2099:
Fixes ##2341 
 @flybayer I came up with this regex. If you have another solution please suggest. 

Thanks!


flybayer:
@Patil2099 thank you! Sorry for the long delay here, was on vacation.

Can you please add at least one test case to test file for this function? See https://github.com/blitz-js/blitz/issues/2341#issuecomment-842663083

Also next time, please don't delete the PR template.

flybayer:
@Patil2099 are you able to finish this?

Patil2099:
Sorry. I forgot about it I will make the required changes soon.

flybayer:
This was fixed in https://github.com/blitz-js/blitz/pull/2502

PR #2304, opened 2021-05-04
Adding JIT mode for Tailwind recipe

Anjianto:
Closes: -

Discussions: [RFC] Adding JIT mode for Tailwind recipe https://github.com/blitz-js/blitz/discussions/2303

## Bug Checklist

- [x] Slow in Dev Mode, in web browser, Dev Tools, etc

## Feature Checklist

- [x] Adding JIT mode to tailwind recipe https://github.com/blitz-js/blitz/discussions/2303

PR #2383, opened 2021-05-23
Fix Regression on Tailwind Recipe purging in production

benbender:
Fix Regression on Tailwind Recipe for css-purging in production (originally fixed in #930, broken since #2304) and remove the aggressive, non-standard purge-settings as they may result in over-purging.


flybayer:
Perfect, thank you!!

PR #2392, opened 2021-05-25
Fix queries & mutations to work with trailing slash config (patch)

timbooker:
Closes: #2294 

This PR adds the implementation to consider the trailingSlash configuration in the API requests from either await direct function calls inside useEffects or useQuery.

Checklist
 Added trailingSlash to the blitz runtime dat
 Updated getApiUrlFromResolverFilePath to add trailing slash if getBlitzRuntimeData().trailingSlash is true.


flybayer (approved):
Thank you!!

Issue #2387, opened 2021-05-25
Error: Cannot find module 'next/constants' in Vercel deployment


remjx:
### What is the problem?

`blitz build` fails in Vercel but suceeds locally and in CI.

Similar issue: https://github.com/blitz-js/blitz/issues/2195

### Paste all your error logs here:

Relevant logs from Vercel:

```
20:48:45.120 | &gt; blitz@0.35.0 postinstall /vercel/path0/packages/website/node_modules/blitz
-- | --
20:48:45.120 | &gt; node scripts/postinstall.js
20:48:45.470 | internal/modules/cjs/loader.js:883
20:48:45.470 | throw err;
20:48:45.470 | ^
20:48:45.470 | Error: Cannot find module 'next/constants'
20:48:45.470 | Require stack:
20:48:45.471 | - /vercel/path0/packages/website/node_modules/@blitzjs/config/dist/blitzjs-config.cjs.dev.js
20:48:45.471 | - /vercel/path0/packages/website/node_modules/@blitzjs/config/dist/blitzjs-config.cjs.js
20:48:45.471 | - /vercel/path0/packages/website/node_modules/@blitzjs/core/dist/blitz-data-f63b7a88.cjs.dev.js
20:48:45.471 | - /vercel/path0/packages/website/node_modules/@blitzjs/core/document/dist/blitzjs-core-document.cjs.dev.js
20:48:45.471 | - /vercel/path0/packages/website/node_modules/@blitzjs/core/document/dist/blitzjs-core-document.cjs.js
20:48:45.471 | - /vercel/path0/packages/website/node_modules/blitz/dist/blitz.cjs.dev.js
20:48:45.472 | - /vercel/path0/packages/website/node_modules/blitz/dist/blitz.cjs.js
20:48:45.472 | at Function.Module._resolveFilename (internal/modules/cjs/loader.js:880:15)
20:48:45.472 | at Function.Module._load (internal/modules/cjs/loader.js:725:27)
20:48:45.472 | at Module.require (internal/modules/cjs/loader.js:952:19)
20:48:45.472 | at require (internal/modules/cjs/helpers.js:88:18)
20:48:45.472 | at Object.getConfig (/vercel/path0/packages/website/node_modules/@blitzjs/config/dist/blitzjs-config.cjs.dev.js:114:18)
[...]

PR #2402, opened 2021-05-27
fix bug: cannot find module `next/constants` (patch)

flybayer:
Closes: #2387

This adds a workaround for an npm bug where the `next` package is sometimes missing. It's an npm bug but happens because of the `"next": "npm:@blitzjs/next@10.1.3-0.35.0"` syntax we use to install our fork at `node_modules/next`


remjx:
Awesome!

Issue #2354, opened 2021-05-17
Build config and custom server during `blitz build`


andreasasprou:
### What do you want and why?

When creating a production Dockerfile, I don't want to include all my source files required to build `server.ts` (or `blitz.config.ts`!).

### Possible implementation(s)

**For custom server**
Build custom server during `blitz build` and use those built files when running `blitz start`:

- modify `getCustomServerPath` to include `.blitz/custom-server.js`
- modify `startCustomServer` to run `.blitz/custom-server.js` if it exists and we're running in production
- modify [build script](https://github.com/blitz-js/blitz/blob/bf7eed62bd1aaf8a23779eb02e19a06cb5ffc0ff/packages/server/src/build.ts) to build custom-server and copy files (see `saveBlitzVersion` and adjust to copy to `.blitz/custom-server.js`)

Reflect on why the files currently reside in `.blitz` and not `.blitz/build` for the above use case.

### Additional context

After docker run:
![image](https://user-images.githubusercontent.com/8077469/118475336-08c7dc00-b704-11eb-9226-604020c5ffbf.png)

flybayer:
Makes sense!

We put blitz.config.js in `.blitz/` because we need the config before `.blitz/build/` is created.

The custom server could go inside `.blitz/build/` but that folder will go away once the next.js fork migration is complete. Which is why I put it just inside `.blitz/`.

### Implementation as I'm seeing it

1. Skip the build config step if command is `blitz start` here: https://github.com/blitz-js/blitz/blob/canary/packages/cli/src/index.ts#L13
2. Extract build code from `startCustomServer` into a new `buildCustomServer` function in https://github.com/blitz-js/blitz/blob/canary/packages/server/src/next-utils.ts#L247 (so startCustomServer will not build unless watch:true
3. Change `blitz dev` to run `buildCustomServer` and `startCustomServer` here: https://github.com/blitz-js/blitz/blob/canary/packages/server/src/dev.ts#L43-L44
4. Run `buildCustomServer` on this line: https://github.com/blitz-js/blitz/blob/canary/packages/server/src/build.ts#L39

mabadir:
Hey all, I am picking up this issue, and will work on it.

PR #2408, opened 2021-05-28
Fix to not build custom server during `blitz start`

mabadir:
Closes: #2354

- Building Custom Server during Blitz Build
- `startCustomServer` function, now doesn't build by default.
- Introduce a separate `buildCustomServer` fn

## Feature Checklist

### DIFF
--- a/packages/server/src/build.ts
+++ b/packages/server/src/build.ts
@@ -3,7 +3,7 @@ import {copy} from "fs-extra"
 import {resolve} from "path"
 import {saveBlitzVersion} from "./blitz-version"
 import {normalize, ServerConfig} from "./config"
-import {nextBuild} from "./next-utils"
+import {buildCustomServer, customServerExists, nextBuild} from "./next-utils"
 import {configureStages} from "./stages"
 
 export async function build(config: ServerConfig) {
@@ -36,7 +36,7 @@ export async function build(config: ServerConfig) {
   })
 
   await saveBlitzVersion(buildFolder)
-
+  if (customServerExists()) await buildCustomServer({watch})
   await nextBuild(nextBin, buildFolder, manifest, config)
 
   const rootNextFolder = resolve(rootFolder, ".next")
--- a/packages/server/src/next-utils.ts
+++ b/packages/server/src/next-utils.ts
@@ -244,12 +244,45 @@ interface CustomServerOptions {
   watch?: boolean
 }
 
+const getEsbuildOptions = (): esbuild.BuildOptions => {
+  const pkg = readJSONSync(path.join(pkgDir.sync()!, "package.json"))
+  return {
+    entryPoints: [getCustomServerPath()],
+    outfile: getCustomServerBuildPath(),
+    format: "cjs",
+    bundle: true,
+    platform: "node",
+    external: [
+      "blitz",
+      "next",
+      ...Object.keys(require("blitz/package").dependencies),
+      ...Object.keys(pkg.dependencies),
+      ...Object.keys(pkg.devDependencies),
+    ],
+  }
+}
+
+export function buildCustomServer({watch}: CustomServerOptions = {}) {
+  const esbuildOptions = getEsbuildOptions()
+  if (watch) {
+    esbuildOptions.watch = {
+      onRebuild(error) {
+        if (error) {
+          log.error("Failed to re-build custom server")
+        } else {
+          log.newline()
+          log.progress("Custom server changed - rebuilding...")
+        }
+      },
+    }
+  }
+  return esbuild.build(esbuildOptions)
+}
 export function startCustomServer(
   cwd: string,
   config: ServerConfig,
   {watch}: CustomServerOptions = {},
 ) {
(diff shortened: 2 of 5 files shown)


packages/server/src/next-utils.ts
```diff
@@ -248,6 +248,7 @@ export function startCustomServer(
   cwd: string,
   config: ServerConfig,
   {watch}: CustomServerOptions = {},
+  buildOnly: boolean = false,
```
flybayer, line 251: Move `buildOnly` into the option object

packages/server/src/next-utils.ts
```diff
@@ -296,7 +297,7 @@ export function startCustomServer(
         "next",
         ...Object.keys(require("blitz/package").dependencies),
         ...Object.keys(pkg.dependencies),
-        ...Object.keys(pkg.devDependencies),
+        // ...Object.keys(pkg.devDependencies),
```
flybayer, line 300: Why?

packages/server/src/build.ts
```diff
@@ -36,9 +36,14 @@ export async function build(config: ServerConfig) {
...
 
   await saveBlitzVersion(buildFolder)
-
   await nextBuild(nextBin, buildFolder, manifest, config)
 
+  try {
+    await startCustomServer(rootFolder, config, {watch, buildOnly: true})
+  } catch (e) {
+    console.log("No custom server found, proceeding")
+  }
```
flybayer, line 45: - I think we should first check if custom server exists instead of doing try/catch

flybayer (commented):
Not sure if you are done? Mark as not-draft if you are done.

However, this doesn't fully address the original issue. The custom server is still being built on `blitz start`. I think we should separate the function into `buildCustomServer` and `startCustomServer`.

mabadir:
Yes sorry, I didn't fully understand the main concern from the original issue. I will review this PR today and will push few updates. Sorry about that.

packages/server/src/next-utils.ts
```diff
@@ -317,9 +338,6 @@ export function startCustomServer(
       }
     }
 
-    // eslint-disable-next-line @typescript-eslint/no-floating-promises
-    esbuild.build(esbuildOptions).then(() => {
-      spawnServer()
-    })
+    spawnServer()
```
flybayer, line 341: Looking good! This is a regression for `blitz dev` use case.

Here's the full list. Sorry, I should have put this earlier.

- [x] `blitz build` builds custom server
- [x] `blitz start` runs but does not build custom server
- [ ] `blitz dev` should both build and run custom server (this is working correctly in existing code, but it's not on this PR)

mabadir: I think adding:
`await buildCustomServer()` to this line will solve it:
https://github.com/blitz-js/blitz/blob/8eb1509b9ad45b647a702381b4f693ed3b8aefe6/packages/server/src/dev.ts#L48

flybayer: That handles initial build, but not watch. Did you test `blitz dev`?

I think we should conditionally do `spawnServer()` vs 

```
esbuild.build(esbuildOptions).then(() => {
      spawnServer()
})
```

based on watch flag. Currently `esbuildOptions` in this function are not used and so hot reload doesn't work.

mabadir: You're right, I rushed this one in. This is what I am coming up with now.

```javascript

export function startCustomServer(
    cwd: string,
    config: ServerConfig,
    {watch}: CustomServerOptions = {},
) {
  const serverBuildPath = getCustomServerBuildPath()

  let spawnEnv = getSpawnEnv(config)
  if (config.env === "prod") {
    spawnEnv = {...spawnEnv, NODE_ENV: "production"}
  }

  return new Promise<void>((res, rej) => {
    let process: ChildProcess

    const RESTART_CODE = 777777

    const spawnServer = () => {
      process = spawn("node", [serverBuildPath], {
        cwd,
        env: spawnEnv,
        stdio: "inherit",
      })
          .on("exit", (code: number) => {
            if (code === 0) {
              res()
            } else if (watch && code === RESTART_CODE) {
              spawnServer()
            } else {
              rej(`server.js failed with status code: ${code}`)
            }
          })
          .on("error", (err) => {
            console.error(err)
            rej(err)
          })
    }

    if (watch) {
      // Handle build & Starting server
      const esbuildOptions = getEsbuildOptions()
      esbuildOptions.watch = {
        onRebuild(error) {
          if (error) {
            log.error("Failed to re-build custom server")
          } else {
            log.newline()
            log.progress("Custom server changed - restarting...")
            log.newline()
            //@ts-ignore -- incorrect TS type from node
            process.exitCode = RESTART_CODE
            process.kill("SIGABRT")
          }
        },
      }
      // eslint-disable-next-line @typescript-eslint/no-floating-promises
      esbuild.build(esbuildOptions).then(() => {
        spawnServer()
      })
    } else {
      // No build required, Start server
      spawnServer()
    }
  })
}
```

flybayer:
yay!

Issue #2372, opened 2021-05-20
Error: The cookie prefix used has invalid characters.


jesseweed:
### What is the problem?

When creating a new app that has a `.` in the name (i.e.  `test.app`),  setting this as the `cookiePrefix` in `blitz.config.ts` throw the error `Error: The cookie prefix used has invalid characters. Only alphanumeric characters, "-"  and "_" character are supported`.

I was able to get it working but just changing the value of  `cookiePrefix` to just `testapp`, so not a big deal. But should consider adding some kind of filter to strip out any chars that aren't allowed in cookies if that key is going to be set automatically.

### Paste all your error logs here:

```
ready - started server on 0.0.0.0:3000, url: http://localhost:3000
info  - Loaded env from /Volumes/Web/projects/test.app/.env
Error: The cookie prefix used has invalid characters. Only alphanumeric characters, "-"  and "_" character are supported
    at assert$1 (/Volumes/Web/projects/test.app/node_modules/@blitzjs/core/server/dist/blitzjs-core-server.cjs.dev.js:488:25)
    at sessionMiddleware (/Volumes/Web/projects/test.app/node_modules/@blitzjs/core/server/dist/blitzjs-core-server.cjs.dev.js:630:3)
    at Object.<anonymous> (/Volumes/Web/projects/test.app/.blitz/blitz.config.js:24:40)
    at Module._compile (node:internal/modules/cjs/loader:1102:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1131:10)
    at Object.newLoader [as .js] (/Volumes/Web/projects/test.app/node_modules/pirates/lib/index.js:104:7)
    at Module.load (node:internal/modules/cjs/loader:967:32)
    at Function.Module._load (node:internal/modules/cjs/loader:807:14)
    at Module.require (node:internal/modules/cjs/loader:991:19)
    at require (node:internal/modules/cjs/helpers:92:18)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

### Paste all relevant code snippets here:

```
import { sessionMiddleware, simpleRolesIsAuthorized } from "blitz"

module.exports = {
  middleware: [
    sessionMiddleware({
      // this is the code that throws the error
      cookiePrefix: "test.app",
      isAuthorized: simpleRolesIsAuthorized,
    }),
  ],
  /* Uncomment this to customize the webpack config
[...]

PR #2427, opened 2021-05-31
Fix so `blitz new` always generates valid cookiePrefix

swinner2:
Closes: #2372 

Adds `safeNameSlug` which is the app name with any special chars replaced by `-`, and uses it for the `cookiePrefix`.

## Bug Checklist

- N/A Integration test added (see [test docs](https://blitzjs.com/docs/contributing#running-tests) if needed)

## Feature Checklist

- N/A Integration test added (see [test docs](https://blitzjs.com/docs/contributing#running-tests) if needed)
- N/A Documentation added/updated (submit PR to [blitzjs.com repo](https://github.com/blitz-js/blitzjs.com))


packages/generator/src/generators/app-generator.ts
```diff
@@ -39,6 +39,7 @@ export class AppGenerator extends Generator<AppGeneratorOptions> {
   async getTemplateValues() {
     return {
       name: this.options.appName,
+      safeNameSlug: this.options.appName.replace(/[^a-zA-Z0-9-_]/, "-"),
```
flybayer, line 42:
```suggestion
      safeNameSlug: this.options.appName.replace(/[^a-zA-Z0-9-_]/g, "-"),
```

flybayer (approved):
Thank you! Had a small bug, but I fixed it.

PR #2441, opened 2021-06-03
Add `validateZodSchema()` API for better form errors

### LINKED ISSUE #2412: Update our form templates to use new Zod error feature
### What do you want and why?

Currently [our `<Form>` templates](https://github.com/blitz-js/blitz/tree/canary/packages/generator/templates/app/_forms) use [`error.formErrors.fieldErrors`](https://github.com/blitz-js/blitz/blob/canary/packages/generator/templates/app/_forms/formik/Form.tsx#L40) to report zod form errors. 

But this doesn't work with nested objects in the form, so we need to update it to use `error.format()`. Example: https://twitter.com/colinhacks/status/1394326038570799105

For the other two, change to use `error.format()`.

swinner2:
Closes: #2412

adds a util to format zod error messages to be compatible with react-final-form and formik

## Feature Checklist


swinner2:
@flybayer Looking for some feedback on this one. Also, I'm not sure how we want to import this function into the form.

packages/core/src/utils/index.test.ts
```diff
@@ -0,0 +1,50 @@
...
+import {formatZodErrors} from "./index"
+
+describe("formatZodErrors", () => {
+  const MySchema = z.object({
+    test: z.string(),
+  })
+
+  it("formats the zod error", () => {
+    const result = MySchema.safeParse({})
+    expect(formatZodErrors(result.error.format())).toEqual({test: "Required"})
```
swinner2, line 11: > Property 'error' does not exist on type '{ success: true; data: { test: string; }; } | { success: false; error: ZodError<{ test: string; }>; }'.

I'm not sure how to fix this.

flybayer: Adding this line before that will fix it.

```ts
if (result.success) throw new Error("Schema should not return success")
```

Typescript will use this to eliminate the `{ success: true; data: { test: string; }; }` from the next line

flybayer:
Upon initial inspection, this looks correct!

flybayer:
> Also, how do we get the function into the form components? We need to use the function here https://github.com/blitz-js/blitz/blob/canary/packages/generator/templates/app/_forms/finalform/Form.tsx#L33

Add `formatZodErrors ` export to here: `packages/core/src/index.ts`

And then add `import {formatZodErrors} from 'blitz'` in the form files.

---

Actually after thinking about this a bit more, can we set up the API to work like this? (can keep the `formatZodErrors` internally if it helps).

```tsx
    <Formik
      initialValues={initialValues || {}}
      validate={validateZodSchema(Login)}
```

swinner2:
I'm not sure why the Example Apps failure is happening. Can you check that out @flybayer ?

flybayer:
@swinner2 should be fixed now! For some unknown reason the form was being submitted during pre rendering.

I also tweaked the validateZodSchema api a bit.

The only thing left is to open a PR to add docs for this! I think let's add a new utilities page under the Basics section and add this to it. Here's where you add to the menu: https://github.com/blitz-js/blitzjs.com/blob/main/app/core/navs/documentation.json#L27-L42

swinner2:
That tweak is awesome! I wanted to do that, but didn't know how in JS @flybayer

swinner2:
Added the docs here @flybayer https://github.com/blitz-js/blitzjs.com/pull/498

flybayer:
Yay!!

Issue #2295, opened 2021-05-03
Cannot run Blitz with Webpack 5


ssledorze:
### What is the problem?

Blitzjs does not work (yet) using webpack 5.

### Paste all your error logs here:

Try to access the App once compiled results in this crash:

```
23:31:19.561 ERROR Error while processing the request 

 Error  Cannot find module '../../webpack-runtime.js'
Require stack:
- /Users/stephaneledorze/Documents/projects/opteameo/cms/ideation-cms/.next/blitz/db.js
- /Users/stephaneledorze/Documents/projects/opteameo/cms/ideation-cms/node_modules/@blitzjs/core/server/dist/blitzjs-core-server.cjs.dev.js
- /Users/stephaneledorze/Documents/projects/opteameo/cms/ideation-cms/node_modules/@blitzjs/core/server/dist/blitzjs-core-server.cjs.js
- /Users/stephaneledorze/Documents/projects/opteameo/cms/ideation-cms/node_modules/blitz/dist/blitz.cjs.dev.js
- /Users/stephaneledorze/Documents/projects/opteameo/cms/ideation-cms/node_modules/blitz/dist/blitz.cjs.js
- /Users/stephaneledorze/Documents/projects/opteameo/cms/ideation-cms/blitz.config.js
- /Users/stephaneledorze/Documents/projects/opteameo/cms/ideation-cms/next.config.js
- /Users/stephaneledorze/Documents/projects/opteameo/cms/ideation-cms/node_modules/next/dist/next-server/server/config.js
- /Users/stephaneledorze/Documents/projects/opteameo/cms/ideation-cms/node_modules/next/dist/next-server/server/next-server.js
- /Users/stephaneledorze/Documents/projects/opteameo/cms/ideation-cms/node_modules/next/dist/server/next.js
- /Users/stephaneledorze/Documents/projects/opteameo/cms/ideation-cms/node_modules/next/dist/server/lib/start-server.js
- /Users/stephaneledorze/Documents/projects/opteameo/cms/ideation-cms/node_modules/next/dist/cli/next-dev.js
- /Users/stephaneledorze/Documents/projects/opteameo/cms/ideation-cms/node_modules/next/dist/bin/next
- /Users/stephaneledorze/Documents/projects/opteameo/cms/ideation-cms/node_modules/@blitzjs/server/bin/next-patched
details:
{
  code: 'MODULE_NOT_FOUND',
  requireStack: [
[...]

PR #2453, opened 2021-06-04
Fix webpack 5 support (Webpack 5 now used by default) (patch)

flybayer:
Closes: https://github.com/blitz-js/blitz/issues/2295

This removes our `with-blitz` wrapper by moving our default config items into nextjs core


nextjs/packages/next/build/webpack-config.ts
```diff
@@ -997,6 +1014,18 @@ export default async function getBaseWebpackConfig(
...
+          ? []
+          : [
+              {
+                issuer: /(mutations|queries)(?!.*\.client)/,
+                resource: /_resolvers/,
+                use: { loader: 'null-loader' },
+              },
+              // Because of packages/core/src/prisma-utils.ts
+              { test: /[\\/]npm-which[\\/]/, use: { loader: 'null-loader' } },
+              { test: /[\\/]cross-spawn[\\/]/, use: { loader: 'null-loader' } },
```
merceyz, line 1027:
```suggestion
                use: { loader: require.resolve('null-loader') },
              },
              // Because of packages/core/src/prisma-utils.ts
              { test: /[\\/]npm-which[\\/]/, use: { loader: require.resolve('null-loader') } },
              { test: /[\\/]cross-spawn[\\/]/, use: { loader: require.resolve('null-loader') } },
```

These should be resolved up front so the correct version is used / avoid relying on hoisting

flybayer: Thanks!

Issue #2380, opened 2021-05-23
name conflict


fleck:
### What is the problem?

building a wiki like application on blitz, the primary model for the application is "page". This resulted in the error:

```
../../node_modules/etag/index.js:22:0
Module not found: Can't resolve 'fs'
null
```

found out it was because my queries/mutations folder were inside of pages (name collision with the special pages folder). Would it be worth adding a warning to the generator or adding some special suffix if there's a name collision with a "magic" folder when generating?

### Paste all your error logs here:

```
../../node_modules/etag/index.js:22:0
Module not found: Can't resolve 'fs'
null
```

### Paste all relevant code snippets here:

```
PASTE_HERE (leave the ``` marks)
```

### What are detailed steps to reproduce this?

1. blitz generate all item name:string url:string content:string generic_name:string type:string standalone:boolean
2. goto http://localhost:3000/pages

### Run `blitz -v` and paste the output here:

```
macOS Catalina | darwin-x64 | Node: v14.16.0

blitz: 0.35.0 (local)

  Package manager: yarn 
  System:
    OS: macOS 10.15.7
    CPU: (16) x64 Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
    Memory: 10.64 GB / 32.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 14.16.0 - ~/.volta/tools/image/node/14.16.0/bin/node
    Yarn: 1.22.10 - ~/.volta/tools/image/yarn/1.22.10/bin/yarn
    npm: 6.14.11 - ~/.volta/tools/image/node/14.16.0/bin/npm
    Watchman: Not Found
  npmPackages:
    @prisma/client: ~2.20 => 2.20.1 
    blitz: 0.35.0 => 0.35.0 
    prisma: ~2.20 => 2.20.1 
    react: 0.0.0-experimental-6a589ad71 => 0.0.0-experimental-6a589ad71 
    react-dom: 0.0.0-experimental-6a589ad71 => 0.0.0-experimental-6a589ad71 
    typescript: ~4.2 => 4.2.4 
```

mabadir:
Hello @fleck; thanks for reporting this issue, I was able to replicate it as well. Your proposal to throw error when using a reserved name for model name, sounds like a good approach.

mabadir:
Here is a proposed implementation:

[...]

PR #2455, opened 2021-06-05
Prevent using Reserved Keywords in `generate` command

mabadir:
Closes: #2380

Comparing model name against reserved keywords list.

## Bug Checklist

- [x] Unit test added

### DIFF
--- a/packages/cli/src/commands/generate.ts
+++ b/packages/cli/src/commands/generate.ts
@@ -184,6 +184,15 @@ export class Generate extends Command {
     }
   }
 
+  checkReservedKeyword(modelName: string) {
+    const reservedKeywords = ["page"]
+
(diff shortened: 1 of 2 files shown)


flybayer (commented):
Thanks Mina! After reviewing this, I think it would actually be better to not have any reserved keywords.

Instead we just change the generated paths to accommodate. 

`blitz generate all page` -> generates to `app/page-model/`

Same for `api`, `query`, and `mutation`.

What do you think?

mabadir:
Makes sense, that will impact our auto generated routes I assume?

flybayer:
> Makes sense, that will impact our auto generated routes I assume?

Yes

mabadir:
Considering, it's completely different approach, I am closing this PR and opening a new one.

Issue #2436, opened 2021-06-02
Dependencies missing prevents pnpm working


vjpr:
### What is the problem?

When `blitz` is install as a dev dependency in a project using `pnpm`, there are many dependencies missing.

This is probably because blitz relies on dependency hoisting in `yarn`.

These need to be added:

- minimist
- debug
- tslib
- babylon

### Paste all your error logs here:

```
Error: Cannot find module 'debug'
Require stack:
- /Users/Vaughan/dev-mono/thirtyfive/node_modules/.pnpm/@blitzjs+config@0.36.3/node_modules/@blitzjs/config/dist/blitzjs-config.cjs.dev.js
- /Users/Vaughan/dev-mono/thirtyfive/node_modules/.pnpm/@blitzjs+config@0.36.3/node_modules/@blitzjs/config/dist/blitzjs-config.cjs.js
- /Users/Vaughan/dev-mono/thirtyfive/node_modules/.pnpm/@blitzjs+cli@0.36.3_typescript@4.2.4/node_modules/@blitzjs/cli/lib/src/index.js
- /Users/Vaughan/dev-mono/thirtyfive/node_modules/.pnpm/blitz@0.36.3_5567bc033954b9273438214653b4e1b9/node_modules/blitz/cli/dist/blitz-cli.cjs.dev.js
- /Users/Vaughan/dev-mono/thirtyfive/node_modules/.pnpm/blitz@0.36.3_5567bc033954b9273438214653b4e1b9/node_modules/blitz/cli/dist/blitz-cli.cjs.js
```

### Paste all relevant code snippets here:

...

### What are detailed steps to reproduce this?

pnpx 

### Run `blitz -v` and paste the output here:

```
macOS Big Sur | darwin-x64 | Node: v14.2.0

blitz: 0.36.3 (local)

  Package manager: npm
  System:
    OS: macOS 11.2.1
    CPU: (16) x64 Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
    Memory: 409.48 MB / 64.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.2.0 - ~/nvm/versions/node/v14.2.0/bin/node
    Yarn: 1.22.4 - ~/.yarn/bin/yarn
    npm: 6.14.4 - ~/nvm/versions/node/v14.2.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  npmPackages:
[...]

PR #2464, opened 2021-06-08
Fix one part of pnp issues (patch)

flybayer:
Related: https://github.com/blitz-js/blitz/issues/2436

Fix to only patch npm install process if npm is actually being used.

### DIFF
--- a/packages/blitz/scripts/postinstall.js
+++ b/packages/blitz/scripts/postinstall.js
@@ -21,6 +21,9 @@ try {
   // noop
 }
 
+const isUsingNpm =
+  Boolean(process.env.npm_execpath) && /node_modules[\\/]npm/.test(process.env.npm_execpath)
+
 /*
   Adapted from https://github.com/prisma/prisma/blob/974cbeff4a7f616137ce540d0ec88a2a86365892/src/packages/client/scripts/postinstall.js
 */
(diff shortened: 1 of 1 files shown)


packages/blitz/scripts/postinstall.js
```diff
@@ -21,6 +21,9 @@ try {
   // noop
 }
 
+const isUsingNpm =
+  Boolean(process.env.npm_execpath) && /node_modules[\\/]npm/.test(process.env.npm_execpath)
```
merceyz, line 25: The path isn't guaranteed to contain `node_modules` (it doesn't on my system) so it would be better to check `npm_config_user_agent`
```
npm_config_user_agent=yarn/1.22.10 npm/? node/v16.3.0 win32 x64
npm_config_user_agent=yarn/3.0.0-rc.2 npm/? node/16.3.0 win32 x64
npm_config_user_agent=npm/6.14.7 node/v16.3.0 win32 x64
npm_config_user_agent=pnpm/5.4.11 npm/? node/v16.3.0 win32 x64
```

flybayer: Ah cool, thanks! I had taken this from an existing npm library + stack overflow. I've switched to your approach.

Issue #2466, opened 2021-06-09
BlitzConfig is missing typings for some nextjs configurations


lksnmnn:
### What is the problem?

When extending the `BlitzConfig` with nextjs specific configurations like `images`, `env` or external configs (`pwa`), TS throws type errors.

Quoting from Dan Greaves  [comment on Discord](https://discord.com/channels/802917734999523368/814627656611397692/852072311073931274)

> @lksnmnn annoyingly, next doesn't actually type all the config options, even though they work just fine. In the `BlitzConfig` type (https://github.com/blitz-js/blitz/blob/canary/packages/config/src/index.ts#L15), note that it extends `PublicNextConfig` (https://github.com/blitz-js/blitz/blob/canary/nextjs/packages/next/next-server/server/config-shared.ts#L12). I think `BlitzConfig` should actually extend `NextConfig` (https://github.com/blitz-js/blitz/blob/canary/nextjs/packages/next/next-server/server/config-shared.ts#L69) which includes the` Record<string, any>` which allows you to put any other key in, which matches Next. Either that or the missing types need to be added into `BlitzConfig`. I would probably open an issue for it and see what the blitz guys think.

With a temporary solution on top:
> In the mean time, you can change your config type to `BlitzConfig & Record<string, any>` - that will allow you to add the images key without typescript complaining about it.
It's referenced this way in the actual Blitz documentation too, so it should work: https://blitzjs.com/docs/image-optimization#domains

### Paste all your error logs here:

```
type images does not exist on BlitzConfig
```

### Paste all relevant code snippets here:

```
const config: BlitzConfig = {
   ...
  images: {
    domains: ["localhost"],
  },
 ...
}
```

### What are detailed steps to reproduce this?

n/a

### Run `blitz -v` and paste the output here:

```
acOS Catalina | darwin-x64 | Node: v14.17.0

blitz: 0.36.4 (global)
blitz: 0.37.0 (local)

  Package manager: yarn 
  System:
    OS: macOS 10.15.7
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 29.16 GB / 64.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 14.17.0 - ~/.nvm/versions/node/v14.17.0/bin/node
    Yarn: 1.22.10 - ~/.nvm/versions/node/v14.17.0/bin/yarn
[...]

PR #2471, opened 2021-06-10
fix `BlitzConfig` type to allow extra fields (patch)

dangreaves:
Closes: #2466

There are a number of Next.js configuration options, like the `images` key which are not included in the [PublicNextConfig](https://github.com/blitz-js/blitz/blob/canary/nextjs/packages/next/next-server/server/config-shared.ts#L12) type. In Next.js, they handle this by appending `Record<string, any>` via the [NextConfig](https://github.com/blitz-js/blitz/blob/canary/nextjs/packages/next/next-server/server/config-shared.ts#L69) type.

To allow all the documented Next.js config options to be passed into BlitzConfig, without typescript complaining, this PR updates BlitzConfig to extend [NextConfig](https://github.com/blitz-js/blitz/blob/canary/nextjs/packages/next/next-server/server/config-shared.ts#L69), rather than [PublicNextConfig](https://github.com/blitz-js/blitz/blob/canary/nextjs/packages/next/next-server/server/config-shared.ts#L12), which includes the `Record<string, any>` escape hatch.

## Bug Checklist

## Feature Checklist

### DIFF
--- a/packages/config/src/index.ts
+++ b/packages/config/src/index.ts
@@ -1,21 +1,21 @@
 import * as esbuild from "esbuild"
 import fs from "fs"
 import {existsSync, readJSONSync} from "fs-extra"
-import {PublicNextConfig} from "next/dist/next-server/server/config"
+import {NextConfig} from "next/dist/next-server/server/config"
 import path, {join} from "path"
 import pkgDir from "pkg-dir"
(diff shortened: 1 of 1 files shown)


flybayer (approved):
Sweet, thank you!

Issue #2449, opened 2021-06-04
Add a setPublicData(userId, publicData) API to update active user sessions (i.e. update a user's role)


lksnmnn:
### What do you want and why?
I want to be able to update publicData in users' sessions, such that the changes are reflected and used, when the user is doing the next request.

My example use case is changing a user's role. Assuming a user has the role ADMIN and I want to change that user's role to USER, then ideally I could update the user's session such that the new role is being enforced in future requests done by said user.

Right now I need to delete all active sessions of that user to make sure that the user has no ADMIN access anymore. As the user does not expect to be forcefully logged out, the UX for this is quite bad.

### Possible implementation(s)

How might we do this?
Write a helper function which updates all active sessions of a user with the provided session data.

```ts
export const setPublicData = async (userId, publicData) {
   await db.session.update({ where: { userId }, data: { publicData } }) 
}
```

### Additional context

See discord discussion: https://discord.com/channels/802917734999523368/843912213399601192/850068347453440030

lksnmnn:
Ha, I've hacked this together in our app and this seems to work.

```
import db from "db"

export const setPublicData = async (userId: string, publicData) => {
  const publicDataString = JSON.stringify(publicData)
  // update all of the user's sessions
  await db.session.updateMany({ where: { userId }, data: { publicData: publicDataString } })
}
```

flybayer:
Yes!! Definitely need this. Your code looks pretty much like we need with one exception. 

Currently all `setPublicData` calls **merge** values with existing publicData. So probably we should take the same approach here? Our current code to do this is here: https://github.com/blitz-js/blitz/blob/canary/packages/core/src/server/auth/sessions.ts#L927-L935


For whoveever wants to work on this:

- Add this new API to https://github.com/blitz-js/blitz/blob/canary/packages/core/src/server/auth/sessions.ts
- Add new API export to this file: https://github.com/blitz-js/blitz/blob/canary/packages/core/src/server/index.ts

lksnmnn:
I've added this rudimentary version of a `setPublicData` function. It *works* but the corresponding user will expierence weird CSRF / Invalid Session issues. So maybe there is another bug, that will surface during development.

Since all sessions of a user could potentially be in use, we need to update each session with the new data. An additional merging step is not needed (?).

Here is my current code:

```ts
import db from "db"

export const setPublicData = async (userId: string, newPublicData) => {
[...]

PR #2473, opened 2021-06-10
Add `setPublicDataForUser()` — useful for updating the role of another user

lksnmnn:
Closes: https://github.com/blitz-js/blitz/issues/2449
Documentation PR: https://github.com/blitz-js/blitzjs.com/pull/500

This change adds a new API to update a user's `publicData` across all of the user's active sessions. Among other things this feature solves the common usecase, where a user's role is changed and the change should be reflected immediatly, i.e., when the user is denoted from "admin" to "user".

## Bug Checklist

## Feature Checklist


flybayer (approved):
Thank you!!

lksnmnn:
There is one thing I might want to test with this, that is actually being logged in with both users and check what happens to the user, when his publiData is being updated.

In my version of this function, which I posted in the linked issue, it worked, but when I changed browsers to quickly after updating the role, the user got weird invalid session and CSRF errors. Which is probably a bug in the session logic, but using this new API could trigger this.

flybayer:
@lksnmnn ok good to know. You are going to test?

lksnmnn:
I could not reproduce this bug in the auth example. So I guess the issue was my code doing something weird. 🚀

flybayer:
Ok thank you, we'll roll with this and address the bug later if it shows up

Issue #2440, opened 2021-06-02
Image domain error when using SSR


acornellier:
### What is the problem?
https://codesandbox.io/s/heuristic-frog-qzvhy?file=/blitz.config.js

When I use `getServerSideProps`, and I render an `<Image />`, an error is thrown that says 
```
Error: Invalid src prop (https://storage.googleapis.com/[...].jpg) on `next/image`, hostname "storage.googleapis.com" is not configured under images in your `next.config.js`
```

But I definitely have the domain configured in `blitz.config.js`
```
  images: {
    domains: [`storage.googleapis.com`],
  },
```

Interestingly, it works after deploying with Vercel...

### Paste all your error logs here:

```
Error: Invalid src prop (https://storage.googleapis.com/coleuspics/photos/user-Monkey/IMG_2756.jpg) on `next/image`, hostname "storage.googleapis.com" is not configured under images in your `next.config.js`
See more info: https://nextjs.org/docs/messages/next-image-unconfigured-host
    at defaultLoader (/Users/acornellier/git/Leaflet/node_modules/next/dist/client/image.js:43:830)
    at defaultImageLoader (/Users/acornellier/git/Leaflet/node_modules/next/dist/client/image.js:17:260)
    at /Users/acornellier/git/Leaflet/node_modules/next/dist/client/image.js:11:410
    at Array.map (<anonymous>)
    at generateImgAttrs (/Users/acornellier/git/Leaflet/node_modules/next/dist/client/image.js:11:396)
    at Image (/Users/acornellier/git/Leaflet/node_modules/next/dist/client/image.js:32:1038)
    at processChild (/Users/acornellier/git/Leaflet/node_modules/react-dom/cjs/react-dom-server.node.development.js:3367:14)
    at resolve (/Users/acornellier/git/Leaflet/node_modules/react-dom/cjs/react-dom-server.node.development.js:3284:5)
    at ReactDOMServerRenderer.render (/Users/acornellier/git/Leaflet/node_modules/react-dom/cjs/react-dom-server.node.development.js:3767:22)
    at ReactDOMServerRenderer.read (/Users/acornellier/git/Leaflet/node_modules/react-dom/cjs/react-dom-server.node.development.js:3704:29)
    at renderToString (/Users/acornellier/git/Leaflet/node_modules/react-dom/cjs/react-dom-server.node.development.js:4349:27)
[...]

PR #2474, opened 2021-06-10
Fix bug with <Image> and SSR for configured domains

mabadir:
Closes: #2440

Use `next/image` instead of `@blitz/core/image` when exporting `Image`

## Bug Checklist


nextjs/packages/next/client/image.tsx
```diff
@@ -241,7 +241,7 @@ function removePlaceholder(
   }
 }
 
-export default function Image({
```
flybayer, line 244: You need to add `export default Image` to keep previous behavior too

flybayer (commented):
Also you can now remove the `core/image` entrypoint. To do this, remove the packages/core/image directory and the image files from packages/core/package.json

flybayer:
@mabadir I'd like to get this in the next release tomorrow if possible

mabadir:
So @flybayer, just wondering, are those test cases still relevant:
https://github.com/blitz-js/blitz/tree/canary/test/integration/export-image-default

mabadir:
Now that core/image is removed from blitz. Do we need to export next/image instead? 🤔

flybayer:
@mabadir I just pushed some fixes.

1. Yeah we still needed to export Image stuff from `blitz` for case where babel doesn't rewrite. I debugged this by running `blitz build` inside the misc integration app.
2. I created a new `integration/images` app for your new ssr test because that addition to misc caused `blitz export` to fail since it doesn't work with images. I debugged this by running `blitz export` inside misc integration app.

PR #2475, opened 2021-06-10
(newapp) move dev dependencies to devDependencies in package.json (better for Docker)

kivi:

Prior to this commit devDependencies were moved to dependencies, in favor to fix issues with builds on some cloud server providers. But this results into a very large production container. If build as Docker image just the directory of node_modules is larger than 550MB. 

There is significant reduction by moving following dependencies back to devDependencies

- @types/preview-email
- @types/react
- eslint
- husky
- lint-staged
- prettier-plugin-prisma
- prettier
- pretty-quick
- preview-email
- prisma
- typescript"

## Testing

```
git clone https://github.com/kivi/blitz-docker-example ./testBlitzDevdeps
cd .\testBlitzDevdeps
yarn
```

Build the image:
```
docker build . -t blitz-devdeps
```

check if image was build. Check size
```
docker images | grep blitz-devdeps
```

Test run and open in browser
```
docker run --name blitz-devdeps-running -p 3031:3000 blitz-devdeps
```

### DIFF
--- a/packages/generator/templates/app/package.json
+++ b/packages/generator/templates/app/package.json
@@ -25,22 +25,24 @@
   },
   "dependencies": {
     "@prisma/client": "2.x",
-    "@types/react": "17.x",
-    "@types/preview-email": "2.x",
     "blitz": "canary",
+    "react-dom": "0.0.0-experimental-0eea57724",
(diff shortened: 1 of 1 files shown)


flybayer:
Thank you!!

Issue #2481, opened 2021-06-11
Generated Edit Page Has Buggy Form Behavior


dbachrach:
### What is the problem?

The generated Edit page for a model has the following buggy behavior: If a user types into the form, clicks outside of the browser window and then clicks back into the browser window, the form will be reverted to the original content. I've tracked down why this happens. The edit page has a `react-query` query with default behavior. This means the query is re-fetched frequently (e.g. window refocus). Each time that query runs, the resulting value gets sent into the form as it's `initialValues`. The issue is that this payload object is not shallow equal after each query re-fetch. So the form thinks it gets a new `initialValues` and it wipes what the user has typed in and resets it. 

react-query has some form of query result equality checking built in to prevent this. However, the equality they perform does not handle date equality so it seems. Since the default Prisma models contain  `createdAt` and `updatedAt` date fields, react-query incorrectly thinks the response is different each time. You can actually tweak your `getProjects` blitz query to only return `id` and `name`, and the form behaves properly. Change:

```
const project = await db.project.findFirst({
    where: { id }
})
```

to:

```
const project = await db.project.findFirst({
    where: { id },
    select: { name: true, id: true },
})
```

So one option is to try to work with the react-query folks to see if this is intentional or if we can augment the default equality check (some discussion here: https://github.com/tannerlinsley/react-query/discussions/433).

The other option, which may in fact be more proper, is to simply add `{ staleTime: Infinity }` in the Edit page template when calling `useQuery`. Since the user is editing the data, it's probably correct to "freeze" the data in time when this page loads and never refetch it.

Happy to make a PR if we want to go with the `staleTime` fix.

### Paste all your error logs here:

```
n/a
```

### Paste all relevant code snippets here:

```
n/a
```

### What are detailed steps to reproduce this?

1. `blitz new newapp`
2. `cd newapp`
3. `blitz generate all project name:string`
4. `blitz dev`
5. Login so you can access localhost:3000/projects
6. Create a new project with name "Test"
7. Navigate to edit the project (http://localhost:3000/projects/1/edit)
8. Start typing in the name field to change the name to "Test123"
9. Click your mouse outside the browser window
10. Click your mouse inside the browser window
11. The text field is changed from "Test123" to "Test"

### Run `blitz -v` and paste the output here:

```
macOS Mojave | darwin-x64 | Node: v14.16.0

blitz: 0.37.0 (global)
blitz: 0.37.0 (local)

  Package manager: yarn
  System:
[...]

PR #2487, opened 2021-06-14
Fix edit page generated by `blitz generate` to not have buggy behavior

dbachrach:
Closes: #2481

The generated Edit Page now has `{ staleTime: Infinity }`, so that the query never refreshes and resets the form (which would lose user edited changes).

## Bug Checklist

This is just a change to the template, so I don't think any tests are required.


flybayer:
Thank you!!

PR #2495, opened 2021-06-17
upgrade internal typescript version

flybayer:

upgrade internal typescript version

### DIFF
--- a/nextjs/package.json
+++ b/nextjs/package.json
@@ -129,7 +129,7 @@
     "tailwindcss": "1.1.3",
     "taskr": "1.1.0",
     "tree-kill": "1.2.2",
-    "typescript": "4.1.5",
+    "typescript": "4.3.3",
     "wait-port": "0.2.2",
     "web-streams-polyfill": "2.1.1",
     "webpack-bundle-analyzer": "4.3.0",
--- a/nextjs/packages/next/package.json
+++ b/nextjs/packages/next/package.json
@@ -237,7 +237,7 @@
     "taskr": "1.1.0",
     "terser": "5.5.1",
     "text-table": "0.2.0",
-    "typescript": "4.1.5",
+    "typescript": "4.3.3",
     "unistore": "3.4.1",
     "web-vitals": "0.2.4",
     "webpack": "4.44.1",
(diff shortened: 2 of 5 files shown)


flybayer:
Houston we have a problem.

In `packages/core/src/auth/auth-types.ts` we do this trick so that user's can extend built in types:

```ts
export interface Session {
  // isAuthorize can be injected here
  // PublicData can be injected here
}

export type PublicData = "PublicData" extends keyof Session
  ? Session["PublicData"]
  : {userId: unknown}
```

This worked great until this attempt to upgrade 4.1 to 4.3. Now we get this build error:

```
packages/core/src/auth/auth-types.ts:18:13 - error TS2538: Type '"PublicData"' cannot be used as an index type.

18   ? Session["PublicData"]
```

Note: I've tested TS 4.3 in a normal blitz app and it works fine. The only issue is this type error in our internal typecheck step.

flybayer:
I'm also getting this new error too:

```ts
packages/core/src/server/resolver.ts:295:50 - error TS2589: Type instantiation is excessively deep and possibly infinite.

295   zod<Schema extends ZodSchema<any, any>, Type = zInfer<Schema>>(schema: Schema) {
```

JuanM04:
> I'm also getting this new error too:
> 
> ```ts
> packages/core/src/server/resolver.ts:295:50 - error TS2589: Type instantiation is excessively deep and possibly infinite.
> 
> 295   zod<Schema extends ZodSchema<any, any>, Type = zInfer<Schema>>(schema: Schema) {
> ```

Try upgrading Zod to v3.2 ([zod issue](https://github.com/colinhacks/zod/issues/473#issuecomment-860144275))

flybayer:
@JuanM04 we already have that version installed

JuanM04:
Try changing this:

```ts
import type {input as zInput, output as zOutput, ZodSchema} from "zod"

// ...

  zod<Schema extends ZodSchema<any, any>, InputType = zInput<Schema>, OutputType = zOutput<Schema>>(schema: Schema) {
    return (input: InputType): OutputType => schema.parse(input)
  },
```

Zod has a `z.input` and a `z.output` method because they can be different. Maybe that solves the issue

flybayer:
Superseded by https://github.com/blitz-js/blitz/pull/2669

PR #2501, opened 2021-06-17
Fix `blitz generate` /new page does not use `Routes`

dbachrach:

The `new.tsx` generated page calls `Router.push` but used a string literal for the path instead of the `Routes` object.


flybayer:
Sweet, thank you!!

PR #2503, opened 2021-06-17
[RFC] Page Generator: Use [thingId]/index.tsx instead of [thingId].tsx

dbachrach:

Currently, the page generation template outputs a directory structure like:

```
.
├── [projectId]
│   └── edit.tsx
├── [projectId].tsx
├── index.tsx
└── new.tsx
```

This PR changes it to: 
```
.
├── [projectId]
│   ├── edit.tsx
│   └── index.tsx
├── index.tsx
└── new.tsx
```

We have moved `[projectId].tsx` to `[projectId]/index.tsx`. The benefit of this is that the ShowProjectPage and the EditProjectPage which are inherently linked now live in the same directory. I think this makes the structure cleaner and more obvious.


flybayer:
Thank you! We previously had a PR for this same thing that we rejected because there wasn't clear support for it: https://github.com/blitz-js/blitz/pull/1243

The main argument is that it's best to have as few files named `index.ts` as possible since it's so ambiguous.

That said, we can consider adding a config item that switches this behavior for people who care about it.

MrLeebo:
Closing this because there is not a consensus that it is an improvement over the existing style.

PR #2504, opened 2021-06-17
Fix `blitz generate` to regenerate the Routes manifest

dbachrach:

`blitz generate` can create new page files. Right now, the user has to run `blitz codegen` manually after a `blitz generate` to get the `Routes` data updated. This change runs the codegen phase automatically after generate.

### DIFF
--- a/packages/cli/src/commands/generate.ts
+++ b/packages/cli/src/commands/generate.ts
@@ -208,14 +208,6 @@ export class Generate extends Command {
     }
   }
 
-  async codegen() {
-    const {generate} = await import("@blitzjs/server")
-    await generate({
-      rootFolder: process.cwd(),
-      env: "dev",
-    })
-  }
-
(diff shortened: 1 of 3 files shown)


flybayer:
Sweet, thank you!!

flybayer:
Actually, I think we should change this to only run the generate step if the pages generator has ran. Because we don't need it to run for `blitz generate model`, `blitz generate crud`, etc.

dbachrach:
That's fair. I'll make the `codegen` dependant on which generators actually ran. I wasn't sure if longterm there will be more things that run during codegen beyond just `Routes` generation.

dbachrach:
This is now implemented in `PageGenerator.postWrite()`, so it is only executed if a page generation occurred.

flybayer:
Yay, thanks!!

PR #2512, opened 2021-06-19
Fix prefetched dehydrated state to support Dates, Maps, etc (patch)

Skn0tt:
Closes: ?

`dehydratedState` will now be passed through SuperJSON, adding support for `Date`, `Map`, `Set`, ...

## Bug Checklist


nextjs/run-tests.js
```diff
@@ -65,7 +65,9 @@ async function main() {
   }
 
   console.log('Running tests with concurrency:', concurrency)
-  let tests = process.argv.filter((arg) => arg.endsWith('.test.js'))
+  let tests = process.argv.filter(
+    (arg) => arg.endsWith('.test.js') || arg.endsWith('.test.ts')
```
Skn0tt, line 69: @flybayer could this become a problem when merging Next.js upstream? Should I contribute this to Next.js instead?

flybayer: this is fine

flybayer:
@Skn0tt we need to update docs for this change?

Skn0tt:
Don't think so, the docs don't mention any restrictions in terms of data types.

Skn0tt:
The CI failure seems to be unrelated.

flybayer:
@Skn0tt I just pushed a commit that cleans up your prefetch test a bit and adds /prefetching to the test precompile list.

But it still fails the test. Running it directly outside the test fails too

Skn0tt:
8a17785 fixed it.

Issue #2493, opened 2021-06-16
trying to add a prisma middleware throws error in the browser


tafelito:
### What is the problem?

when using the prisma client to add a prisma middleware in `db/index.ts` throws error `Uncaught TypeError: prisma.$use is not a function` in the browser console.

A workaround now is to check whether is running in the browser or not

### Paste all relevant code snippets here:

db/index.ts
```
const isBrowser = typeof window !== "undefined"

if (!isBrowser) {
  prisma.$use(async (params, next) => {
    ...
    return next(params)
  })
}
```

flybayer:
Thanks for reporting this!

This is an easy fix. We need to change this line: https://github.com/blitz-js/blitz/blob/canary/packages/core/src/prisma-utils.ts#L25

to return `{$use:()=>{}}` instead of `{}`

And while we are it we should check the prisma API and see if we need to add anything else there.

benjakugler96:
Hi @flybayer @tafelito . I've just created a pr as per above comment suggestion. I started using blitz for a new project so I'll be happy to help whenever I can

PR #2514, opened 2021-06-19
Fix browser error when using `prisma.$use()`

benjakugler96:
Closes: #2493 

Return object with $use method if in the browser


flybayer:
Sweeet, thank you!!

Issue #2448, opened 2021-06-04
antiCSRF cookie gets set to "undefined" when using $setPublicData inside of getServerSidePorps 


lksnmnn:
### What is the problem?

When updating a user's session data inside of `getServerSideProps` using `session.$setPublicData` somehow the antiCSRF cookie value is being set to `"undefined"` resulting in a CSRF error which can only be resolved by deleting the session cookie and log in again.

Reproduction repository: https://github.com/lksnmnn/blitz-gssp-csrf/commit/39ccd2a379eb2893e5362e9e0fb52ecdac421689

```tsx
export const getServerSideProps = async ({ req, res }) => {
  const session = await getSession(req, res)
  session.$setPublicData({ role: "ADMIN" })

  return {
    props: {},
  }
}
``` 

### Paste all your error logs here:

```
[Fast Refresh] rebuilding hot-dev-client.js:196:6
CSRFTokenMismatchError: 
query.js:321:19
    onError query.js:321
    reject retryer.js:62
    run retryer.js:126
Uncaught CSRFTokenMismatchError: 
constants-d9c5e473.esm.js:115
    construct self-hosted:2716
    _construct constants-d9c5e473.esm.js:115
    Wrapper constants-d9c5e473.esm.js:135
    CSRFTokenMismatchError constants-d9c5e473.esm.js:201
    promise blitzjs-core.esm.js:862
The above error occurred in the <UserInfo> component:

UserInfo@webpack-internal:///./pages/index.tsx:42:105
Suspense
div
main
div
Home
WithSuperJSON@webpack-internal:///../../node_modules/babel-plugin-superjson-next/dist/tools.js:110:26
BlitzInnerRoot@webpack-internal:///../../node_modules/@blitzjs/core/dist/blitzjs-core.esm.js:697:19
Layout@webpack-internal:///./app/core/layouts/Layout.tsx:13:15
ErrorBoundary@webpack-internal:///../../node_modules/react-error-boundary/dist/react-error-boundary.umd.js:71:37
Suspense
App@webpack-internal:///./pages/_app.tsx:25:19
Hydrate@webpack-internal:///../../node_modules/react-query/es/hydration/react.js:26:18
QueryClientProvider@webpack-internal:///../../node_modules/react-query/es/react/QueryClientProvider.js:37:16
BlitzProvider@webpack-internal:///../../node_modules/@blitzjs/core/dist/blitzjs-core.esm.js:476:16
[...]

Issue #2096, opened 2021-03-09
CSRFTokenMismatchError with passportAuth


hirviid:
### What is the problem?

The user gets a CSRFTokenMismatchError when starting (but not finishing) a login flow with oauth2 (using passportAuth).
I also reproduced this in the official Blitz passport example. 

### Paste all your error logs here:

```
401 | CSRFTokenMismatchError
```

### Paste all relevant code snippets here:

My [auth].ts configuration

```typescript
// app/api/auth/[...auth].ts

import { passportAuth } from "blitz"
import { Strategy as OAuth2Strategy } from "passport-oauth2"

export default passportAuth({
  successRedirectUrl: "/",
  errorRedirectUrl: "/",
  // Tried with this option set to true / false, no difference there
  // secureProxy: true,
  strategies: [
    {
      strategy: new OAuth2Strategy(
        {
          authorizationURL: process.env.OAUTH2_AUTH_URL,
          tokenURL: process.env.OAUTH2_TOKEN_URL,
          clientID: process.env.OAUTH2_CLIENT,
          clientSecret: process.env.OAUTH2_SECRET,
          callbackURL: process.env.OAUTH2_CALLBACK_URL,
          scope: ["userinfo", "fulluserinfo", "groupinfo", "sendmessage"],
        },
        async function (accessToken, refreshToken, profile, _, done) {
          // Code below doesn't really matter, because user doesn't finish the login flow
          return done(new Error("TODO"))
        }
      ),
    },
  ],
})
```

### What are detailed steps to reproduce this?

1. User clicks "Login with oauth2"
2. User is redirected to /api/auth/oauth2?redirectUrl=...
3. Passport redirects the user to the oauth2 provider
4. User DOESN'T login, but goes back to my blitz app
5. User sees a 401 | CSRFTokenMismatchError. There seems no way to get away from this screen, only by removing cookies

### Run `blitz -v` and paste the output here:

```
macOS Mojave | darwin-x64 | Node: v12.14.1

blitz: 0.31.1 (global)
blitz: 0.31.1 (local)

  Package manager: yarn 
  System:
    OS: macOS Mojave 10.14.6
    CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
    Memory: 417.69 MB / 16.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 12.14.1 - ~/.nvm/versions/node/v12.14.1/bin/node
[...]

Issue #2534, opened 2021-06-27
intermittent CSRFTokenMismatchError


remjx:
### What is the problem?

My app is throwing CSRFTokenMismatchError during queries.

I haven't been able to narrow down why this is happening:
- One time I found that the `myAppCookiePrefix_sAntiCsrfToken` was set to "undefined". Clearing the cookies and refreshing fixed the problem
- Sometimes after waiting a few minutes and refreshing it fixes itself
- Opening the page in an incognito window does not fix the problem

So far I've seen this happen on Chrome for Android and Firefox on Windows.

### Run `blitz -v` and paste the output here:

My app is on v0.38.0.

### Environment
```
Vercel

  npmPackages:
    @prisma/client: 2.24.1 => 2.24.1
    blitz: 0.38.0 => 0.38.0
    prisma: 2.24.1 => 2.24.1
    react: ^18.0.0-alpha-c96b78e0e => 18.0.0-alpha-e6be2d531
    react-dom: ^18.0.0-alpha-c96b78e0e => 18.0.0-alpha-e6be2d531
    typescript: ^4.3.2 => 4.3.4
```

flybayer:
@remjx could this be the same issue as https://github.com/blitz-js/blitz/issues/2448?

remjx:
@flybayer I did find a situation where my sAntiCsrfToken was undefined, but I am not using $setPublicData at all, and furthermore I have experienced this issue even when sAntiCsrfToken is defined.

flybayer:
@remjx do you notice it happens more on certain pages than others? And if so, what's unique about those pages? Using many queries? Or using the nextjs data hooks? Or API routes?

flybayer:
So the core bug fixed by #2448 affected any call to `getSessionKernel` for anonymous users. So it's still very possible that is the same bug. I'll leave this open until I get that published and you can test it.

remjx:
The only similarity I've been able to identify is that all the problematic pages use the getCurrentUser query.

I'm following that issue now so I'll be sure to test it, although I haven't found a way to replicate. It's just been happening intermittently.

flybayer:
That fix is released: https://github.com/blitz-js/blitz/releases/tag/v0.38.3

remjx:
@flybayer after upgrading to 0.38.4 I am still encountering this issue. Here are the error logs from Vercel. You'll notice it is logging `Internal error: antiCSRFToken is being set to undefined`

```
[...]

PR #2541, opened 2021-07-03
Fix CSRF error when using `session.$setPublicData`

flybayer:
Closes: https://github.com/blitz-js/blitz/issues/2448
Closes: https://github.com/blitz-js/blitz/issues/2096
Related: https://github.com/blitz-js/blitz/issues/2534

Fix CSRF error when using `session.$setPublicData`. Issue was that the csrf cookie would get set to `undefined`

## Bug Checklist


lksnmnn:
This looks great, I'll try to test this fix with our app to see if this also fixes (our) issue when setting public data.

Issue #2527, opened 2021-06-23
redirectAuthenticatedTo for authenticated users redirects but doesn't call queries


Swapnull:
### What is the problem?

When sending users to my app, I always send them to `/auth`, which contains 
```ts
LoginPage.redirectAuthenticatedTo = Routes.DashboardPage().pathname;
```
When they get there and the user is logged in, it redirects to `/dashboard` as it should, but the dashboard components never actually calls any of its queries, so the user gets stuck in the suspense fallback component. 

### Paste all your error logs here:

```
PASTE_HERE (leave the ``` marks)
```

### Paste all relevant code snippets here:

```tsx
// Login Component
import AuthForm from "app/auth/components/Form";
import { BlitzPage, useRouter } from "blitz";
import React from "react";

const LoginPage: BlitzPage = () => {
  const router = useRouter();

  return (
    <AuthForm
      onSuccess={() => {
        const next = (router.query.next as string) ?? "/";
        router.push(next);
      }}
    />
  );
};

LoginPage.redirectAuthenticatedTo = "/dashboard";

export default LoginPage;
```

### What are detailed steps to reproduce this?

1. Use code like above (taken from `auth` example) to set up redirectAuthenticatedTo 
2. Ensure page redirecting to is a page which includes a `useQuery` and contains a suspense fallback with something like `<div>Loading...</div>`. 
3. Visit `/auth` as a logged in user 

You will see the fallback component and the page won't load unless refreshed, or another action taken. 

### Run `blitz -v` and paste the output here:

```
macOS Big Sur | darwin-x64 | Node: v15.5.1

blitz: 0.30.5 (global)
blitz: 0.38.1 (local)

  Package manager: yarn 
  System:
    OS: macOS 11.2.2
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 1.48 GB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 15.5.1 - /usr/local/bin/node
    Yarn: 1.22.10 - /usr/local/bin/yarn
    npm: 7.3.0 - /usr/local/bin/npm
    Watchman: Not Found
  npmPackages:
    @prisma/client: ^2.24.0 => 2.24.0 
    blitz: 0.38.1 => 0.38.1 
    prisma: ^2.24.0 => 2.24.0 
    react: 0.0.0-experimental-3310209d0 => 0.0.0-experimental-3310209d0 
[...]

PR #2548, opened 2021-07-05
(newapp) Remove `<Suspense>` from _app.js because it causes buggy redirect behavior

flybayer:
Related: https://github.com/blitz-js/blitz/issues/2527, https://github.com/blitz-js/blitz/discussions/2533

Remove `<Suspense>` from _app.js because it causes buggy redirect behavior


flybayer:
I've opened a React issue here: https://github.com/facebook/react/issues/21802

remjx:
@flybayer should we avoid using <Suspense> at all in our _app.js? Simply moving it inside a component fixes it?

flybayer:
@remjx it does work if you have Suspense inside the ErrorBoundary, but test the UX for redirects to see if it meets your expectations.

Issue #2557, opened 2021-07-07
Found a change in blitz.config.js


medihack:
### What is the problem?

Since v0.38.3 a new `.blitz.config.compiled.js` is generated in the project root folder. Also, this message is printed during startup of the dev server: `Found a change in blitz.config.js. Restart the server to see the changes in effect.`. This message is printed every time the server is started. Also, the compiled file sometimes changes a bit its structure (even if `.blitz.config.js` stays the same).

### Paste all your error logs here:

```
Found a change in blitz.config.js. Restart the server to see the changes in effect.
```

### Paste all relevant code snippets here:

No extra code necessary.

### What are detailed steps to reproduce this?

Simply create a new Blitz project with v0.38.3 and start the dev server with `yarn dev`.

### Run `blitz -v` and paste the output here:

```
$ /workspace/radreporter/node_modules/.bin/blitz -v
Linux 5.4 | linux-x64 | Node: v14.17.0

blitz: 0.38.3 (local)

  Package manager: yarn 
  System:
    OS: Linux 5.4 Ubuntu 20.04.2 LTS (Focal Fossa)
    CPU: (16) x64 Intel(R) Xeon(R) CPU
    Memory: 8.94 GB / 62.81 GB
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 14.17.0 - /tmp/yarn--1625665781789-0.10486586535605058/node
    Yarn: 1.22.10 - /tmp/yarn--1625665781789-0.10486586535605058/yarn
    npm: 6.14.13 - ~/.nvm/versions/node/v14.17.0/bin/npm
    Watchman: Not Found
  npmPackages:
    @prisma/client: 2.25.0 => 2.25.0 
    blitz: 0.38.3 => 0.38.3 
    prisma: 2.26.0 => 2.26.0 
    react: alpha => 18.0.0-alpha-73ffce1b6-20210624 
    react-dom: alpha => 18.0.0-alpha-73ffce1b6-20210624 
    typescript: ~4.3 => 4.3.4
```

### Please include below any other applicable logs and screenshots that show your problem:

-

flybayer:
Thanks for the report!

PR #2559, opened 2021-07-07
Fix false "Found a change in blitz.config.js" message

flybayer:
Closes: https://github.com/blitz-js/blitz/issues/2557

Fix false "Found a change in blitz.config.js" message

### DIFF
--- a/nextjs/packages/next/bin/next.ts
+++ b/nextjs/packages/next/bin/next.ts
@@ -118,13 +118,17 @@ commands[command]()
   })
 
 if (command === 'dev') {
-  const { CONFIG_FILE } = require('../next-server/lib/constants')
-  const { watchFile } = require('fs')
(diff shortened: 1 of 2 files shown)


nextjs/packages/next/bin/next.ts
```diff
@@ -118,17 +118,19 @@ commands[command]()
...
+      )
+    }
+  })
+  watchFile(`${process.cwd()}/blitz.config.ts`, (cur: any, prev: any) => {
+    if (cur.size > 0 || prev.size > 0) {
+      console.log(
+        `\n> Found a change in blitz.config.ts. Restart the server to see the changes in effect.`
+      )
+    }
+  })
```
webdeb, line 135: Why are there two identical `watchFile` expressions?

webdeb: oh, just saw ts & js

PR #2573, opened 2021-07-14
Replace keyof check

beerose:
I replaced `"X" extends of keyof Y` with `Y extends of {X: unknown}` to solve `Type "X" cannot be used as an index type.`. Why? Turns out TypeScript 4.3 doesn't narrow the types of an object after a keyof check.

Closes: —

## Bug Checklist

## Feature Checklist


flybayer:
Thanks!!

PR #2593, opened 2021-07-23
(newapp) Fix image warning in index page

bravo-kernel:

This PR replace `` to solve the following warning message:

```
Do not use <img>. Use Image from 'next/image' instead. See https://nextjs.org/docs/messages/no-img-element.
``


flybayer:
Thank you @bravo-kernel!!

PR #2597, opened 2021-07-25
Fix duplicate image types (requires adding line to global.d.ts)

sam3d:
This PR attempts to fix the duplicate image types present in `next/types/global` that have since been stripped out and migrated to `next/image-types/global`. This, as a byproduct, uses the correct new svg typings in `image-types` that uses an `any` type instead of `StaticImageData`.

```ts
/// <reference types="next" />
/// <reference types="next/types/global" />
/// <reference types="next/image-types/global" />
```

Correct `next/types/global`: https://github.com/vercel/next.js/blob/canary/packages/next/types/global.d.ts

Correct `next/image-types/global`: https://github.com/vercel/next.js/blob/canary/packages/next/image-types/global.d.ts


flybayer:
Sweet, thank you @sam3d!!

PR #2343, opened 2021-05-14
Fix `Routes.X` is not a function when inside a monorepo 

Skn0tt:
Closes: #2272

### What are the changes and their implications?
Fixes #2272 by detecting the location of `node_modules/blitz` and emitting besides it.

## Bug Checklist

- [ ] Integration test added (see [test docs](https://blitzjs.com/docs/contributing#running-tests) if needed)

## Feature Checklist

- [ ] Integration test added (see [test docs](https://blitzjs.com/docs/contributing#running-tests) if needed)
- [ ] Documentation added/updated (submit PR to [blitzjs.com repo](https://github.com/blitz-js/blitzjs.com))

Issue #2272, opened 2021-04-25
"Routes.SignupPage is not a function" when using monorepo


niklasgrewe:
### What is the problem?

When i try run my blitz app inside my yarn workspace, i get the following webpack error:

```
_blitzjs_core__WEBPACK_IMPORTED_MODULE_6__.Routes.SignupPage is not a function
```

### Paste all your error logs here:

```
_blitzjs_core__WEBPACK_IMPORTED_MODULE_6__.Routes.SignupPage is not a function

pages/index.tsx (37:27) @ UserInfo

  35 | return (
  36 |   <>
> 37 |     <Link href={Routes.SignupPage()}>
     |                       ^
  38 |       <a className="button small">
  39 |         <strong>Sign Up</strong>
  40 |       </a>
```

### Paste all relevant code snippets here:

my root `package.json` looks like this:

```
{
  "name": "blitz-monorepo",
  "private": "true",
  "version": "1.0.0",
  "workspaces": {
    "packages": [
      "packages/*"
    ]
  },
  "scripts": {
    "web": "yarn workspace web"
  }
}
```

### What are detailed steps to reproduce this?

1. create a new folder like `blitz-monorepo`
2. create a subfolder `packages`
3. install a new blitz app inside `packages`
4. remove `node_modules` folder from blitz app
5. create a root `package.json` like above
6. run `yarn install && yarn workspace <blitzAppName> dev`

### Run `blitz -v` and paste the output here:

```
macOS Big Sur | darwin-arm64 | Node: v15.14.0

blitz: 0.34.3 (global)
blitz: 0.34.3 (local)

  Package manager: yarn 
  System:
    OS: macOS 11.2.3
    CPU: (8) arm64 Apple M1
    Memory: 1.05 GB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 15.14.0 - /opt/homebrew/bin/node
    Yarn: 1.22.10 - /opt/homebrew/bin/yarn
    npm: 7.7.6 - /opt/homebrew/bin/npm
    Watchman: Not Found
  npmPackages:
    @prisma/client: Not Found
    blitz: Not Found
    prisma: Not Found
    react: Not Found
    react-dom: Not Found
    typescript: Not Found
```

Skn0tt:
@niklasgrewe thanks for reporting!

It seems to be caused by: 
[...]

PR #2599, opened 2021-07-26
Fix a bug when just 1 level deep inside yarn workspace

kenkuan:
Origin fix https://github.com/blitz-js/blitz/pull/2343 can't handle yarn workspace with 1 level structure properly.

It will return the parent folder of the project root and result in https://github.com/blitz-js/blitz/issues/2272 .

This PR looks up `node_modules` in each level of parent folders and returns the `node_modules` folder with `.blitz` correctly. 

The sample root `package.json`
```
{
  "name": "OneLevelMonoRepo",
  "version": "0.0.1",
  "main": "index.js",
  "private": true,
  "workspaces": [
    "contract",
    "web"
  ]
}
```

## Bug Checklist

## Feature Checklist


packages/server/src/stages/route-import-manifest/route-import-manifest.ts
```diff
@@ -149,9 +149,13 @@ export function parseParametersFromRoute(
  * We're not fooled by you, `yarn workspace`!
  */
 function findNodeModulesRoot(src: string) {
-  const nodeModules = join(src, "node_modules")
-  const includesBlitzPackage = fs.existsSync(join(nodeModules, "blitz"))
-  return includesBlitzPackage ? nodeModules : join(nodeModules, "../../../node_modules")
+  let nodeModules = join(src, "node_modules")
+  let includesBlitzPackage = fs.existsSync(join(nodeModules, "blitz"))
+  while (!includesBlitzPackage) {
+    nodeModules = join(nodeModules, "../../node_modules")
```
flybayer, line 155: Thanks @kenkuan! (And sorry for the delay!!)

Shouldn't this line only go up one directory?
```suggestion
    nodeModules = join(nodeModules, "../node_modules")
```

Also we should have a limit on this while loop to prevent an infinite loop.

kenkuan: Hi @flybayer ,

I added a simple check 47b09db to prevent the infinite loop. The join is two-level up because the root is `xxx/node_modules`, so `../node_modules` will go back to the same `node_modules`.

flybayer: ah yes, thanks!

packages/server/src/stages/route-import-manifest/route-import-manifest.ts
```diff
@@ -149,9 +149,18 @@ export function parseParametersFromRoute(
...
-  return includesBlitzPackage ? nodeModules : join(nodeModules, "../../../node_modules")
+  let nodeModules = join(src, "node_modules")
+  let includesBlitzPackage = fs.existsSync(join(nodeModules, "blitz"))
+  let count = 0
+  while (!includesBlitzPackage) {
+    nodeModules = join(nodeModules, "../../node_modules")
+    includesBlitzPackage = fs.existsSync(join(nodeModules, "blitz"))
+    count++
+    if (count > 5) {
+      throw new Error("Can't find node_modules of blitz! Check report an issue. Thanks!")
```
flybayer, line 160:
```suggestion
      throw new Error("Oops. We can't find the blitz node_modules package. Are you sure it's installed?")
```

packages/server/src/stages/route-import-manifest/route-import-manifest.ts
```diff
@@ -149,9 +149,18 @@ export function parseParametersFromRoute(
  * We're not fooled by you, `yarn workspace`!
  */
 function findNodeModulesRoot(src: string) {
-  const nodeModules = join(src, "node_modules")
-  const includesBlitzPackage = fs.existsSync(join(nodeModules, "blitz"))
-  return includesBlitzPackage ? nodeModules : join(nodeModules, "../../../node_modules")
+  let nodeModules = join(src, "node_modules")
+  let includesBlitzPackage = fs.existsSync(join(nodeModules, "blitz"))
+  let count = 0
+  while (!includesBlitzPackage) {
```
flybayer, line 155:
```suggestion
  while (!includesBlitzPackage) {
    // Check for node_modules at the next level up
```

packages/server/src/stages/route-import-manifest/route-import-manifest.ts
```diff
@@ -149,9 +149,19 @@ export function parseParametersFromRoute(
...
+    // Check for node_modules at the next level up
+    nodeModules = join(nodeModules, "../../node_modules")
+    includesBlitzPackage = fs.existsSync(join(nodeModules, "blitz"))
+    count++
+    if (count > 5) {
+      throw new Error("Oops. We can't find the blitz node_modules package. Are you sure it's installed?")
```
flybayer, line 161: @kenkuan So the test in `packages/server` are failing (run `yarn test` inside `packages/server` to see). I think instead of throwing an error here, we should log a warning and default to `<src>/node_modules`. This will make tests pass and have a fallback like before.

kenkuan: @flybayer Cool. I made the changes in 65880b1.

packages/server/src/stages/route-import-manifest/route-import-manifest.ts
```diff
@@ -158,7 +159,11 @@ function findNodeModulesRoot(src: string) {
     includesBlitzPackage = fs.existsSync(join(nodeModules, "blitz"))
     count++
     if (count > 5) {
-      throw new Error("Oops. We can't find the blitz node_modules package. Are you sure it's installed?")
+      log.warning(
+        "Oops. We can't find the blitz node_modules package. Are you sure it's installed?",
```
flybayer, line 163:
```suggestion
        "We couldn't determine your actual node_modules location, so defaulting to normal location",
```

PR #2607, opened 2021-07-28
(patch) Fix bug when no dependencies are defined in app package.json

zenhob:
Closes: ?

Allow a custom server to start even without dependencies defined in the app package.json

## Bug Checklist


zenhob:
this seems like the intended behavior based on code found elsewhere like this `getConfig` stanza:
https://github.com/blitz-js/blitz/blob/f8ce548f4a119cc0b3b458f7b283e66c63742168/packages/config/src/index.ts#L95

flybayer:
Thank you!!

Issue #2578, opened 2021-07-18
Allow resolver.zod to parseAsync


LucasCoderT:
### What do you want and why?

`resolver.zod` only parses using the `sync` method, it'd be helpful if resolver.zod had a 2nd parameter for mode: `"async" | "sync"` to allow for parsing that requires it. Modeled after the resolver from ``@hookform/resolvers`` package.

The more information the better!

### Possible implementation(s)

How might we do this?

https://github.com/blitz-js/blitz/blob/e799b088477a693f8555bc8ee16ce4127978e019/packages/core/src/server/resolver.ts#L295-L297

```js
  zod<Schema extends ZodSchema<any, any>, Type = zInfer<Schema>>(schema: Schema, mode: "sync" | "async" = "sync") {
    return (input: Type): Type => mode === "sync" ? schema.parse(input) : schema.parseAsync(input)
  },
```

flybayer:
Thanks! I'm thinking we can actually just switch to parseAsync. Do you see any reason not to?

LucasCoderT:
I don't see any reason not to parseAsync no.

flybayer:
Ok cool.

Let's add the argument, make it optional, and default to async.

Roesh:
I was thinking of picking this up, but I want to clarify what is being targeted here. Based on @LucasCoderT Lucas' comment, it looks like we will want to do something like 
`export default resolver.pipe(resolver.zod(GetEvent, "sync")` and if that argument isn't specified, we do an async parse?

From the file @flybayer linked, I see an additional function that is also commonly used, and I assume this should also be updated?
```
syncTypes = "sync" | "async"
export const validateZodSchema = (schema: any, syncType: syncTypes ) => async (values: any): any => {
 <based on synctype do the correct parse>
```

Do we want to do a performance test on this though? The only time I've heard that async may be worse than syncs are if the work done is quick, where it might take longer to spawn the threads to do the work than it is to start working on the problem right away. That's from a .NET context though, so I don't know if that's the same here

flybayer:
@Roesh so sorry for the delay here but yes that's correct. And perf isn't an issue here since it's a quick one time thing vs being inside a for loop.

Thanks for the PR, I'll review soon!

PR #2616, opened 2021-07-31
Use parseAsync by default and add argument to parse sync to common zod validations

Roesh:
Closes: #2578

Adding an optional _parserType_ argument, defaulting to "async" to 
1) `validateZodSchema` found in `packages\core\src\utils\index.ts` and 
2) `resolver.zod` found in `packages\core\src\server\resolver.ts`. 

Added a `ParserType` type in that same resolver.ts file.

In addition, I added tests that are the same as the ones that exist right now, with the only difference being the presence of the newly added `ParserType` optional argument

I think the validateZodSchema update might be a breaking change for projects because we need to use await validateZodSchema() to resolve the promise, regardless of whether we are using the sync or async argument, since the functions were marked async to work. I think the resolver.zod function is fine within the resolver.pipe function, but if users are using it outside of resolver.pipe, this will likely be a breaking change for them as well?

I'm also wondering if it is an issue that both functions are marked async, should we have two different functions (one returns promise and on doesn't?) 
In addition, does marking the function as async make the "sync" argument seem a bit misleading?

I haven't added documentation as yet since I'm not sure this is the best way to implement this, so will wait for feedback.

## Bug Checklist

## Feature Checklist


flybayer:
@Roesh woot! Looks good!

Good point on potential breaking for some cases.

The way to change return type based on parameter values is [function overloads](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads). Give this a shot, and if you are having trouble share the code you have, and I can help. May need to ping me in discord for a timely response

Roesh:
Thanks! Im on a break till mid week next week, so this might take until then, but gonna learn and use those overloads once Im back :)))

Roesh:
Ok, I'm back! I've made the validator have a sync and async overload, and intellisense is giving me the correct return types based on the parser type argument. If its unspecified, it defaults to returning the error as a promise.

The resolver is failing because of issues with the overload not externally accessible, I'm looking into resolving that now

packages/core/src/server/resolver.ts
```diff
@@ -290,10 +290,33 @@ const authorize: ResolverAuthorize = (...args) => {
...
+  schema: Schema,
+): (input: Type) => Promise<Type>
+function zod<Schema extends ZodSchema<any, any>, Type = zInfer<Schema>>(
+  schema: Schema,
+  parserType: ParserType = "async",
+){
+  if(parserType === "sync"){
+    return (input: Type) : Type => schema.parse(input)
+  }else{
+    return async (input: Type) : Promise<Type> => schema.parseAsync(input)
```
flybayer, line 313: Don't have to mark as async since we don't use `await` inside the body
```suggestion
    return (input: Type) : Promise<Type> => schema.parseAsync(input)
```

flybayer:
@Roesh awesome, looks great! Last thing for you is to make a PR to update our docs :)

flybayer:
@Roesh you have a couple lint errors too

Roesh:
Cool! I'm almost done with the docs PR and I'm having issues testing the validatezodschema in the FinalForm template. I'm going to uninstall node_modules and try again but might take some time.
The main question I'm trying to answer is whether 

```
<FinalForm
  initialValues={initialValues}
  validate={validateZodSchema(schema)}
  ...
  ```
should actually be
```
<FinalForm
  initialValues={initialValues}
  validate={await validateZodSchema(schema)}
  ...
```
If so, we may have to update our templates as well.

In addition, I think we will need to include the `💥breaking changes` disclaimer for users who may be using validateZodSchema(schema) outside it and expecting it to run synchronously since by default were returning a promise now

Roesh:
Alright, reinstalling node_modules fixed the issue. Our form validator does work with the new async validator, which I confirmed using the auth example by typing in an invalid email address into the signup box.

I added PR https://github.com/blitz-js/blitzjs.com/pull/543 documenting these changes.

Going to stress again that we should likely add the `💥Breaking Changes` tag but specify clearly that the change only affects users if they have used that code in areas that are not the usual resolver or Form template.

flybayer:
Yay, thank you!!

Issue #2601, opened 2021-07-26
dehydrateState does not apply server side props


clgeoio:
### What is the problem?

The [Prefetching](https://blitzjs.com/docs/query-usage#prefetching) suggests that if set up correctly, adding `dehydratedState` should render the result of a query in the server-side render.

The [test case](https://github.com/blitz-js/blitz/blob/334a771c867334f76123d9c4c0b9b1f5d39c9eb2/test/integration/queries/test/index.test.ts#L79) for the dehydrated state is invalid as this line should not `waitForElementByCss`, but instead should expect that this element is present immediately as a result of the SSR.

If the above line is removed, the test fails, when running the test with the browser enabled, the Suspense fallback text can be seen as the server-side render result.

### Paste all your error logs here:

N/A

### Paste all relevant code snippets here:

```
import getDate from "app/queries/getDate"
import {
  dehydrate,
  getQueryKey,
  GetServerSideProps,
  invokeWithMiddleware,
  QueryClient,
  useQuery,
} from "blitz"
import {Suspense} from "react"

export const getServerSideProps: GetServerSideProps = async (ctx) => {
  const queryClient = new QueryClient()
  const queryKey = getQueryKey(getDate, null)
  await queryClient.prefetchQuery(queryKey, () => invokeWithMiddleware(getDate, undefined, ctx))

  return {
    props: {
      dehydratedState: dehydrate(queryClient),
    },
  }
}

function Content() {
  const [date] = useQuery(getDate, undefined)
  return <p id="content">date is Date: {"" + (date instanceof Date)}</p>
}

function Page() {
  return (
    <Suspense fallback="Loading ...">
      <Content />
    </Suspense>
  )
}

export default Page
```

### What are detailed steps to reproduce this?

N/A

### Run `blitz -v` and paste the output here:

```
macOS Big Sur | darwin-x64 | Node: v14.16.1

blitz: 0.38.5 (local)

  Package manager: yarn
  System:
    OS: macOS 11.5
    CPU: (12) x64 Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
    Memory: 442.04 MB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.16.1 - /var/folders/hl/wl16kswj4dd9jvkvz2ykm16c0000gn/T/yarn--1627293289668-0.31100445211523553/node
[...]

PR #2617, opened 2021-07-31
Fix dehydrateState does not apply server side props

clgeoio:
Closes: https://github.com/blitz-js/blitz/issues/2601

The query that was setup in the dehydrated state did not have matching params (`null` vs `undefined`). Additionally, the SuperJSON deserialisation seemed to be causing some issues with the cached query being used on the SSR.

However, this change looks to undo: https://github.com/blitz-js/blitz/pull/2512

## Bug Checklist

## Feature Checklist


packages/core/src/blitz-app-root.tsx
```diff
@@ -110,7 +109,7 @@ export function withBlitzAppRoot(UserAppRoot: React.ComponentType<any>) {
     }, [])
 
     const dehydratedState = props.pageProps.dehydratedState
-      ? SuperJSON.deserialize(props.pageProps.dehydratedState)
+      ? props.pageProps.dehydratedState
```
Skn0tt, line 112: It's interesting to me that removing the deserialisation helps in this case, since babel-plugin-superjson-next should be serializing `dehydratedState`: https://github.com/blitz-js/blitz/blob/a28216d067ff03bbd954a553275972e6f9340bc8/packages/babel-preset/src/index.ts#L16 🤔

Skn0tt: I pushed an update in 98bdcf8, which should fix the behaviour while retaining SuperJSON support. The problem basically was that `SuperJSON.deserialize` was a totally wrong call - see the 98bdcf8's commit message for more context.

clgeoio: Awesome, thanks for making that change!

I think I will add a test for some of the more complex types, like Map!

test/integration/queries/pages/dehydrated-state.tsx
```diff
@@ -11,7 +11,7 @@ import {Suspense} from "react"
 
 export const getServerSideProps: GetServerSideProps = async (ctx) => {
   const queryClient = new QueryClient()
-  const queryKey = getQueryKey(getDate, null)
+  const queryKey = getQueryKey(getDate, undefined)
```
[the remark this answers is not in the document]
clgeoio: Interesting suggestion, I like it. It took me a while to figure it out and there was a small piece in the react-query docs that made me realize there was a mismatch.

Anyway we can make it easier for users to form the right code I am in support of :)

packages/core/src/blitz-app-root.tsx
clgeoio: Interestly @Skn0tt, if I swap out `getDate` with `getComplexTypes` and return something like:
```
export default async function getComplexTypes() {
  return {
    date: new Date(0),
    map: new Map().set("foo", "bar"),
    obj: {
      one: 1,
      child: {
        two: 2,
      },
    },
  }
}
```

When `map: new Map()` is in the returned object, the test fails:
```
 NoSuchElementError: no such element: Unable to locate element: {"method":"css selector","selector":"#content"}
      (Session info: headless chrome=92.0.4515.107)

      77 |     it("should work", async () => {
      78 |       const browser = await webdriver(context.appPort, "/dehydrated-state")
    > 79 |       const content = await browser.elementByCss("#content")
```

But when the `map` key is omitted from the return object, the test passes.

Any ideas here, should this be supported?

Skn0tt: Yes, it should definitely be supported. That's some weird behaviour! Will look into it.

test/integration/queries/pages/dehydrated-state.tsx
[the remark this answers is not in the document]
flybayer: Yeah I knew from the start we needed to make this better.

How about this?

```ts
import {prefetchQuery} from 'blitz'

export const getServerSideProps = async ctx => {

  await prefetchQuery(getData, undefined)

  return { props: {} }
}
```

And then we automatically add `new QueryClient()`, `dehydrate` etc via a babel plugin.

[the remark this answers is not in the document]
clgeoio: A small suggestion would be to require that `dehydratedState` be passed into `props: {}` to avoid the edge case of a user declaring a prop `dehydratedState` and having overridden without them knowing. Perhaps we could rename the prop to `prefetchedQuery` for clarity and then transform in `blitz-root`?

```
export const getServerSideProps = async ctx => {
  const dehydratedState = await prefetchQuery(getData, undefined)
  return { props: {
    dehydratedState
  } }
}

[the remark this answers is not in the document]
Comment (flybayer) [reply test/integration/queries/pages/dehydrated-state.tsx:14]: The babel plugin can throw a helpful error or something in that case, right? I think having `dehydratedState` exposed is exposing unnecessary implementation details.

File: packages/core/src/blitz-app-root.tsx (hunk 1 above)
Comment (clgeoio) [reply packages/core/src/blitz-app-root.tsx]: Thank you @Skn0tt.
Let me know if I can lend a hand.
I'll park this PR until you've had a look 👍

Comment (mikealche) [reply packages/core/src/blitz-app-root.tsx]: Hey hey, I'm just starting to learn and poke around the blitz codebase, but I tried to debug this issue. I've found that after 98bdcf8, a possibility that the test is failing is that the query keys don't match.

The dehydrated state on `blitz-app-root`is: 
```
{ dehydratedState: 
   { mutations: [],
     queries: 
      [ { state: 
           { data: Wed Dec 31 1969 21:00:00 GMT-0300 (Argentina Standard Time),
             dataUpdateCount: 1,
             dataUpdatedAt: 1628558585243,
             error: null,
             errorUpdateCount: 0,
             errorUpdatedAt: 0,
             fetchFailureCount: 0,
             fetchMeta: null,
             isFetching: false,
             isInvalidated: false,
             isPaused: false,
             status: 'success' },
          queryKey: 
           [ '/api/queries/getDate',
             { json: null, meta: { values: [ 'undefined' ] } } ],
          queryHash: '["/api/queries/getDate",{"json":null,"meta":{"values":["undefined"]}}]' } ] } }
```

while if you log `getQueryKey(getDate, null)` in the client,you get: 
```
{ queryKey: [ '/api/queries/getDate', { json: null } ] }
```


If you look closely you'll see that the latter code is missing the `meta` part of the key: 
```
queryKey: 
           [ '/api/queries/getDate',
             { json: null, meta: { values: [ 'undefined' ] } } ]

vs

{ queryKey: [ '/api/queries/getDate', { json: null } ] }
```

I'm not sure what should be done to fix this, but I hope this helps and also I'm available if you need any helping hand @Skn0tt

flybayer:
@Skn0tt what's the status on this PR?

Skn0tt:
> @Skn0tt what's the status on this PR?

looking at it now, let's hope I find something :)

PR #2634, opened 2021-08-05
Add support for `Page.redirectAuthenticatedTo` to be a function with access to `session`

### LINKED ISSUE #2575: Add session info (or maybe context) to the redirectAuthenticatedTo
### What do you want and why?

Based on the conversation we had over discord, it would be nice to have a a way to pull the session data when using `redirectAuthenticatedTo` to make it easier to redirect to different places depending on the user's session 

After implementing this, redirectAuthenticatedTo would look like something like this `Page.redirectAuthenticatedTo = (session) => session.onboarded ? '/dashboard' : '/onboarding'`

### Additional context

This can prevent duplicating getServerSideProps logic just for this use case

s-r-x:
Hello ^-^
This PR closes #2575
redirectAuthenticatedTo can now be also a function. It accepts a context, which includes
```typescript
type RedirectAuthenticatedToFnCtx = {
  publicData: EmptyPublicData
  session: ClientSession
}
```
i added a publicData because in my use case i need access to some user metadata on the first render, when session is empty.

## Bug Checklist

## Feature Checklist

### DIFF
--- a/packages/core/src/blitz-app-root.tsx
+++ b/packages/core/src/blitz-app-root.tsx
@@ -46,17 +46,21 @@ const NoPageFlicker = () => {
 export function withBlitzInnerWrapper(Page: BlitzPage) {
   const BlitzInnerRoot = (props: ComponentPropsWithoutRef<BlitzPage>) => {
     // We call useSession so this will rerender anytime session changes
-    useSession({suspense: false})
+    const session = useSession({suspense: false})
 
     useAuthorizeIf(Page.authenticate === true)
 
     if (typeof window !== "undefined") {
+      const publicData = publicDataStore.getData()
       // We read directly from publicDataStore.getData().userId instead of useSession
       // so we can access userId on first render. useSession is always empty on first render
-      if (publicDataStore.getData().userId) {
+      if (publicData.userId) {
         clientDebug("[BlitzInnerRoot] logged in")
         let {redirectAuthenticatedTo} = Page
         if (redirectAuthenticatedTo) {
+          if (typeof redirectAuthenticatedTo === "function") {
+            redirectAuthenticatedTo = redirectAuthenticatedTo({publicData, session})
+          }
           if (typeof redirectAuthenticatedTo !== "string") {
             redirectAuthenticatedTo = formatWithValidation(redirectAuthenticatedTo)
           }
--- a/packages/core/src/blitz-app-root.tsx
+++ b/packages/core/src/blitz-app-root.tsx
@@ -56,17 +56,17 @@ export function withBlitzInnerWrapper(Page: BlitzPage) {
       // so we can access userId on first render. useSession is always empty on first render
       if (publicData.userId) {
         clientDebug("[BlitzInnerRoot] logged in")
-        let {redirectAuthenticatedTo} = Page
-        if (typeof redirectAuthenticatedTo === "function") {
-          redirectAuthenticatedTo = redirectAuthenticatedTo({session: publicData})
-          // redirectAuthenticatedTo is string/RouteUrlObject/false now
-        }
-        if (redirectAuthenticatedTo && typeof redirectAuthenticatedTo !== "string") {
-          redirectAuthenticatedTo = formatWithValidation(redirectAuthenticatedTo)
-        }
+        const redirectAuthenticatedTo =
+          typeof Page.redirectAuthenticatedTo === "function"
+            ? Page.redirectAuthenticatedTo({session: publicData})
+            : Page.redirectAuthenticatedTo
         if (redirectAuthenticatedTo) {
-          clientDebug("[BlitzInnerRoot] redirecting to", redirectAuthenticatedTo)
-          const error = new RedirectError(redirectAuthenticatedTo)
+          const redirectUrl =
+            typeof redirectAuthenticatedTo === "string"
+              ? redirectAuthenticatedTo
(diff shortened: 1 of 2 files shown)


packages/core/src/blitz-app-root.tsx
```diff
@@ -46,17 +46,21 @@ const NoPageFlicker = () => {
...
       // so we can access userId on first render. useSession is always empty on first render
-      if (publicDataStore.getData().userId) {
+      if (publicData.userId) {
         clientDebug("[BlitzInnerRoot] logged in")
         let {redirectAuthenticatedTo} = Page
         if (redirectAuthenticatedTo) {
+          if (typeof redirectAuthenticatedTo === "function") {
+            redirectAuthenticatedTo = redirectAuthenticatedTo({publicData, session})
+          }
           if (typeof redirectAuthenticatedTo !== "string") {
```
flybayer, line 64:
```suggestion
          } else if (typeof redirectAuthenticatedTo !== "string") {
```

flybayer:
Awesome, thank you!

> i added a publicData because in my use case i need access to some user metadata on the first render, when session is empty.

We can't do this, because it will cause React hydration errors. This is the reason useSession is empty on first render. This is a fundamental law of React that we have to live with. The solution is use `Page.suppressFirstRenderFlicker = true` to hide the first render.

So remove publicData, and then the last thing is opening a PR to the docs repo to update the docs for this :)

s-r-x:
@flybayer done 
[Docs PR](https://github.com/blitz-js/blitzjs.com/pull/534)

s-r-x:
@flybayer here is one more addition. redirectAuthenticatedTo now gets called only when session is not isLoading. i think this is a better option instead of letting users to do some extra checks in their code:
```typescript
if(!session.isLoading) {
  return session.role === 'admin' ? '/admin' : '/';
}
```

flybayer:
@s-r-x ah.. I was mistaken — I'm sorry! In this case, since we are redirecting, we don't have to wait for second render.

So we should do like you first had it, except pass `publicDataStore.getData()` as `session` instead of both that and `useSession` result. Like this:

```ts
redirectAuthenticateTo({session: publicDataStore.getData()})
```

And the TS type should be `session: PublicData`. Because in this case, `isLoading` is not needed and doesn't have any effect.

Make sense?

s-r-x:
@flybayer yep

s-r-x:
@flybayer done.
i also added false as a possible redirectAuthenticatedTo value
possible use case for abstract admin route
```typescript
Page.redirectAuthenticatedTo = ({ session }) => {
  if(session.role === 'admin') return false;
  return '/'
}
```

packages/core/src/blitz-app-root.tsx
```diff
@@ -51,16 +51,19 @@ export function withBlitzInnerWrapper(Page: BlitzPage) {
...
+      // We read directly from publicData.userId instead of useSession
       // so we can access userId on first render. useSession is always empty on first render
-      if (publicDataStore.getData().userId) {
+      if (publicData.userId) {
         clientDebug("[BlitzInnerRoot] logged in")
         let {redirectAuthenticatedTo} = Page
+        if (typeof redirectAuthenticatedTo === "function") {
+          redirectAuthenticatedTo = redirectAuthenticatedTo({session: publicData})
+        }
+        if (redirectAuthenticatedTo && typeof redirectAuthenticatedTo !== "string") {
```
flybayer, line 63: Should this be an `else if`?

s-r-x: dont't think so. the second if block would never fire(in case if it's "else if") if redirectAuthenticatedTo is a function. but one of the possible return values from this function is an object, which one we should transform to a string inside the second if block on line 63. correct me if i wrong.

erikshestopal: @s-r-x I think @flybayer is correct because `redirectAuthenticatedTo` will be truthy since it's a function and `typeof redirectAuthenticatedTo !== "string"` will also be `true` because `"function" !== "string"`. 

Probably adding a test case would surface this? 

```ts
if (typeof redirectAuthenticatedTo === "function") {
  redirectAuthenticatedTo = redirectAuthenticatedTo({session: publicData})
  
}

// `redirectAuthenticatedTo` will be truthy since it's a function
//  and "function" !== "string" will equal true, so the second `if` block will run. 

if (redirectAuthenticatedTo && typeof redirectAuthenticatedTo !== "string") {
  redirectAuthenticatedTo = formatWithValidation(redirectAuthenticatedTo)
}
```

flybayer: Technically the code will work fine as is, but it's better to be an `else if` from a code readability standpoint. Easier to instantly see that only one of these branches will run.

s-r-x: @erikshestopal,
@flybayer

https://user-images.githubusercontent.com/41614937/129136186-67c72eaf-30ca-4804-b706-a35227391d53.mp4

this is still fine because RedirectError accepts not only a string, but also RouteUrlObject(as far as i understand. at least in video it looks like it does) .
but the second if block is never fire in this case, and tbh i'm not quite sure how important that formatWithValidation call is.

s-r-x: @flybayer i simplified the code, make everything immutable

flybayer:
Awesome, thank you!!

Issue #2613, opened 2021-07-30
npx blitz new prints an error but resumes creating app


Talor-A:
### What is the problem?

running `npx blitz new` (no global `blitz` installed, so fetching the latest version 

### Paste all your error logs here:

```
$ npx blitz new the-magic-scheduler 
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received null
    at validateString (internal/validators.js:120:11)
    at Object.join (path.js:1039:7)
    at main (/Users/ta/.npm/_npx/15781/lib/node_modules/blitz/scripts/postinstall.js:84:43) {
  code: 'ERR_INVALID_ARG_TYPE'
}
? Pick a form library (you can switch to something else later if you want) … 
❯ React Final Form (recommended)
  React Hook Form
  Formik





```

### Paste all relevant code snippets here:

N/A

### What are detailed steps to reproduce this?

1. `npx blitz new myapp` with no global `blitz` installed
2. error is printed
3. project is initialized fine

### Run `blitz -v` and paste the output here:

```
$ npx blitz version

@blitzjs/cli/0.38.6 darwin-x64 node-v12.20.0
```

flybayer:
Ah, looks like a bug! Thanks for the report.

For whoever wants to look into this, here's the broken line of code: https://github.com/blitz-js/blitz/blob/canary/packages/blitz/scripts/postinstall.js#L84

I think probably we need to also check that `root` is not null in that if statement. If null, skip that if.

meehawk:
@flybayer I'd love to take this up if thats okay?

flybayer:
You bet!

PR #2637, opened 2021-08-06
Fix case where `blitz new` prints an error but continues creating app

meehawk:
Closes: #2613 

Changed 
`if (isUsingNpm && !fs.existsSync(path.join(root, "node_modules", "next")))`

to handle situation when `root` is null as 
`if (isUsingNpm && root && !fs.existsSync(path.join(root, "node_modules", "next")))`

We skip the if block in case `root` is null

## Bug Checklist

## Feature Checklist


flybayer:
Sweet, thank you!!

PR #2649, opened 2021-08-11
Failing test dehydrated map

Skn0tt:
As mentioned in https://github.com/blitz-js/blitz/pull/2617#issuecomment-896124193, query dehydration doesn't seem to work correctly with objects like `Map` or `Set`. This PR initially adds a failing test to document that behaviour, and intends to fix it.


Skn0tt:
Debugged this a little, and it appears that this is caused by `babel-plugin-superjson-next` being executed twice during SSR. The first pass mutates the props for the second pass, which results in non-superjson-ified props being passed into `__NEXT_DATA`.

Skn0tt:
This is the same bug as https://github.com/blitz-js/babel-plugin-superjson-next/issues/63, and it was fixed in SuperJson@[v1.7.4](https://github.com/blitz-js/superjson/releases/tag/v1.7.4).

Skn0tt:
@flybayer I'm trying to run `cd nextjs/packages/next && yarn add superjson@latest`, but it throws an error (missing `execa`). Could you try updating SuperJSON? I'm pretty confident that should fix this error.

flybayer:
@Skn0tt to upgrade, bump the superjson version number in the package.json, then run yarn from repo root.

flybayer:
yes lol. Thank you!

PR #2652, opened 2021-08-12
Fix indentation of json on `blitz generate` edit pages

bravo-kernel:
Currently, the generated Edit pages show one large line of json instead of indented json [as done on this line](https://github.com/blitz-js/blitz/blob/canary/packages/generator/templates/page/__modelIdParam__.tsx#L24). 

This PR makes the generated Edit pages look identical to the View pages.


flybayer:
Cool, thank you!

PR #2659, opened 2021-08-19
Fix serialization of RedirectError

akbo:
Closes: -

RedirectError is registered incorrectly with SuperJson. It is probably a copy+paste oversight. This commit fixes that. 

## Bug Checklist

This probably doesn't require a test. If it does, please let me know and I will add one.


akbo:
Same as last PR, I ran the failing integration test locally twice and both times it is successful. Is there a problem with the speed of the CI machines?

MrLeebo (approved):
Thank you!

I agree, I don't think the CI error has anything to do with this change.

PR #2663, opened 2021-08-21
Convert app template Jest config to TypeScript

ratson:

This enables code completion.

## Feature Checklist

### DIFF
--- a/packages/generator/templates/app/README.md
+++ b/packages/generator/templates/app/README.md
@@ -119,7 +119,7 @@ __name__
 ├── README.md
 ├── babel.config.js
 ├── blitz.config.js
-├── jest.config.js
+├── jest.config.ts
 ├── package.json
 ├── tsconfig.json
 ├── types.d.ts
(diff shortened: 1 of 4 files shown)


flybayer:
Thanks @ratson! This seems reasonable for TS users. However, as is, this breaks support for JS users (`blitz new myapp --js`). You'll need to take the [same approach here as we do for jsconfig.json and tsconfig.json](https://github.com/blitz-js/blitz/blob/canary/packages/generator/src/generators/app-generator.ts#L32-L37)

packages/generator/templates/app/README.md
```diff
@@ -119,7 +119,7 @@ __name__
 ├── README.md
 ├── babel.config.js
 ├── blitz.config.js
-├── jest.config.js
+├── jest.config.ts
```
flybayer, line 122:
```suggestion
├── jest.config.js
```

packages/generator/templates/app/README.md
```diff
@@ -147,7 +147,7 @@ These files are:
 
 - `blitz.config.js` is for advanced custom configuration of Blitz. It extends [`next.config.js`](https://nextjs.org/docs/api-reference/next.config.js/introduction).
 
-- `jest.config.js` contains config for Jest tests. You can [customize it if needed](https://jestjs.io/docs/en/configuration).
+- `jest.config.ts` contains config for Jest tests. You can [customize it if needed](https://jestjs.io/docs/en/configuration).
```
flybayer, line 150:
```suggestion
- `jest.config.js` contains config for Jest tests. You can [customize it if needed](https://jestjs.io/docs/en/configuration).
```
keep these as js since that's the base case. Later we can optimize to set correctly based on js/ts.

flybayer:
Awesome, thank you!!

Issue #2038, opened 2021-03-02
Multiple fields forms using templates during generation


Maastrich:
### What do you want and why?

I was thinking of a feature that allow the generator to use less "static" templates for exemple for the Form which, after generation, only contains a `Name` field whether or not it exists.
The goal would be to complete the Form generation to a complete one allowing multiple variable as the CLI allow it.

After running command: `blitz g all test variableOne:string variableTwo:int` all the files ends up with the correct configuration

as `db/schema.prisma` which contains a complete new `Test` model:
```prisma
model Test {
  id        Int      @default(autoincrement()) @id
  createdAt DateTime @default(now())
  updatedAt DateTime @updatedAt
  variableOne String 
  variableTwo Int    
}
```

but not `app/tests/components/TestForm.tsx` which is still completely generic.

```typescript
import { Form, FormProps } from "app/core/components/Form"
import { LabeledTextField } from "app/core/components/LabeledTextField"
import * as z from "zod"
export { FORM_ERROR } from "app/core/components/Form"

export function TestForm<S extends z.ZodType<any, any>>(props: FormProps<S>) {
  return (
    <Form<S> {...props}>
      <LabeledTextField name="name" label="Name" placeholder="Name" />
    </Form>
  )
}
```

### Possible implementation(s)

Add to the templates file some specific comments that allow the generator to fill in the corrects information like:

The template file for form would be:

```typescript
import {Form, FormProps} from "app/core/components/Form"
import {LabeledTextField} from "app/core/components/LabeledTextField"
import * as z from "zod"
export {FORM_ERROR} from "app/core/components/Form"

export function __ModelName__Form<S extends z.ZodType<any, any>>(props: FormProps<S>) {
  return (
    <Form<S> {...props}>
     /* blitz_template {"subTemplate":"LabeledTextField","allowMultiple":true,"keysToChange":["__FieldName__Label","__FieldName__Name","__FieldName__Placeholder"]} */
    </Form>
  )
}
```

and the template file for the LabeledTextField subTemplate would be something like

```typescript
<LabeledTextField name="__FieldName__Name" label="__FieldName__Label" placeholder="__FieldName__Placeholder" />
```

Obviously, if you are interested in such feature, I would love myself to contribute to this project.

Thanks, Maastrich

flybayer:
Oh yes!! I've definitely wanted to add this, so I'd love if you'd work on it!

The subtemplate idea is a good one.

I wonder if we can do even better, maybe like this?

[...]

PR #2679, opened 2021-08-27
[legacy-framework] Subtemplate management

Roesh:
Closes: blitz-js/blitz#2038 

Continuing PR blitz-js/blitz#2134 , using @Maastrich 's regex functions in the generator.

This PR improves the generator scaffolding process, allowing files like mutations, pages and forms to include fields specified by the user, rather than have a default "name" field.

Sub tasks

- [x] Support the following replacement cases: field_name, Field_name, Field_Name 
- [x] Update form to use this format: `{/* template: <__FieldComponent__ name="__fieldName__" label="__Field_Name__" placeholder="__Field_Name__" /> */}`
- [x] [Abstract out](https://github.com/blitz-js/blitz/pull/2134#discussion_r602575611) getTemplateValues into the base Generator class, and allow subclasses to add or override the default values.
- [x] support // template: ... in addition to /* template: ... */
    Note: Supported, but only either one can exist in one file. Works like the previous logic we had, but we may want to support multiple template comments per file actually, so once that is decided, we can check this box. (EDIT: Leaving this as it works presently)
- [x] Don't hardcode a display component in the template comment, rather use `__FieldComponent__`, then provide a map in blitz.config.js or in templates/config
    - [x] Research if the new template key should be placed under the experimental key, or if it can be one of the top level keys
    - [x] Determine how expensive `await loadConfigAtRuntime()` is. If it is expensive, create a proxy to hold the typeToComponentMap so that multiple fields can use the config fetched during the first call to get the config object.
- [x] Determine if the logic that sets the `BLITZ_APP_DIR` variable is correct. The following files check if the variable is already set, and if not, sets it to the current directory using "."
    - [x] `packages\generator\src\generators\template-builders\builder.ts` (New abstract builder class)
    - [x] `packages\generator\src\generators\page-generator.ts` (Existing page generator class)
- [x] Ensure that subsequent runs of `blitz generate` use the existing file as a template if it exists.

## Feature Checklist

    - [x] form-generator test added
    Docs PR is https://github.com/blitz-js/blitzjs.com/pull/577


Roesh:
For the display component aspect of this, I was wondering if we could strongly type the field mappings in blitz.config.ts
```
import {LabeledTextField} from "app/core/components/LabeledTextField"

fieldComponents: {
    string: {
        name: LabeledTextField,
        importStatement: `import {LabeledTextField} from "app/core/components/LabeledTextField"`
    }
} 
```
Based on my understanding of how JS works and some searching, this does not seem possible (In c#, there is a nameof operator, but JS does not seem to have such a concept.)

Shall we leave these untyped and in the documentation inform users to ensure that the value corresponding to the type should have the correct name and path?

Im adding the path key because I haven't found an ideal way to infer the path to an object based on its reference. I'll search for this some more, but if there is a known way to do this, please let me know.

So we could do:

fieldComponents: {
    string: {
        name: LabeledTextField,
        importStatement: `import {LabeledTextField} from "app/core/components/LabeledTextField"`
    }
}

Roesh:
Progress update 09/03/2021:

Working on `support // template: ... in addition to /* template: ... */`

Currently in generator.ts, we have a single `fieldTemplateRegExp`. 

My current solution is create a separate regex for the `// template: ` 
Previous regex: `/{?\/\* template: (.*) \*\/}?/`
New regex: `/\/\/ template: (.*)/`

The alternative is to use a single regex but have multiple capture `(/{?\/\* template: (.*) \*\/}?|\/\/ template: (.*))/`. I may go that path because we later use the `fieldTemplateRegExp`.

Will continue to update once every couple days, feel free to pitch in with suggestions!

Roesh:
Today's progress update:

Working on making the base Generator class define the `getTemplateValues` by unmarking it as an abstract function.

Next steps were to identify the template values that are common among all generators so that we can return an object that contains the maximum number of overlapping values instead of reusing code.
Not all these values have been created already, so I will have to go through each existing generator, see what it will need, and once that is done, update the base Generator's `getTemplateValues` to return them.

One potential issue is that the app-generator does not require any model or model fields, so it won't be using any of the values that the others do. 
Another issue to solve is updating the `GeneratorOptions` base interface in Generator.ts to contain any properties that are needed by the base Generator to generate those values. This is because child generators, like `FormGenerator` extend the base `GeneratorOptions` with the properties like `modelName` and add those to the object returned by their `getTemplateValues` method

Roesh:
Progress update 09/12:

When creating the object returned by `getTemplateValues` we rely on the options being passed in containing properties such as `extraArgs` and so on. Will need to add these to the base Generator's options so that it can access these and build the options

Roesh:
Progress update 09/13:

Abstraction of `getTemplateValues` is complete. The base Generator class has a default implementation that children can override. Initially, these problems were present

- One generator (model generator) does not implement any getTemplateValues
- Another generator (app generator) have a completely different set of values

All other generators did output several common fields (like parentModels, modelId, etc.) 
This could have been solved by making the base generator class return the most common values while app generator and model generator would override conditionally

The issue here is that in order to generate the above options, we need the "extra args" and other models provided into the options argument passed in. This would require adding all of those options into the base generator's options interface. 

I personally did not like that the base generator would have a ton of parameters that would have to be set as optional (because not all generators (like the app generator) would provide "modelName", "modelNames" etc.), while their children when extending the base options, would require them to be passed in.

My solution was to create a separate `Builder` class that would be responsible for generating these field template values. The base generator class would now "have" a public `templateValuesBuilder`, and the base class' `getTemplateValues()` will call this builder's `getTemplateValues`.
Now, each child class would specify the builder it wanted to use.
Because everything other than the app generator and model generator share ~90% of the values, there are three concrete values builders, which inherit from an abstract template builder:

1) `AppValuesBuilder`: Used when generating `templateValues` during new app generation
2) `NullBuilder`: Used by default in the base Generator class, and also used by `ModelGenerator` which does not override the base
2) `FieldValuesBuilder`: Used by all other Generators, including `FormGenerator`, `MutationsGenerator`, etc.

The base abstract builder class also holds methods like `getId`, `getParam`, etc. which were originally duplicated across generators themselves.

Roesh:
Progress update 09/14/2020:

Didn't have as much time to work on this today, but I haven't found the best solution for populating the `process.env.BLITZ_APP_DIR` variable before calling the `loadConfigAtRuntime`

Doing `process.env.BLITZ_APP_DIR = "."` works, but I might need to ask Brandon if there is a better way this variable needs to be initialized so that we can cover all use cases?

Roesh:
Progress update 09/16/2021:

Added the template to component map support in biltz.config.ts. This required updating the `NextConfig` type in the `nextjs\packages\next\server\config-shared.ts` file.

The issue is whether this should be nested under the experimental property or if it can stay at the top level. The other issue is understanding how expensive `await loadConfigAtRuntime()` is. If it is expensive, we may need to create a proxy to hold the typeToComponentMap so that multiple fields can use the config fetched during the first call to get the config object.

nextjs/packages/next/server/config-shared.ts
```diff
@@ -168,6 +168,9 @@ export type NextConfig = { [key: string]: any } & {
     pageDataCollectionTimeout?: number
     isrMemoryCacheSize?: number
     concurrentFeatures?: boolean
+  }  
+  template?:{
```
Roesh, line 172: OK to be left here, or should this be moved into the experimental key?

packages/generator/src/generator.ts
```diff
@@ -211,6 +218,22 @@ export abstract class Generator<
     if (codeFileExtensions.test(pathEnding)) {
       templatedFile = this.replaceConditionals(inputStr, templateValues, prettierOptions || {})
     }
+    // templatedFile.match
+    const fieldTemplateRegExp = new RegExp(/({?\/\* template: (.*) \*\/}?|\/\/ template: (.*))/)
```
Roesh, line 222: - [ ] Support 0-many spaces between // and template for // template comments

Roesh:
Need to rewrite the form-generator test. It runs locally but fails during CI, I'm assuming because the builder that gets the configuration uses `await loadConfigAtRuntime()` which doesn't work in CI, causing `config.template` to be undefined.

Roesh:
OK, the fix for issue https://github.com/blitz-js/blitz/issues/2727 should resolve the tasks regarding BLITZ_APP_DIR for this PR. Once that is closed, I can check them off here.

flybayer:
@Roesh you are a gem for working on this!!

Roesh:
Thanks haha! Just noticed one of the CI failing tests was one I wrote, let me fix that. I noticed the failing Windows one but I think that test is being worked on elsewhere?

flybayer:
@Roesh ok awesome!! I will review next week.

@beerose can you take first pass through reviewing and testing this? Then I can do a more final review.

packages/generator/templates/mutations/delete__ModelName__.ts
```diff
@@ -3,7 +3,8 @@ import db from "db"
 import {z} from "zod"
 
 const Delete__ModelName__ = z.object({
-  id: z.number(),
+  id: z.number(), //TODO: Infer Id type from schema
```
beerose, line 6: Reminder: we can create an issue for that after merging this PR (or earlier?)

packages/generator/templates/app/blitz.config.ts
```diff
@@ -7,6 +7,15 @@ const config: BlitzConfig = {
       isAuthorized: simpleRolesIsAuthorized,
     }),
   ],
+  template:{
+    typeToComponentMap: {
```
beerose, line 11: I found this comment from the previous PR: https://github.com/blitz-js/blitz/pull/2134#issuecomment-808513567

Was it discussed for this PR as well (I might have missed it)? If not, that seems like something we can create an issue for afterwards.

nextjs/packages/next/server/config-shared.ts
```diff
@@ -171,6 +171,9 @@ export type NextConfig = { [key: string]: any } & {
...
+    typeToComponentMap: any
```
beerose, line 176: Could it be something like `typeToComponentMap: [key in PossibleTypes]: string`?

beerose: If `PossibleTypes` (for generator, I only found `zodPossibleTypes`) doesn't exist, we could add one — what do you think?

packages/generator/src/generators/template-builders/builder.ts
```diff
@@ -0,0 +1,106 @@
...
+import {GeneratorOptions} from "../../generator"
+import {
+  addSpaceBeforeCapitals,
+  camelCaseToKebabCase,
+  singleCamel,
+  singlePascal,
+} from "../../utils/inflector"
+
+export interface IBuilder<T> {
+  getTemplateValues(Options: T): Promise<any>
```
beerose, line 11: I know it's not the subject of this PR, but wondering if we can narrow it. From what I see `getTemplateValues` mostly returns a promise with something like:
```
    return {
      name: this.options.name,
      Name: this.options.Name,
    }
```

beerose (commented):
I noticed that after generating a new model, the template line remains in the form component. Is it expected?
Example of a generated file:
```js
export function PrtestForm<S extends z.ZodType<any, any>>(props: FormProps<S>) {
  return (
    <Form<S> {...props}>
      <LabeledTextField name="age" label="Age" placeholder="Age" />
      <LabeledTextField name="info" label="Info" placeholder="Info" />
      {/* template: <__FieldComponent__ name="__fieldName__" label="__Field_Name__" placeholder="__Field_Name__" /> */}
    </Form>
  )
}
```

flybayer:
> I noticed that after generating a new model, the template line remains in the form component. Is it expected?

Yes, that way after can run `blitz generate` a second time to add more fields to a model

Roesh:
Update 10/06/2021:

Pulled in the changes from the uuid support PR. 

Finished supporting uuid in the mutations template, need to do the same for the prisma generator (currently that PR allows non-id fields to be uuids, but does not affect the primary key)

flybayer (commented):
Ok finally I was able to thoroughly review today! Overall this is amazing!! Can't wait to ship it!

I do see one significant problem with how subsequent runs of `blitz g` work. Currently for subsequent runs, we use the base template file again. But that's a problem because if I make ANY changes to the generated file (for example, adding `console.log()` inside the ProductForm render function), the next blitz g run will remove that custom code. This is a show stopper for people using `blitz g` to update their code. So essentially what we need to do in this case is use the actual code in the project as the template file. That way any customizations are preserved. The idea is that ONLY a single line is inserted above the `{/* template:xxx*/` line. Everything else should be untouched.

I am very sorry, I didn't communicate that vision clear enough on how it should work. Hopefully it's not too much work to change this.


- [x] `blitz g test name:string?` does not use correct zod type. Generates `.any()` instead of `string`
- [ ] update and delete mutations are left with an extra line space. Can that be fixed? (maybe will be fixed with the rework mentioned above)
```ts
const UpdateProduct = z.object({
  id: z.number(),

  name: z.string(),
  price: z.number(),
  // template: __fieldName__: z.__zodTypeName__(),
})
```
- [x] In the template syntax, can we use `__zodType__` instead of `__zodTypeName__`?
- [x] I think we should update the config syntax to something like below. Open to ideas to improve even further
```ts
  templates: {
    fieldTypeMap: {
      string: {
        component: "LabeledTextField",
        inputType: "text",
        zodType: "string",
      },
      boolean: {
        component: "LabeledTextField",
        inputType: "text",
        zodType: "boolean",
      },
      int: {
        component: "LabeledTextField",
        inputType: "number",
        zodType: "number",
      },
      bigint: {
        component: "LabeledTextField",
        inputType: "number",
        zodType: "number",
      },
      float: {
        component: "LabeledTextField",
        inputType: "number",
        zodType: "number",
      },
      decimal: {
        component: "LabeledTextField",
        inputType: "number",
        zodType: "number",
      },
      datetime: {
        component: "LabeledTextField",
        inputType: "string",
        zodType: "string",
      },
      uuid: {
        component: "LabeledTextField",
        inputType: "text",
        zodType: "string().uuid",
      },
    },
  }
```
- [x] The idea with the above config format is that any new key/values added inside the object will be available for use in templates. 
- The above default config is quite lengthy. Should we omit it and just have those as defaults? I'm conflicted because this is something that seems that if it's easy to discover (by generating all this in blitz config) people will use it way more. Otherwise if it's all hidden by default, they won't realize they can customize it. So I'm thinking generate it all for now. We can always make them default and remove from config later if we get feedback that direction.

Roesh:
Progress update 10/17/2021:

Consolidated the types used across resource generators.

TODO:
- [ ] Noticed that the model generator isn't adding a relation field if we use the --parent flag. Looks like this is because it isn't part of the extra args which are what are used when those are generated. Need to create a separate issue to track this

Roesh:
Progress update 10/21/2021:

Tried to access prisma's functionality that performs `prisma format`, but wasn't very successful. I was rummaging through what the prisma node_modules folders exposed, and none of them have the Format function I am looking for.
- [ ] Research to see if there is a way to make a command line call from a generator function

Roesh:
Good question. I've got to check, but I assumed we didn't because in https://github.com/blitz-js/blitz/issues/2864, someone is running into an issue which would have been solved if prisma format was run after model generation.

Roesh:
Progress update 10/25/2021:

Continuing to investigate the copy issue, looks like its coming from 
`var diskFiles = globby.sync(fromGlob, globOptions);`

`fromGlob` is set on all paths, but diskFiles is only set for paths that dont end with "[" and "]"

flybayer:
@Roesh you are a champ!! I'm so thankful you are working on this.

Adding fast-glob as a dep sounds goods.

Roesh:
Ayy, thank you :)))

Cool, I'll do that

Roesh:
So I think that's all of the functionality that we needed so far. I'm going to start writing tests, including integration ones where we generate files and examine the output to make sure everything's OK. I might uncover things during that process, but hope to have some tests done by end of week

flybayer:
@Roesh awesome!

Roesh:
OK, so almost done, based on our conversation about integration testing, it looks like we'll want to wait for the CLI cypress integration tester to be complete, so all that's left is the accompanying documentation for this

The other thing to note are improvements that can be made to the incremental generation of fields. Right now, we're defaulting to using an existing file and adding fields specified in the new command, but there could be additional generation "modes" which we can consider supporting:

1) "From scratch" mode - will wipe the existing resources and replace them with the resources generated by the generate command, useful to quickly discard existing work
2) "Template override" mode - will use the existing resources if they exist, but will delete all field specific lines (things like the form fields and zod inputs, basically anything generated by the output of the 'Builder' classes. Useful if someone wants to keep their pages etc., but just want to both delete and/or update resources
3) "Add" mode, which is what we do now, which uses an existing file if present and adds any new fields to it, and if the file doesnt exist, generates it

imcodingideas:
This is amazing work, thank you for this.

flybayer:
@Roesh ugh, so sorry for delay in testing this out again. I've offloaded a lot of my other blitz work now. So I can follow up here more quickly now.

- yes you can update that doc you mentioned
- [ ] looks like you have a failing unit test: https://github.com/blitz-js/blitz/runs/4245479178?check_suite_focus=true
- [ ] Adding a new field after first generation fails
   - cd to `examples/auth/`
   - Run `blitz g all test name:string`
   - Run `blitz g all test description:string`
   - Fails with file doesn't exist, but the file does exist
   ```
  ✕ Error generating __modelIdParam__/edit.tsx
  ✕ AssertionError [ERR_ASSERTION]: Trying to copy from a source that does not exist: 
  /Users/b/c/blitz/examples/auth/app/pages/tests/\[testId\]/edit.tsx
  ```
- [ ] [Previous comment](https://github.com/blitz-js/blitz/pull/2679/files#r729407701) still pending. I tried adding that and looks like `__inputType__` is not used as a template variable. My intention was that every key inside the codegen config (`inputType`, `zodType`, `component`, etc) would be available as a template variable. I guess this could be added later, but I think generating a `number` field type is common enough that it would be good to have in initial release.
- [ ] I also just realized there's an opportunity for standardization because we use `zodType` in both config and template, but we use `component` in config and `FieldComponent` in template. So maybe we should use `FieldComponent` in the config too?
- [ ] Prisma type is not being read from blitz.config correctly
  - I added a `number` type to blitz.config fieldTypeMap:
  ```ts
      number: {
        component: "LabeledTextField",
        inputType: "number",
        zodType: "number",
        prismaType: "Int",
      },
   ```
   - Then `blitz g model apple price:number` results in prisma `Number` instead of `Int`
  ```
  $ blitz g model sauce price:number
  ✔ Model 'Sauce' created in schema.prisma:
  >
  > model Sauce {
  >   id        Int      @id @default(autoincrement())
  >   createdAt DateTime @default(now())
  >   updatedAt DateTime @updatedAt
  >   price     Number
  > }
  >
  ```
- [ ] Update mutation is still left with an extra line in the zod schema.  (generated with ` blitz g all test 'name:string?' price:number`) Not a big deal, but would be nice to be fixed
  ```ts
  const UpdateProduct = z.object({
    id: z.number(),
  
    name: z.string(),
    price: z.number(),
    // template: __fieldName__: z.__zodTypeName__(),
  })
  ```

Roesh:
Thats alright! Totally understandable,

I'll get to a few of these at once, and once you verify they're complete, you can check them off.

__inputType__ issue: I made a function that gets the input type just for this, but totally forgot to add the key to the getFieldTemplateValues function. Also updated __ModelName__Form.tsx so it now generates this:
```
<Form<S> {...props}>
      <LabeledTextField name="name" label="Name" placeholder="Name" type="text" />
      {/* template: <__FieldComponent__ name="__fieldName__" label="__Field_Name__" placeholder="__Field_Name__"  type="__inputType__" /> */}
    </Form>
```

Failing unit test: Updated it to take into account the __inputType__ variable that was added. Going to wait until github actions re-runs, haven't been able to resolve the running of tests locally.

Roesh:
11/21/2021:

Ok, so I'm having issues with getting the config from the `field.ts` file, unsure why it isn't picking the config up, hmm
```
blitz:config Loading config from  D:\QuickDocs\blitz\examples\auth\.blitz.config.compiled.js +2s
  blitz:config Did not find custom config file +241ms
```

Roesh:
Adding a new field after first generation fails 

Hm, this is a weird one, but after some debugging I think I know what happened. Basically, I don't think the escape characters function is working to fix the special characters issue, and when I worked on the fix, I was looking at the Form.tsx output which _was_ working, but I was ignoring the Edit.tsx and other special character files because I wasn't escaping the line that checked whether to use an existing file or not.
Before committing, I probably looked at that line and caught that, but didn't test it, so it looks like this is still TODO

beerose (commented):
I did some testing and marked a few things as done in Brandon's comment: https://github.com/blitz-js/blitz/pull/2679#issuecomment-972384027.

packages/generator/templates/app/blitz.config.ts
```diff
@@ -15,5 +15,69 @@ const config: BlitzConfig = {
     return config
   },
   */
+  codegen:{
```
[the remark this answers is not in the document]
flybayer: Yeah after further thought, let's make this config default so that it doesn't have to be in blitz.config. This will be better for people upgrading too.

Roesh:
Title: Blitzjs progress update 11/30/2021
TODO:
            After some consideration Brandon felt that codegen was simpy too long to fit, so we're going to remove that from the blitz.config.ts file and have it defaulted in code.

Next steps:
- [x] remove from generated code"
- [x] Make sure everything works like it did before"
- [ ] New issue - make sure uuid type works and specifies the uuid decorator"

Roesh:
Based on @Vandivier 's opened request, I realized that there is some additional default handling code that is needed. This is for two cases so far:
1) unspecifed fieldType
`blitz g all model field` doesn't pull a default component. Should we default in string in this case?
```
<Form<S> {...props}>
      <__component__ name="field" label="Field" placeholder="Field" type="__inputType__" />
      {/* template: <__component__ name="__fieldName__" label="__Field_Name__" placeholder="__Field_Name__"  type="__inputType__" /> */}
    </Form>
```

Logic for belongsTo must be specified, must intercept this "special name" before its processed as a template field. Right now this happens:

```
<Form<S> {...props}>
      <__component__
        name="belongsTo"
        label="Belongs To"
        placeholder="Belongs To"
        type="__inputType__"
      />
      {/* template: <__component__ name="__fieldName__" label="__Field_Name__" placeholder="__Field_Name__"  type="__inputType__" /> */}
    </Form>
```

packages/generator/src/utils/get-codegen.ts
```diff
@@ -0,0 +1,101 @@
...
+
+export const getResourceConfigFromCodegen = async (
+  fieldType: keyof CodegenConfig["fieldTypeMap"],
+): Promise<CodegenConfig["fieldTypeMap"][string]> => {
+  const codegen = (await getCodegen()).codegen as CodegenConfig
+  const config = codegen.fieldTypeMap[fieldType] || {}
+  return config
+}
+
+export const getCodegen = async (): Promise<Pick<NextConfigComplete, "codegen">> => {
```
beerose, line 87: We could simplify it to return `codegen` directly instead of an `{codegen: codegen}` object

beerose:
We're loading codegen a few times — I'm wondering if we can load it once somewhere top-level and pass it down as we do for other commands with `loadConfigProduction`.

Roesh:
> We're loading codegen a few times — I'm wondering if we can load it once somewhere top-level and pass it down as we do for other commands with `loadConfigProduction`. E.g. we could pass it to generators that need it in `generate.ts` file.

Yeah, I initally was caching it in the builder, but I moved that logic into get-codegen, and I didn't know a good spot to add a cache variable since it wasn't a class

Roesh:
So I was doing some testing on the belongs to field to make sure it was similar to the one specified with the --parent flag, I don't see the model being generated with a User[] field, but if I use the belongsTo key, I do see those fields.

The field.ts file generates the schema values, so I'll take a look at that once I'm done with syncing up the generator behavior for the two

https://blitzjs.com/docs/cli-generate#options

nextjs/packages/next/server/config-shared.ts
```diff
@@ -176,6 +189,7 @@ export type NextConfig = { [key: string]: any } & {
     isrMemoryCacheSize?: number
     concurrentFeatures?: boolean
   }
+  codegen?: CodegenConfig
```
beerose, line 192: todo: should be places next to Blitz's stuff starting at `// -- Blitz start`

imcodingideas: blocker or nitpick?

imcodingideas:
> So I was doing some testing on the belongs to field to make sure it was similar to the one specified with the --parent flag, I don't see the model being generated with a User[] field, but if I use the belongsTo key, I do see those fields.
> 
> The field.ts file generates the schema values, so I'll take a look at that once I'm done with syncing up the generator behavior for the two
> 
> https://blitzjs.com/docs/cli-generate#options
> 
> Ok, so I broke my node_modules again. donig a global search while yarn dev is running always messes things up lol. I was able to update the parent zod type to use the new codegen utility, but will have to test it once the repo is re-installed.

Let me know if you need support wrapping this up. This looks fun.

beerose:
> Let me know if you need support wrapping this up. This looks fun. I know there is the belongsTo field case and the todo from @beerose

Thank you @imcodingideas! 
I'll fix the conflicts shortly. And if you'd want to test it briefly, we'd very much appreciate that! 

I also think that the `belongTo` thing can be covered later (unless it's a regression).

Roesh:
Yeah feel free to jump in! I have less time these days so am not be able to contribute as much as I used to :(((

Jarrodsz:
Any updates on this? Will this be released in short terms? Or else, is there a prisma scaffold generator that anyone could advise. Kind regards

flybayer:
@Jarrodsz yeah we're trying to ship it very soon.

Roesh:
Ok, so I should have some free time now. I'm making a virtual machine and installing nvm as well so hopefully between the stability offered by nvm and the ability to take machine snapshots my npm installation should be stable.

I can handle the belongsTo flag in a couple days time I think

packages/generator/src/generators/template-builders/field-values-builder.ts
```diff
@@ -0,0 +1,110 @@
...
+        }
+        if (valueName === "belongsTo") {
+          // TODO: Determine how this is done. The model will generate with a field with the id name
+          // and type of the parent of this model, and forms etc. should
+          // In addition, need to do the same logic that the options.parentModel != undefined below does
+          specialArgs[arg] = "present"
+        }
+      })
+      await Promise.all(processSpecialArgs)
+      // Filter out special args by makins sure the argument isn't present in the list
```
beerose, line 56:
```suggestion
      // Filter out special args by making sure the argument isn't present in the list
```

Roesh:
Aleksandra helped setting up a stable dev environment by recommending Volta and the following package versions:
```
"volta": {
    "node": "14.18.1",
    "yarn": "1.18.0",
    "npm": "6.14.15"
  },
```

I began working on the belongsTo:ModelName feature. I was processing special args in the field-values-builder, but a lot what I'm trying to do is done in `packages\cli\src\commands\generate.ts`, around line 248:

```
const generators = generatorMap[args.type]
      for (const GeneratorClass of generators) {
        const generator = new GeneratorClass({
...
        rawParentModelName: flags.parent,
        parentModel: modelName(flags.parent),
        parentModels: modelNames(flags.parent),
        ParentModel: ModelName(flags.parent),
        ParentModels: ModelNames(flags.parent),
```

I copied those into the field values generator for now, but we might want to make it a separate utility down the line

The prisma schema gets generated fine but the create model which is supposed to have the parent model id as a requirement doesn't work. The template file uses `if (process.env.parentModel) {` to render the parentModelId field conditionally. I added `process.env.parentModel = typeName` to field values builder but its still not being set, need to troubleshoot that next

siddhsuresh:
Hey, @Roesh I have ported this PR to the toolkit in the branch [port-subtemplate](https://github.com/blitz-js/blitz/tree/port-subtemplate) do you want to continue to work on this there? If you are busy with other commitments, would it be okay, if I were to continue working on this?

Roesh:
Hello!

Yeah feel free to continue work on this in the toolkit, I'm now working fulltime so don't have as much time as before.

flybayer:
continued here: https://github.com/blitz-js/blitz/pull/3869

PR #2698, opened 2021-09-08
Fix custom-server hot reload on first run

andreasasprou:
We want to build custom-server, but only run watch if specified.

Closes: ?

## Bug Checklist

## Feature Checklist

### DIFF
--- a/packages/server/src/next-utils.ts
+++ b/packages/server/src/next-utils.ts
@@ -253,30 +253,25 @@ export function startCustomServer(
         })
     }
 
-    if (watch) {
-      // Handle build & Starting server
-      const esbuildOptions = getEsbuildOptions()
-      esbuildOptions.watch = {
-        onRebuild(error) {
-          if (error) {
-            log.error("Failed to re-build custom server")
-          } else {
-            newline()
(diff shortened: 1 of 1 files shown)


packages/server/src/next-utils.ts
```diff
@@ -253,30 +253,25 @@ export function startCustomServer(
...
+          process.exitCode = RESTART_CODE
+          process.kill("SIGABRT")
+        }
+      },
+    } : undefined
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    esbuild.build(esbuildOptions).then(() => {
       spawnServer()
-    }
+    })
```
flybayer, line 275: Thanks @andreasasprou!

This change has a slight regression.
- Before: `blitz start` will not build custom server (it was built with `blitz build`)
- After: `blitz start` will build custom server, requiring dev deps in production environment.

So we need to skip esbuild if `config.env === 'prod'` in this function.

andreasasprou: I'd added a small change to mitigate this in the most recent commit in this PR

flybayer:
Thanks bro!!

Issue #2674, opened 2021-08-26
Fix `blitz install` to work with pnpm


flybayer:
### What do you want and why?

Currently `blitz install` only knows about yarn or npm. So when you run `blitz install tailwind` in a pnpm repo, the code doesn't see yarn, so installs all the deps with npm. 

This code here needs fixed: https://github.com/blitz-js/blitz/blob/canary/packages/cli/src/commands/install.ts#L245-L249

P.s. general pnpm support is just now working with https://github.com/blitz-js/blitz/pull/2673. You can test with `blitz@0.40.0-danger.0a8df2b8a.10`

PR #2702, opened 2021-09-08
fix(cli): blitz install to work with pnpm

meehawk:
Closes: #2674 

Earlier when you run `blitz install tailwind` in a pnpm repo, the code doesn't see `yarn`, so defaults to `npm`. With this PR, blitz recognizes the presence of a `pnpm-lock.yaml` file and so installs with `pnpm`


meehawk:
Apologies for the delay @flybayer

flybayer:
Awesome, thank you!!

Issue #2727, opened 2021-09-22
Internal Blitz Error: process.env.BLITZ_APP_DIR is not set


kenneho:
### What is the problem?

Hi.

Steps to reproduce:

1. Install 0.40.0-canary.8
2. `blitz new myapp`
3. `cd myapp`
4. `blitz generate all something`. Throws error message `Error: Internal Blitz Error: process.env.BLITZ_APP_DIR is not set`

### Error logs:

`blitz generate all something` output:
```
CREATE    app/pages/somethings/[somethingId].tsx
CREATE    app/pages/somethings/[somethingId]/edit.tsx
CREATE    app/pages/somethings/index.tsx
CREATE    app/pages/somethings/new.tsx
CREATE    app/somethings/components/SomethingForm.tsx
CREATE    app/somethings/queries/getSomething.ts
CREATE    app/somethings/queries/getSomethings.ts
CREATE    app/somethings/mutations/createSomething.ts
CREATE    app/somethings/mutations/deleteSomething.ts
CREATE    app/somethings/mutations/updateSomething.ts
✕ Error: Internal Blitz Error: process.env.BLITZ_APP_DIR is not set
```

Roesh:
I think I have the fix for this, working on it in PR https://github.com/blitz-js/blitz/pull/2679

Wondering if we should add a separate quick fix for this:
packages/generator/src/generators/page-generator.ts
```
async postWrite() {
    if (!process.env.BLITZ_APP_DIR) {
      process.env.BLITZ_APP_DIR = "."
    }
    const {loadConfigProduction} = await import("next/dist/server/config-shared")
    const {saveRouteManifest} = await import("next/dist/build/routes")
```

Roesh:
I actually was wondering where the BLITZ_APP_DIR should be set. 

1) Should this be done by the abstract generator's constructor instead of at the postWrite hook in that one page generator? 
I don't mind leaving it here until we add a feature to all generators that requires the config from blitz.config

flybayer:
@Roesh yes let's get a quick fix for this.

It should be set in https://github.com/blitz-js/blitz/blob/canary/packages/cli/src/commands/generate.ts, same as we are doing in https://github.com/blitz-js/blitz/blob/canary/packages/cli/src/commands/routes.ts#L32

Roesh:
Ok, I've made the following change:

```
constructor(protected readonly options: T) {
    super()

    this.options = options
    this.store = createStore()
[...]

PR #2732, opened 2021-09-23
Fix blitz app dir

Roesh:
Closes: #2727 

Sets the BLITZ_APP_DIR env variable to process.cwd() to prevent `loadConfigProduction` from throwing an error.

## Bug Checklist

// TODO


flybayer:
Thanks @Roesh, but this fix should be in the generate cli command as I commented here: https://github.com/blitz-js/blitz/issues/2727#issuecomment-925081155

I want to get this out in a release today, so I've already fixed in https://github.com/blitz-js/blitz/pull/2736

Roesh:
Sounds good. I'll pull it into the PR I'm working on.

Issue #2636, opened 2021-08-06
`redirectAuthenticatedTo` does not call queries when redirecting from a dynamic route


CarterMcAlister:
### What is the problem?

When redirecting from a dynamic route, the queries are not called on the page you are redirected to.
`redirectAuthenticatedTo` works fine if the page is not a dynamic route.

I've tried redirecting to a non-dynamic route first as a workaround, and this did not work.

I believe this [this issue](https://github.com/blitz-js/blitz/issues/2527) may be related

### Paste all relevant code snippets here:

Reproduction repo: https://github.com/CarterMcAlister/blitzjs-redirect-bug-example

### What are detailed steps to reproduce this?

1. Pull down the reproduction repo
2. Sign up and log in
3. go to `/test/[example]`
4. You will be redirected to the home page
5. `getCurrentUser` call hangs, and the `<UserInfo>` component will stay in a suspense state

CarterMcAlister:
Note: the `useRedirectAuthenticated` hook **does** work properly on dynamic routes

flybayer:
Ok, thanks @CarterMcAlister!

PR #2741, opened 2021-09-24
Fix redirectAuthenticatedTo does not call queries when redirecting fr…

s-r-x:
…om a dynamic route

Closes: #2527, #2636 

The evil raises [here](https://github.com/blitz-js/blitz/blob/canary/nextjs/packages/next/data-client/react-query.tsx#L129). react-query throws a promise... and that's the end of the story. all because the router isReady state is forever false, since the moment of redirect...
on [this line](https://github.com/blitz-js/blitz/blob/canary/nextjs/packages/next/shared/lib/router/router.ts#L649) next seeds the initial isReady state of the router. there are many flags that i don't understand, but autoExportDynamic looked like our patient. 
[this](https://github.com/blitz-js/blitz/blob/canary/nextjs/packages/next/shared/lib/router/router.ts#L839) is the only place where isReady can potentially become true later(as far as i understand). 
[here](https://github.com/blitz-js/blitz/blob/canary/nextjs/packages/next/client/index.tsx#L221) next container could possibly set isReady to true, but only if isSsr prop of the router was true on componentDidMount. the problem is - when redirect happens before this hook,  isSsr becomes false on didMount, so [this one](https://github.com/blitz-js/blitz/blob/canary/nextjs/packages/next/client/index.tsx#L245) never gets called...

[the first solution](https://github.com/blitz-js/blitz/compare/canary...s-r-x:canary#diff-543a24f6b2c2a2b3a5be806641dfbd3699f73409da4cb56551bb59b4abcb3a45) is a bit hacky. i'm not quite familiar with next.js internals so here may be some edge cases that i don't know about.
[the second one](https://github.com/blitz-js/blitz/compare/canary...s-r-x:canary#diff-f2e285ca72ba6ef0d8da9f6e7928f79da842b7296f1d20b5a3c8f33f3a5cde10) is a clean one, and also fixes server/client markup mismatch error on redirects. the downside is that the user will see the initial page for a moment, but i guess it can be fixed in the same manner as suppressFirstRenderFlicker... and also additional rerenders on router state change

## Bug Checklist

## Feature Checklist

### DIFF
--- a/nextjs/packages/next/client/index.tsx
+++ b/nextjs/packages/next/client/index.tsx
@@ -214,19 +214,32 @@ let CachedApp: AppComponent, onPerfEntry: (metric: any) => void
 class Container extends React.Component<{
   fn: (err: Error, info?: any) => void
 }> {
+  wasSsrBeforeMount: boolean | undefined = undefined
+  constructor(props: any) {
+    super(props)
+    if (router) {
+      this.wasSsrBeforeMount = router.isSsr
+    }
+  }
   componentDidCatch(componentErr: Error, info: any) {
     this.props.fn(componentErr, info)
   }
 
   componentDidMount() {
+    // on routes with dynamic params the router's initial isReady status is false
+    // and it will be settled to true in the code below
+    // but if there were any redirects before componentDidMount hook(this one)
+    // isSsr becomes false and isReady status will be false forever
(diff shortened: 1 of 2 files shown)


MadcowD:
still broken :( @beerose

Issue #2746, opened 2021-09-25
CSRFTokenMismatchError when logout


Vahor:
**Edit :** It's maybe not related to logout, I got this same error message on first load, just after starting production. And this multiples times in a row. And after some time, while I was looking in local if I hadn't broke something, it worked again ( I've cleared my cache directly after the first crash )

### What is the problem?

Hello, I have a `CSRFTokenMismatchError` error with the logout method
It's a bit random but it always happens during/after a logout

- I have already tried deleting the node_module and the .next folder and rebuilding.
- Clearing cache does not change anything
- Using development / production ( on different machines ) gives the same error

### Paste all your error logs here:

As It is a bit random, I have added logs from before the crash ( removed dates for readability )

```
login() Finished: resolver:442ms serializer:2ms total:447ms 

getCurrentUser() Starting with input: null 
getCurrentUser() Finished: resolver:89ms serializer:2ms total:91ms 
 
getNotifications() Starting with input: null 
getNotifications() Finished: resolver:30ms serializer:1ms total:31ms 

logout() Starting with input: undefined 
logout() Finished: resolver:151ms serializer:1ms total:153ms 
 
getCurrentUser() Starting with input: null 
getCurrentUser() Finished: resolver:0ms serializer:0ms total:1ms 
 
getNotifications() Starting with input: null 
getNotifications() Finished: resolver:0ms serializer:1ms total:1ms 
 
getCurrentUser() Starting with input: null 
getCurrentUser() Finished: resolver:0ms serializer:0ms total:1ms 
 
getNotifications() Starting with input: null 
getNotifications() Finished: resolver:0ms serializer:1ms total:1ms 

2021-09-25 20:41:52.813 ERROR Error while processing the request 

 CSRFTokenMismatchError  
details:
{
  name: 'CSRFTokenMismatchError',
  statusCode: 401
} 
[CSRFTokenMismatchError] { statusCode: 401 }
events.js:377
      throw er; // Unhandled 'error' event
      ^

Error [ERR_STREAM_WRITE_AFTER_END] [ERR_STREAM_WRITE_AFTER_END]: write after end
    at writeAfterEnd (_http_outgoing.js:694:15)
    at ServerResponse.end (_http_outgoing.js:815:7)
    at ServerResponse.end (/home/site/node_modules/next/dist/compiled/compression/index.js:1:4179)
    at ServerResponse.apiRes.end (/home/site/node_modules/next/dist/server/api-utils.js:83:25)
[...]

PR #2761, opened 2021-09-29
Fix CSRFTokenMismatchError on logout

flybayer:
Closes: https://github.com/blitz-js/blitz/issues/2746

Fix CSRFTokenMismatchError on logout

We now create a new anonymous session on logout, which prevents race conditions when queries refresh right after logout. So end result is the same, but now one anon session is created ahead of time.


iDavidB:
Any idea when this will be released? Been having this issue occur a lot.

flybayer:
@iDavidB I'm trying for today, but if not, Monday

PR #2775, opened 2021-10-02
Fix Yarn Berry gitignore for new apps

JuanM04:
Now, it commits, for example, `install-state.gz`. This gitignore was taken from the [Yarn berry documentation](https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored).

Also, Prettier will ignore `.pnp*` and the Yarn folder


flybayer:
Sweet, thank you!

Issue #2782, opened 2021-10-02
Expose the Blitz logger and chalk for easy usage


jmcelreavey:
### What do you want and why?
Currently in order to include the logger and chalk we have to 
```
import { log } from "@blitzjs/display"
import chalk from "chalk"
```

These are not currently exposed or documented, we should expose these in the blitz module

jelling:
+1 for documenting `log`

beerose:
Right now, the logging utility is here: https://github.com/blitz-js/blitz/blob/canary/nextjs/packages/next/server/lib/logging.ts. We removed the `@blitz/display` package.

PR #2783, opened 2021-10-02
exposing @blitzjs/display package

jmcelreavey:
Closes: #2782

### DIFF
--- a/packages/blitz/src/index.ts
+++ b/packages/blitz/src/index.ts
@@ -4,6 +4,7 @@ export * from "@blitzjs/core/dynamic"
 export * from "@blitzjs/core/head"
 export * from "@blitzjs/core"
 export * from "@blitzjs/core/server"
+export * from "@blitzjs/display"
 
 /*
  * IF YOU CHANGE THE BELOW EXPORTS


packages/blitz/src/index.ts
```diff
@@ -4,6 +4,7 @@ export * from "@blitzjs/core/dynamic"
 export * from "@blitzjs/core/head"
 export * from "@blitzjs/core"
 export * from "@blitzjs/core/server"
+export * from "@blitzjs/display"
```
flybayer, line 7: Hey thanks @jmcelreavey!

So this actually doesn't expose the core logger.

1. What we need to expose is [`baseLogger`](https://github.com/blitz-js/blitz/blob/canary/nextjs/packages/next/server/lib/logging.ts#L31), although I think we should export it as `logger` instead of `baseLogger`.
2. And to export it, we should move that `logging.ts` file into this directory: https://github.com/blitz-js/blitz/tree/canary/nextjs/packages/next/stdlib-server
3. And then re-export `logger` from `next/stdlib-server/index.ts`
4. We already export `* from 'next/stdlib-server'` here in `blitz/src/index.ts`
5. But you need to add `logger` to this import rewrite list: https://github.com/blitz-js/blitz/blob/canary/packages/babel-preset/src/rewrite-imports.ts#L28-L38

Lastly, next time please don't remove the PR template from the description :) It has a critical checklist, namely ensure changes are covered by tests (this change doesn't need a new test) and opening a PR to the blitz docs repo to document any new features. And you are doing that, so a doc PR will be required.

flybayer (changes requested):

packages/blitz/src/index.ts
jmcelreavey: Apologies! Re-added the PR template and will update this PR accordingly. Thanks for all the helpful pointers :)

jmcelreavey: @flybayer Are we sure we want to expose baseLogger? `log` within `@blitzjs/display` gives us a lot more functionality such as spinners and chalk.

flybayer: To start with, let's do just baseLogger. Users can always import and use chalk directly. I've found baseLogger is almost always what you want in app dev. Chalk is more useful when building CLIs.

We can always add more later if there is demand

flybayer:
@jmcelreavey awesome! Could you also resolve the merge conflicts? (run git fetch --all locally then git merge upstream/canary into your branch).

beerose:
Hey @jmcelreavey! Do you think you'll have the time to finish this PR?

beerose:
Closing due to inactivity. Feel free to reopen!

Issue #2781, opened 2021-10-02
delay between `blitz new` prompts with slow internet


beerose:
### What is the problem?

When a user has a slow network, there is a delay between `"install deps?"` prompt and `"form library"` prompt. 
It's because after `install deps` step, we are fetching the latest blitz version to see if cli needs to be updated. It looks like cli is stuck because of no activity indicator. 

**Possible implementations:**

* show a loading indicator,
* do the network request in the background while the user is going through the prompt

### Paste all your error logs here:

n/a

### Paste all relevant code snippets here:

n/a

### What are detailed steps to reproduce this?

1. Downgrade Blitz to an older version if you're on the latest.
2. Run `blitz new`.
3. See what happens after `install deps?` prompt.

### Run `blitz -v` and paste the output here:

n/a

flybayer:
Here is the code that handles the upgrade check, and that either needs a spinner or, maybe even better, network request moved to the background: https://github.com/blitz-js/blitz/blob/canary/packages/cli/src/commands/new.ts#L105-L112

g3offrey:
I'm working on this 😄

g3offrey:
I tried to think about both of the solutions (the spinner, and the background task one). I'm not sure the second one is the best in terms of user experience. Indeed, what happen if the user is in the middle of the prompting part when the request finish. Should we take it back at the start? Record his answers? 
What happen if the prompting questions change between the version used initially and the upgraded one? 

I think the spinner solution is a good one in terms of UX per developing time.

Hope my code is good enough. Don't hesitate to request a change to the pull request, I'll be glad to update it.

On a second hand I want to thank you @flybayer and the Blitz team for doing such an awesome work on the project. I really love using Blitz and contributing to it is really nice. The codebase is very clean ✌️

Issue #2118, opened 2021-03-16
blitz new --dry-run not respected


CGamesPlay:
### What is the problem?

The documentation for `blitz new` indicates a `--dry-run` flag which will "show what files will be created without writing them to disk". It appears to have no effect.

### Paste all your error logs here:

```
$ ls
node_modules  package.json  yarn.lock

$ blitz new foo --dry-run
You are using beta software - if you have any problems, please open an issue here:
      https://github.com/blitz-js/blitz/issues/new/choose

✔ Pick a form library (you can switch to something else later if you want) · React Final Form

Hang tight while we set up your new Blitz app!

✔ The file "/Users/rpatterson/Projects/foo/package.json" has conflicts. What do you want to do? · overwrite
CREATE    .env
CREATE    .env.local
CREATE    .env.test.local
CREATE    .eslintrc.js
CREATE    .gitignore
CREATE    .npmrc
CREATE    .prettierignore
CREATE    README.md
CREATE    app/api/.keep
CREATE    app/auth/components/LoginForm.tsx
CREATE    app/auth/components/SignupForm.tsx
CREATE    app/auth/mutations/changePassword.ts
CREATE    app/auth/mutations/forgotPassword.test.ts
CREATE    app/auth/mutations/forgotPassword.ts
CREATE    app/auth/mutations/login.ts
CREATE    app/auth/mutations/logout.ts
CREATE    app/auth/mutations/resetPassword.test.ts
CREATE    app/auth/mutations/resetPassword.ts
CREATE    app/auth/mutations/signup.ts
CREATE    app/auth/pages/forgot-password.tsx
CREATE    app/auth/pages/login.tsx
CREATE    app/auth/pages/reset-password.tsx
CREATE    app/auth/pages/signup.tsx
CREATE    app/auth/validations.ts
CREATE    app/core/components/Form.tsx
CREATE    app/core/components/LabeledTextField.tsx
CREATE    app/core/hooks/useCurrentUser.ts
CREATE    app/core/layouts/Layout.tsx
CREATE    app/pages/404.tsx
CREATE    app/pages/_app.tsx
CREATE    app/pages/_document.tsx
CREATE    app/pages/index.test.tsx
CREATE    app/pages/index.tsx
CREATE    app/users/queries/getCurrentUser.ts
CREATE    babel.config.js
CREATE    blitz.config.js
CREATE    db/index.ts
CREATE    db/migrations/.keep
CREATE    db/schema.prisma
[...]

PR #2791, opened 2021-10-03
add a spinner during the blitz new version check

g3offrey:
Closes: #2781 

I added a spinner to let the user know that something is happening in the background.
I also added two more log lines to know exactly what is the result of the upgrading process.

## Bug Checklist

## Feature Checklist


mrbrunelli (approved):
:+1:

packages/cli/src/commands/new.ts
```diff
@@ -104,11 +104,18 @@ export class New extends Command {
...
+        .start()
       const wasUpgraded = await this.maybeUpgradeGloballyInstalledBlitz()
+      spinner.succeed()
       if (wasUpgraded) {
+        log.success("Blitz was sucessfully upgraded to latest version")
         this.rerunButSkipUpgrade(flags, args)
         return
       }
+
+      log.success("Blitz already running the latest version")
```
flybayer, line 118:
```suggestion
      log.success("You already have the latest version")
```

flybayer (commented):
Awesome, thank you! I concur with your comments in the original issue.

One change I think we should make here though is move the upgrade check before `determinePkgManagerToInstallDeps`. I think that will make for a smoother CLI flow.

g3offrey:
Yeah, I think you totally right @flybayer, I also think it makes more sense to check update before every prompt. 
I swapped the two first step so. You can find the update in my latest commit. 😄

beerose (approved):
Works great! Thanks!

beerose:
> Works great! Thanks!

I was too fast 😅 Looks like `new.test.ts` is failing. Could you take a look?

packages/cli/src/commands/new.ts
```diff
@@ -14,7 +14,6 @@ import {runPrisma} from "./prisma"
 
 export interface Flags {
   js: boolean
-  "skip-install": boolean
```
g3offrey, line 17: I removed this and related code, as this is dead code that now we do update before asking for dependency installation.

packages/cli/src/commands/new.ts
```diff
@@ -99,26 +92,34 @@ export class New extends Command {
...
+
     try {
       const destinationRoot = require("path").resolve(args.name)
       const appName = require("path").basename(destinationRoot)
 
       const form = await this.determineFormLib(flags)
 
       const {"dry-run": dryRun, "no-git": skipGit} = flags
-      const needsInstall = dryRun || !shouldInstallDeps
+      const needsInstall = !dryRun && shouldInstallDeps
```
g3offrey, line 122: This condition was wrong isn't it ? 
We should install if we are not dry running and if `shouldInstallDeps` is truthy

packages/cli/test/commands/new.test.ts
```diff
@@ -137,9 +137,14 @@ describe("`new` command", () => {
       })
 
       const newAppDir = fs.mkdtempSync(path.join(tempDir, "full-install-"))
-      await whileStayingInCWD(() => New.run([newAppDir, "--skip-upgrade"]))
-
-      expect(getStepsFromOutput()).toStrictEqual([`cd ${newAppDir}`, "blitz dev"])
+      await whileStayingInCWD(() => New.run([newAppDir, "--pnpm", "--skip-upgrade"]))
```
g3offrey, line 140: I force `pnpm` to be sure that we are running `pnpm install`

packages/cli/test/commands/new.test.ts
```diff
@@ -137,9 +137,14 @@ describe("`new` command", () => {
...
+
+      expect(getStepsFromOutput()).toStrictEqual([
+        `cd ${newAppDir}`,
+        "pnpm install",
+        "blitz prisma migrate dev (when asked, you can name the migration anything)",
```
g3offrey, line 145: There were not all the steps. I think this test was a false positive because of the `dryRun` condition bug.

packages/cli/src/commands/new.ts
beerose: Here's an issue related to `dry run`: https://github.com/blitz-js/blitz/issues/2118. I think that are more problems with dry run flag, but yeah, looks like the condition was wrong

packages/cli/src/commands/new.ts
beerose: Aren't those two separate things? If you pass `skip-install` it should assume automatically that you don't want to install dependencies and don't show the `install deps?` prompt.

beerose: I think this line: https://github.com/blitz-js/blitz/pull/2791/files#diff-6290fc15ecf346e1efd5672845e50c464b898dd25364403fe918314da4d838ffR95 should go after `determinePkgManagerToInstallDeps`

g3offrey: Yes I totally agree about the `skip-install` part, however this was mainly used when the upgrade Blitz process finished to rerun the whole `new` command without the `install deps?` prompt. As now the upgrade process is before we don't need to restart the `new` command without this prompt. Also the `skip-install` was hidden, so not used by user.

g3offrey: Indeed, I'll swipe these lines.

g3offrey:
@beerose I'll revert my changes on `skip-installs` as we need to skip dependency installation in tests to keep the test duration below the jest timeout.

g3offrey:
@flybayer @beerose I updated the way I displayed the spinner. In the previous version some log messages were duplicated, or were wrongly displayed (eg: when user refused the upgrade, it displayed user has the latest version). I think now it should be clearer.

packages/cli/src/commands/new.ts
```diff
@@ -289,19 +291,22 @@ export class New extends Command {
...
             return true
           }
         }
+
         this.error(
           "Unable to upgrade blitz, please run `blitz new` again and select No to skip the upgrade",
         )
       }
+    } else {
+      spinner.succeed(log.withBrand("You already have the latest version"))
```
beerose, line 307: Since the previous message disappears, and by `You already have the latest version`, users might not know what it was about.
```suggestion
      spinner.succeed(log.withBrand("You have the latest Blitz version"))
```

Issue #2751, opened 2021-09-27
Loading a mutation via `blitz console` fails silently due to a missing ENV var


beerose:
### Discussed in https://github.com/blitz-js/blitz/discussions/2745

> The mutation in question was not being loaded because it referenced an environment variable that was missing. This meant the code would compile via `npm run build` but when console loads mutations it apparently evaluates the code in a run-time fashion even prior to the mutation. Unfortunately, though, if it fails due a missing ENV var it does so silently.

_Originally posted by @jelling in https://github.com/blitz-js/blitz/discussions/2745#discussioncomment-1387880_

g3offrey:
I am working on a solution, that enables the debug mode when run with the verbose flag.

jelling:
@g3offrey what do you think of making it display errors by default, meaning even without a debug flag? My thinking is that the console is designed almost exclusively for technical users. At a minimum, I don't think a console command (or module load) should ever fail silently. A middle ground might be writing errors to a log but only displaying an abbreviated trace message unless the debug flag is set.

g3offrey:
Yes that what I did finally. I always display the errors. I also documented the previously existing DEBUG flags, I think they could have helped in your initial case. You can find my contributions here : 
https://github.com/blitz-js/blitzjs.com/pull/560
https://github.com/blitz-js/blitz/pull/2792

PR #2792, opened 2021-10-03
display an error if loading a module fails in blitz console

g3offrey:
Closes: #2751 

I decided to simply log the error that appears at the loading of the module.
I initially thought about a verbose mode, but I finally think this error should always be displayed.
If expert users need more details they can still enable additional logging using the DEBUG env var.

I'll check if there is some documentation about the DEBUG scopes, if not I'll be glad to create it.

Edit : I created it here : https://github.com/blitz-js/blitzjs.com/pull/560

## Bug Checklist

## Feature Checklist

### DIFF
--- a/packages/repl/src/utils/load-blitz.ts
+++ b/packages/repl/src/utils/load-blitz.ts
@@ -55,7 +55,7 @@ export const loadBlitz = async () => {
           [name]: contextObj,
         }
       } catch (error) {
-        log.error(`Failed loading ${modulePath}: ${error}`)
+        log.error(`Failed to load ${modulePath}: ${error}`)
         debug("Failed to load module", error)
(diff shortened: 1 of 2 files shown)


packages/repl/src/utils/load-blitz.ts
```diff
@@ -54,6 +55,7 @@ export const loadBlitz = async () => {
           [name]: contextObj,
         }
       } catch (error) {
+        log.error(`Failed loading ${modulePath}: ${error}`)
```
beerose, line 58:
```suggestion
        log.error(`Failed to load ${modulePath}: ${error}`)
```

beerose (approved):
Thanks for your PR!

Issue #2324, opened 2021-05-10
Support async `initServer` hook


flybayer:
### What do you want and why?

Support async `experimental.initServer` hook so that the server is not started until the promise returned from initServer resolves.

Original discussion: https://github.com/blitz-js/blitz/pull/2256#discussion_r629454953

### Possible implementation(s)

Move to [this](https://github.com/blitz-js/blitz/blob/canary/nextjs/packages/next/next-server/server/next-server.ts#L255-L257) to [here](https://github.com/blitz-js/blitz/blob/canary/nextjs/packages/next/server/lib/start-server.ts#L12)

g3offrey:
I am working on that.

PR #2798, opened 2021-10-04
enable user to do async initServer

g3offrey:
Closes: #2324 

I did the solution proposed by @flybayer on the issue #2324 

To do so I had to load the config before the server is listening. Indeed today, the config loading is done at first on `prepare` method, which is run on the server listening.

The `loadConfig` method was private so I had to set it a public 😢
Another drawback of this is that the `loadConfig` is now done two times: before the listening and when we call `getServer` (at the listening).

I was thinking about two things doing that, moving the `initServer` logic inside a `NextServer` method (`onServerInit` ?), so I can set back `loadConfig` as private. 

Or maybe instead of calling `loadConfig`, I can call `getServer` (setting it public before so) which has a `loadConfig` inside of it. This has advantages of preparing the `serverPromise` per advance and won't do a duplicate call to `loadConfig`.

What do you think ? 🤔 

## Bug Checklist

## Feature Checklist


flybayer:
@g3offrey thank you so much!

Upon review, I *think* we could run initServer just before [this line in getServer](https://github.com/blitz-js/blitz/blob/canary/nextjs/packages/next/server/next.ts#L133). What do you think?

g3offrey:
@flybayer Don't worry, I'm glad to contribute to such a project 😄 
Sure, I think this is possible, however I have to do some modification to do so.

First, at the moment the `getServer` method is called first on `prepare`, so after listening. (This is not what we want)

With my modification, we can call it once in `getRequestHandler` instead of at every request (in request handler).
Also this method is called before the listening so it is perfect for our use-case.

```ts
 async getRequestHandler() {
    const requestHandler = await this.getServerRequestHandler()

    return (
      req: IncomingMessage,
      res: ServerResponse,
      parsedUrl?: UrlWithParsedQuery
    ) => {
      return requestHandler(req, res, parsedUrl)
    }
  }
```

As now `getRequestHandler` is simply an empty wrapper to `getServerRequestHandler`, I deleted it and simply called `getServerRequestHandler`.

With these modifications, I can implement your solution. What do you think? Do you prefer this way of doing things or the previous one? 😄 

Check out the last commit. I hope I am clear enough, that's not easy by text 😅

Issue #2793, opened 2021-10-03
Spinner from `ink-spinner` causes `blitz install` to fail


Talor-A:
### What is the problem?

blitz chakra-ui recipe seems to be broken. maybe a dependency update breaking part of the installer cli?

### Paste all your error logs here:

```
$ yarn blitz install chakra-ui

 # after hitting "enter" to continue......


ERROR  Cannot read property 'dots' of undefined

 .blitz/recipe-install/node_modules/ink-spinner/build/index.js:12:43

  9:  */
 10: const Spinner = ({ type = 'dots' }) => {
 11:     const [frame, setFrame] = react_1.useState(0);
 12:     const spinner = cli_spinners_1.default[type];
 13:     react_1.useEffect(() => {
 14:         const timer = setInterval(() => {
 15:             setFrame(previousFrame => {

 - Spinner (.blitz/recipe-install/node_modules/ink-spinner/build/index.js:12:43)
 -renderWithHoo (.blitz/recipe-install/node_modules/react-reconciler/cjs/react-reconc
  s            iler.development.js:6036:18)
 -mountIndeterminateCo (.blitz/recipe-install/node_modules/react-reconciler/cjs/react
  ponent              -reconciler.development.js:8570:13)
 -beginWork$ (.blitz/recipe-install/node_modules/react-reconciler/cjs/react-reconcile
            r.development.js:9938:16)
 -Object.invokeGuardedCal (.blitz/recipe-install/node_modules/react-reconciler/cjs/re
  backImpl               act-reconciler.development.js:11563:10)
 -invokeGuardedCall (.blitz/recipe-install/node_modules/react-reconciler/cjs/react-re
  ack              conciler.development.js:11740:31)
 -beginWork$$ (.blitz/recipe-install/node_modules/react-reconciler/cjs/react-reconcil
  1          er.development.js:15778:7)
 -performUnitOfW (.blitz/recipe-install/node_modules/react-reconciler/cjs/react-recon
  rk            ciler.development.js:14696:12)
 -workLoopSyn (.blitz/recipe-install/node_modules/react-reconciler/cjs/react-reconcil
  c          er.development.js:14669:22)
 -performSyncWorkOn (.blitz/recipe-install/node_modules/react-reconciler/cjs/react-re
  oot              conciler.development.js:14265:11)

✕ TypeError: Cannot read property 'dots' of undefined
✨  Done in 59.73s.
```

### Paste all relevant code snippets here:

```
[...]

PR #2799, opened 2021-10-05
Disable ink-spinner as a temporary measure

9j:
#2793

(This is a temporary fix for this issue, so I just linked the issue. )

### What are the changes and their implications?

Temporarily fix ink-spinner's dependency issue on cli-spinner
I know this is not a good thing
However, it quickly restores important features. 

## Bug Checklist

- [ ] Integration test added (see [test docs](https://blitzjs.com/docs/contributing#running-tests) if needed)

## Feature Checklist

- [ ] Integration test added (see [test docs](https://blitzjs.com/docs/contributing#running-tests) if needed)
- [ ] Documentation added/updated (submit PR to [blitzjs.com repo](https://github.com/blitz-js/blitzjs.com))

PR #2801, opened 2021-10-05
Replace spinner 2793

johnmurphy01:
Closes: #2793

- Removed `ink-spinner` dependencies
- Replaced `ink-spinner` with `log.spinner` from `@blitzjs`

NOTE: I don't know how to test a recipe locally, so this hasn't been tested. I did run the test scripts however.

## Bug Checklist

## Feature Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,7 +106,6 @@
     "@types/get-port": "4.2.0",
     "@types/gulp-if": "0.0.33",
     "@types/htmlescape": "^1.1.1",
-    "@types/ink-spinner": "3.0.0",
     "@types/jest": "26.0.20",
     "@types/jsonwebtoken": "8.5.0",
     "@types/lodash": "4.14.149",
--- a/packages/installer/package.json
+++ b/packages/installer/package.json
@@ -42,7 +42,6 @@
     "fs-extra": "^9.1.0",
     "globby": "11.0.2",
     "ink": "3.0.8",
-    "ink-spinner": "4.0.1",
     "ink-testing-library": "2.1.0",
     "jscodeshift": "0.11.0",
     "recast": "0.20.4"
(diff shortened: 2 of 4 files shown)


beerose:
As for the testing, this is what you can do (note: I'm not sure that this is the best way, but it works for me):
1. in root, run `yarn build` to build packages
2. go to `packages/installer` and run `yarn pack` to get a tarball with the installer package
3. go to, for example, `recipes/render` and install it via `yarn add ../../packages/installer/<PATH TO .tar FILE>`.
4. you can then install the recipe, by running `blitz install <path to blitz repo>/recipes/render/index.ts`.

beerose (commented):
I'm afraid it won't work because `log.spinner` logs messages to the console, and you use it as a component. 
Here you can find an example usage: https://github.com/blitz-js/blitz/blob/canary/packages/cli/src/commands/install.ts#L238.
I think replacing `Spinner` with `log.spinner` would requires changing the way we handle notifications in installer package — it probably means removing usages of Ink components.

9j:
> I'm afraid it won't work because `log.spinner` logs messages to the console, and you use it as a component. Here you can find an example usage: https://github.com/blitz-js/blitz/blob/canary/packages/cli/src/commands/install.ts#L238. I think replacing `Spinner` with `log.spinner` would requires changing the way we handle notifications in installer package — it probably means removing usages of Ink components.

That's why I made a temporary plan. #2799
It's being used in too many places in this project to take away the ink
If you want to remove the ink, you need to modify the files that are combined with those two.
It takes an insignificant amount of time compared to the impact.

beerose:
> > I think replacing `Spinner` with `log.spinner` would requires changing the way we handle notifications in installer package — it probably means removing usages of Ink components.
> 
> That's why I made a temporary plan. #2799 It's being used in too many places in this project to take away the ink If you want to remove the ink, you need to modify the files that are combined with those two. Compared to impact, it takes a lot of time.

@9j ee will merge your fix as a temporary workaround, and then we merge @johnmurphy01's PR when it's ready :)

johnmurphy01:
Cool, thanks for the testing steps. This will help.

beerose:
I'm closing this PR — if you or anyone else wants to work on it in the future, feel free to open a new one!

Issue #2800, opened 2021-10-05
`blitz generate` should check if model name is passed correctly


beerose:
### What is the problem?

When you forget to pass a model name in `blitz generate` command (for example: `blitz g all name:string age:int`), it assumes that `name:string` is the model and proceeds to generate invalid code. It would be nice if it was caught before generating the files. We could prompt an error that the model name is not correct.

Related code: https://github.com/blitz-js/blitz/blob/canary/packages/cli/src/commands/generate.ts#L82

Slightly related issue: https://github.com/blitz-js/blitz/issues/2380

### Paste all your error logs here:

```
CREATE    app/name:strings/queries/getName:string.ts
CREATE    app/name:strings/queries/getName:strings.ts
> ⚠️  Failed trying to run prettier:SyntaxError: Unexpected token, expected ";" (14:38)
  12 |   async (input) => {
  13 |     // TODO: in multi-tenant app, you must add validation to ensure correct tenant
> 14 |     const name:string = await db.name:string.create({data: input})
     |                                      ^
  15 | 
  16 |     return name:string
  17 |   },
> ⚠️  Failed trying to run prettier:SyntaxError: Unexpected token, expected ";" (15:38)
  13 |   async ({id}) => {
  14 |     // TODO: in multi-tenant app, you must add validation to ensure correct tenant
> 15 |     const name:string = await db.name:string.deleteMany({where: {id}})
     |                                      ^
  16 | 
  17 |     return name:string
  18 |   },
> ⚠️  Failed trying to run prettier:SyntaxError: Unexpected token, expected ";" (15:38)
  13 |   async ({id, ... data}) => {
  14 |     // TODO: in multi-tenant app, you must add validation to ensure correct tenant
> 15 |     const name:string = await db.name:string.update({where: {id}, data})
     |                                      ^
  16 | 
  17 |     return name:string
  18 |   },
CREATE    app/name:strings/mutations/createName:string.ts
CREATE    app/name:strings/mutations/deleteName:string.ts
CREATE    app/name:strings/mutations/updateName:string.ts
 
✔ Model 'Name:string' created in schema.prisma:

> 
> model Name:string {
>   id        Int      @id @default(autoincrement())
>   createdAt DateTime @default(now())
>   updatedAt DateTime @updatedAt
>   age       Int
> }
> 
 
✔ Run 'prisma migrate dev' to update your database? (Y/n) · false
 
✨  Done in 72.21s.
```

### Paste all relevant code snippets here:

[...]

PR #2802, opened 2021-10-05
add err when model name not passed to cli generate

martinsaxa:
Closes: #2800

Blitz-cli generate options will check if model name contains `:`. If it does, it will throw an error. Usually this happens when someone forget to pass model name e.g. `blitz generate all name:string`

## Bug Checklist

## Feature Checklist

### DIFF
--- a/packages/cli/src/commands/generate.ts
+++ b/packages/cli/src/commands/generate.ts
@@ -203,6 +203,13 @@ export class Generate extends Command {
       }
     }
 
+    if (modelName.includes(":")) {
+      console.warn(
+        "Model name has not been passed. Use blitz generate [type] [modelName] [field:type]...",
+      )
+      this.exit(0)
+    }
+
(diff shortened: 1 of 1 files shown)


beerose (changes requested):
Thanks for the contribution!

Instead of using `console.warn`, we should use Blitz's `log` function `log.error(...)`.  It's already imported into this file.

Apart from that, we should check if the model name contains only valid characters instead of checking for particular invalid symbols, e.g. `:` inside of the model name is only one of the invalid characters. Model names should satisfy the following expression: `[A-Za-z][A-Za-z0-9_]*`.

martinsaxa:
Thanks for review, I will look into these issues.

martinsaxa:
I just looked at that PR and function and I thing you have an great idea. I'm going to work on
this weekend.

PR #2823, opened 2021-10-08
[legacy-framework] Improve error message on a duplicate page name error

p-siriphanthong:
Closes: blitz-js/legacy-framework#232

### DIFF
--- a/examples/cypress/package.json
+++ b/examples/cypress/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@examples/cypress",
-  "version": "0.41.0",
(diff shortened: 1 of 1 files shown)


flybayer:
> I don't think it fixes the issue — it doesn't close the process, but it's stuck so the user needs to restart the dev server. Maybe we can fix it in a separate PR as it seems to be more complex?

@beerose 🤔 it is working for me, I think this was the correct solution.

I've pushed a commit with a few UX tweaks including not throwing error in development.

Good to merge from my end.

PR #2844, opened 2021-10-15
[legacy-framework] `createMuiTheme` deprecation update

aleksanderbrymora:
The `createMuiTheme` has been deprecated and a simple rename of the import is needed to fix that. `createTheme` is the new name used by Mui.

No implications, it's a simple rename of the import and a function. There is not much to add.
Here is some info related to the change:
- https://github.com/mui-org/material-ui/pull/25992
- https://github.com/mui-org/material-ui/pull/26004


aleksanderbrymora:
Already fixed what mentioned in this pull request and more in #2853. Closing due to overlap

PR #2875, opened 2021-10-24
Add e2e cli tests setup

Redaloukil:
Closes: https://github.com/blitz-js/legacy-framework/issues/37

## Bug Checklist

## Feature Checklist


packages/cli/test/e2e/cypress/integration/build.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  cy.readFile("../blitz/package.json").then((packageJson) => {
+    cy.exec(
+      `cd ${appName} && ${packageManager} add blitz@${packageJson.dependencies["@blitzjs/cli"]} --save`,
+    )
+    cy.readFile(`./${appName}/package.json`).then((testAppPackageJson) => {
+      expect(testAppPackageJson.dependencies["blitz"]).eqls(
+        packageJson.dependencies["@blitzjs/cli"],
+      )
+    })
```
flybayer, line 35: I have some doubts about whether this is doing what we want it too. Ultimately what we want is that none of the `blitz`, `@blitzjs/*`, or `next` packages should be inside the `<test-app-dir>/node_modules/` folder. All of these packages are already linked in `<monorepo-root>/node_modules`. So if those packages are not in the test app node_modules, they will automatically be resolved to the monorepo root.

We need to carefully confirm this. I suspect what's happening with the current code is that it's installing the given version from NPM, which is not what we want.

packages/cli/test/e2e/cypress/integration/build.ts
```diff
@@ -0,0 +1,65 @@
...
+}
+
+describe("blitz new command", () => {
+  beforeEach(() => {
+    cy.exec(`rimraf ${TEST_APP_NAME}`)
+  })
+
+  afterEach(() => {
+    cy.exec(`rimraf ${TEST_APP_NAME}`)
+  })
```
flybayer, line 48: It's common practice to not remove the test after it finished. That way you can easily inspect the test files after the test finishes.

packages/cli/test/e2e/cypress/integration/build.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  it("Should create a new app", () => {
+    buildAppCommand(TEST_APP_NAME, "yarn", "full", "react-hook-form", "typescript")
+    cy.writeFile(
+      `./${TEST_APP_NAME}/.env`,
+      'DATABASE_URL="file:./db.sqlite\nSESSION_SECRET_KEY="ThisIsAKeyForTestingProductionBuildStart"',
+      {encoding: "utf8", flag: "a+"},
+    )
+    cy.exec(`cd ${TEST_APP_NAME} && yarn add pm2`)
+    cy.exec(`cd ${TEST_APP_NAME} && pm2 --name testBuild start yarn -- start`)
```
flybayer, line 58: why is pm2 being used? 

A test command of `blitz build && blitz start` should be sufficient, right?

flybayer (commented):
Thanks for your work on this!! 

Some comments below.

Also, this initial test is good, but at some point we also want to run the commands from the new app getting started page `blitz generate` etc. Can add that in this PR or we can open an issue for someone else to add

packages/cli/test/e2e/cypress/integration/build.ts
Redaloukil: The main reason behind installing pm2, is the fact that `blitz start` does not run as a detached process from the command line. why I need to run a detached process ? to pass the step of starting the built server app and continue my tests down bellow. 

Cypress documentation gives an indication that running the server inside cypress is considered as an anti-pattern [see here](https://docs.cypress.io/guides/references/best-practices#Web-Servers). 

I think I took a wrong path imo, generating, building and starting  a new generated app could be done separately from outside cypress via a script or something. After the project is built and ready to go, we `yarn run start` the server,wait for the server till it starts and  after that run cypress to do the e2e test like singup/login..

packages/cli/test/e2e/cypress/integration/build.ts
Redaloukil: totally agree

packages/cli/test/e2e/cypress/integration/build.ts
Redaloukil: It's installing the version from npm indeed. my attention behind this was to follow according to steps that you mentioned, I probably miss understood what you meant by : 
- get blitz version from packages/blitz/package.json version field
- set blitz version in the new test app package.json to match the version from last step. ( i'm installing the corresponding blitz version in this step )

flybayer: Ok yeah. Maybe another approach is to use the `file:...` syntax and point it to the blitz package?

Redaloukil: > Ok yeah. Maybe another approach is to use the `file:...` syntax and point it to the blitz package?

Not sure I get totally what you mean, as I understood this was made because we want blitz version in the repo to be the same as the generated app that we are testing, since the app is generated using `packages/blitz`, should I assume that the generated app is with same version as `packages/blitz` ? 

And what do you mean exactly by `file:...`

beerose: I think what Brandon meant is local paths (more here: https://docs.npmjs.com/cli/v7/configuring-npm/package-json#local-paths), so having something like:
```
{
  "name": "tests",
  "dependencies": {
    "blitz": "file:../../packages/blitz"
  }
}
```
You can run `npm i <path to package>` and it will add it in the above way.

packages/cli/test/e2e/cypress/integration/helpers/blitz-cli-version.js
```diff
@@ -0,0 +1,6 @@
+const path = require("path")
```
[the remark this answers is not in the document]
crazyguero: Oscar

Issue #2840, opened 2021-10-13
Environment Variable Loading


beerose:
### Discussed in https://github.com/blitz-js/blitz/discussions/2444

<div type='discussions-op-text'>

<sup>Originally posted by **flybayer** June  3, 2021</sup>
## Problem

Blitz/Next use basically [dotenv-flow](https://github.com/kerimdzhanov/dotenv-flow) to load env var files like `.env`, `.env.test`, `.env.development.local`, etc. 

That behavior relies on setting NODE_ENV to the appropriate value before invoking a CLI command.

Problems:

1. Nextjs internals don't allow changing NODE_ENV. The `build` command must use `NODE_ENV=production` and `dev` must use `NODE_ENV=development`.
2. There's a [new Nextjs bug](https://github.com/vercel/next.js/issues/25491) that breaks previous usage of `NODE_ENV=test blitz build`
3. Now it's impossible to use `NODE_ENV` inside your app code to conditionally do things based on different environments.

## Proposed Solution

Use `APP_ENV` to determine app environment instead of `NODE_ENV`. `NODE_ENV` will only be used for `blitz dev` vs `blitz build`.

The value of `APP_ENV` will determine which `.env.[ENV].*` files will be loaded.

1. Defaults:
   1. `APP_ENV=development` for `blitz dev` 
   2. `APP_ENV=production` for `blitz build` and `blitz start`
   3. `APP_ENV=test` for Jest tests
2. Ways to manually specify APP_ENV:
   1. Set it before invoking the CLI like `APP_ENV=test blitz start`
   2. Set it with a new `-e` flag like `blitz start -e test`

With this solution, you can set `APP_ENV` to any string. Example: `APP_ENV=staging` which would load `.env.staging`

---

Any and all feedback is welcome!</div>

fleck:
Nice, just spent the last hour trying to figure out why NODE_ENV was getting set to production when I thought it should be test!

PR #2878, opened 2021-10-25
Load env variables based on `APP_ENV` or `-e` flag

beerose:
Closes: https://github.com/blitz-js/blitz/issues/2840

## Bug Checklist

## Feature Checklist

- [x] Documentation added/updated (submit PR to [blitzjs.com repo `canary` branch](https://github.com/blitz-js/blitzjs.com/tree/canary)): https://github.com/blitz-js/blitzjs.com/pull/617


beerose:
@flybayer I added the ability to set app env to `dev, start, build, export`. As for the other commands:
- `new` doesn't load anything from `.env` (or did I miss something)?
- `db` is deprectaed
- `prisma` — loads env variables inside of the `prisma` library, so that would need more changes right?
- `generate, codegen, install, routes` — doesn't need it?
- `console` — from what I see it doesn't load env variables, is that right?

flybayer:
Ok this is looking good, but need a few more changes.

1. Set `APP_ENV` if `-e` present in [next cli entrypoint](https://github.com/blitz-js/blitz/blob/canary/nextjs/packages/next/bin/next.ts).
2. Replace every call to `dotenv-flow` with a call to the next-env package
3. Remove `dotenv-flow` dependency

Prisma loads `.env` but none of the other files. So we previously called `dotenv-flow` which loaded the other files. Now we are replacing that with next-env.

nextjs/packages/next-env/package.json
```diff
@@ -1,7 +1,6 @@
 {
-  "private": true,
   "name": "@next/env",
-  "version": "11.1.0",
+  "version": "0.42.0",
```
beerose, line 3: update to latest

beerose (commented):
* Add integration tests
* Test locally the next package
* Brandon has to do a release with new `@blitzjs/env` package
* `loadEnvConfig` should default to `process.cwd()`

nextjs/packages/next-env/index.ts
```diff
@@ -78,8 +78,8 @@ export function loadEnvConfig(
   if (combinedEnv) return { combinedEnv, loadedEnvFiles: cachedLoadedEnvFiles }
 
   const isTest = process.env.NODE_ENV === 'test'
-  const mode = isTest ? 'test' : dev ? 'development' : 'production'
-  const dotenvFiles = [
+  let mode = isTest ? 'test' : dev ? 'development' : 'production'
```
beerose, line 81:
```suggestion
  const mode = isTest ? 'test' : dev ? 'development' : 'production'
```

nextjs/test/integration/env-config-app-env/app/next.config.js
```diff
@@ -0,0 +1,17 @@
+module.exports = {
+  cleanDistDir: false,
+  target: 'experimental-serverless-trace',
```
flybayer, line 3: This line shouldn't be here

nextjs/test/integration/env-config-app-env/test/index.test.js
```diff
@@ -0,0 +1,163 @@
...
+      '.env.staging.local',
+      '.env.production',
+    ].map((file) => join(appDir, file))
+
+    beforeAll(async () => {
+      nextConfigContent = await fs.readFile(nextConfigPath, 'utf8')
+      await fs.writeFile(
+        nextConfigPath,
+        nextConfigContent.replace(
+          '// update me',
```
flybayer, line 128: use this instead in config by default

flybayer (approved):

Issue #2880, opened 2021-10-26
Propagate `prisma generate` errors when using `blitz codegen`


beerose:
### What is the problem?

Currently, if a user runs `blitz codegen`, and there there is Prisma installed, Blitz also runs `primsa generate`. If there are any errors during the `prisma generate` step, they aren't displayed in the terminal even when using `DEBUG=blitz:*`. This way some useful information returned from Prisma is omitted. 

Relevant code: https://github.com/blitz-js/blitz/blob/canary/packages/cli/src/commands/codegen.ts#L32

### Paste all your error logs here:

`blitz codegen` output:

```
✔ Generating route manifest
✖ Generating Prisma client
```

While `prisma generate` returns the following:

```
Error: 
You don't have any models defined in your schema.prisma, so nothing will be generated.
You can define a model like this:

model User {
  id    Int     @id @default(autoincrement())
  email String  @unique
  name  String?
}

More information in our documentation:
https://pris.ly/d/prisma-schema
```

### Paste all relevant code snippets here:

n/a

### What are detailed steps to reproduce this?

1. Try to run `blitz codegen` without any models in `schema.prisma` (or `schema.prisma` containing some errors). 

### Run `blitz -v` and paste the output here:

```
macOS Big Sur | darwin-x64 | Node: v14.17.6

blitz: 0.41.1 (local)

  Package manager: yarn 
  System:
    OS: macOS 11.1
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 73.24 MB / 16.00 GB
    Shell: 5.0.11 - /usr/local/bin/bash
  Binaries:
    Node: 14.17.6 - /var/folders/xt/37ypr1bj3s565bjrdzx3znd00000gn/T/yarn--1635260280228-0.10240775555445136/node
    Yarn: 1.19.2 - /var/folders/xt/37ypr1bj3s565bjrdzx3znd00000gn/T/yarn--1635260280228-0.10240775555445136/yarn
    npm: 8.1.0 - ~/workspace/emojis/emojis-lib/bin/npm
    Watchman: Not Found
  npmPackages:
    @prisma/client: ^3.3.0 => 3.3.0 
    blitz: ^0.41.1 => 0.41.1 
    prisma: ^3.3.0 => 3.3.0 
    react: 17.0.2 => 17.0.2 
    react-dom: 17.0.2 => 17.0.2 
    typescript: 4.4.4 => 4.4.4 
```

PR #2891, opened 2021-10-29
Show the error message from `prisma generate` when `blitz codegen` fails

lucasvazq:
Closes: #2880 

When the `blitz codegen` command fails for prisma, the cause of the error is not displayed. This PR changes that, showing the cause of the problem.

- [x] I replaced the return type of a function, so I updated all the places where it is used.


packages/cli/src/commands/prisma.ts
```diff
@@ -23,18 +24,29 @@ export const runPrisma = async (args: string[], silent = false) => {
   }
 
   const cp = require("cross-spawn").spawn(prismaBin, args, {
-    stdio: silent ? "ignore" : "inherit",
+    stdio: silent ? "pipe" : "inherit",
```
lucasvazq, line 27: This function takes one parameter named `silent`.

Before the changes, this parameter suppresses the stderr and stdout at all. Also, the function returns a boolean that represents the status of the execution.

Now, when `silent` is `true`, the stderr is captured and saved, and the function returns an object with two keys:
\- success: This represents the execution status (the original return of the function).
\- stderr: the captured stderr, that only applies when `silent` is `true`.

packages/cli/src/commands/codegen.ts
```diff
@@ -30,8 +30,10 @@ export class CodeGen extends Command {
 
       if (hasPrisma && hasPrismaSchema) {
         let prismaSpinner = log.spinner(`Generating Prisma client`).start()
-        void runPrisma(["generate"], true).then((success) =>
-          success ? prismaSpinner.succeed() : prismaSpinner.fail(),
+        void runPrisma(["generate"], true).then((result) =>
+          result.success
+            ? prismaSpinner.succeed()
+            : prismaSpinner.fail() && console.log("\n" + result.stderr)
```
lucasvazq, line 36: Read first: [packages/cli/src/commands/prisma.ts:27](https://github.com/blitz-js/blitz/pull/2891/files#diff-28ae2cc5cebe012f80c6c602a99837c65b147f773ca87d1b029691e11b9b98c7R27)

When prisma fails, we take the captured stderr and print them using `console.log`.
Why `console.log` and not `log.*` alternatives?
Because `console.log` keeps the stderr generated by prisma and I think it's very well formated.

Here is a comparison of different output formats: [view comparison](https://user-images.githubusercontent.com/38964964/139505247-3ce439f3-2bbc-4584-828b-e423e395cc05.png)

The final result: 
Screenshot_2021-10-29_17-18-46

beerose: Wow! Thanks so much for evaluating all the options!

beerose, line 36: It works okay, but I think we can slightly improve it — since there was an error, we should exit with code 1. What about this:

```
          if (result.success) {
            prismaSpinner.succeed()
          } else {
            prismaSpinner.fail()
            console.log("\n" + result.stderr)
            process.exit(1)
          }
```

beerose (commented):
Thanks so much for this PR! It's amazing how you described in detail your implementation decisions!

Left one comment.

beerose (approved):

Issue #2904, opened 2021-11-03
Update graphql-apollo-server and run prisma generate


maxall41:
### What is the problem?

after running `blitz install graphql-apollo-server` and then going to `http://localhost:3000/api/graphql` i get `Internal Server Error` on the page and in the terminal i get the logs below.

### Paste all your error logs here:

```
TypeError: Cannot read properties of undefined (reading '$name')
    at eval (webpack-internal:///./app/graphql/types/user-type.ts:12:56)
    at Object../app/graphql/types/user-type.ts (/Users/maxcampbell/Desktop/app/.next/server/pages/api/graphql.js:52:1)
    at __webpack_require__ (/Users/maxcampbell/Desktop/app/.next/server/webpack-runtime.js:33:42)
    at eval (webpack-internal:///./app/graphql/types/query-type.ts:8:68)
    at Object../app/graphql/types/query-type.ts (/Users/maxcampbell/Desktop/app/.next/server/pages/api/graphql.js:42:1)
    at __webpack_require__ (/Users/maxcampbell/Desktop/app/.next/server/webpack-runtime.js:33:42)
    at eval (webpack-internal:///./app/graphql/types/index.ts:2:69)
    at Object../app/graphql/types/index.ts (/Users/maxcampbell/Desktop/app/.next/server/pages/api/graphql.js:32:1)
    at __webpack_require__ (/Users/maxcampbell/Desktop/app/.next/server/webpack-runtime.js:33:42)
    at eval (webpack-internal:///./app/graphql/api/graphql.ts:14:64)
```

### Paste all relevant code snippets here:

Just the default graphQL stuff

### What are detailed steps to reproduce this?

1. create new blitz project
2. run `blitz install graphql-apollo-server`
3. go to `http://localhost:3000/api/graphql` 

### Run `blitz -v` and paste the output here:

```
macOS Big Sur | darwin-arm64 | Node: v16.11.0

blitz: 0.41.1 (global)
blitz: 0.41.1 (local)

  Package manager: yarn 
  System:
    OS: macOS 11.5.2
    CPU: (8) arm64 Apple M1
    Memory: 82.70 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.11.0 - ~/.nvm/versions/node/v16.11.0/bin/node
    Yarn: 1.22.17 - /opt/homebrew/bin/yarn
[...]

PR #2905, opened 2021-11-03
Add info about a required manual step to graphql-apollo-server recipe

noxify:
Closes: #2904 

Added a new step to the recipe which informs the user to run the `blitz prisma generate` command.

## Bug Checklist

## Feature Checklist


beerose (commented):
Thanks for the contribution. 
However, I'm not sure if that's a good approach, because if a user keeps pressing enter, they won't see the message as it disappears after the installation. I think, we can instead add it to the Readme for now

flybayer:
Let's keep this addition and also add it to the recipe README

flybayer:
Actually, the best solution is to run `blitz prisma migrate` automatically. Do we have a recipe step for running arbitrary node code? 🤔

noxify:
I have added the manual step also to the readme.

> Do we have a recipe step for running arbitrary node code?

Currently, there is no step available in the `RecipeBuilder` to do this.

https://github.com/blitz-js/blitz/blob/457015ae5b300753dc64e1032fae4e7c8ab133ee/packages/installer/src/recipe-builder.ts#L8

Sounds like a new feature request for me :)

beerose:
While `runBlitzCommand` is a nice improvement, I think `runCommand` can be useful so that recipes creators have full flexibility. E.g. let's say we have a new recipe adding some library, and as one of the setup steps, users need to run `<new library> init` or something similar. Having `runCommand` and running `<new library> init` inside of the recipe could be nice. As for the `rm -rf` — someone should spot it in the review, and this code should never be merged.

recipes/graphql-apollo-server/index.ts
```diff
@@ -39,4 +39,10 @@ export default RecipeBuilder()
         assignments: [{type: "assignment", key: "provider", value: '"nexus-prisma"'}],
       }),
   })
+  .printMessage({
+    stepId: "manualStep",
+    stepName: "Informs the user about a manual step",
+    explaination: "After the graphql recipe has been installed successfully, the user has to run `blitz prisma generate` manually.",
+    message: "Please run `blitz prisma generate` in your console/terminal to generate the graphql schema."
+  })
```
beerose, line 47:
```suggestion
  .printMessage({
    stepId: "manualStep",
    stepName: "Informs the user about a manual step",
    message:
      "Please run `blitz prisma generate` in your console/terminal to generate the graphql schema.",
  })
```

beerose:
I'm going to merge this PR, and when we have #2941 implemented, we can add it to this recipe.

PR #2920, opened 2021-11-06
ci(workflow): add 'yarn' cache to setup-node workflows and remove unn…

oscard0m:
Context: https://github.com/blitz-js/blitz/pull/2837

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,19 +20,7 @@ jobs:
         uses: actions/setup-node@v2
         with:
           node-version: "14"
-      - name: Get yarn cache directory path
-        id: yarn-cache-dir-path
-        run: echo "::set-output name=dir::$(yarn cache dir)"
-      - name: Cache node_modules
-        id: yarn-cache
-        uses: actions/cache@v2
-        with:
-          path: |
-            ${{ steps.yarn-cache-dir-path.outputs.dir }}
-            **/node_modules
-          key: ${{ runner.os }}-${{ runner.node_version}}-yarn-v13-${{ hashFiles('yarn.lock') }}
-          restore-keys: |
-            ${{ runner.os }}-${{ runner.node_version}}-yarn-v13-
+          cache: yarn
       - name: Install dependencies
         run: yarn install --frozen-lockfile --silent
         env:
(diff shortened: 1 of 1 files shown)


oscard0m:
Applied changes required in https://github.com/blitz-js/blitz/pull/2837#issuecomment-942055008

Regarding the improvements, I would not expect any further improvements since `actions/setup-node` with `cache` option uses `actions/cache` internally. I would say this is an abstraction which simplifies workflow's code but it should stay the same.

Resources:
- Here you can read more about `setup-node`'s team goals when added this feature: https://github.com/actions/setup-node/issues/271
- Here you can read an explanation on how `setup-node's` cache option is working: https://github.com/actions/setup-node#caching-packages-dependencies

flybayer:
Thanks! This change looks good, except I'm not sure it it's working or configured properly.

As you can see in the screenshot here from the CI run, the cache is being restored **after** yarn install. Know how to fix?

image

oscard0m:
> Thanks! This change looks good, except I'm not sure it it's working or configured properly.
> 
> As you can see in the screenshot here from the CI run, the cache is being restored **after** yarn install. Know how to fix?
> 
> image

Good point. I didn't notice there were jobs depending on `build-linux` job so, in my understanding, we can get rid of the step `actions/cache` in those.

.github/workflows/main.yml
```diff
@@ -280,12 +201,6 @@ jobs:
       matrix:
         group: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20]
     steps:
-      - uses: actions/cache@v2
-        id: restore-build
-        with:
-          path: ./*
-          key: ${{ runner.os }}-${{ github.sha }}
```
flybayer, line 287: we can't remove this section from the jobs because it restores the actual built packages which are needed for the tests

oscard0m: Good catch @flybayer! The moment you need to use `actions/cache` for other jobs (like the one you pointed for the `test` jobs) I think it makes the migration to `actions/setup-node`'s `cache` option not worth it. `setup-node` uses `actions/cache` internally when enabling `cache` option.

Unless you see a clean way to use `cache` option in `setup-node` and test jobs can use packages from that cache folder (the default one), we should close this issue.

What do you think @flybayer

beerose: > Unless you see a clean way to use cache option in setup-node and test jobs can use packages from that cache folder (the default one), we should close this issue.

Yeah, I think that makes sense to close it. Thanks for your effort though, we appreciate you looking into this!

beerose:
Closing because of https://github.com/blitz-js/blitz/pull/2920#discussion_r758320999

Issue #2519, opened 2021-06-20
Allow RouteURLObject in GetServerSideProps redirect


remjx:
### What do you want and why?

[Currently](https://blitzjs.com/docs/redirects#on-the-server), the expected type of GetServerSideProps' `redirect.destination` is `string`. 

It would be nice if we could pass one of our Route Manifest objects into this field.

### Possible implementation(s)

- ~Add a `toString()` method to the route manifest~ (will be covered by https://github.com/blitz-js/blitz/issues/2829)
- Update type of `redirect.destination` to handle RouteURLObject

### Additional context

N/A

flybayer:
Great idea! Let's modify redirect.destination type

To implement we need to modify the following and maybe more:
- https://github.com/blitz-js/blitz/blob/canary/nextjs/packages/next/types/index.d.ts#L45
- https://github.com/blitz-js/blitz/blob/canary/nextjs/packages/next/next-server/server/next-server.ts#L1527
- https://github.com/blitz-js/blitz/blob/canary/nextjs/packages/next/build/webpack/loaders/next-serverless-loader/page-handler.ts#L367
- https://github.com/blitz-js/blitz/blob/canary/nextjs/packages/next/next-server/server/render.tsx#L324

esemeniuc:
This is very much requested as well! I am doing manual string concatenation as a workaround

PR #2932, opened 2021-11-09
Support RouteUrlObject as redirect.destination in GetServerSideProps/GetStaticProps

s-r-x:
basically it just converts RouteUrlObject to string, using resolveHref function from next/router module. resolveHref requires next router, but it only uses pathname and aspath props from one. so i created a server version, which mocks router as  {pathname: '/', aspath: '/'}. this version always passes resolveAs: true to resolveHref, in order to normalize routes like /blog/post/[id]?id=1 to /blog/post/1.
i ran the auth example from examples folder with new updates, and it seems to work fine, although some tests are definitely required

## Bug Checklist

## Feature Checklist


beerose:
Does this close this issue: https://github.com/blitz-js/blitz/issues/2519?

s-r-x:
@beerose oh sorry, i forgot to link it. yes, that's the issue this pr is aiming for.

nextjs/packages/next/shared/lib/router/router.ts
```diff
@@ -269,12 +269,17 @@ function omitParmsFromQuery(query: ParsedUrlQuery, params: string[]) {
...
+  return resolved[1] || resolved[0]
+}
+
 /**
  * Resolves a given hyperlink with a certain router state (basePath not included).
  * Preserves absolute urls.
  */
 export function resolveHref(
-  router: NextRouter,
+  router: Pick<NextRouter, 'asPath' | 'pathname'>,
```
beerose, line 282: Why was this change needed?

nextjs/packages/next/server/render.tsx
```diff
@@ -381,6 +382,14 @@ function checkRedirectValues(
   }
 }
 
+function normalizeRedirectValues(redirect: Redirect) {
+  const dest = redirect.destination
+  if (dest && typeof dest === 'object') {
+    const resolvedDest = resolveHrefOnServer(dest)
```
beerose, line 388: We can re-use `formatWithValidation` function here — https://github.com/blitz-js/blitz/blob/canary/nextjs/packages/next/shared/lib/utils.ts#L567
```suggestion
    const resolvedDest = formatWithValidation(dest)
```

nextjs/packages/next/shared/lib/router/router.ts
```diff
@@ -269,12 +269,17 @@ function omitParmsFromQuery(query: ParsedUrlQuery, params: string[]) {
   return filteredQuery
 }
 
+export function resolveHrefOnServer(href: Url): string {
```
beerose, line 272: Changes in this file can be reverted if we use `formatWithValidation` in `normalizeRedirectValues` function

beerose (changes requested):
Thanks for this PR! 🙏 And sorry for the delay with the review!

nextjs/packages/next/server/render.tsx
s-r-x: actually, using formatWithValidation was my first idea. it works, but with a catch! i'll capture a video later to demonstrate you

s-r-x: with formatWithValidation:
https://user-images.githubusercontent.com/41614937/142560305-21131556-e348-4430-8fc0-1ab1c76967f6.mp4

with resolveHrefOnServer(i don't really like the name though):
https://user-images.githubusercontent.com/41614937/142560313-589eafcf-1977-4d0e-89a4-c1859b59eebb.mp4

beerose: Okay, got it. Thanks for the recordings! 

@flybayer It looks like we actually need to use `resolveHref` function as we're probably interested in the `interpolatedAs` value: https://github.com/blitz-js/blitz/blob/canary/nextjs/packages/next/shared/lib/router/router.ts#L336

flybayer: @s-r-x ok makes sense. Is there any issue with hardcoding `{ pathname: '/', asPath: '/' }`? Is there any edge cases this could cause a problem for? It would be good to add a clear comment explaining this for future folks.

s-r-x: @flybayer, @beerose  i replaced that hacky router mock with real server router(basically i just remove that resolveHrefOnServer util and revert the type of resolveHref router from { pathname: string, asPath: string } back to NextRouter). final result is basically the same, but the solution is less hacky. and i think there is no need for additional comments anymore.

beerose:
Could you also add a mention of that to the docs? (to the `getStaticProps` and `getServerSideProps` sections)

beerose (commented):
~It looks like Nexjs test #1 fails consistently (`server-side only compilation errors`). It works locally though~

PR #2946, opened 2021-11-12
Add vanilla-extract recipe

JuanM04:
Added [vanilla-extract](https://vanilla-extract.style/) recipe


beerose:
I think there's a bug with wrapping the config with `createVanillaExtractPlugin` — looks like the session is undefined, and there are multiple errors when accessing fields of the session context, e.g.: `Cannot read property '$create' of undefined`.

Here's also a related issue: https://github.com/blitz-js/blitz/issues/2955, where wrapping the config causes the same problems.

beerose (approved):

Issue #2935, opened 2021-11-10
0.42.4 .blitz folder not resolved by @next/stdlib with pnpm


90dy:
### What is the problem?

When using blitz 0.42.4 (latest) with pnpm 6.16.0, .blitz folder is not resolved by @next/stdlib
It works with blitz 0.40.0

### Paste all your error logs here:

```
Uncaught TypeError: next_stdlib__WEBPACK_IMPORTED_MODULE_6__.Routes.SignupPage is not a function
    at UserInfo (index.tsx?39bb:38)
    at renderWithHooks (react-dom.development.js?bf59:15958)
    at mountIndeterminateComponent (react-dom.development.js?bf59:20783)
    at beginWork (react-dom.development.js?bf59:22270)
    at HTMLUnknownElement.callCallback (react-dom.development.js?bf59:4139)
    at Object.invokeGuardedCallbackDev (react-dom.development.js?bf59:4188)
    at invokeGuardedCallback (react-dom.development.js?bf59:4252)
    at beginWork$1 (react-dom.development.js?bf59:26991)
    at performUnitOfWork (react-dom.development.js?bf59:26160)
    at workLoopConcurrent (react-dom.development.js?bf59:26146)
    at renderRootConcurrent (react-dom.development.js?bf59:26108)
    at performConcurrentWorkOnRoot (react-dom.development.js?bf59:25382)
    at workLoop (scheduler.development.js?4f06:256)
    at flushWork (scheduler.development.js?4f06:229)
    at MessagePort.performWorkUntilDeadline (scheduler.development.js?4f06:523)
UserInfo @ index.tsx?39bb:38
renderWithHooks @ react-dom.development.js?bf59:15958
mountIndeterminateComponent @ react-dom.development.js?bf59:20783
beginWork @ react-dom.development.js?bf59:22270
callCallback @ react-dom.development.js?bf59:4139
invokeGuardedCallbackDev @ react-dom.development.js?bf59:4188
invokeGuardedCallback @ react-dom.development.js?bf59:4252
beginWork$1 @ react-dom.development.js?bf59:26991
performUnitOfWork @ react-dom.development.js?bf59:26160
workLoopConcurrent @ react-dom.development.js?bf59:26146
renderRootConcurrent @ react-dom.development.js?bf59:26108
performConcurrentWorkOnRoot @ react-dom.development.js?bf59:25382
workLoop @ scheduler.development.js?4f06:256
flushWork @ scheduler.development.js?4f06:229
[...]

PR #2963, opened 2021-11-16
Fix routes manifest location when using pnpm 

beerose:
Closes: https://github.com/blitz-js/blitz/issues/2935

## Bug Checklist

## Feature Checklist


packages/blitz/scripts/postinstall.js
```diff
@@ -21,6 +22,45 @@ try {
   // noop
 }
 
+async function findNodeModulesRoot(src) {
```
beerose, line 25: So that's not perfect, because I copy pasted this function from `routes.ts` with some changes like: removed types, used different imports, `isInBlitzMonorepo`. Open for suggestions how to make it better.

PR #2965, opened 2021-11-16
Fix typo: "serialiazable" should be "serializable"

tmcw:
This is just a simple spelling fix: the documentation has a non-word in it, serialiazable


beerose:
Thanks for the PR! That's a valid change, but it's next's docs. It looks like Next team fixed it in their repo: https://github.com/vercel/next.js/blob/canary/docs/api-routes/response-helpers.md, so this change will get pulled once we sync the latest Next. 

However, we do have this typo in Blitz docs 😅 You can submit a PR to this repo: https://github.com/blitz-js/blitzjs.com/edit/main/app/pages/docs/api-routes.mdx.

I'm closing this one then.

Issue #2941, opened 2021-11-12
[Feature] Allow to run additional commands in the recipe installer


noxify:
### What do you want and why?

> This feature request is based on comments in #2905 

@beerose explained it very good in the PR:

> While runBlitzCommand is a nice improvement, I think runCommand can be useful so that recipes creators have full flexibility. E.g. let's say we have a new recipe adding some library, and as one of the setup steps, users need to run <new library> init or something similar. Having runCommand and running <new library> init inside of the recipe could be nice.

### Possible implementation(s)

I had in mind to implement only something like `runBlitzCommand` ( Example: `builder.runBlitzCommand('prisma migrate')` ).

But with the comment from @beerose, I think it's better to implement it more generic via `runCommand` ( Example: `builder.runCommand('blitz prisma migrate')` ).

This allows us to run whatever we want - We could also run something like `builder.runCommand('rm -rf /')` but @beerose  commented already to this:

> As for the rm -rf — someone should spot it in the review, and this code should never be merged.

### Additional context

I'm not sure if there are other ways to run a command from a script as with:
https://nodejs.org/api/child_process.html

beerose:
Thanks for creating this issue! We use [`cross-spawn`](https://www.npmjs.com/package/cross-spawn) package in other places, so that's something we can use here as well.

noxify:
If it's not urgent, I can give it a try after my vacation ( 2 weeks ).

beerose:
Awesome!

noxify:
Reminder for myself: 

Docs to update: 
* https://blitzjs.com/docs/writing-recipes 
* https://github.com/blitz-js/blitzjs.com/blob/main/app/pages/docs/writing-recipes.mdx

PR #3006, opened 2021-11-29
Add runCommand as option to the recipe builder

noxify:
This PR adds a new option to the recipe builder which allows the installer to run additional commands.

Closes: #2941 

## Feature Checklist

- [x] Documentation added/updated (submit PR to [blitzjs.com repo `canary` branch](https://github.com/blitz-js/blitzjs.com/tree/canary))


packages/installer/src/executors/run-command-executor.tsx
```diff
@@ -0,0 +1,121 @@
...
+  React.useEffect(() => {
+    if (commandInstalled) {
+      handleChangeCommitted()
+    }
+  }, [commandInstalled, handleChangeCommitted])
+
+  const childProps: CommitChildProps = {
+    commandInstalled,
+    handleChangeCommitted,
+    step,
```
noxify, line 98: Not sure why, but currently the `step` definitions refers to `ExecutorConfig` instead of `Config`.
I checked the code from `add-dependency-executor.tsx`, but can't find a difference in the interface definition.

image

noxify:
@beerose / @flybayer - Spent some hours to create a first prototype.

Defined tests are running fine - Do you miss something? 

Maybe you could also take a look to the problem with the wrong used type ( see my review comment ).

If the implementation is not what you had in mind, please let me know and I will try my best to update it.

My todos: 
- [ ] I have to run some manual tests to make sure that the code is working also in real life :) 
- [ ] Docs are missing

packages/installer/src/executors/run-command-executor.tsx
beerose: Would removing `as Config` assertion from here: https://github.com/blitz-js/blitz/pull/3006/files/2d38191343de415a576bc00f35323bf1b40a837c..8947f0f1616ad19c93a86bed00cdd1951d36adc7#diff-7c110c48e211b614d1134a448d8200e1550607cee14aabee969f35d2e9e5cedeR79 make it work?

noxify: Unfortunately not - If I remove `as Config`, `step` refers to `ExecutorConfig`

beerose: Okay, got it. I guess you'll have to make an assertion for that (in line 98). It's not ideal, we could improve the types in the future.

JuanM04: I think that's intended. Try passing the `command` through `proposalData` instead of `step`

noxify: Hi,

thanks for the feedback.

> I think that's intended. Try passing the command through proposalData instead of step

To be honest, I have no glue how to do this 🙈 Maybe there is an example - but I wasn't able to find it, yet

I pushed some changes, which passes the command to execute as dedicated param.
Also I use now `step: step as Config`

Not sure if this is the best/correct way, but it solves the errors.

packages/installer/src/executors/run-command-executor.tsx
```diff
@@ -0,0 +1,123 @@
...
+import {Executor, ExecutorConfig, getExecutorArgument} from "./executor"
+
+export interface CliCommand {
+  command: string
+  commandArgs?: string[]
+}
+
+export interface Config extends ExecutorConfig {
+  command: CliCommand
+}
```
JuanM04, line 18: What about this way of passing the command?

```ts
export type CliCommand = string | [string, ...string[]]

export interface Config extends ExecutorConfig {
  command: CliCommand
}
```

And you can split the command into arguments with this regex:

```ts
export async function executeCommand(command: CliCommand) {
  // from https://stackoverflow.com/a/43766456/9950655
  const argsRegex = /("[^"\\]*(?:\\[\S\s][^"\\]*)*"|'[^'\\]*(?:\\[\S\s][^'\\]*)*'|\/[^/\\]*(?:\\[\S\s][^/\\]*)*\/[gimy]*(?=\s|$)|(?:\\\s|\S)+)/g
  const parsedCommand: string[] = Array.isArray(command) ? command : command.match(argsRegex) || []

  if (parsedCommand.length === 0) {
    throw new Error("The command is too short: " + JSON.stringify(command))
  }

  await new Promise((resolve) => {
    const cp = spawn(parsedCommand[0], parsedCommand.slice(1), {
      stdio: ["inherit", "pipe", "pipe"],
    })
    cp.on("exit", resolve)
  })
}
```

packages/installer/src/recipe-builder.ts
```diff
@@ -16,6 +17,7 @@ export interface IRecipeBuilder {
   addAddDependenciesStep(step: Omit<AddDependencyExecutor.Config, "stepType">): IRecipeBuilder
   addNewFilesStep(step: Omit<NewFileExecutor.Config, "stepType">): IRecipeBuilder
   addTransformFilesStep(step: Omit<TransformFileExecutor.Config, "stepType">): IRecipeBuilder
+  runCommand(step: Omit<RunCommandExecutor.Config, "stepType">): IRecipeBuilder
```
JuanM04, line 20: Rename to `addRunCommandStep`

JuanM04 (changes requested):

packages/installer/src/recipe-builder.ts
noxify: done

packages/installer/src/executors/run-command-executor.tsx
noxify: needed a minute to understand what the snippet is doing - but yeah, looks good to me.

I updated the code and the unit tests - seems to work as expected.

JuanM04: One last thing: the syntax of `[string, ...string[]]` is a trick for TypeScript so you don't get an empty array:

```ts
type NormalArray = string[]

const test1: NormalArray = [] // valid
const test2: NormalArray = ["ls"] // valid
const test3: NormalArray = ["ls", "-a"] // valid
const test4: NormalArray = ["ls", "-a", "-l"] // valid

type TrickArray = [string, ...string[]]

const test5: TrickArray = [] // error
const test6: TrickArray = ["ls"] // valid
const test7: TrickArray = ["ls", "-a"] // valid
const test8: TrickArray = ["ls", "-a", "-l"] // valid
```

So it isn't necessary to write the command as `["ls", ...["-a", "-l"]]`, just `["ls", "-a", "-l"]` will work

noxify: i have updated the test script to test valid commands and also the thrown error to ensure everything is working as expected.

packages/installer/src/executors/run-command-executor.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+export const type = "run-command"
+
+function Command({command, loading}: {command: CliCommand; loading: boolean}) {
+  return (
+    <Text>
+      {`   `}
+      {loading ? <Spinner /> : "✅"}
+      {` ${command}`}
```
JuanM04, line 31:
```suggestion
      {` ${typeof command === "string" ? command : command.join(' ')}`}
```

JuanM04:
All seem good, but it threw me an error when testing a full recipe:

```
 ERROR  Cannot read properties of undefined (reading 'Propose')

 871:       step = _ref4.step,
 872:       status = _ref4.status;
 873:   var _ExecutorMap$step$ste = ExecutorMap[step.stepType],
 874:       Propose = _ExecutorMap$step$ste.Propose,
 875:       Commit = _ExecutorMap$step$ste.Commit;
 876:   var dispatch = React__default['default'].useContext(DispatchContext);
 877:   var handleProposalAccepted = React__default['default'].useCallback(function (msg) {
 ```

**Comment by beerose** [comment]
Hi @noxify! How's it going with this PR? Do you need any help?

**Comment by noxify** [comment]
Hi @beerose ,

Thanks for the reminder.

Last days were more stressful as expected 🙈

I could need some guidance with testing it locally.

Based on my calendar it seems I could start tomorrow with it.

Maybe you have some free time in the next days? If so, I would give you a ping via discord.

**Comment by beerose** [comment]
> Last days were more stressful as expected 🙈

Hope you're taking care of yourself! There's no rush with this PR. 

You can ping me on Discord.

**Comment by noxify** [comment]
@JuanM04 - i was able to reproduce the issue and pushed the fix.

Tested it locally with the following combinations:

```
  .addRunCommandStep({
    stepId: "command1",
    stepName: "Run Command 1",
    explanation: "Runs the command `ls -al`",
    command: "ls -al",
  })
  .addRunCommandStep({
    stepId: "command2",
    stepName: "Run Command 2",
    explanation: "Runs the command `blitz help`",
    command: "blitz help",
  })
  .addRunCommandStep({
    stepId: "command3",
    stepName: "Run Command 3",
    explanation: "Runs the command `blitz routes`",
    command: "blitz routes",
  })
  .addRunCommandStep({
    stepId: "command4",
    stepName: "Run Command 4",
    explanation: "Runs the command `blitz prisma generate`",
    command: "blitz prisma generate",
  })
  .addRunCommandStep({
    stepId: "command5",
    stepName: "Run Command 5",
    explanation: "Runs the command `npm list`",
    command: "npm list",
  })
```

I checked also that the result of `blitz prisma generate` is what we expect and ... 🥁 ... it works! ( i deleted the previous created `node_modules/.prisma/client/` directory to make sure it's generated from the recipe command)

Let me know if there is something else.

JuanM04 (dismissed):
It seems good!

beerose:
There are a few conflicts because we recently moved the installer package from `./packages`  to `nextjs/packages` (https://github.com/blitz-js/blitz/pull/3044). We haven't done any other changes to those files (maybe except for some import updates) so it should be okay to just copy the new content of these files to the new location. 

I'm sorry for the inconvenience. Let me know if you want to resolve the conflicts instead!

noxify:
Not sure what I did, but it's not what i wanted.

I created the new branch based on the latest canary and added my changes.
New PR is #3090

Issue #3022, opened 2021-12-05
Tests fail out of the box on Windows


Vandivier:
### What is the problem?

As mentioned in #3021 `yarn test` fails out of the box on Windows at [this canary SHA](https://github.com/blitz-js/blitz/commit/8f5cad91d47a87cc53690ae40284afe9bca72f16), which I think is equivalent to blitz 0.44. This issue adds detail for a specific test example within `yarn testonly:packages`

### Paste all your error logs here:

```
$ yarn testonly:packages
yarn run v1.22.10
$ ultra -r --filter "packages/*" --concurrency 15 test
√ @blitzjs/server at packages\server 13.829s
  √ test 10.979s
    √ $ jest --passWithNoTests 6.604s
      │ No tests found, exiting with code 0
      │ 
/ @blitzjs/generator at packages\generator
  / test
    / $ jest
      │  PASS  test/utils/inflector.test.ts (9.399 s)
      │  PASS  test/prisma/field.test.ts (11.767 s)
      │
× @blitzjs/installer at packages\installer 23.983s
  × test 23.958s
    × $ jest  23.891s
      │  FAIL  test/transforms/transform-blitz-config.test.ts
      │   ● Test suite failed to run
      │
      │     Cannot find module '@blitzjs/display' from 'dist/blitzjs-installer.cjs.dev.js'
      │
      │     Require stack:
      │       dist/blitzjs-installer.cjs.dev.js
      │       dist/blitzjs-installer.cjs.js
      │       test/transforms/transform-blitz-config.test.ts
      │
      │       3 | Object.defineProperty(exports, '__esModule', { value: true });
      │       4 |
      │     > 5 | var display = require('@blitzjs/display');
      │         |               ^
      │       6 | var ink = require('ink');
      │       7 | var React = require('react');
      │       8 | var crossSpawn = require('cross-spawn');
      │
      │       at Resolver.resolveModule (../../node_modules/jest-resolve/build/index.js:311:11)
      │       at Object.<anonymous> (dist/blitzjs-installer.cjs.dev.js:5:15)
      │
      │  FAIL  test/transforms/update-babel-config.test.ts
      │   ● Test suite failed to run
      │
      │     Cannot find module '@blitzjs/display' from 'dist/blitzjs-installer.cjs.dev.js'
      │
      │     Require stack:
      │       dist/blitzjs-installer.cjs.dev.js
      │       dist/blitzjs-installer.cjs.js
[...]

Issue #3021, opened 2021-12-05
TS Incompatible with typescript-estree


Vandivier:
### What is the problem?

1. Checking out [this commit](https://github.com/blitz-js/blitz/commit/8f5cad91d47a87cc53690ae40284afe9bca72f16), which I think is equivalent to 0.44 release, `yarn && yarn test` fails out of the box without modification
2. I get errors which are described below, including one that says `WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.`
3. note that I'm on windows

### Paste all your error logs here:

```
$ yarn test
yarn run v1.22.10
$ yarn run lint && yarn run build && ultra -r test
$ eslint --ext ".js,.ts,.tsx" .
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=3.3.1 <4.4.0

YOUR TYPESCRIPT VERSION: 4.5.2

Please only submit bug reports when using the officially supported version.

=============

D:\workspace\github\blitz\test\jest-environment.js
  1:0  error  Parsing error: D:\workspace\github\blitz\test\jest-environment.js: Unexpected token (1:0)

> 1 | ../nextjs/test/jest-environment.js
    | ^

D:\workspace\github\blitz\test\jest-global-setup.js
  1:0  error  Parsing error: D:\workspace\github\blitz\test\jest-global-setup.js: Unexpected token (1:0)

> 1 | ../nextjs/test/jest-global-setup.js
    | ^

D:\workspace\github\blitz\test\jest-global-teardown.js
  1:0  error  Parsing error: D:\workspace\github\blitz\test\jest-global-teardown.js: Unexpected token (1:0)

> 1 | ../nextjs/test/jest-global-teardown.js
    | ^

D:\workspace\github\blitz\test\lib\next-webdriver.d.ts
  1:0  error  Parsing error: Declaration or statement expected

D:\workspace\github\blitz\test\lib\next-webdriver.js
  1:0  error  Parsing error: D:\workspace\github\blitz\test\lib\next-webdriver.js: Unexpected token (1:0)

> 1 | ../../nextjs/test/lib/next-webdriver.js
    | ^

✖ 5 problems (5 errors, 0 warnings)

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
[...]

PR #3030, opened 2021-12-06
Fix windows incompatibility with nested dependency

flybayer:
Closes: https://github.com/blitz-js/blitz/issues/3022
Closes: https://github.com/blitz-js/blitz/issues/3021

Fix windows incompatibility by upgrading typescript-eslint dependencies


Vandivier:
kudos once again @flybayer

PR #3031, opened 2021-12-06
Console - Load Nested Mutations and Queries

skotchpine:
## The Problem

When I run `blitz c`, my queries and mutations aren't loaded. This is because in `next-console.ts` mutations and queries are only loaded if they are directly inside a mutations or queries directory. I nest mine one directory further. I'll explain why, but feel free to skip to the features at the bottom.

## Rationale

I like conventions. I'm structuring my project like this:
```
/app/api/feature/example.js
/app/components/feature/Example.js
/app/integrations/feature/featureExample.js
/app/mutations/feature/exampleMutation.js
/app/pages/feature/example.js
/app/queries/feature/exampleQuery.js
/app/schemas/feature/exampleSchema.js
/app/utils/example.js
/app/db.js
```

There are a few reasons why this is better:
1. The file structure of `pages` matches that of other type classes
2. All type classes are visible from inside `/app`
3. Other monoliths are structured this way
4. (filenames) Using type class in filenames makes autocomplete easier
5. (schemas) Exporting one schema per file makes adding and removing imports less work
6. (integrations & db) Using `"./app"` as `"baseUrl"` makes imports shorter `import exampleQuery from 'queries/feature/exampleQuery.js'`

## Feature Checklist

- [x] In auth example: add `"console"` script & `currentUserQuery` for dev-testing
- [x] Modify glob in `next-console.ts` to load nested queries & mutations


skotchpine:
This category of problem would be solved by loading or generating those globs from `blitz.config`

beerose (approved):
Awesome! Thanks for the PR

Issue #3042, opened 2021-12-08
Blitz new command won't generate a SQLite database


marraleks:
### What is the problem?

When i run blitz new ... it's fails at Initializing SQLite database and I have to migrate manually after.

### Paste all your error logs here:

```
✔ Retrieving the freshest of dependencies
✔ Resolving packages
✔ Fetching packages
✔ Linking dependencies
✔ Building fresh packages
✖ Initializing SQLite database.     <-- this
✔ Committing your app

Your new Blitz app is ready! Next steps:

   1. cd testapp
   2. blitz prisma migrate dev (when asked, you can name the migration anything).   <-- this
   3. blitz dev
```

### Paste all relevant code snippets here:

```
blitz new ...
```

### What are detailed steps to reproduce this?

1. Run; blitz new ...

### Run `blitz -v` and paste the output here:

```
blitz -v
macOS Big Sur | darwin-x64 | Node: v17.0.1

blitz: 0.44.3 (global)
blitz: undefined (local)

  Package manager: yarn
  System:
    OS: macOS 11.6
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 4.57 GB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 17.0.1 - ~/.nvm/versions/node/v17.0.1/bin/node
    Yarn: 1.22.17 - /usr/local/bin/yarn
    npm: 8.1.0 - ~/.nvm/versions/node/v17.0.1/bin/npm
    Watchman: Not Found
  npmPackages:
    @prisma/client: Not Found
    blitz: Not Found
    prisma: Not Found
    react: Not Found
    react-dom: Not Found
    typescript: Not Found


❯ blitz -v
macOS Big Sur | darwin-x64 | Node: v17.0.1

blitz: 0.44.3 (global)
blitz: undefined (local)

  Package manager: yarn
  System:
    OS: macOS 11.6
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 3.43 GB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 17.0.1 - ~/.nvm/versions/node/v17.0.1/bin/node
    Yarn: 1.22.17 - /usr/local/bin/yarn
    npm: 8.1.0 - ~/.nvm/versions/node/v17.0.1/bin/npm
    Watchman: Not Found
[...]

PR #3043, opened 2021-12-08
Load env vars for blitz new command - ignore cached variables

beerose:
Closes: https://github.com/blitz-js/blitz/issues/3042

Not sure if that would be the best way to handle that (I might have missed something), so would love some suggestions if it's not. 

## Bug Checklist

## Feature Checklist


nextjs/packages/next-env/index.ts
```diff
@@ -67,15 +67,16 @@ export function processEnv(
 
 export function loadEnvConfig(
   dir: string = process.cwd(),
-  dev?: boolean,
-  log: Log = console
+  log: Log = console,
+  ignoreCache = false
```
flybayer, line 71: Why did you remove `dev`?

flybayer: So `ignoreCache` is what fixes the issue right? Makes sense.

It would be slightly better to add an object as the new param as `{ignoreCache}`. I always fine unnamed function params error prone, especially when we are in a fork situation.

beerose: > Why did you remove dev?

Not used in the function body

flybayer: > Not used in the function body

Normally I would agree, but in the context of a fork, this makes upstream merges more difficult

PR #3048, opened 2021-12-09
Install latest stable Blitz version

JuanM04:

Install the latest stable Blitz version instead of the unstable canary one for new apps. (Also, `latest` points to v0.44.3 and `canary` points to v0.41.2-canary.5)


beerose (approved):
Looks like we didn't consider it? https://github.com/blitz-js/blitz/blob/canary/packages/generator/src/utils/get-blitz-dependency-version.ts#L5 (it depends on the cli version)

PR #3060, opened 2021-12-11
Use free tier by default in Render recipe

Vandivier:
1. default to free tier
2. remove obsolete guidance to click generate button
3. build DB prior to building server


beerose (approved):
Thanks!

PR #3063, opened 2021-12-13
Fix inconsistent formatting in LoginForm and Login page 

frankiesardo:
This commit https://github.com/blitz-js/blitz/commit/4384ce0514557c2fe7905b0c022caa8b63afff83#diff-90f545371ea059291383a0ec101e86b455616c68fe05777f7554c69bb431c23f introduced inconsistent formatting for the login form and the login pages, which causes users of a brand new app to have unneeded "unstaged changes" after they format their files.

This is a quick fix that adds back the space inside the curly braces.


frankiesardo:
Tangential question: why is is that the source of code of blitz itself uses `{this, convention}` but the generated app uses `{ this, convention, instead }` ?

beerose:
> Tangential question: why is it that the source of code of blitz itself uses `{this, convention}` but the generated app uses `{ this, convention, instead }` ?

I don't really know. @flybayer do you remember why is that?

beerose (approved):
Thanks for the PR!

PR #3068, opened 2021-12-15
Allow passing custom templates folder for `blitz generate`

beerose:
Related to: https://github.com/blitz-js/blitz/issues/236

## Bug Checklist

## Feature Checklist


packages/cli/src/commands/generate.ts
```diff
@@ -116,6 +117,10 @@ export class Generate extends Command {
       char: "d",
       description: "Show what files will be created without writing them to disk",
     }),
+    templateDir: flags.string({
+      char: "t",
+      description: "Specify a custom template directory",
+    }),
```
flybayer, line 123: @beerose lets make this an option in blitz.config.ts instead of a CLI flag, so we don't have to pass this flag every time.

Other than that, it looks great

Issue #3098, opened 2021-12-27
Throwing RedirectError on server side (in queries and mutations) does not redirect on the frontend


dineshgadge:
### What is the problem?

https://codesandbox.io/s/fervent-morning-qtvzm

Throwing RedirectError from server (queries/mutations) does not redirect on the frontend. It seems this case is not handled, as the ErrorBoundary expects the error to have have a url property but raising RedirectError like
```typescript
// app/projects/queries/getProjects.ts

resolver.pipe((input) => throw new RedirectError("/edit-profile"))
```

from server sets the message property to /edit-profile instead of setting url to /edit-profile leading an unhandled error where it expects a url but gets undefined/null.
My use case is redirecting the user to edit-profile page if profile is not at least 50% complete if they try to access projects list page

### Paste all your error logs here:

```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'auth')
    at Object.formatUrl (format-url.js?2477:32)
    at Object.formatWithValidation (utils.js?e766:109)
    at resolveHref (router.js?4a0b:169)
    at prepareUrlAs (router.js?4a0b:226)
    at Router.push (router.js?4a0b:490)
    at Object.instance.<computed> [as push] (router.js?9ce1:175)
    at ErrorBoundary.eval (error-boundary.js?e551:84)
    at Generator.next (<anonymous>)
    at asyncGeneratorStep (error-boundary.js?e551:11)
    at _next (error-boundary.js?e551:29)
    at eval (error-boundary.js?e551:34)
    at new Promise (<anonymous>)
    at eval (error-boundary.js?e551:26)
    at ErrorBoundary.componentDidCatch (error-boundary.js?e551:88)
    at ErrorBoundary.callback (react-dom.development.js?61bb:18669)
    at callCallback (react-dom.development.js?61bb:13146)
    at commitUpdateQueue (react-dom.development.js?61bb:13167)
    at commitLayoutEffectOnFiber (react-dom.development.js?61bb:23057)
    at commitLayoutMountEffects_complete (react-dom.development.js?61bb:24328)
    at commitLayoutEffects_begin (react-dom.development.js?61bb:24314)
    at commitLayoutEffects (react-dom.development.js?61bb:24252)
    at commitRootImpl (react-dom.development.js?61bb:26411)
    at commitRoot (react-dom.development.js?61bb:26280)
    at finishConcurrentRender (react-dom.development.js?61bb:25509)
[...]

PR #3100, opened 2021-12-28
Fix `RedirectError` not redirecting on the client #3098

dineshgadge:
Closes: #3098 

Added `url` to the list of allowedErrorProps for SuperJson.
This will propagate the url parameter to the frontend. The frontend can then handle the redirection to url (as `string` or `UrlObject` with `pathname` and optionally `href` properties). Earlier, since `url` property was not serialized, such redirection could not be correctly handled by the frontend.

On https://blitzjs.com/docs/error-handling#built-in-errors the documentation mentions 
```
You can throw this error from a render function if you want to redirect the user ...
```
which is a bit confusing since it makes one believe that `RedirectError` can be thrown only on the frontend because of the `render function` bit. That documentation can be updated too with correct language. However, I am not the best person to suggest what the final wordings there should be, so I am not submitting a corresponding documentation PR.

Also, I could not find any integration / unit tests for `nextjs/packages/stdlib/errors.ts`, so I have not added the corresponding tests.

## Bug Checklist

## Feature Checklist

### DIFF
--- a/nextjs/packages/next/stdlib/errors.ts
+++ b/nextjs/packages/next/stdlib/errors.ts
@@ -1,7 +1,7 @@
 import SuperJson from 'superjson'
 import type { UrlObject } from 'url'
 
-const errorProps = ['name', 'message', 'code', 'statusCode', 'meta']
+const errorProps = ['name', 'message', 'code', 'statusCode', 'meta', 'url']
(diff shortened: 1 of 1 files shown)


beerose (approved):
Thanks!

PR #3129, opened 2022-01-16
Set up fancy new monorepo for blitz toolkit with pnpm, turborepo, unbuild, vitest

flybayer:
This is officially the start of the new blitz toolkit codebase! Toolkit development will run on the `main` branch.

We may later break out the @blitzjs/auth and @blitzjs/rpc packages into standalone repos, as I find that helps make it clearer to folks that they are standalone packages that can be used directly.


eric-burel:
Won't you have trouble with client-only import? Like importing Leaflet? You may want an intermediate "shared" index as well to support this

PR #3147, opened 2022-01-25
Migrate auth code to Blitz toolkit, add server & client setup

beerose:
Todo:

- [x] Move auth tests to auth package
- [x] Move plugin middleware skeleton to auth package
- [x] Test it in a sample next app

## Bug Checklist

## Feature Checklist


package.json
```diff
@@ -24,7 +24,7 @@
     "prettier": "^2.5.1",
     "prettier-plugin-prisma": "3.8.0",
     "pretty-quick": "3.1.3",
-    "turbo": "latest",
+    "turbo": "1.0.24",
```
beerose, line 27: Related issue: https://github.com/vercel/turborepo/issues/549

JuanM04: That issue was (supposedly) fixed in v1.1.0 ([ref](https://github.com/vercel/turborepo/issues/549#issuecomment-1022284349))

beerose: Cool! thanks for letting me know!

packages/blitz-auth/package.json
```diff
@@ -18,17 +18,40 @@
...
     "typescript": "^4.5.3",
-    "unbuild": "0.6.7",
+    "unbuild": "0.6.9",
     "watch": "1.0.2"
   },
+  "peerDependencies": {
+    "next": "12.0.7",
+    "react": "^17.0.2",
+    "react-dom": "17.0.2"
+  },
```
flybayer, line 38: I think we shouldn't use peerDeps — they are annoying for users at best and at worst, they screw up npm v7+

packages/blitz-auth/src/shared/index.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export function addMinutes(date: Date, amount: number): Date {
+  const cleanAmount = toInteger(amount)
+  return addMilliseconds(date, cleanAmount * MILLISECONDS_IN_MINUTE)
+}
+
+export function addYears(date: Date, amount: number): Date {
+  const cleanAmount = toInteger(amount)
+  return addMinutes(date, cleanAmount * MINUTES_IN_YEAR)
+}
```
flybayer, line 38: Utils like these should eventually go in the core `blitz` package

packages/blitz-next/src/index-server.ts
```diff
@@ -1,3 +1,72 @@
...
+      middleware(req, res, result =>
+        result instanceof Error ? reject(result) : resolve(result),
+      );
+    });
+    return [...acc, promise];
+  }, [] as Promise<unknown>[]);
+
+  await Promise.all(promises);
+}
+
```
flybayer, line 38: We need to somehow support error middleware for things like capturing to Sentry:

```ts
async (req, res, next) => {
  try {
    await next()
  } catch (error) {
    Sentry.captureException(error)
    throw error
  }
}
```

I'm not sure if legacy blitz middleware supported the above or not, but for sure it didn't work for RPC errors, because the old RPC middleware would catch the error and turn it into a 200 response.

So we may need to add an extra error middleware type or event type so we can capture RPC resolver errors.

PR #3159, opened 2022-02-09
Add email verification

0xsamrath:
Features:
- Adds email verification, using tokens
- Prevents users from seeing content without verifying their emails

The method for implementation is as follows
- A new field in the user model in the database, called `verified` this defaults to false and isn't provided when users sign up
- When a user signs up, a token(generated the same way as a reset password token) is sent to their email(uses previewMail)
- The user clicks on the link from their email, it opens a link
- The link is `/confirm-email?token=[token]`, the user then clicks on a button on the page
- This triggers a mutation called `confirmEmail`, which verifies the token
- If the token is valid, and the user is logged in(they are allowed to log in even if their email isn't verified), they are successfully verified
- They can now view the whole site
- View to the site is blocked by changes in they layout(prevents the user from seeing the content, if they aren't verified or a noVerification prop is passed)

### DIFF
--- a/packages/generator/templates/app/app/auth/mutations/confirmEmail.ts
+++ b/packages/generator/templates/app/app/auth/mutations/confirmEmail.ts
@@ -1,0 +1,41 @@
+import { resolver, SecurePassword, hash256 } from "blitz"
+import db from "db"
+import login from "./login"
+
+import { ConfirmEmail } from "../validations"
+
+export class ResetPasswordError extends Error {
+  name = "ConfirmEmailError"
+  message = "Email confirmation link is invalid or it has expired."
+}
+
+export default resolver.pipe(resolver.zod(ConfirmEmail), async ({ token }, ctx) => {
+  // 1. Try to find this token in the database
+  const hashedToken = hash256(token)
+  const possibleToken = await db.token.findFirst({
+    where: { hashedToken, type: "CONFIRM_EMAIL" },
+    include: { user: true },
+  })
+
+  // 2. If token not found, error
+  if (!possibleToken) {
+    throw new ResetPasswordError()
+  }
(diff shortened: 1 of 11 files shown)


beerose (changes requested):
Thanks for the PR!

I have a few comments/questions.

1. When generating a new app, you get a React error, because there's a `useQuery` usage in `Layout.tsx` and no suspense specified to wrap it.
Screenshot 2022-02-09 at 11 08 54

I'm not sure if wrapping the whole app (in `_app.tsx`) by default is a good idea — can the verification logic be moved to `index.tsx` instead of layout?

2. So I think that currently, it's really awesome that you can create a new Blitz app and signup & login instantly. By adding `email confirmation` users won't be able to do it unless they implement the mailer and/or manually set `verified` to true in the database. Some of the users could ditch Blitz because of that or get discouraged. Could this be an opt-in feature? Or could we provide a way to improve this experience?

3. One more reason for making this opt-in is that I think that there are many people who don't want email confirmation in their apps (e.g. they only use social login)  — having it by default forces them to remove a lot of code at start. I also think that a common approach nowadays is to let user create an account, and then prompt them with `Please confirm your email` in the app. For example:

Screenshot 2022-02-09 at 11 21 00

0xsamrath:
@beerose, would having this as an opt-in from the `blitz.config.js` be a good ideas?
And to the other points, I could create a second component which would display in a popup format that hovers over the Layout and is then called in the layout, this would remove the Suspense issue

0xsamrath:
I'll have to change a lot of the code, I'll PR it in soon, once I manage to get the time

0xsamrath:
I've read through the docs and decided a recipe would definitely be a better fit, I'll close this PR and create a recipe soon

PR #3177, opened 2022-02-19
`secureheaders` recipe: add `Content-Security-Policy` header and `includeSubDomains; preload` to `Strict-Transport-Security`

apexkoder:
Related: #2019

## Bug Checklist

## Feature Checklist


recipes/secureheaders/templates/core/secureheaders.ts
```diff
@@ -10,7 +10,7 @@ export const cspHashOf = (text: string) => {
 export const computeCsp = (props: Readonly<DocumentProps>) => {
   const scriptHash = cspHashOf(BlitzScript.getInlineScriptSource(props))
 
-  return `default-src 'self'; style-src 'self' 'unsafe-inline' fonts.googleapis.com; font-src 'self' data: fonts.gstatic.com; script-src 'self' ${
-    process.env.NODE_ENV === "production" ? "" : "'unsafe-eval'"
-  } ${scriptHash}`
+  return `default-src 'self'; img-src 'self' res.cloudinary.com data:; style-src 'self' 'unsafe-inline' fonts.googleapis.com; font-src 'self' data: fonts.gstatic.com; script-src 'self' ${
```
MrLeebo, line 13: Is this addition necessary? Not every app requires cloudinary, and allowing the data: protocol gives a malicious user a method to inject XSS attacks https://www.w3.org/TR/CSP3/#csp-directives

MrLeebo (changes requested):

recipes/secureheaders/templates/core/secureheaders.ts
PaulEbose: > Not every app requires cloudinary

It can be removed. I put in there as an example, since also not every app require Google fonts.

> allowing the data: protocol gives a malicious user a method to inject XSS attacks https://www.w3.org/TR/CSP3/#csp-directives

I'm adding `data:` protocol to the `img-src` directive. From the link you sent:

```md
To mitigate the risk of cross-site scripting attacks, web developers SHOULD include directives that regulate sources of script and plugins. They can do so by including:

• Both the [script-src](https://www.w3.org/TR/CSP3/#script-src) and [object-src](https://www.w3.org/TR/CSP3/#object-src) directives, or
• a [default-src](https://www.w3.org/TR/CSP3/#default-src) directive
```

This [web.dev](https://web.dev/csp-xss/#csp-targets-xss) article hints that the `script-src`, `object-src`, and `base-uri` directives are relevant for XSS

MrLeebo: I'm fine with keeping the data: protocol for img-src then, but I think cloudinary can be left out. Unlike cloudinary, Google fonts is used by the default app template and the secureheaders recipe should generate a policy that doesn't produce errors on a fresh blitz app.

It would be a good idea to add a comment next to that line letting users know they can remove the Google font if they've replaced the default CSS though.

MrLeebo (approved):
Great work on this 👍

Nisarga-Developer (approved):

beerose (approved):
Thanks for this PR!

siddhsuresh:
Closing since required headers added in #3813

PR #3178, opened 2022-02-20
don't suggest faker.js in code comments

Vandivier:

This PR is purely a comment change. Previously, seed files:
1. suggested two data gen libraries, which is possibly confusing and as an opinionated framework it is perhaps better to pick just one
2. one of the suggested libraries is no longer maintained, no longer works, and it has been identified as a security vulnerability:
https://jaxenter.com/software-security-sharma-176708.html
https://github.com/Marak/Faker.js

Both of those concerns are solved by simply removing the comments suggesting faker.js


MrLeebo (approved):
Thanks!

MrLeebo:
Did you happen to check if there were references to faker.js in the blitzjs.com repo?

beerose (approved):

Vandivier:
> Did you happen to check if there were references to faker.js in the blitzjs.com repo?

Yes I did and yes it does and I thought I submitted a PR for that but now I can't find it #oof

there's one reference on the below URL I'll take care of it later this weekend if no one else takes care of it today:
https://blitzjs.com/docs/database-seeds#example

prisis:
> ### What are the changes and their implications?
> 
> This PR is purely a comment change. Previously, seed files:
> 
> 1.  suggested two data gen libraries, which is possibly confusing and as an opinionated framework it is perhaps better to pick just one
> 2.  one of the suggested libraries is no longer maintained, no longer works, and it has been identified as a security vulnerability:  
>     [jaxenter.com/software-security-sharma-176708.html](https://jaxenter.com/software-security-sharma-176708.html)  
>     [Marak/Faker.js](https://github.com/Marak/Faker.js)
> 
> Both of those concerns are solved by simply removing the comments suggesting faker.js

This is not completely true, there is a overtake(alternative) of the old faker https://github.com/faker-js/faker

PR #3263, opened 2022-04-02
toolkit: fix some various package and build issues

flybayer:
`next build` was failing because of prisma import type, so fixed that by fixing the tsconfig symlink property.

Then `next build` was failing because of `TypeError: Cannot destructure property 'inAmpMode' of '(0 , _react).useContext(...)' as it is null.` error. Upgrading both nextjs and react apparently fixed that.

Fixed a bunch of "inline dependency" warnings by adding to peer deps using "*" version range

And added `next build` to CI


flybayer:
FINALLY!

Issue #3250, opened 2022-03-29
cli: `blitz dev` should automatically choose next empty port


flybayer:
like legacy blitz, will pass`-p` to the next cli

Issue #3248, opened 2022-03-29
cli: add `blitz start` and `blitz build` etc


flybayer:
- [ ] `blitz next CMD` will be special cased, such that all nextjs commands like `blitz dev`, `blitz lint`, etc will run `blitz next dev`, etc.
- [ ] Set `process.env.__NEXT_PROCESSED_ENV = 'true` to skip nextjs env loading

dillonraphael:
Idea:
Commands like `blitz dev` should check if user is using `next` or `remix`, then fetch cli commands from appropriate separate package. This will keep the cli agnostic and consistent with blitz toolkit vision.

beerose:
https://github.com/blitz-js/blitz/pull/3272

Issue #3247, opened 2022-03-29
🔥cli: add `blitz NAME CMD`


flybayer:
**The minimum that we need is `blitz prisma` to work, we can hardcode that for now if needed.**
---
- [ ] Any `blitz CMD` will run some logic (load env variables like we are currently with dotenv-flow), then run `node_modules/.bin/CMD` and pass through all arguments
- [ ] should work for yarn v1
- [ ] should work for yarn v3
- [ ] should work for pnpm
- [ ] should work for npm

PR #3270, opened 2022-04-06
Cli

dillonraphael:
## Feature Checklist

- [x] Setups cli boilerplate & sets bin in package.json
- [x] Migrate over legacy blitz `new` cli command
- [ ] Ensure every blitz command gets env variables passed (ie. #3247)
- [ ] Add blitz commands that check for framework & fetch commands to spawn from plugin packages ( ie. #3248 )
- [ ] `blitz dev` command should automatically find empty port and pass it to correct framework runner. For example port should be passed to `next dev` (ie. #3250)


packages/blitz/src/cli/commands/new.ts
```diff
@@ -0,0 +1,302 @@
...
+}
+enum TLanguage {
+  "typescript" = "TypeScript",
+  "javascript" = "Javascript",
+}
+type TPkgManager = "npm" | "yarn" | "pnpm"
+enum TTemplate {
+  "full" = "full",
+  "minimal" = "minimal",
+}
```
beerose, line 24: Let's use literal types or objects instead of enums — they compile to something complicated and there are some other problems with enums:
Screenshot 2022-04-07 at 10 30 50

We can do something like this instead:

```ts
const forms = {
  "react-final-form": "React Final Form",
  "react-hook-form": "React Hook Form",
  "formik": "Formik",
}
type Form = keyof typeof forms

const language = {
  "typescript": "TypeScript",
  "javascript": "Javascript",
}
type Lang = keyof typeof language

type TPkgManager = "npm" | "yarn" | "pnpm"
type Template = "full" | "minimal"
```

packages/blitz/src/cli/index.ts
```diff
@@ -0,0 +1,121 @@
+#!/usr/bin/env node
+import {NON_STANDARD_NODE_ENV} from "./utils/constants"
+import arg from "arg"
+import packageJson from "../../package.json"
+
+const defaultCommand = "dev"
+export type cliCommand = (argv?: string[]) => void
```
beerose, line 7:
```suggestion
export type CliCommand = (argv?: string[]) => void
```

PR #3272, opened 2022-04-08
Support `blitz next CMD`

beerose:
Closes: https://github.com/blitz-js/blitz/issues/3248

## Bug Checklist

## Feature Checklist


packages/blitz/package.json
```diff
@@ -41,15 +41,21 @@
...
+    "@types/fs-extra": "9.0.13",
     "@types/hasbin": "1.2.0",
     "@types/node-fetch": "2.6.1",
     "@types/npm-which": "3.0.1",
     "@types/prompts": "2.0.14",
     "@types/react": "17.0.43",
     "@types/react-dom": "17.0.14",
     "@types/test-listen": "1.1.0",
+    "debug": "4.3.3",
+    "detect-port": "1.3.0",
```
flybayer, line 56: Needs to be production dep, right?

packages/blitz/package.json
```diff
@@ -41,15 +41,21 @@
...
     "express": "4.17.3",
+    "fs-extra": "10.0.1",
```
flybayer, line 58: Also needs to be prod dep

packages/blitz/src/cli/utils/config.ts
```diff
@@ -0,0 +1,157 @@
...
+    "cypress/**/*",
+    "test/**/*",
+    "tests/**/*",
+    "spec/**/*",
+    "specs/**/*",
+    "**/*.test.*",
+    "**/*.spec.*",
+    "**/.yalc/**/*",
+  ],
+  includePaths: ["**/*"],
```
flybayer, line 114: I'm pretty sure all of this is unused and can be removed.

packages/blitz/src/cli/index.ts
```diff
@@ -1,31 +1,82 @@
...
+  next: async () => async (argv) => {
+    process.env.__NEXT_PROCESSED_ENV = "true"
+
+    const nextArgs = arg(
+      {
+        ...commonArgs,
+        "--port": Number,
+        "--hostname": String,
+        "--inspect": Boolean,
+        "--no-incremental-build": Boolean,
```
flybayer, line 33: I don't think we are using this

packages/blitz/src/cli/index.ts
```diff
@@ -1,31 +1,82 @@
...
+        permissive: true,
+      },
+    )
+
+    const config: ServerConfig = {
+      rootFolder: process.cwd(),
+      port: nextArgs["--port"],
+      hostname: nextArgs["--hostname"],
+      inspect: nextArgs["--inspect"],
+      clean: nextArgs["--no-incremental-build"],
```
flybayer, line 48: I don't think we are using this

flybayer:
Clarification on #3248, sorry for bad communication on my part.

Since we are adding the ability to run any npm binary with `blitz CLI_NAME ...ARGS`, with that code, `blitz next dev` etc should just work without this PR.

However, we have some special things for next like setting the port, and then also handing custom server. **So this PR only needs to handle `next dev`, `next build`, and `next start`**. All other nextjs commands should pass straight through and be handled with the general `blitz CLI_NAME ...ARGS` feature.

Additionally, there's one more optimization we want to do: We don't have devs to have to write out the entire `blitz next dev` command, so all nextjs commands should work like `blitz dev`, `blitz build`, `blitz lint`, etc.

packages/blitz/package.json
```diff
@@ -41,14 +44,18 @@
...
     "@types/express": "4.17.13",
+    "@types/fs-extra": "9.0.13",
     "@types/hasbin": "1.2.0",
     "@types/node-fetch": "2.6.1",
     "@types/npm-which": "3.0.1",
     "@types/prompts": "2.0.14",
     "@types/react": "17.0.43",
     "@types/react-dom": "17.0.14",
     "@types/test-listen": "1.1.0",
+    "debug": "4.3.3",
```
flybayer, line 58: this probably needs to be in production deps

PR #3278, opened 2022-04-10
Add RPC package to toolkit, including magical imports 🎉

flybayer:
Closes: https://github.com/blitz-js/blitz/issues/3196


flybayer:
@dillonraphael can you fix the test timeout?

PR #3308, opened 2022-04-19
Update react-query.tsx

fleck:
Getting runtime errors from attempting to read properties on undefined, this change should prevent that!

Context: During SSR useQuery returns undefined until the query resolves.

### DIFF
--- a/apps/web/pages/users.tsx
+++ b/apps/web/pages/users.tsx
@@ -7,7 +7,7 @@ function UsersPage() {
     <div>
       Users:
       <ul>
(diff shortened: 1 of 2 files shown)


fleck:
Just pushed a fix for this PR, it built locally so it should build in CI now 🤞

flybayer:
Hey @fleck, thank you so much for the contribution! We have a bug with nextjs & react 18 for suspense that is part of this issue. So we need to fix the underlying issue vs this PR so that suspense and loading screens work properly.

PR #3321, opened 2022-04-27
change password updated with try/catch

the-bayer:
Closes: n/a

Adds try/catch with invalid password error to the change password mutation- this prevents an invalid password from logging a user out while attempting to change their password.

## Bug Checklist

## Feature Checklist

### DIFF
--- a/packages/generator/templates/app/app/auth/mutations/changePassword.ts
+++ b/packages/generator/templates/app/app/auth/mutations/changePassword.ts
@@ -12,7 +12,11 @@ export default resolver.pipe(
     const user = await db.user.findFirst({ where: { id: ctx.session.userId as number } })
     if (!user) throw new NotFoundError()
 
-    await authenticateUser(user.email, currentPassword)
+   try {
+      await authenticateUser(user.email, currentPassword)
+    } catch (error: any) {
+      throw new Error("Invalid Password")
+    }
 
     const hashedPassword = await SecurePassword.hash(newPassword.trim())
     await db.user.update({


packages/generator/templates/app/app/auth/mutations/changePassword.ts
```diff
@@ -12,7 +12,11 @@ export default resolver.pipe(
     const user = await db.user.findFirst({ where: { id: ctx.session.userId as number } })
     if (!user) throw new NotFoundError()
 
-    await authenticateUser(user.email, currentPassword)
+   try {
+      await authenticateUser(user.email, currentPassword)
+    } catch (error: any) {
+      throw new Error("Invalid Password")
```
beerose, line 18: Maybe let's throw `AuthenticationError` from `blitz`?
```
throw new AuthenticationError()
```

In a similar way as in `authenticateUser` function.

the-bayer: The issue with the Authentication error is that actually logs the user out if they enter the wrong password within the change password form, and if they're accessing their dashboard then they should have already logged in.

beerose: Okay I see. I'm not sure if we should return "invalid password" though. What if that error in `authenticateUser` didn't come from the password being invalid? E.g. if there was an error unrelated to the password or if email was incorrect or if there was an internal error (bug in our code, or something wrong with the database). Maybe we can have a new error instance (like AuthenticationError), e.g. InvalidPassword and throw it from `authenticateUser` explicitly?

packages/generator/templates/app/app/auth/mutations/changePassword.ts
```diff
@@ -12,7 +12,14 @@ export default resolver.pipe(
...
     if (!user) throw new NotFoundError()
 
-    await authenticateUser(user.email, currentPassword)
+   try {
+      await authenticateUser(user.email, currentPassword)
+    } catch (error: any) {
+      if (error instanceof AuthenticationError) {
+      throw new Error("Invalid Password")
+      }
+      throw new Error(error)
```
beerose, line 21: Since the `error` is already of the Error instance, we can just throw it:
```suggestion
      throw error
```

packages/generator/templates/app/app/auth/mutations/changePassword.ts
```diff
@@ -12,7 +12,14 @@ export default resolver.pipe(
     const user = await db.user.findFirst({ where: { id: ctx.session.userId as number } })
     if (!user) throw new NotFoundError()
 
-    await authenticateUser(user.email, currentPassword)
+   try {
+      await authenticateUser(user.email, currentPassword)
+    } catch (error: any) {
+      if (error instanceof AuthenticationError) {
+      throw new Error("Invalid Password")
```
beerose, line 19: (lint)
```suggestion
        throw new Error("Invalid Password")
```

PR #3337, opened 2022-05-05
Try catch for change password

the-bayer:
Closes:  n/a

adds try/catch for change password and throws error to prevent user from being logged out while trying to change their password

## Bug Checklist

## Feature Checklist


packages/generator/templates/app/app/auth/mutations/changePassword.ts
```diff
@@ -10,7 +10,15 @@ export default resolver.pipe(
...
     if (!user) throw new NotFoundError()
 
-    await authenticateUser(user.email, currentPassword)
+   try {
+      await authenticateUser(user.email, currentPassword)
+    } catch (error: any) {
+      if (error instanceof AuthenticationError) {
+        throw new Error("Invalid Password")
+      }
+      throw new Error(error)
```
beerose, line 19: Since the `error` is already of the Error instance, we can just throw it:
```suggestion
      throw error
```

Issue #3199, opened 2022-03-09
Pass queryClient into gSSP and gSP - `blitz-next`


flybayer:
Idea is that instead of this:
```ts
export async function getServerSideProps(ctx: GetServerSidePropsContext) {
  const queryClient = new QueryClient()

  const queryKey = getQueryKey(getCurrentOrganization, null)
  await queryClient.prefetchQuery(queryKey, () =>
    invokeWithMiddleware(getCurrentOrganization, null, ctx)
  )

  return {
    props: {
      dehydratedState: dehydrate(queryClient),
    },
  }
}
```

We can do this:

```ts
export const getServerSideProps = gSSP(async ({ctx}) => {
  await ctx.prefetchBlitzQuery(getCurrentOrganization, null)

  return {props: {}}
}
```

So gSSP:
1. Creates queryClient and passes it in via `ctx.queryClient`
2. Passes in `ctx.prefetchBlitzQuery`
3. After the user's function returns, handle `dehydratedState: dehydrate(queryClient)` automatically for the user

PR #3341, opened 2022-05-09
Support `prefetchBlitzQuery` in gSP and gSSP

beerose:
Closes: https://github.com/blitz-js/blitz/issues/3199

## Bug Checklist

## Feature Checklist


packages/blitz-rpc/src/server-plugin.tsx
```diff
@@ -0,0 +1,88 @@
...
+    await queryClient.prefetchQuery(queryKey, () => fn(input, res.blitzCtx))
+
+    return {
+      dehydratedState: dehydrate(queryClient),
+    }
+  }
+}
+
+export interface RpcServerPluginOptions {
+  prefetchQueryOptions?: DefaultOptions
```
flybayer, line 56: I think we can call this `reactQueryOptions` since we can use it for more than just prefetch?

flybayer (commented):
Nice, this all looks really great! One comment below on naming and then I think we should make it so the user doesn't have to return `dehydratedState` as an explicit prop. If you need help figuring that out, let me know :)

packages/blitz-rpc/src/server-plugin.tsx
beerose: Actually, we only use it for prefetch here. User can pass react query options in **client** plugin config and we use them for all the react-query hooks. So I was thinking about the naming and figured that it can be confusing if we use the same name in both client and server.

packages/blitz-next/src/index-server.ts
```diff
@@ -50,13 +58,47 @@ export const setupBlitzServer = ({plugins}: SetupBlitzOptions) => {
...
+      }
+
+      const result = await handler({req, res, ctx, ...rest})
+      if ("props" in result) {
+        const props = {
+          ...result.props,
+          ...(queryClient ? {dehydratedProps: dehydrate(queryClient)} : {}),
+        }
+        return {...result, props}
+      }
```
flybayer, line 77: why do we only add dehydratedProps if `"props" in result`?

packages/blitz-rpc/src/server-plugin.tsx
flybayer: @beerose ok yeah that makes sense to have it named prefetch.

flybayer:
Overall looks great!!

packages/blitz-next/src/index-server.ts
beerose: Ah good point, will change that

PR #3350, opened 2022-05-09
Added db seed cli command

dillonraphael:
Closes: #3318


packages/blitz/src/cli/commands/db.ts
```diff
@@ -0,0 +1,90 @@
...
+    permissive: true,
+  },
+)
+
+const runSeed = async (seedBasePath: string) => {
+  if (!process[REGISTER_INSTANCE]) {
+    // During blitz interal dev, oclif automaticaly sets up ts-node so we have to check
+    require("ts-node").register({compilerOptions: {module: "commonjs"}})
+  }
+  require("tsconfig-paths/register")
```
beerose, line 29: Why is it needed? We didn't have it in legacy blitz. I'm getting `MODULE_NOT_FOUND` errors with that line, and looks like it works fine without it

PR #3380, opened 2022-05-25
Added resolverBasePath option to rpc loaders

iDavidB:
This PR adds a property to the Blitz config that allows you to change the way rpc routes are generated from being relative to the queries/mutations folder to being from the repository root. 

Ex. `app/queries/getCurrentUser.ts`
"queries/mutations" result: `api/rpc/getCurrentUser`
 "root" result: `api/rpc/app/queries/getCurrentUser`

This is only my first version of this PR as I feel it may require some structure changes as you'll read below.

- Move `Loader` type for client/server loader to shared utils file.
- Move `installWebpackConfig` to `blitz-next` from `blitz-rpc`
	- Required me to export `loaderServer` & `loaderClient` from blitz-rpc to import it in blitz-next. This is currently my biggest "issue" as I feel that this is not the preferred setup.
- Add `options` object containing `baseResolverPath` option to every webpack use rule.
	- These options can be retrieved inside the loader by accessing `this.query` which is a little strange considering the docs say `this.query` is an alias to `this.options`, yet `this.options` outputs regular Webpack options.   https://webpack.js.org/configuration/module/#ruleoptions--rulequery


iDavidB:
Additionally, we could allow `resolverBasePath` to be a function which gets passed the filePath so users can add their own logic to how they would like to generate their paths.

beerose (approved):

Schotsl:
LGTM

Issue #3383, opened 2022-05-28
`blitz dev` throws error when used directly (only is someone uses pnpm)


dillonraphael:
The error thrown: `Module not found: Can't resolve 'styled-jsx/style'`.

Copy code from https://github.com/blitz-js/blitz/blob/canary/packages/cli/src/utils/is-blitz-root.ts

dillonraphael:
Seems the issue only happens when using pnpm. `blitz dev` works when using npm.

PR #3398, opened 2022-06-06
Update generator template npmrc & next versions

dillonraphael:
Closes: #3383

### DIFF
--- a/packages/generator/templates/app/npmrc
+++ b/packages/generator/templates/app/npmrc
@@ -1,6 +1,7 @@
 save-exact=true
 legacy-peer-deps=true
-
+strict-peer-dependencies=false 
+side-effects-cache=false
 public-hoist-pattern[]=next
 public-hoist-pattern[]=secure-password
 public-hoist-pattern[]=*jest*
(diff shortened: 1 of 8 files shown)


beerose (commented):
Let's add changeset

packages/generator/templates/app/npmrc
```diff
@@ -1,7 +1,9 @@
 save-exact=true
 legacy-peer-deps=true
-
+strict-peer-dependencies=false 
+side-effects-cache=false
 public-hoist-pattern[]=next
 public-hoist-pattern[]=secure-password
 public-hoist-pattern[]=*jest*
 public-hoist-pattern[]=@testing-library/*
+# Dont change above this
```
beerose, line 9: hm?

packages/generator/templates/app/package.ts.json
```diff
@@ -27,7 +27,7 @@
     "@blitzjs/rpc": "latest",
     "@prisma/client": "3.9.0",
     "blitz": "alpha",
-    "next": "12.1.1",
+    "next": "latest",
```
beerose, line 30: Are we sure about it? What if they push something that breaks Blitz? Shouldn't we pin it?

packages/generator/templates/app/npmrc
dillonraphael: We should let the user know not to modify npmrc file above the comment. Thoughts?

beerose (approved):

Issue #3347, opened 2022-05-09
Add cli shortcuts for every command


flybayer:
`blitz b` -> `blitz build`
`blitz p` -> `blitz prisma`
`blitz c` -> `blitz console`

etc

damilolarandolph:
Hey @flybayer! I would like to handle this

damilolarandolph:
@beerose  @flybayer I just realized that most of the commands already have aliases. Only the `new` command doesn't. I thinking `$blitz -h` should list all commands as well as their aliases to make this more obvious.

beerose:
> I thinking `$blitz -h` should list all commands as well as their aliases to make this more obvious.

Yeah, we should have that. Let's make the `-h` info better then.

datner:
took initiative and went ahead closing this, I hope it's ok :)

beerose:
Closed in https://github.com/blitz-js/blitz/pull/3410

PR #3410, opened 2022-06-08
Add aliases for Blitz CLI commands

datner:
Closes: #3347

continuing the good work of @damilolarandolph , just on the main branch instead of canary :)

Adding of aliases for commands, also wrote a description for each entry

I also removed some seemingly dead code on a separate commit __if this is a mistake, drop the commit__
I also let my prettier run on the file on a separate commit __if this is a mistake, drop the commit__

## Bug Checklist

## Feature Checklist


packages/blitz/src/cli/index.ts
```diff
@@ -167,8 +166,14 @@ async function main() {
...
   
       Available commands
-        ${Object.keys(commands).join(", ")}
-  
+        dev, d          Start a development server 🪄
+        build, b        Create a production build 🏗️
+        start, s        Start the production server 🐎
+        new, n          Create a new Blitz project ✨
+        generate, g     Generate new files for your Blitz project 🤠
+        codegen         Manually trigger the blitz codegen 🤖
```
beerose, line 174: Commenting `generate` as we don't have it working yet
```suggestion
        // generate, g     Generate new files for your Blitz project 🤠
        codegen         Run the blitz codegen 🤖
```

beerose (commented):
Looks amazing! Thank you!

packages/blitz/src/cli/index.ts
datner: not sure that the cli would understand that you don't want to show that to the user hahaha, i'll push a change that deletes it,

datner: Pushed

packages/blitz/src/cli/index.ts
```diff
@@ -36,23 +35,22 @@ const commands: {[command: string]: () => Promise<CliCommand>} = {
...
 }
-const foundCommand = Boolean(commands[args._[0] as string])
+
+const aliases: typeof commands = {
+  d: commands.dev!,
+  b: commands.build!,
+  s: commands.start!,
+  n: commands.new!,
+  g: commands.generate!,
+}
```
beerose, line 45: We need to improve this a little — `aliases` is not a type of `commands`, because it has different keys. What we want instead is to say that `aliases` is an object with custom keys and values from the `commands` object. 

In this case, we don't have to write anything because the type will be inferred and TS will know the proper types.  In many cases — the fewer types the better. What's more, I'm also going to push PR to main that removes the type of `const commands` — we can let TS infer the types for us. We do it because sometimes if we type "too much", we accidentaly hide/override what TS would infer.

beerose (changes requested):

packages/blitz/src/cli/index.ts
beerose: (I took the liberty to push this change when resolving conflicts)

beerose (approved):
I took the liberty to push a few small changes — wanted to get it merged today (though they are kicking me out from a coffee shop so will probably get back to it later 😅)

Again THANK YOU for the PR!

PR #3421, opened 2022-06-09
Expand & rename baseResolverPath

iDavidB:

- Renamed `baseResolverPath` to `resolverPath`
	- As it is now allowed to be a function, it doesn't feel logical to prefix it with `base`. I'm not 100% keen on `resolverPath` either, so i'm open to suggestions. Maybe `resolveResolverPath`?
- Allow `resolverPath` to be a function which gets ran for every filePath that needs to be converted
	- This allows people to write their own custom logic for converting file paths to RPC routes. Should this allow async actions? 
- Update TS types to be a little more specific


packages/blitz-next/src/index-server.ts
```diff
@@ -133,59 +139,13 @@ export const setupBlitzServer = ({plugins}: SetupBlitzOptions) => {
 
 export interface BlitzConfig extends NextConfig {
   blitz?: {
-    resolverBasePath?: ResolverBasePath
+    resolveResolverPath?: ResolverPathOptions
```
beerose, line 142: I think I prefer `resolverPath`. `resolveResolverPath` strongly suggests that it can be a function while in most cases (I think) it will be a string value.

beerose (commented):
Left one comment about the name — otherwise looks good to me.

packages/blitz-next/src/index-server.ts
iDavidB: I agree that in most cases it will likely be a string, but I don't necessarily agree that it suggests it should be a function. In my eyes it's just a more clear way to say how you want each file path to be resolved. 

But I don't really mind what which one it should be, won't really make a difference to me. 

I am currently on vacation and not in the position to make any changes to this PR, so if you want to make this chance so it can at least be merged and added to a new alpha, then feel free.

Issue #3432, opened 2022-06-12
Fix codemod for nested pages directory


dillonraphael:
There seems to be an issue with nested directories consolidating to the root page directory. (eg. `app/consultants/pages/consultants/[consultantId]/profile.tsx`)

PR #3443, opened 2022-06-14
Codemod fixes

itsdillon:
Closes: #3433, #3432, #3435

- Fix codemod for wrapping _app step that uses an arrow function
- Fix codemod for nested pages directory
- Copy over cookiePrefix from blitz config to blitz client file


.changeset/fuzzy-jars-admire.md
```diff
@@ -0,0 +1,5 @@
+---
+"@blitzjs/codemod": patch
+---
+
+fix codemod for wrapping \_app arrow function & fix codemod for nested pages directory
```
beerose, line 5: Can we rephrase it to be more user-facing? Ie what did it change from the user's perspective

Issue #3200, opened 2022-03-09
Support RPC error middleware


flybayer:
Make `onError` in rpcHandler work

PR #3460, opened 2022-06-23
Hook up onError function for rpc server and next

Zeko369:
Closes: #3422, #3200

- Adds `onError` for next server init
- Hooks up `onError` from `rpcHandler`

### DIFF
--- a/.changeset/late-steaks-give.md
+++ b/.changeset/late-steaks-give.md
@@ -1,0 +1,6 @@
+---
+"@blitzjs/next": patch
+"@blitzjs/rpc": patch
+---
+
+Support RPC error middleware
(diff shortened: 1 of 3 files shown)


beerose (dismissed):
Perfect! Thank you a lot for adding this!

Zeko369:
Hmm but we have 2 onErrors, one for BlitzServer and one for RPCHandler?

The only thing actually missing is that BlitzServer one is only used in API routes not in SSR/SSG and that might me useful?

Zeko369:
Update on the last comment, I added error handling into GSSP / GSP / API functions

beerose:
Ok, after further thought, I think it's good. Thanks again for working on that — it's super helpful !!

beerose (approved):

PR #3461, opened 2022-06-26
Correctly type GSSP, again

Zeko369:
Continuation of #3457 since I figured out I didn't pass the generic all types 🤦

now getServerSideProps will be correctly typed when passing a specific type instead of `ParsedUrlQuery`


beerose (approved):
Thank you!

Issue #3517, opened 2022-07-07
There is an issue with prefetching queries in SSR/SSG


Zeko369:
### What is the problem?

Related to this #3199 / #3357

1st this infinite/not infinite query logic is wrong, it should be reversed
[queryKey](https://github.com/blitz-js/blitz/blob/830b51d0dc3674591c5df912322d443860706e54/packages/blitz-next/src/index-server.ts#L107)

2nd `getQueryKey` is returning undefined as the queryName when run on backend (i.e. gSSP)

### Paste all your error logs here:

It's undefined

### Paste all relevant code snippets here:

`console.log(getQueryKey(getProject));` work on the client and not on the server

### What are detailed steps to reproduce this?

```
import React, { Suspense } from 'react';
import { BlitzPage } from '@blitzjs/next';
import { getQueryKey, useQuery } from '@blitzjs/rpc';
import { gSSP } from 'app/blitz-server';

import getProject from 'app/data/projects/queries/public/getProject';

const Thingy: React.FC = () => {
  const [project] = useQuery(getProject, { slug: 'sizzy' });
  return <code style={{whiteSpace: "pre"}}>{JSON.stringify(project, null, 2)}</code>;
};

const ThingyPage: BlitzPage = (props) => {
  console.log(props);
  console.log(getQueryKey(getProject));

  return (
    <Suspense fallback={<h1>Loading...</h1>}>
      <Thingy />
    </Suspense>
  );
};

export const getServerSideProps = gSSP(async (ctx) => {
  await ctx.ctx.prefetchInfiniteQuery(getProject, { slug: 'sizzy' });

  return {
    props: {},
  };
});

export default ThingyPage;
```

### Run `blitz -v` and paste the output here:

yarn run v1.22.17
$ ./node_modules/blitz/bin/blitz -v
Loaded env from /Users/franzekan/zekit/appjourney/.env.local
Loaded env from /Users/franzekan/zekit/appjourney/.env
Blitz version: 2.0.0-alpha.44 (local)
macOS Monterey | darwin-arm64 | Node: v16.15.1

 Package manager: npm

  System:
    OS: macOS 12.4
    CPU: (10) arm64 Apple M1 Pro
    Memory: 133.58 MB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
[...]

PR #3518, opened 2022-07-07
Fix SSP / SP not prefetching queries correctly

Zeko369:
Closes: #3517

## Bug Checklist

- [x] Infinite and non infinite getQueryKey were reversed
- [x] Added a Webpack loader that transforms resolvers and adds metadata to them so they can be used in getQueryKey on the server as well as on the client
- [x] Fix key error when loading `dehydratedState`

### DIFF
--- a/.changeset/nine-birds-confess.md
+++ b/.changeset/nine-birds-confess.md
@@ -1,0 +1,7 @@
+---
+"@blitzjs/next": patch
+"@blitzjs/rpc": patch
+"blitz": patch
+---
+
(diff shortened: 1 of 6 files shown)


dillonraphael:
@Zeko369 Thanks for this. Some nice fixes. I believe @flybayer wants to thoroughly review and to have some sort of unit test.

Zeko369:
I agree that we need some unit tests for this, I'll write them later today

flybayer (approved):
Looks good, thank you!

Issue #3521, opened 2022-07-08
blitz new . choses the wrong name


Zeko369:
### What is the problem?

Running `blitz new .` should use the name of the current folder instead of . (which breaks package.json)

### Paste all your error logs here:

```
✔ Retrieving the freshest of dependencies
package.json: Name can't start with a dot
✔ Committing your app
```

### Paste all relevant code snippets here:

```
blitz new .
```

### What are detailed steps to reproduce this?

1. create a folder `foobar` and in it
2. run `blitz new .` and select yarn as the package manager (didn't test with others)

### Run `blitz -v` and paste the output here:

```
Loaded env from /Users/franzekan/sandbox/toolkit-test/.env.local
Loaded env from /Users/franzekan/sandbox/toolkit-test/.env
Blitz version: 2.0.0-alpha.53 (global)
Blitz version: 2.0.0-alpha.53 (local)
macOS Monterey | darwin-arm64 | Node: v16.15.1


 Package manager: npm

  System:
    OS: macOS 12.4
    CPU: (10) arm64 Apple M1 Pro
    Memory: 90.36 MB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.15.1 - ~/.nvm/versions/node/v16.15.1/bin/node
    Yarn: 1.22.17 - /opt/homebrew/bin/yarn
    npm: 8.11.0 - ~/.nvm/versions/node/v16.15.1/bin/npm
  npmPackages:
    @prisma/client: 3.9.0 => 3.9.0 
    blitz: 2.0.0-alpha.53 => 2.0.0-alpha.53 
    next: 12.2.0 => 12.2.0 
    prisma: 4.0.0 => 4.0.0 
    react: 18.0.0 => 18.0.0 
    react-dom: 18.0.0 => 18.0.0 
    typescript: ^4.5.3 => 4.7.4 
```

PR #3523, opened 2022-07-08
If project name argument is set to "." change it to current folder name

Zeko369:
Closes: #3521

## Bug Checklist

- [x] `blitz new .` used to name the project `.` causing yarn to fail, now it reads the directory name


dillonraphael (approved):

beerose:
Can we make sure that we cover these things in the docs? That could be just a single sentance

PR #3537, opened 2022-07-15
fix(blitz-rpc): fix 6th pipe resolver not returning resolved value

datner:

exact same as #3538 but for the canary branch

No idea how this wasn't caught by now, but the 6th function in a given pipe will return the `Ctx` instead of the value yielded by the function (according to the type info) 🤷‍♂️

```ts
// this yields Ctx
export default resolver.pipe(
  fn1,
  fn2,
  fn3,
  fn4,
  fn5,
  fn6
)
```

## Bug Checklist

## Feature Checklist

### DIFF
--- a/.changeset/perfect-eyes-repeat.md
+++ b/.changeset/perfect-eyes-repeat.md
@@ -1,0 +1,5 @@
+---
+"@blitzjs/rpc": patch
+---
+
+Fix pipe resolver return type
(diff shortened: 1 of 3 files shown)


datner:
No idea why it would fail, literally changed just the type info from broken to working 🤔

beerose (approved):
I guess no one has used 6-level pipe so far 😅

Thank you for the fix! 

I'll handle the failing CI thing — it looks like pnpm.lock it outdated.

PR #3538, opened 2022-07-15
fix(blitz-rpc): fix 6th pipe resolver not returning resolved value -- canary

datner:

exact same as #3537 but for the canary branch

No idea how this wasn't caught by now, but the 6th function in a given pipe will return the `Ctx` instead of the value yielded by the function (according to the type info) 🤷‍♂️

```ts
// this yields Ctx
export default resolver.pipe(
  fn1,
  fn2,
  fn3,
  fn4,
  fn5,
  fn6
)
```

## Bug Checklist

## Feature Checklist


beerose:
Since we recently moved the framework code to https://github.com/blitz-js/legacy-framework/, I'll close this PR. Thank you for contributing, though, and thank you for catching this error!

beerose:
No, actually the toolkit/Blitz 2.0 code is in the `main` branch currently

Issue #3550, opened 2022-07-17
prefetchQuery in ssr/static only keeps last query


Zeko369:
Credit to @iDavidB for finding the bug, thread on discord 998248548854399047

### What is the problem?

When prefetching multiple queries only the last one is actually saved

### Paste all your error logs here:

Doens't work

### Paste all relevant code snippets here:

[Reproduction](https://codesandbox.io/s/prefetch-repro-5b25q9?file=/pages/prefetch.tsx)

### What are detailed steps to reproduce this?

1. codesandbox

### Run `blitz -v` and paste the output here:

```
latest
```

PR #3551, opened 2022-07-17
Fix prefetching multiple queries causes only the last one to be passed to page

Zeko369:
Closes: #3550

## Bug Checklist

- [x] Initialise QueryClient only the first time (instead of on every prefetch)


beerose (approved):
Thanks!

PR #3562, opened 2022-07-18
Add redis storage implementation in Auth

StringKe:
Closes: ?

Add redis storage implementation in Auth,

Supports both [ioredis](https://www.npmjs.com/package/ioredis) and [redis](https://www.npmjs.com/package/redis)

**Since Pipeline doesn't have a redis backend set up, I don't know how unit tests for redis can be added to the project and I need some help.**

change:
1. Separate the prisma implementation from the [auth-plugin](https://github.com/StringKe/blitz/tree/48b30eeb9c7e42585869326e0a00eb813af93f25/packages/blitz-auth/src/server/auth-plugin.ts) (move it to the [plugins](https://github.com/StringKe/blitz/tree/48b30eeb9c7e42585869326e0a00eb813af93f25/packages/blitz-auth/src/server/plugins) directory) and implement redis in a new directory ([plugins](https://github.com/StringKe/blitz/tree/48b30eeb9c7e42585869326e0a00eb813af93f25/packages/blitz-auth/src/server/plugins))

## Bug Checklist

## Feature Checklist


packages/blitz-auth/package.json
```diff
@@ -31,10 +31,12 @@
     "cookie-session": "2.0.0",
     "debug": "4.3.3",
     "http": "0.0.1-security",
+    "ioredis": "5.2.1",
```
beerose, line 34: I think we should find a solution where we avoid adding Redis to the dependencies. Currently, everyone using blitz-auth will end up having ioredis and redis in their deps even if they aren't using it.

beerose (commented):
Thanks a lot for starting to work on that!

I left one comment with something up for a discussion

packages/blitz-auth/package.json
StringKe: Both of these dependencies can be added to devDependencies , taking into account the typescript definitions.

Maybe it is possible to provide an abstraction layer up to the user concrete implementation to decide which one to use?

Zeko369: Maybe have `@blitzjs/auth-backend-redis`?

beerose: > Maybe have `@blitzjs/auth-backend-redis`?

Yeah, I was thinking about `@blitzjs/auth/redis` or something. Not sure yet what would be best here

flybayer: Can we keep it in `@blitzjs/auth` but move redis stuff to devDeps and have users install it themselves?

That's what we do with Prisma and Zod, and it works well

lhocke:
@Zeko369 pointed me to this and I've been testing it out with ioredis, ran into a few issues when trying to actually create the session and add it to the blitz context.

1. ```expires``` needs to be converted to an integer before it can be passed into ```redis.set``` or it throws an error
2. ```getSession``` needs to return a ```Promise<SessionModel>``` to properly serialize a session on login or else it just creates an entry in the database that is inaccessible on the client.
3. attempting to resolve the parsed data from a redis query into blitz-context will throw ```date.getTime() is not a function``` unless ```expiresAt``` is first converted to a Date object (values are stored as strings in Redis)

I've written a potential replacement for [packages/blitz-auth-ioredis/src/index-server.ts](https://github.com/blitz-js/blitz/pull/3562/files#diff-45460f1d865d1d5a47e427e749e38b9ad61331844bbedc1b7d11adf1d36f21cc) that resolves these issues, that I can either include as a code snippet or fork and do a PR, whichever you'd prefer

packages/blitz-auth/src/server/plugins/auth-primsa.ts
```diff
@@ -0,0 +1,50 @@
+import {PublicData, SessionConfigMethods, SessionModel} from "../../shared"
```
beerose, line 1: There's a type in the filename (primsa -> prisma)

packages/blitz-auth-redis/test/sample.test.ts
```diff
@@ -0,0 +1,13 @@
+import {assert, expect, test} from "vitest"
```
beerose, line 1: Are you planning to add tests for the auth-redis?

PR #3578, opened 2022-07-21
Fix prefetching, again...

Zeko369:
Will this finally fix it or not? :(

### DIFF
--- a/.changeset/silly-apricots-share.md
+++ b/.changeset/silly-apricots-share.md
@@ -1,0 +1,5 @@
+---
+"@blitzjs/next": patch
+---
+
+Avoid reassigning queryClient in prefetch methods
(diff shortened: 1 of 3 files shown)


beerose:
Could you explain briefly what was wrong and why this change fixes it? Because at first glance, it looks like it should do the same thing

Zeko369:
@beerose That's why I though the original one would work without testing, but it didn't 😅.

So, the old version returns the reference to the variable which CAN be reassigned in the prefetchQuery functions -> causing the returned variable to have the old value (null) instead of the new QueryClient.
Having a lambda that can be called after the fact will in turn always return the current value.

beerose:
Okay, thanks. Have you tested it or should I?

Zeko369:
I tested this and it works

beerose (approved):

Issue #3579, opened 2022-07-21
pnpm: Brand new app fails with Error: No QueryClient set


flybayer:
1. Generate a brand new app with `2.0.0-alpha.56` 
   - have to use `blitz new testblitz --skip-upgrade` flag because of https://github.com/blitz-js/blitz/issues/3566
   - select pnpm
2. Run `blitz build`

Get failed build with this:

```
Error: No QueryClient set, use QueryClientProvider to set one
    at useQueryClient (/private/tmp/testblitz/node_modules/.pnpm/react-query@3.39.0_zpnidt7m3osuk7shl3s4oenomq/node_modules/react-query/lib/react/QueryClientProvider.js:36:11)
    at useBaseQuery (/private/tmp/testblitz/node_modules/.pnpm/react-query@3.39.0_zpnidt7m3osuk7shl3s4oenomq/node_modules/react-query/lib/react/useBaseQuery.js:24:61)
    at Object.useQuery (/private/tmp/testblitz/node_modules/.pnpm/react-query@3.39.0_zpnidt7m3osuk7shl3s4oenomq/node_modules/react-query/lib/react/useQuery.js:14:41)
    at useQuery (/private/tmp/testblitz/node_modules/.pnpm/@blitzjs+rpc@2.0.0-alpha.56_mwbxcnklvkvmbhor4xlfqgl32e/node_modules/@blitzjs/rpc/dist/index-browser.cjs:29:45)
    at useCurrentUser (/private/tmp/testblitz/.next/server/pages/index.js:64:37)
    at UserInfo (/private/tmp/testblitz/.next/server/pages/index.js:97:25)
    at Hc (/private/tmp/testblitz/node_modules/.pnpm/react-dom@18.0.0_react@18.0.0/node_modules/react-dom/cjs/react-dom-server.browser.production.min.js:64:191)
    at Kc (/private/tmp/testblitz/node_modules/.pnpm/react-dom@18.0.0_react@18.0.0/node_modules/react-dom/cjs/react-dom-server.browser.production.min.js:66:253)
    at Z (/private/tmp/testblitz/node_modules/.pnpm/react-dom@18.0.0_react@18.0.0/node_modules/react-dom/cjs/react-dom-server.browser.production.min.js:71:89)
    at Lc (/private/tmp/testblitz/node_modules/.pnpm/react-dom@18.0.0_react@18.0.0/node_modules/react-dom/cjs/react-dom-server.browser.production.min.js:73:98)
[...]

PR #3585, opened 2022-07-22
Hoist up react-query

Zeko369:
Closes: #3579

## Bug Checklist

- [x] When using `pnpm` builds / server renders of anything using ReactQuery (BlitzRPC) would fail because it would be using 2 different react-query packages


beerose (approved):
Perfect! thank you!

Issue #3555, opened 2022-07-17
Hook up react-query fetch cancellation


flybayer:
### What do you want and why?

With just a small change, we can hook up request cancelation with react-query.

Docs: https://react-query-v2.tanstack.com/guides/query-cancellation

### Possible implementation(s)

- Pass through the `signal` to each of our react-query hook uses, first one here: https://github.com/blitz-js/blitz/blob/main/packages/blitz-rpc/src/data-client/react-query.tsx#L88
- Update [our fetch function](https://github.com/blitz-js/blitz/blob/main/packages/blitz-rpc/src/data-client/rpc.ts#L60) to accept `signal` and pass it to `fetch`

PR #3590, opened 2022-07-22
Hook up react-query fetch cancellation

Zeko369:
Closes: #3555

## Feature Checklist

- [x] Pass `signal` from `useQuery` to our `__internal_buildRpcClient` to be able to cancel queries on unmount

### DIFF
--- a/packages/blitz-rpc/src/data-client/rpc.ts
+++ b/packages/blitz-rpc/src/data-client/rpc.ts
@@ -39,7 +39,7 @@ export interface EnhancedRpc {
 }
 
 export interface RpcClientBase<Input = unknown, Result = unknown> {
-  (params: Input, opts?: RpcOptions): Promise<Result>
+  (params: Input, opts?: RpcOptions, signal?: AbortSignal): Promise<Result>
 }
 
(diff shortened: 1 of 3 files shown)


packages/blitz-rpc/src/data-client/rpc.ts
```diff
@@ -93,9 +93,6 @@ export function __internal_buildRpcClient({
       serialized = serialize(params)
     }
 
-    // Create a new AbortController instance for this request
-    const controller = new AbortController()
-
```
beerose, line 98: Hmm, shouldn't we still use this in case `signal` argument is undefined?

Zeko369: Unless we want to wire it up to something, this won't do anything by itself so 🤷‍♂️

beerose (approved):

beerose:
I forgot to mention that before merging, but we also need docs for that. We ideally shouldn't change the PR template

beerose:
Hey @Zeko369, are you going to add docs for that, or you want me to do it?

Issue #3598, opened 2022-07-25
Next Static Revalidate from mutation/query


Zeko369:
### What do you want and why?

It would be amazing to call `ctx.revalidate('/foo')` or even better to support our custom URL generation thingy `ctx.revalidate(Routes.Foo())`

### Possible implementation(s)

In the following PR

PR #3599, opened 2022-07-25
Next revalidate page in mutation/query

Zeko369:
Closes: #3598

## Feature Checklist

- [x] decorate `blitzCtx` with `.revalidatePage`
- [x] make our `.revalidatePage` accept Routes
- [x] throw error if called in SP (could lead to recursion and bugs + would be hard to implement)
- [x] DOCS: https://github.com/blitz-js/blitzjs.com/pull/717


packages/blitz/src/middleware.ts
```diff
@@ -17,7 +26,9 @@ export async function handleRequestWithMiddleware<
   } = {},
 ) {
   if (!(res as unknown as MiddlewareResponse).blitzCtx) {
-    ;(res as unknown as MiddlewareResponse).blitzCtx = {} as Ctx
+    ;(res as unknown as MiddlewareResponse).blitzCtx = {
+      revalidatePage: revalidateFactory(res as unknown as NextApiResponse),
```
beerose, line 30: Can we avoid this assertion here? Since we already have a type assertion to `MiddlewareResponse` this feels kinda hacky

packages/blitz/src/middleware.ts
```diff
@@ -1,5 +1,14 @@
 import {IncomingMessage, ServerResponse} from "http"
 import {compose, Ctx, RequestMiddleware, MiddlewareNext, MiddlewareResponse} from "./index-server"
+import {UrlObject} from "url"
+import {NextApiResponse} from "next"
+import {resolveHref} from "next/dist/shared/lib/router/router"
+
+const revalidateFactory =
+  (res: NextApiResponse) =>
+  (url: UrlObject | string, opts?: Parameters<NextApiResponse["revalidate"]>[1]) => {
+    return res.revalidate(resolveHref({} as any, url, false), opts)
```
beerose, line 10: Why is `as any` needed?

packages/blitz/src/middleware.ts
```diff
@@ -1,5 +1,14 @@
 import {IncomingMessage, ServerResponse} from "http"
 import {compose, Ctx, RequestMiddleware, MiddlewareNext, MiddlewareResponse} from "./index-server"
+import {UrlObject} from "url"
+import {NextApiResponse} from "next"
+import {resolveHref} from "next/dist/shared/lib/router/router"
+
+const revalidateFactory =
```
beerose, line 7:
```suggestion
const revalidateFn =
```

beerose (changes requested):

packages/blitz/src/cli/index.ts
```diff
@@ -144,6 +144,7 @@ async function main() {
     console.warn(NON_STANDARD_NODE_ENV)
   }
 
+  // @ts-expect-error
```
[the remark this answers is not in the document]
Zeko369: image

packages/blitz/src/middleware.ts
Zeko369: It expects an instance of next/router but only uses it to resolve `#` paths, so we need to pass something here

packages/blitz/src/middleware.ts
Zeko369: Technically it would be better to abstract this whole thing to our `MiddlewareResponse` and move this logic into `@blitzjs/next` since this is NEXT specific (so converting all `MiddlewareResponse` to be `NextApiResponse` compliant seams bad since it's gonna make it harder for us to adopt i.e. nuxt/sveltkit...).

I'd keep this for now and handle this with better type overrides when we switch to true GENERIC TOOLKIT instead of NextToolkit

packages/blitz/src/middleware.ts
beerose: Can you change it to `as NextRouter`?

`NextRouter` imported as:
```
import {resolveHref, NextRouter} from "next/dist/shared/lib/router/router"
```

packages/blitz/src/middleware.ts
```diff
@@ -27,8 +27,8 @@ export async function handleRequestWithMiddleware<
 ) {
   if (!(res as unknown as MiddlewareResponse).blitzCtx) {
     ;(res as unknown as MiddlewareResponse).blitzCtx = {
-      revalidatePage: revalidateFactory(res as unknown as NextApiResponse),
-    } as Ctx
+      revalidatePage: revalidateFn(res as unknown as NextApiResponse),
```
beerose, line 30: Why are we assuming that `res` is assignable to the NextApiResponse type and has `revalidate` property?

In the handleRequestWithMiddleware, the response is typed as ServerResponse. It can be overwritten and extended, but the minimal accepted type is ServerResponse. Hence I don't think we can assume that it's NextApiResponse. Especially since this is `blitz` package which should remain next-agnostic (that's why we have own types, and we're not using Next's types anywhere in this file.

beerose: One solution that I can see here is moving this logic to blitz-next — i.e. having a middleware there.

beerose (changes requested):

packages/blitz/src/middleware.ts
Zeko369: Added to blitz-next

flybayer (changes requested):
@Zeko369 looks good, but need to fix CI and to add an integration test

Issue #3611, opened 2022-07-26
Env flag not working with blitz CLI


dillonraphael:
### What is the problem?

When running `blitz build -e test` it doesn't set `process.env.APP_ENV` to `process.env.APP_ENV=test`

PR #3629, opened 2022-07-28
Fix APP_ENV not being set before running loadEnvConfig

dillonraphael:
Closes: #3611

`loadEngConfig` was being run before checking for the `--env` flag. This moves the call location after the `--env` flag check.

### DIFF
--- a/.changeset/fuzzy-bees-warn.md
+++ b/.changeset/fuzzy-bees-warn.md
@@ -2,4 +2,4 @@
 "blitz": patch
 ---
 
-APP_ENV wasn't being set when using the -e flag. This fixes it.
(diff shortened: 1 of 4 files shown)


beerose (dismissed):

.changeset/fuzzy-bees-warn.md
```diff
@@ -0,0 +1,5 @@
+---
+"blitz": patch
+---
+
+APP_ENV wasn't being set when using the -e flag. This fixes it.
```
beerose, line 5: Can you rephrase it? It will be in in release notes, and it's not clear what "this" in the second sentence refers to.

Issue #3490, opened 2022-07-04
getQueryData function


edrickleong:
### What do you want and why?

I want an exported `getQueryData` function that takes in a resolver. 

Instead of having to get the query key manually, such as below
```ts
const queryKey = getQueryKey(getTasks, { taskListId })
const tasks = queryClient.getQueryData<PromiseReturnType<typeof getTasks>>(queryKey)
```

I want to be able to write
```ts
const tasks = getQueryData(getTasks, { taskListId }))
```

### Possible implementation(s)

Add a function `getQueryData` similar to the `setQueryData` function defined in `react-query-utils` as below. https://github.com/blitz-js/blitz/blob/e1fc695d24f6dd85eba8ffce9f88e357edfb562d/nextjs/packages/next/data-client/react-query-utils.ts#L166

### Additional context

The main use case I have for this is to do optimistic updates. See [example](https://react-query.tanstack.com/guides/optimistic-updates#updating-a-list-of-todos-when-adding-a-new-todo) from react query docs.

pivotrooky:
Hi!
I would like to work this issue since it is marked as a good first issue.
Would that be ok?

abuuzayr:
absolutely @pivotrooky ! give us a shout if you need any help

Himanxu1:
Hey , I want to work on this issue ..
Assign this to me .
@abuuzayr

pivotrooky:
Oh, ok. @Himanxu1 I'd only just started, so fine by me :D @abuuzayr 

> Hey , I want to work on this issue .. Assign this to me . @abuuzayr

Himanxu1:
OK @pivotrooky .I am also a beginner

beerose:
Hey @Himanxu1, I think @pivotrooky meant that he was already assigned to this issue :) In general, it's helpful to look at the discussion and see if someone was already assigned or created a PR or mentioned that they are going to work on an issue. This way, we can avoid working on the same things and any future confusion. I hope you can decide between you two who's going to work on this one :)

pivotrooky:
@beerose I'd like to free up this issue, there'll be plenty of issues to work on in the future.
Could you please unassign me? 
I'm not sure if the option to do it myself is available as I'm not yet a collaborator but, if it is, I can't find seem to find the button. 
Thank you very much!

edrickleong:
I have talked to @Himanxu1 about picking up this ticket on discord and he's ok with passing this to me. Could this be assigned to me instead?

[...]

PR #3648, opened 2022-08-01
add getQueryData

edrickleong:
Closes: https://github.com/blitz-js/blitz/issues/3490

A new function `getQueryData` will be exposed. 

## Feature Checklist

- [x] Documentation added/updated (https://github.com/blitz-js/blitzjs.com/pull/721)


integration-tests/trailing-slash/test/index.test.ts
```diff
@@ -24,6 +24,25 @@ const runTests = (mode?: string) => {
...
+
+        browser.waitForElementByCss("#new-data", 0)
+        const newText = await browser.elementByCss("#new-data").text()
+        expect(newText).toMatch(/basic-result/)
+
+        if (browser) await browser.close()
+      },
+      5000 * 60 * 2,
+    )
+  })
```
beerose, line 45: This is awesome, but is it possible to have this test in `integration-tests/rpc` or in a new folder? Asking because tests in this folder should only be related to trailing slash.

edrickleong: Sure. I’ll have a look at this tomorrow

edrickleong: Hi @beerose, I need help with this. I'm trying to add this into the `rpc` integration test but that means pulling in `@blitzjs/rpc` and also `@blitzjs/auth` (because of https://github.com/blitz-js/blitz/issues/3663). Pulling in `@blitzjs/auth` means I have to add prisma. Unfortunately, serverless mode which is part of the tests, doesn't work with this. What should I do here?

beerose: Ah, I see.

Hmm, there's only one test in the `trailing-slash`, and to be honest, I'm not even sure if it's testing the trailing slash thing correctly (https://nextjs.org/docs/api-reference/next.config.js/trailing-slash) 🤷‍♀️

I see two options:
- Renaming this directory `trailing-slash` -> `react-query-utils`. But then we lose the trailing-slash context.
- Copying the whole trailing slash directory into `integration-tests/react-query-utils` and putting this test there (over time, we should have more tests there).

edrickleong: Thanks. I've duplicated the trailing slash directory into a new `react-query-utils` and the tests are working locally now.

beerose (approved):
Thank you a lot for your work on this PR!

PR #3671, opened 2022-08-04
Export `BlitzServerMiddleware` utility function

dillonraphael:

Exports a `BlitzServerMiddleware()` function from `blitz` that allows you to wrap legacy blitz middleware functions.

An example:
```ts
BlitzServerMiddleware(BlitzGuardMiddleware({
  excluded: [
   '/api/rpc/signup'
  ]
})),
```

## Feature Checklist

- [X] Documentation added/updated (https://github.com/blitz-js/blitzjs.com/pull/723)

### DIFF
--- a/.changeset/great-terms-rescue.md
+++ b/.changeset/great-terms-rescue.md
@@ -1,0 +1,5 @@
+---
+"blitz": patch
+---
+
+Add BlitzServerMiddleware utility function to wrap middleware in blitz server file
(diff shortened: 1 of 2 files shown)


flybayer:
@dillonraphael what about adding `middleware` as an option to `setupBlitzServer` instead of requiring a plugin adapter here?

dillonraphael:
> @dillonraphael what about adding `middleware` as an option to `setupBlitzServer` instead of requiring a plugin adapter here?

I think this will get confusing for the user to differentiate between nextjs middleware and middleware that is specific for blitz.

Issue #3566, opened 2022-07-19
(CLI version check error) Cannot find module create noop.js


DmacMcgreg:
### What is the problem?

blitz@alpha version 55 fails to create new project

### Paste all your error logs here:

```
➜  6_Code blitz new test-project
Error: Cannot find module 'blitz'
Require stack:
- /Users/d/Documents/6_Code/noop.js
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:933:15)
    at resolveFileName (/Users/d/.nvm/versions/node/v16.13.0/lib/node_modules/blitz/node_modules/resolve-from/index.js:29:39)
    at resolveFrom (/Users/d/.nvm/versions/node/v16.13.0/lib/node_modules/blitz/node_modules/resolve-from/index.js:43:9)
    at module.exports (/Users/d/.nvm/versions/node/v16.13.0/lib/node_modules/blitz/node_modules/resolve-from/index.js:46:47)
    at findNodeModulesRoot (/Users/d/.nvm/versions/node/v16.13.0/lib/node_modules/blitz/dist/chunks/check-latest-version.cjs:34:10)
    at Object.checkLatestVersion (/Users/d/.nvm/versions/node/v16.13.0/lib/node_modules/blitz/dist/chunks/check-latest-version.cjs:46:35)
    at async newApp (/Users/d/.nvm/versions/node/v16.13.0/lib/node_modules/blitz/dist/chunks/new.cjs:200:5) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [ '/Users/d/Documents/6_Code/noop.js' ]
}
```

### Paste all relevant code snippets here:

```
npm i -g blitz@alpha
blitz new test-project
```

### What are detailed steps to reproduce this?

1. npm i -g blitz@alpha
2. blitz new test-project
3. error

OR

1. npx blitz@alpha new test-project

### Run `blitz -v` and paste the output here:

```
➜  6_Code blitz -v
Blitz version: 2.0.0-alpha.55 (global)
macOS Monterey | darwin-x64 | Node: v16.13.0


 Package manager: npm

  System:
    OS: macOS 12.4
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 278.05 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.13.0 - ~/.nvm/versions/node/v16.13.0/bin/node
[...]

Issue #3565, opened 2022-07-19
Improve CLI version check


beerose:
The current output:
<img width="501" alt="CleanShot_2022-07-19_at_11 53 262x" src="https://user-images.githubusercontent.com/9019397/179727481-e44c4877-028b-4ecf-9e20-26ec2e29ffe0.png">

- [x] We shouldn't suggest upgrading to latest if someone is using stable
- [x] The current version is missing
- [x] Consider printing info about version mismatch (e.g. "you're using blitz v 54, auth v50")

`blitz -v` output:

```
Blitz version: 2.0.0-alpha.55 (global)
Blitz version: 2.0.0-alpha.55 (local)
macOS Monterey | darwin-arm64 | Node: v16.15.0


 Package manager: npm

  System:
    OS: macOS 12.2
    CPU: (10) arm64 Apple M1 Pro
    Memory: 944.30 MB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.15.0 - /opt/homebrew/opt/node@16/bin/node
    Yarn: 1.22.18 - ~/.volta/bin/yarn
    npm: 8.5.5 - /opt/homebrew/opt/node@16/bin/npm
  npmPackages:
    @blitzjs/auth: 2.0.0-alpha.55 => 2.0.0-alpha.55 
    @blitzjs/next: 2.0.0-alpha.55 => 2.0.0-alpha.55 
    @blitzjs/rpc: 2.0.0-alpha.55 => 2.0.0-alpha.55 
    @prisma/client: 3.9.0 => 3.9.0 
    blitz: 2.0.0-alpha.55 => 2.0.0-alpha.55 
    next: 12.2.0 => 12.2.0 
    prisma: 4.0.0 => 4.0.0 
    react: 18.0.0 => 18.0.0 
    react-dom: 18.0.0 => 18.0.0 
    typescript: ^4.5.3 => 4.7.4 
```

dillonraphael:
As reference https://github.com/blitz-js/blitz/issues/3625

PR #3675, opened 2022-08-05
CLI Version check fixes

Zeko369:
Closes: #3566 #3565

- Fix types
- Add a `getGlobalPkgManager` helper function
- Run update check only on `blitz new`
- Remove checking local package manager

The issue before regarding noop is because when running `blitz new` it wasn't able to find the local package manager. But now with `getGlobalPkgManager` helper function we can determine if the user is using pnpm or yarn globally.

## Bug Checklist

## Feature Checklist


Zeko369 (commented):
Approve (I created this so I can't approve)

packages/blitz/src/cli/utils/check-latest-version.ts
```diff
@@ -23,153 +21,43 @@ function getUpdateString(packageName: string, tag: string, isGlobal?: boolean) {
...
           if (version[0] === "globalVersion") {
-            const versionType = resolveVersionType(version[1] as string)
-
-            if (remoteBlitzVersions.hasOwnProperty("beta") && versionType !== "beta") {
-              errors.push({
-                message: `blitz(global) (current) ${version[1]} -> (latest) ${remoteBlitzVersions["beta"]}`,
-                instructions: `${getUpdateString("blitz", "beta", true)}`,
-              })
-            } else if (remoteBlitzVersions[versionType] !== version[1]) {
+            if (remoteBlitzVersions["latest"] !== version[1]) {
```
beerose, line 40: `version[1]` can be empty:
CleanShot 2022-09-20 at 14 31 37@2x

Someone can decide not to install blitz globally (e.g. they only use local version or through npx). We shouldn't print this message in this case.

packages/blitz/src/cli/utils/check-latest-version.ts
```diff
@@ -23,153 +21,43 @@ function getUpdateString(packageName: string, tag: string, isGlobal?: boolean) {
...
               errors.push({
-                message: `blitz(global) (current) ${version[1]} -> (latest) ${remoteBlitzVersions[versionType]}`,
-                instructions: `${getUpdateString("blitz", versionType, true)}`,
```
beerose, line 92: Why are we not handling blitz modules anymore?

beerose (changes requested):

packages/blitz/src/cli/utils/check-latest-version.ts
dillonraphael: Since we're only running version check on the `new` command there wasn't a need to check the modules. I remember @flybayer asking for this.

packages/blitz/src/cli/utils/check-latest-version.ts
dillonraphael: Ok just pushed an update @beerose. My original though process was that if they're doing `npx blitz new` they will get the latest version any way.

packages/blitz/src/cli/utils/check-latest-version.ts
beerose: Ok, I see, that makes sense. Though I saw that you added it back. So what's the final decision?

packages/blitz/src/cli/utils/check-latest-version.ts
```diff
@@ -35,11 +36,30 @@ export async function checkLatestVersion() {
...
+                message: `blitz(global) (current) ${versionValue} -> (latest) ${remoteBlitzVersions["latest"]}`,
+                instructions: `${getUpdateString("blitz", "latest", true)}`,
+              })
+            }
+          } else if (
+            versionType === "localVersions" &&
+            typeof versionValue === "object" &&
+            versionValue.blitz !== ""
+          ) {
+            console.log(versionValue)
```
beerose, line 59: To be removed

packages/blitz/src/cli/utils/check-latest-version.ts
```diff
@@ -1,19 +1,17 @@
...
 import {isInternalBlitzMonorepoDevelopment} from "./helpers"
 
 const returnNpmEndpoint = (packageName: string) => {
   return `https://registry.npmjs.org/-/package/${packageName}/dist-tags`
 }
 
 function getUpdateString(packageName: string, tag: string, isGlobal?: boolean) {
-  const pkgManager = getPkgManager()
-  switch (pkgManager) {
+  // const pkgManager = getPkgManager()
```
beerose, line 11:
```suggestion
```

packages/blitz/src/cli/utils/check-latest-version.ts
```diff
@@ -23,153 +21,63 @@ function getUpdateString(packageName: string, tag: string, isGlobal?: boolean) {
...
-      dotBlitzCache = fs.readFileSync(join(nodeModulesRoot, ".blitz", "checkUpdateCache.json"))
-      const now = new Date()
-      const msBetweenTimes = Math.abs(
-        superjson.parse<{lastUpdated: Date}>(dotBlitzCache.toString()).lastUpdated.getTime() -
-          now.getTime(),
-      )
-      const hoursBetweenTimes = msBetweenTimes / (60 * 60 * 1000)
-      shouldRun = hoursBetweenTimes > 24
-    }
+    let shouldRun = true
```
beerose, line 31: It looks like it's redundant (the value is never changed)

packages/blitz/src/cli/utils/check-latest-version.ts
```diff
@@ -23,153 +21,63 @@ function getUpdateString(packageName: string, tag: string, isGlobal?: boolean) {
...
-        const remoteBlitzNextVersions = (await blitzNextResponse.json()) as Record<string, string>
-
-        const blitzAuthResponse = await fetch.default(returnNpmEndpoint("@blitzjs/auth"))
-        const remoteBlitzAuthVersions = (await blitzAuthResponse.json()) as Record<string, string>
-
-        const blitzRpcResponse = await fetch.default(returnNpmEndpoint("@blitzjs/rpc"))
-        const remoteBlitzRpcVersions = (await blitzRpcResponse.json()) as Record<string, string>
+        for (const version of versionsArray) {
+          const versionType: string = version[0]
+          const versionValue: typeof version[1] = version[1]
```
beerose, line 41: This is inferred automatically, you don't have to type it manually. 
```suggestion
          const versionValue = version[1]
```

packages/blitz/src/cli/utils/check-latest-version.ts
dillonraphael: I added it back just for the local blitz package. Incase of the edge case of a user generating another app inside a monorepo and happen use the local blitz package to generate the new project.

packages/blitz/src/cli/utils/check-latest-version.ts
dillonraphael: Ah thanks

beerose (commented):
Looks good, but we should fix the indent too:

CleanShot 2022-09-22 at 12 57 35@2x

Should be aligned. You can fix that by removing a space after `\n` in lines 71 and 73.

beerose (approved):

.changeset/honest-cherries-push.md
```diff
@@ -0,0 +1,5 @@
+---
+"blitz": patch
+---
+
+Fix CLI version check by only checking when generating a new blitz project & by using the global package manager
```
beerose, line 5: Add a global Blitz version check when generating a new Blitz project to ensure users use the latest Blitz.

Issue #3601, opened 2022-07-25
Update codemod step for updating package.json


dillonraphael:
### What is the problem?

Currently we have:

https://github.com/blitz-js/blitz/blob/6559c25f71326df4fa81ebfc0eb25208ecab5d28/packages/codemod/src/upgrade-legacy.ts#L85-L94

But we should only update blitz packages, next packages & zod (because of the requirement for `^3.17.0`.

This means:
```diff
      packageJsonPath.dependencies["react"] = "latest"
      packageJsonPath.dependencies["react-dom"] = "latest"
      packageJsonPath.dependencies["@blitzjs/next"] = "alpha"
      packageJsonPath.dependencies["@blitzjs/rpc"] = "alpha"
      packageJsonPath.dependencies["@blitzjs/auth"] = "alpha"
      packageJsonPath.dependencies["blitz"] = "alpha"
      packageJsonPath.dependencies["next"] = "latest"
+    packageJsonPath.dependencies["zod"] = "latest"
-    packageJsonPath.dependencies["prisma"] = "latest"
-    packageJsonPath.dependencies["@prisma/client"] = "latest"
-    packageJsonPath.devDependencies["typescript"] = isTypescript && "latest"
```

We should also avoid using `latest` & `alpha` tags, and instead should use a static version number.

PR #3678, opened 2022-08-05
Set correct packages versions in package.json with upgrade-legacy codemod

beerose:
Closes: https://github.com/blitz-js/blitz/issues/3601

Instead of hardcoding blitz packages versions to `alpha`, we now fetch the latest version and use it instead.


packages/codemod/src/utils.ts
```diff
@@ -326,3 +328,27 @@ export function replaceIdentifiers(
...
+  if (!packageJson.dependencies) {
+    packageJson.dependencies = {}
+  }
+
+  packageJson.dependencies["@blitzjs/next"] = blitzPkgVersion
+  packageJson.dependencies["@blitzjs/rpc"] = blitzPkgVersion
+  packageJson.dependencies["@blitzjs/auth"] = blitzPkgVersion
+  packageJson.dependencies["blitz"] = blitzPkgVersion
+  packageJson.dependencies["next"] = "12.2.0"
+  packageJson.dependencies["zod"] = "latest"
```
dillonraphael, line 351: I think we should add a hardcoded version for zod as well

beerose: Let's fetch latest for zod? Otherwise, we'd make codemod depending on zod releases

PR #3679, opened 2022-08-05
Fix template path for the generator

dillonraphael:

Corrects the path for the templates directory when using the blitz generator. 

```diff
- process.env.NODE_ENV === "test" ? "./templates" : "./templates",
+ process.env.NODE_ENV === "test" ? "./templates" : "./dist/templates",
```

### DIFF
--- a/.changeset/shaggy-carpets-brake.md
+++ b/.changeset/shaggy-carpets-brake.md
@@ -1,0 +1,5 @@
+---
+"@blitzjs/generator": patch
+---
+
+Fix template path for the generator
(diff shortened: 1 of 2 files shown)


beerose:
Btw. I noticed that locally (when running blitz in the monorepo) it resolves to `dist/templates` while it should use `templates`. It looks like we should have something like:
```
process.env.NODE_ENV === "test" || isBlitzMonorepo ? "./templates" : "./dist/templates",
```

It's not a big deal but leaving it here as a note for future reference.

PR #3690, opened 2022-08-07
update console.log to baseLogger in @blitz-rpc/src/index-server.ts

siddhsuresh:
Closes: #3614

## Bug Checklist

- [ ] Integration test added (see [test docs](https://blitzjs.com/docs/contributing#running-tests) if needed)

## Feature Checklist

- [ ] Integration test added (see [test docs](https://blitzjs.com/docs/contributing#running-tests) if needed)
- [ ] Documentation added/updated (submit PR to [blitzjs.com repo `canary` branch](https://github.com/blitz-js/blitzjs.com/tree/canary))

PR #3693, opened 2022-08-08
Fixes #3614 - changed console logs to standard logger

### LINKED ISSUE #3614: Blitz RPC logging is not using our standard logger and does not match legacy log output
### What do you want and why?

[All these console logs](https://github.com/blitz-js/blitz/blob/main/packages/blitz-rpc/src/index-server.ts#L171-L245) should be changed to use [our standard logger](https://github.com/blitz-js/blitz/blob/main/packages/blitz/src/logging.ts#L32).

All the logging code should be copied [from our legacy code](https://github.com/blitz-js/legacy-framework/blob/main/nextjs/packages/next/server/api-utils.ts#L174-L276). The logging output should match **exactly** the output of legacy blitz.

oltdaniel:
I tried to keep the logging format close to what has been present before, by also including the chalk package for coloring.

Closes: #3614 

As requested, the logging of the RPC package has been moved to the built-in standard logger. Additionally, the `chalk` packages has been used to be as close as possible to the logging as it was before. All tests passed locally.

> **INFO** There are two other PRs open for this issue, which both shouldn't be able to correctly run through the workflow. See #3690 and #3614 .

## Bug Checklist

## Feature Checklist

### DIFF
--- a/.changeset/great-candles-stare.md
+++ b/.changeset/great-candles-stare.md
@@ -1,0 +1,6 @@
+---
+"@blitzjs/rpc": patch
+"blitz": patch
+---
+
+Use internal branded blitz logger for @blitzjs/rpc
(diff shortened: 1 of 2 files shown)


dillonraphael:
Looks good! Thanks @oltdaniel

PR #3701, opened 2022-08-09
fix detecting `blitz.config.(ts|js)` config file

oltdaniel:
the presence of the tsconfig.json is not enough to indicate whether our blitz.config.*  is a typescript or javascript file. this commit fixes the detection and logging of which file will be moved.

Closes: None

The detection of the `blitz.config.*` file is no more precise, by ignoring the presence of the `tsconfig.json`, as we cannot determine based on that if the config file itself is written in Typescript. This fixes this detection and also fixes the logging of which file has been detected.

> **NOTE** I'm not sure if the usage of `[0]` is clean enough, maybe something like a `.reduce` is nicer? Or replace everything with some `if` code.

## Bug Checklist

## Feature Checklist


packages/codemod/src/upgrade-legacy.ts
```diff
@@ -33,8 +33,12 @@ const isInternalBlitzMonorepoDevelopment = fs.existsSync(
 
 const upgradeLegacy = async () => {
   let isTypescript = fs.existsSync(path.resolve("tsconfig.json"))
-  let blitzConfigFile = `blitz.config.${isTypescript ? "ts" : "js"}`
-  let isLegacyBlitz = fs.existsSync(path.resolve(blitzConfigFile))
+  let existingBlitzConfigFiles = ["ts", "js"].map((e) => `blitz.config.${e}`).filter(fs.existsSync)
+  if (existingBlitzConfigFiles.length !== 1 || existingBlitzConfigFiles[0] === undefined) {
+    throw new ExpectedError("Could not identify Legacy Blitz Config file")
+  }
+  let isLegacyBlitz = true
```
dillonraphael, line 40: I think we should check `existingBlitzConfigFiles` to set `isLegacyBlitz` status, since we run a check here https://github.com/blitz-js/blitz/blob/4b8b7182aafb01558649b1ea69f83ef9e6029498/packages/codemod/src/upgrade-legacy.ts#L1190

oltdaniel: With the new detection, the `isLegacyBlitz` will always be true. This is due to the change, that I throw an error if I couldn't find the legacy configuration. In the previous version, this wasn't implemented.

I pushed a new commit with a change to make it a little bit cleaner with the reduce function, but I see now, that this would now also work if there is a `ts` and `js` file present and ignore the other one. Which may not be desired. Is there any behavior you have in mind, that would be the best fit for the config detection?

dillonraphael: > 

Regarding `isLegacyBlitz` will always be true. I get what you mean. What I was trying to avoid is duplicate logic as we're checking `isLegacyBlitz` later in the file and throwing an error.

packages/codemod/src/upgrade-legacy.ts
```diff
@@ -33,8 +33,15 @@ const isInternalBlitzMonorepoDevelopment = fs.existsSync(
...
   let isTypescript = fs.existsSync(path.resolve("tsconfig.json"))
-  let blitzConfigFile = `blitz.config.${isTypescript ? "ts" : "js"}`
-  let isLegacyBlitz = fs.existsSync(path.resolve(blitzConfigFile))
+  let existingBlitzConfigFiles = ["ts", "js"]
+    .map((e) => `blitz.config.${e}`)
+    .map((e) => path.resolve(e))
+    .filter(fs.existsSync)
+  let blitzConfigFile = existingBlitzConfigFiles.reduce((_prev, current) => current, "")
+  if (blitzConfigFile === "") {
+    throw new ExpectedError("Could not identify Legacy Blitz Config file")
```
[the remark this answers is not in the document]
oltdaniel: No problem. As far as I can see, there is no other error later in the file regarding finding the config file.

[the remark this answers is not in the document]
dillonraphael: @oltdaniel This is what I meant https://github.com/blitz-js/blitz/blob/be39286393159dce2279fdfeec3b8da69652afdb/packages/codemod/src/upgrade-legacy.ts#L1211

Since we'll now throw the `ExpectedError()` at the top of the `upgradeLegacy` function like you have it.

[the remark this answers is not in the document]
dillonraphael: All that's left is the actual log.error ^ @oltdaniel

packages/codemod/src/upgrade-legacy.ts
```diff
@@ -323,8 +327,9 @@ const upgradeLegacy = async () => {
           })
           fs.writeFileSync(path.resolve(appDir, filename), program.toSource())
         } catch (e) {
-          log.error(`Error updating imports in the ${file}`)
-          throw new Error(e)
+          const error = e as string
+          log.error(`Error updating imports in the ${filename}`)
+          throw new Error(error)
```
beerose, line 332: Why are you doing this? Can't you instead throw `e`?

packages/codemod/src/upgrade-legacy.ts
```diff
@@ -317,8 +327,8 @@ const upgradeLegacy = async () => {
           })
           fs.writeFileSync(path.resolve(appDir, filename), program.toSource())
         } catch (e) {
-          log.error(`Error updating imports in the ${file}`)
-          throw new Error(e)
+          log.error(`Error updating imports in the ${filename}`)
+          throw new Error(e as string)
```
beerose, line 331: Can you elaborate on why `as string` is needed here?

dillonraphael: Because otherwise it's unknown. For example you can check line 1340

dillonraphael (approved):

PR #3704, opened 2022-08-09
Store error for cookiePrefix if undefined & show error at the end of codemod

dillonraphael:

Noticed a weird user flow that if their `cookiePrefix` is parsed as undefined (due to reasons such as a ternary statement) that the codemod would error and the only way for the user to continue is modify their `blitz.config.ts`.

- Instead this PR introduces an array called `collectedErrors`. So now in a step action if you want to store an error for later, you can push the error as an object, ie. `{messages: string, step: number}`

- The type for the step object has now been updated to add a `stepIndex` parameter to the `action` key:
`type TStep = {name: string; action: (stepIndex?: number) => Promise<void>}`


packages/codemod/src/upgrade-legacy.ts
```diff
@@ -434,9 +431,12 @@ const upgradeLegacy = async () => {
             blitzClientProgram.toSource(),
           )
         } else {
-          throw new ExpectedError(
-            "Cookie Prefix is undefined & not a string. Please set your cookie prefix manually in app/blitz-client",
-          )
+          // Show error at end of codemod
+          storedErrors.push({
+            message:
+              "Cookie Prefix is undefined & not a string. Please set your cookie prefix manually in app/blitz-client",
```
flybayer, line 437:
```suggestion
              "Detected cookiePrefix is undefined. Please set your cookie prefix manually in app/blitz-client",
```

packages/codemod/src/upgrade-legacy.ts
```diff
@@ -30,19 +30,16 @@ class ExpectedError extends Error {
 const isInternalBlitzMonorepoDevelopment = fs.existsSync(
   path.join(__dirname, "../../../blitz-next"),
 )
-
+type TStep = {name: string; action: (stepIndex?: number) => Promise<void>}
```
beerose, line 33: It's not a big deal — I'm only suggesting this because we don't prefix types with `T` in the codebase (we only do it for generic arguments sometimes)

```suggestion
type Step = {name: string; action: (stepIndex?: number) => Promise<void>}
```

packages/codemod/src/upgrade-legacy.ts
```diff
@@ -434,9 +431,12 @@ const upgradeLegacy = async () => {
...
           )
         } else {
-          throw new ExpectedError(
-            "Cookie Prefix is undefined & not a string. Please set your cookie prefix manually in app/blitz-client",
-          )
+          // Show error at end of codemod
+          collectedErrors.push({
+            message:
+              "Cookie Prefix is undefined & not a string. Please set your cookie prefix manually in app/blitz-client",
+            step: stepIndex!,
```
beerose, line 438: Let's avoid unsafe type assertions as much as possible, especially in a new code. Since it can be undefined (as stated in `TStep` definition) but is required for the collectedErrors, can you either:

1) update the TStep type and make `stepIndex` required 
2) change this line to remove unsafe assertion (`!`) by providing a default value
3) modify `collectedErrors` type

beerose (changes requested):

Issue #3691, opened 2022-08-08
Add `blitz generate` back to toolkit


beerose:

noxify:
related: #3685

PR #3707, opened 2022-08-10
Add back generate command

dillonraphael:
Closes: #3691 

- Add back the generate command
- Clean up help message
- Fix `determineContext` and only run if no context is supplied & the `--context` flag is explicitly used
- Run `blitz codegen` after generating pages

_Note: This PR also fixes an issue with generating a new blitz app_

### Todo

- [X] Use `Routes` object in generator templates
- [x] Update generator templates


dillonraphael:
@beerose Do you think we should update the generator templates for mutations & queries to use resolvers?

beerose:
> @beerose Do you think we should update the generator templates for mutations & queries to use resolvers?

Yes, we should.

packages/generator/src/generator.ts
```diff
@@ -302,7 +302,7 @@ export abstract class Generator<
       return path.join(
         __dirname,
         "..",
-        process.env.NODE_ENV === "test" ? "./templates" : "./dist/templates",
+        process.env.NODE_ENV === "test" ? "./templates" : "./templates",
```
beerose, line 305: I think it needs to be `dist/templates` when you're using it outside of monorepo.

dillonraphael: Oops ya, this is a mistake. Thanks for finding.

PR #3718, opened 2022-08-12
Print model added or updated in schema.prisma after running model generator

beerose:
We are currently printing an empty message:
CleanShot 2022-08-12 at 19 40 55@2x

This PR fixes that.


packages/generator/src/generators/model-generator.ts
```diff
@@ -88,8 +89,11 @@ export class ModelGenerator extends Generator<ModelGeneratorOptions> {
           dryRun ? "" : ` ${updatedOrCreated} in schema.prisma`
         }:\n`,
       )
-      ast.printSchema({type: "schema", list: [model]}).split("\n")
-      // .map(log.progress) // todo
+      console.log({model})
```
beerose, line 92:
```suggestion
```

Issue #3725, opened 2022-08-14
invalidateQuery not working without params


medihack:
### What is the problem?

In v2.0.0-alpha.70 `invalideQuery` does not work anymore without providing the optional `params`. So `invalidateQuery(getUsers)` does not work, but `invalidateQuery(getUsers, {})` does work.

### Paste all your error logs here:

None

### Paste all relevant code snippets here:

See above

### What are detailed steps to reproduce this?

See above

### Run `blitz -v` and paste the output here:

Blitz version: 2.0.0-alpha.67 (global)
Blitz version: 2.0.0-danger.67 (local)
Linux 5.15 | linux-x64 | Node: v16.16.0

 Package manager: npm

  System:
    OS: Linux 5.15 Ubuntu 20.04.4 LTS (Focal Fossa)
    CPU: (16) x64 AMD EPYC 7B13
    Memory: 41.35 GB / 62.81 GB
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 16.16.0 - ~/.nvm/versions/node/v16.16.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v16.16.0/bin/yarn
    npm: 8.11.0 - ~/.nvm/versions/node/v16.16.0/bin/npm
  npmPackages:
    @blitzjs/auth: ^2.0.0-alpha.70 => 2.0.0-danger.67 
    @blitzjs/next: ^2.0.0-alpha.70 => 2.0.0-danger.67 
    @blitzjs/rpc: ^2.0.0-alpha.70 => 2.0.0-danger.67 
    @prisma/client: ^4.1.1 => 4.2.1 
    blitz: ^2.0.0-alpha.70 => 2.0.0-danger.67 
    next: ^12.2.2 => 12.2.5 
    prisma: ^4.1.1 => 4.2.1 
    react: ^18.2.0 => 18.2.0 
    react-dom: ^18.2.0 => 18.2.0 
    typescript: 4.6.3 => 4.6.3

beerose:
Thanks for the report! We'll look into it soon

beerose:
For whoever is going to look into this issue: we probably need to look into how queryKey is generated. We probably generate a different query key if `useQuery` doesn't have params vs in `invalidateQuery`.

Zeko369:
The problem here is that if nothing is passed to `invalidateQuery` it assumes undefined as a value and passed that to the serializer.

medihack:
[...]

PR #3728, opened 2022-08-14
Fix queryKeyGeneration when using invalidateQuery

Zeko369:
Closes: #3725

## Bug Checklist

- [x] Fix invalidateQuery called with one param generating `{json: undefined}` as param

## Feature Checklist


beerose (commented):
Couldn't we do something like `queryKey.push(serialize(args || {}) as any)` instead?

Zeko369:
`invalidateQuery(a)` if a is `[queryName]` it revalidates everything regardless of arguments, while `[queryName, params]` revalidates a specific params.

So the `|| {}` would only invalidate that if the params were {} for the query itself, so technically it would work but it wouldn't work the same way as tanstack/query itself

Zeko369:
@beerose 
`invalidateQuery(a)` if a is `[queryName]` it revalidates everything regardless of arguments, while `[queryName, params]` revalidates a specific params.

So the `|| {}` would only invalidate that if the params were {} for the query itself, so technically it would work but it wouldn't work the same way as tanstack/query itself

Zeko369:
Actually what we want is this
```ts
useQuery(getUser, undefined) // get dummy user
useQuery(getUser, 10)
```

Having undefined / 10 and in that case you have
```ts
invalidateQuery(getUser) // all get user queries
invalidateQuery(getUser, undefined) // just the first one -> so dummy user
invalidateQuery(getUser, 10) // invalidate user 10
```

And the PR currently doesn't do that so I'll fix it now and then re-request your review

Jaxenormus:
Is there a workaround for this while we wait for a release with this fix?

packages/blitz-rpc/src/data-client/react-query-utils.ts
```diff
@@ -124,24 +124,26 @@ const sanitize =
 export const sanitizeQuery = sanitize("query")
 export const sanitizeMutation = sanitize("mutation")
 
-export const getQueryKeyFromUrlAndParams = (url: string, params: unknown) => {
-  const queryKey = [url]
-
-  const args = typeof params === "function" ? (params as Function)() : params
-  queryKey.push(serialize(args) as any)
+type BlitzQueryKey = [string] | [string, any]
+export const getQueryKeyFromUrlAndParams = (url: string, ...params: [unknown]): BlitzQueryKey => {
```
beerose, line 128: The type says that it's a single-element array. Why are we spreading then?

Zeko369: So that if it's optional it's an empty array, otherwise we'd get undefined

```ts
foo(1, 2);
foo(1, undefined);
foo(1);
```

If it's a single param first would be 2, the other 2 would be undefined
If it's an array param it's [2], [undefined], [] <- which is used to tell react-query to clear all keys

This could have also been solved by defaulting to something like "BLITZ_EMPTY_KEY_DONT_PASS" but that's kinda hacky 😅

beerose:
Hey @Zeko369, wanted to follow up on this. We were talking about adding tests. Are you still up for adding them?

Zeko369:
@beerose unit tests for getKey and integration tests with actual invalidation 🚀 added

integration-tests/react-query-utils/test/index.test.ts
```diff
@@ -70,20 +106,20 @@ describe("React Query Utils Tests", () => {
...
+  //     } catch (err) {
+  //       console.log(err)
+  //     }
+  //   }, 5000 * 60 * 2)
+  //   afterAll(async () => await killApp(app))
 
-    runTests()
-  })
+  //   runTests()
+  // })
```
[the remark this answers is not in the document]
Zeko369: I removed it as a debugging step, but forgot to uncomment it

PR #3742, opened 2022-08-16
Remove the NODE_ENV development check when patching nextjs

dillonraphael:

Removes the `NODE_ENV=development` check inside the string and replace during the nextjs patch


.changeset/lazy-maps-sort.md
```diff
@@ -0,0 +1,5 @@
+---
+"blitz": patch
+---
+
+Patch Next.js Suspense issue in all node environments. Previously we only patched it in Development env, but now we make sure it gets patched during build.
```
beerose, line 5:
```suggestion
Patch Next.js Suspense issue in all node environments. Previously we only patched it in the `development` environment, but now we make sure it gets patched in the `production` env (with the `blitz build` command) as well.
```

beerose (approved):
Left one comment regarding the changeset. Otherwise LGTM.

beerose (approved):

PR #3753, opened 2022-08-18
Fix #3737 Fix eslint config using the codemod

### LINKED ISSUE #3737: Update eslint config with upgrade-legacy codemod
### What is the problem?

ESLint doesn’t work after migrating to Blitz V2. The [documentation page](https://blitzjs.com/docs/eslint-config) uses the same code snippet that exists in my repo so it’s unclear how to resolve this.

### Paste all your error logs here:

```
$ eslint --ext .js,.ts,.tsx . --fix

Oops! Something went wrong! :(

ESLint: 8.22.0

ESLint couldn't find the config "blitz" to extend from. Please check that the name of the config is correct.

The config "blitz" was referenced from the config file in "./.eslintrc.js".

If you still have problems, please stop by https://eslint.org/chat/help to chat with the team.
```

### Paste all relevant code snippets here:

```ts
module.exports = {
	extends: ['blitz']
}
```

### What are detailed steps to reproduce this?

1. Make an `.eslintrc.js` which attempts to extend from `blitz`

### Run `blitz -v` and paste the output here:

```
Blitz version: 2.0.0-beta.1 (local)
macOS Big Sur | darwin-x64 | Node: v16.16.0


 Package manager: npm

  System:
    OS: macOS 11.6.5
    CPU: (8) x64 Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz
    Memory: 213.43 MB / 32.00 GB
    Shell: 3.3.1 - /usr/local/bin/fish
  Binaries:
    Node: 16.16.0 - /var/folders/qx/3nty7c0x6hzd8_q8nkbqq50m0000gn/T/yarn--1660575023904-0.5016463859620897/node
    Yarn: 1.22.19 - /var/folders/qx/3nty7c0x6hzd8_q8nkbqq50m0000gn/T/yarn--1660575023904-0.5016463859620897/yarn
    npm: 8.11.0 - /usr/local/bin/npm
  npmPackages:
    @blitzjs/auth: Not Found
    @blitzjs/next: Not Found
    @blitzjs/rpc: Not Found
    @prisma/client: Not Found
    blitz: Not Found
    next: Not Found
    prisma: Not Found
    react: Not Found
    react-dom: Not Found
    typescript: ^4.7.4 => 4.7.4 
```

(versions not showing is probably a blitz + lerna interoperability issue)

### Please include below any other applicable logs and screenshots that show your problem:

_No response_

siddhsuresh:
Closes: #3737

- [x] Add a new codemod step to update the .eslintrc.js file to ```module.exports=require("@blitzjs/next/eslint")```

### DIFF
--- a/packages/codemod/src/upgrade-legacy.ts
+++ b/packages/codemod/src/upgrade-legacy.ts
@@ -112,6 +112,24 @@ const upgradeLegacy = async () => {
       fs.writeFileSync(path.resolve("next.config.js"), program.toSource())
     },
   })
+  
+  steps.push({
+    name: "update .eslintrc.js configuration",
+    action: async () =>{
+      const program = getCollectionFromSource(".eslintrc.js")
+      const parsedProgram = program.get()
+      parsedProgram.value.program.body = []
+      const moduleExport = j.expressionStatement(
+        j.assignmentExpression(
+          "=",
+          j.memberExpression(j.identifier("module"), j.identifier("exports")),
+          j.callExpression(j.identifier("require"), [j.identifier(`"@blitzjs/next/eslint"`)]),
+      )
+      )
+      parsedProgram.value.program.body.push(moduleExport)
+      fs.writeFileSync(path.resolve(".eslintrc.js"), program.toSource())
+    }
+  })
 
   steps.push({
     name: "update dependencies in package.json",
--- a/packages/codemod/src/upgrade-legacy.ts
+++ b/packages/codemod/src/upgrade-legacy.ts
@@ -115,7 +115,7 @@ const upgradeLegacy = async () => {
   
   steps.push({
     name: "update .eslintrc.js configuration",
-    action: async () => {
+    action: async (stepIndex) => {
       if (fs.existsSync(path.resolve(".eslintrc.js"))) {
         const program = getCollectionFromSource(".eslintrc.js")
         const parsedProgram = program.get()
@@ -131,7 +131,11 @@ const upgradeLegacy = async () => {
         fs.writeFileSync(path.resolve(".eslintrc.js"), program.toSource())
       }
       else{
-        log.error(".eslintrc.js does not exist")
+        collectedErrors.push({
+          message:
+            ".eslintrc.js does not exist",
+          step: stepIndex,
+        })
       }
     },
   })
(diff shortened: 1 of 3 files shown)


packages/codemod/src/upgrade-legacy.ts
```diff
@@ -112,6 +112,29 @@ const upgradeLegacy = async () => {
...
+            j.memberExpression(j.identifier("module"), j.identifier("exports")),
+            j.callExpression(j.identifier("require"), [j.identifier(`"@blitzjs/next/eslint"`)]),
+          ),
+        )
+        parsedProgram.value.program.body.push(moduleExport)
+        fs.writeFileSync(path.resolve(".eslintrc.js"), program.toSource())
+      }
+      else{
+        log.error(".eslintrc.js does not exist")
+      }
```
dillonraphael, line 135: I don't think we need to show an error if .eslint doesn't exist.

But if you think we should, instead of logging the error here, can we show it at the end?

You can use the `collectedErrors` array, and inside the else statement just do:

```ts
          collectedErrors.push({
            message:
              ".eslintrc.js does not exist",
            step: stepIndex,
          })
```

`stepIndex` is a parameter inside the action function.

dillonraphael (changes requested):

dillonraphael:
Amazing stuff! You're a real codemod pro at this point. Can't thank you enough @siddhsuresh

packages/codemod/src/upgrade-legacy.ts
siddhsuresh: > I don't think we need to show an error if .eslint doesn't exist.
> 
@dillonraphael I had added that check because when I tested it with a legacy repository without the .eslint config 

image
The above error will stop the codemod even though the config is not technically essential. 

image
So I had used the format done when cookie prefix is not defined in the blitz config file

**This is how it would look to a user when they did not have .eslintrc.js file**
image

siddhsuresh: > But if you think we should, instead of logging the error here, can we show it at the end?
> 
> You can use the `collectedErrors` array, and inside the else statement just do:
> 
> ```ts
>           collectedErrors.push({
>             message:
>               ".eslintrc.js does not exist",
>             step: stepIndex,
>           })
> ```
> 
> `stepIndex` is a parameter inside the action function.

Okay I shall add a commit with the requested change.

dillonraphael (approved):

dillonraphael (approved):

PR #3759, opened 2022-08-19
Fix for codemod that removes types from the blitz config file

dillonraphael:
Per discord discussion: https://discord.com/channels/802917734999523368/1010201051145699328/1010207387640156160

When removing all typescript related code from the blitz config during the codemod, an error occurred for situations such as `string[]`. The fix checks if the NodePath is `typeAnnotation` before trying to remove.

### DIFF
--- a/.changeset/fair-kangaroos-clean.md
+++ b/.changeset/fair-kangaroos-clean.md
@@ -1,0 +1,5 @@
+---
+"@blitzjs/codemod": patch
+---
+
+Fix for codemod that removes types from the blitz config file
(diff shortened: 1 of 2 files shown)


.changeset/fair-kangaroos-clean.md
```diff
@@ -0,0 +1,5 @@
+---
+"@blitzjs/codemod": patch
+---
+
+Fix for codemod that removes types from the blitz config file
```
beerose, line 5: Remove TypeScript type annotations from `next.config.js` (a JavaScript file) in the `upgrade-legacy` codemod

beerose (approved):

Issue #3760, opened 2022-08-19
Show file name for when codemod errors with an Unexpected Error


dillonraphael:
### What is the problem?

Currently when running the codemod and an `Unexpected Error` occurs during a step that loops through all files with the `getAllFiles` function it's impossible to determine which file is causing the error.

### Paste all your error logs here:

```
✖ wrap getServerSideProps, getStaticProps and API handlers with gSSP, gSP, and api
TypeError: Cannot read properties of null (reading 'name')
This is an unexpected error. Please ask for help in the discord #general-help channel. https://discord.blitzjs.com
```

PR #3763, opened 2022-08-21
show file name when unexpected error occurs

siddhsuresh:
Closes: #3760 

- [x] add try catch block around the wrap gSP, gSSP and api loops to show the file name that is causing the error.
- [x] created a progress like display showing the file name currently being processed to the used
- [x] fixed an unexpected error during the _document.tsx step when .get() is being called when it could be null

### DIFF
--- a/packages/codemod/src/upgrade-legacy.ts
+++ b/packages/codemod/src/upgrade-legacy.ts
@@ -95,7 +95,7 @@ const upgradeLegacy = async () => {
           "=",
           j.identifier("const { withBlitz }"),
           j.callExpression(j.identifier("require"), [j.identifier(`"@blitzjs/next"`)]),
-        )
+        ),
       )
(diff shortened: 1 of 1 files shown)


packages/codemod/src/upgrade-legacy.ts
```diff
@@ -1078,26 +1081,34 @@ const upgradeLegacy = async () => {
     name: "wrap getServerSideProps, getStaticProps and API handlers with gSSP, gSP, and api",
     action: async () => {
       const pagesDir = path.resolve("pages")
+      console.log("")
```
dillonraphael, line 1084: Can we remove this? It's causing issues with the spinner.

dillonraphael (changes requested):
Just tested locally. Works great! Just needs a little clean up before merging.

dillonraphael (approved):
Great stuff. Thanks again @siddhsuresh

PR #3764, opened 2022-08-21
convert invokeWithMiddleware to resolver call in codemod

siddhsuresh:
Closes: ?

- [x] convert invokeWithMiddleware to resolver call

### Before
```ts
export const getServerSideProps = gSSP<Props>(async ({
  req,
  res,
}) => {
  const user = invokeWithMiddleware(getUserQuery, {},{req,res})
  const user2 = invokeWithMiddleware(getUserQuery2, {},{req,res})
  return {
    props: {
      user,
      user2
    },
  }
})
```
### After
```ts
export const getServerSideProps = gSSP<Props>(async ({
  req,
  res,
}) => {
  const user = getUserQuery({},{req,res});
  const user2 = getUserQuery2({},{req,res});
  return {
  props: {
      user,
      user2
    },
  }
})
```


dillonraphael:
Hey @siddhsuresh I'm so sorry for our mistake, but `invokeWithCtx` should not be a public. This was our fault for not making it clear ☹️. But since `gSP` & `gSSP` already passes the ctx `invokeWithCtx` is not needed.

So essentially the correct codemod would require us to unwrap `invokeWithMiddleware` and call the resolver function directly.

siddhsuresh:
> Hey @siddhsuresh I'm so sorry for our mistake, but `invokeWithCtx` should not be a public. This was our fault for not making it clear ☹️. But since `gSP` & `gSSP` already passes the ctx, `invokeWithCtx` is not needed.
> 
> So essentially the correct codemod would require us to unwrap `invokeWithMiddleware` and call the resolver function directly.

Hey thanks @dillonraphael for letting me know. I shall update the PR with the correct codemod in some time

siddhsuresh:
> So essentially the correct codemod would require us to unwrap `invokeWithMiddleware` and call the resolver function directly.

@dillonraphael I have updated the codemod to this requirement.

packages/codemod/src/upgrade-legacy.ts
```diff
@@ -1198,43 +1198,44 @@ const upgradeLegacy = async () => {
...
+                const {node} = path
+                const {arguments: asyncFunction} = node
+                //@ts-ignore
+                const params = asyncFunction[0]?.params
+                if (params && params.length) {
+                  params[0].properties.push(j.identifier("ctx"))
+                }
+              })
+            }
+          }
```
dillonraphael, line 1230: Not sure we need this

packages/codemod/src/upgrade-legacy.ts
```diff
@@ -1198,43 +1198,44 @@ const upgradeLegacy = async () => {
...
         const program = getCollectionFromSource(file)
-        const invokeWithMiddlewarePath = findCallExpression(program, "invokeWithMiddleware")
-        if (invokeWithMiddlewarePath?.length) {
-          log.error(`\n invokeWithMiddleware found at ${file}. \n`)
-          errors++
+        try {
+          const invokeWithMiddlewarePath = findCallExpression(program, "invokeWithMiddleware")
+          if (invokeWithMiddlewarePath?.length) {
+            invokeWithMiddlewarePath.forEach((path) => {
+              path.value.arguments.push(j.identifier("ctx"))
```
dillonraphael, line 1208: We can also remove this?

dillonraphael (changes requested):
@siddhsuresh 🏎
 
 Other than the changes it seems to work great for the example here https://legacy.blitzjs.com/docs/resolver-server-utilities#invoke-with-middleware

And I tested with a legacy blitz project, works great.

packages/codemod/src/upgrade-legacy.ts
siddhsuresh: okay I will remove this.  
So should it be shown to the user that they need to get the ```ctx``` parameter from gSSP?

packages/codemod/src/upgrade-legacy.ts
siddhsuresh: which part should be removed? because the loop is integral to the codemod step to work. I first get the array of all invocations of invokeWithMiddleware and then add the ctx argument to it so that when the arguments list is sliced, it will exist for the final resolver call

packages/codemod/src/upgrade-legacy.ts
dillonraphael: > okay I will remove this. So should it be shown to the user that they need to get the `ctx` parameter from gSSP?

It should be expected that ctx is being passed already if they're using `invokeWithMiddleware`, as it's the 3rd argument.

packages/codemod/src/upgrade-legacy.ts
dillonraphael: > which part should be removed? because I think the loop is integral to the codemod step to work. The array of all invocations of `invokeWithMiddleware` is received and then add the `ctx` argument to it so that when the arguments list is sliced, it will exist for the final resolver call

Oops my bad. Was supposed to just highlight line `1208`. The `path.value.arguments.push(j.identifier("ctx"))`.

packages/codemod/src/upgrade-legacy.ts
siddhsuresh: @dillonraphael I have removed this section

packages/codemod/src/upgrade-legacy.ts
siddhsuresh: @dillonraphael I am pushing the ```ctx``` into the array so that we remove the first element during

https://github.com/siddhsuresh/blitz/blob/0a8acd2e163a3e6481d759242442b2dc69200485/packages/codemod/src/upgrade-legacy.ts#L1212

```ts
path.value.arguments.slice(1)
```

The array of the ```input,ctx``` is passed into the resolver call.
Do you want be to explicitly create an array like ```[path.value.arguments[1]?.name,j.identifier("ctx")] ```

dillonraphael: @siddhsuresh Hmm I'm not sure I understand. I tested with the change I recommended for:

```ts
  const products = await invokeWithMiddleware(
    getProducts,
    { orderBy: { id: "desc" } },
    { req, res }
  )
 ```

and the output was:

```ts
  const products = await getProducts(
    { orderBy: { id: "desc" } },
    { req, res }
  )
```

siddhsuresh: @dillonraphael okay, I have updated the code with your recommended change.

dillonraphael (approved):

packages/codemod/src/upgrade-legacy.ts
```diff
@@ -1198,43 +1198,30 @@ const upgradeLegacy = async () => {
...
-        const invokeWithMiddlewarePath = findCallExpression(program, "invokeWithMiddleware")
-        if (invokeWithMiddlewarePath?.length) {
-          log.error(`\n invokeWithMiddleware found at ${file}. \n`)
-          errors++
+        try {
+          const invokeWithMiddlewarePath = findCallExpression(program, "invokeWithMiddleware")
+          if (invokeWithMiddlewarePath?.length) {
+            invokeWithMiddlewarePath.forEach((path) => {
+              const resolverExpression = j.callExpression(
+                //@ts-ignore
```
beerose, line 1209: Could we avoid using `@ts-ignore`?

siddhsuresh: Hey @beerose the reason for the //@ts-ignore is that the types provided for the arguments is only 
image
while the return value when logging the value provides other information like the name that is required for replacing as the function
image
So I have now removed the explicit //@ts-ignore by converting the type to of the ```resolverName``` variable to any to avoid type errors

siddhsuresh: @beerose okay I shall try to fix the type mismatch

siddhsuresh: @beerose I have fixed the type error by assigning the type of the returned element, i.e Identifier
https://github.com/siddhsuresh/blitz/blob/a047f589c1f8db5faae0f081fc9d68bda5831d3a/packages/codemod/src/upgrade-legacy.ts#L1209

siddhsuresh: Hey, @dillonraphael I have removed the assertion, instead, I am checking whether the first argument is an Identifier, If so updating. Otherwise, throwing an error with the line number of the failing invokeWithMiddleware call.

siddhsuresh: @dillonraphael Is this change fine?

oyatek:
When I call the resolver directly as shown below 
```
await cancelTransaction({ id: LocalTransactionId }, { req, res } )
```

I get the following error:
```
TS2352: Conversion of type '{ req: NextApiRequest; res: NextApiResponse<any>; }' to type 'Ctx' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.   Type '{ req: NextApiRequest; res: NextApiResponse<any>; }' is missing the following properties from type 'Ctx': session, prefetchQuery, prefetchInfiniteQuery
```

How this can be fixed?

siddhsuresh:
> When I call the resolver directly as shown below
> 
> ```
> await cancelTransaction({ id: LocalTransactionId }, { req, res } )
> ```
> 
> I get the following error:
> 
> ```
> TS2352: Conversion of type '{ req: NextApiRequest; res: NextApiResponse<any>; }' to type 'Ctx' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.   Type '{ req: NextApiRequest; res: NextApiResponse<any>; }' is missing the following properties from type 'Ctx': session, prefetchQuery, prefetchInfiniteQuery
> ```
> 
> How this can be fixed?

Hey @oyatek, sorry seems like I missed this message, for this error you could use the third variable passed by `api` or `gssp` exports, ie `ctx` and pass that to the resolver as the second argument

oyatek:
@siddhsuresh thanks, but I still don't understand where the `ctx` should come from. Could you please explain on the following example? 

```
import { NextApiRequest, NextApiResponse, NextApiHandler } from "next"
import changePassword from "../../auth/mutations/changePassword"

const handler: NextApiHandler = async (req: NextApiRequest, res: NextApiResponse) => {

  // This line gives a TS2345 error, "req" is marked with red color
  await changePassword({currentPassword: "foo", newPassword: "bar"}, {req, res})

  res.statusCode = 200
  res.setHeader("Content-Type", "application/json")
  res.end(JSON.stringify({ status: "OK" }))
}
export default api(handler)
```

oyatek:
@siddhsuresh thank you!

Issue #3758, opened 2022-08-19
codemod replaceIdentifiers replaces all identifiers with NextApiRequest


tmcw:
### What is the problem?

Running the codemod on my codebase produces some odd results. A function signature like

```
async function thumbnail(req: BlitzApiRequest, res: BlitzApiResponse) {
```

Becomes

```
async function thumbnail(req: NextApiRequest, res: NextApiRequest) {
```

The cause looks to be that `newIdentifier` isn't used the method that transforms identifiers, and instead it transforms everything to requests:

https://github.com/blitz-js/blitz/blob/1bf185d6189776eb27eb737760ba8580ef4ea884/packages/codemod/src/utils.ts#L320-L330

### Paste all your error logs here:

```
None
```

### Paste all relevant code snippets here:

```
async function thumbnail(req: BlitzApiRequest, res: BlitzApiResponse) {
}
```

### What are detailed steps to reproduce this?

1. Run the codemod on a codebase that uses the `BlitzApiResponse` type

### Run `blitz -v` and paste the output here:

```
Blitz version: 2.0.0-beta.2 (local)
macOS Monterey | darwin-arm64 | Node: v16.16.0


 Package manager: npm

  System:
    OS: macOS 12.5
    CPU: (8) arm64 Apple M2
    Memory: 554.30 MB / 24.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.16.0 - ~/.volta/tools/image/node/16.16.0/bin/node
    Yarn: 1.22.19 - ~/.volta/tools/image/yarn/1.22.19/bin/yarn
    npm: 8.11.0 - ~/.volta/tools/image/node/16.16.0/bin/npm
  npmPackages:
    @blitzjs/auth: 2.0.0-beta.2 => 2.0.0-beta.2
    @blitzjs/next: 2.0.0-beta.2 => 2.0.0-beta.2
    @blitzjs/rpc: 2.0.0-beta.2 => 2.0.0-beta.2
    @prisma/client: ^4 => 4.1.1
    blitz: 2.0.0-beta.2 => 2.0.0-beta.2
    next: 12.2.0 => 12.2.0
    prisma: ^4 => 4.1.1
    react: ^18.0.0 => 18.2.0
    react-dom: ^18.0.0 => 18.2.0
    typescript: ~4.6 => 4.6.4```


### Please include below any other applicable logs and screenshots that show your problem:

_No response_

dillonraphael:
[...]

PR #3766, opened 2022-08-22
Fix `replaceIdentifiers` codemod utility

beerose:
Closes: https://github.com/blitz-js/blitz/issues/3758

Fixes a hardcoded value in `replaceIdentifiers` codemod utility.


dillonraphael (approved):
Thanks @beerose!

PR #3773, opened 2022-08-22
Replace blitz-env.d.ts with next-env.d.ts

medihack:
Closes: Not associated with an issue

If I understand it correctly (do I?), then there is no `blitz-env.d.ts` anymore in Blitz toolkit. I simply replaced it with `next-env.d.ts`.


beerose (approved):
Thanks a lot!

Issue #3775, opened 2022-08-22
Only generate prisma client if it has not been generated


dillonraphael:
### What is the problem?

Currently the prisma client is being generated on most commands as part of the `codegen-tasks.ts` file (https://github.com/blitz-js/blitz/blob/3d950f77721d3ba766379445dd1e5e4ee5a08080/packages/blitz/src/cli/utils/codegen-tasks.ts#L40-L50)

Since it's a big performance hit, we should only be generating the prisma client once.

siddhsuresh:
Hey @dillonraphael can I work on this?

beerose:
> Hey @dillonraphael can I work on this?

Sure!

PR #3780, opened 2022-08-23
#3775 Only generate prisma client if it has not been generated

siddhsuresh:
Closes: #3775

- [x] create a temporary ```blitz-prisma-status.json``` file in the temp directory
- [x] store the last modified time of the ```db/prisma.schema``` and store it in the json file
- [x] check if the modified time of the schema file has changed since the last prisma client generation by the codegen


packages/blitz/src/cli/utils/codegen-tasks.ts
```diff
@@ -37,15 +38,33 @@ export const codegenTasks = async () => {
...
 
+    const tempDir = os.tmpdir() 
+    const creationDate = await fs.stat(process.cwd())
+    if (!fs.existsSync(join(tempDir, `blitz-${creationDate.birthtimeMs}-prisma-client.json`))) {
+      const lastModified = await fs.stat(join(process.cwd(), "db/schema.prisma"))
+      await fs.writeFile(
+        join(tempDir, `blitz-${creationDate.birthtimeMs}-prisma-client.json`),
+        JSON.stringify({lastModified: lastModified.mtimeMs}),
+      )
+    }
```
dillonraphael, line 49: I don't mean to be picky, but I think we should only be doing this if `hasPrisma` is true.

dillonraphael (changes requested):

packages/blitz/src/cli/utils/codegen-tasks.ts
siddhsuresh: @dillonraphael I have made the requested change

PR #3781, opened 2022-08-23
Revert "#3775 Only generate prisma client if it has not been generated"

dillonraphael:
Reverts blitz-js/blitz#3780

### DIFF
--- a/.changeset/slow-walls-camp.md
+++ b/.changeset/slow-walls-camp.md
@@ -1,5 +1,0 @@
----
-"blitz": patch
----
-
-Only run the prisma generate command when the schema file changes.
--- a/packages/blitz/src/cli/utils/codegen-tasks.ts
+++ b/packages/blitz/src/cli/utils/codegen-tasks.ts
@@ -5,7 +5,6 @@ import {join} from "path"
 import fs from "fs-extra"
 import {getPackageJson} from "./get-package-json"
 import {runPrisma} from "../../utils/run-prisma"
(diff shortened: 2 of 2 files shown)


siddhsuresh:
Hey, @dillonraphael was there some error in #3780?

dillonraphael:
> Hey, @dillonraphael was there some error in #3780?

Unfortunately after we thought about it, there are a couple edge cases where this solution wouldn't work.

1. The user's schema may not always be in the `db` directory. There is a possibility of having it in a monorepo
2. It doesn't account for which blitz project. For example, the user may have 2 blitz projects locally on their machine

siddhsuresh:
> > Hey, @dillonraphael was there some error in #3780?
> 
> @siddhsuresh Unfortunately after we thought about it, there are a couple edge cases where this solution wouldn't work.
> 
> 1. The user's schema may not always be in the `db` directory. There is a possibility of having it in a monorepo
> 2. It doesn't account for which blitz project. For example, the user may have 2 blitz projects locally on their machine

Oh, I had not considered the first case while creating the solution, while ig for the second case I think my solution takes care of the identification of the blitzjs project by taking the creation time of the current directory into account which should not be the same for any two projects

dillonraphael:
Decided not to revert, and build off what @siddhsuresh contributed (which I'm extremely grateful for). I'll close this PR and share the issue for simplifying the solution. https://github.com/blitz-js/blitz/issues/3801

PR #3782, opened 2022-08-23
Delete and ignore next-env.d.ts files

medihack:
Closes: Not associated with any issue

According to the [Next.js TypeScript documentation](https://nextjs.org/docs/basic-features/typescript#existing-projects) (see one of those boxes there) `next-env.d.ts` should not be put under version control.

I am not sure if we should just do this for newly generated apps using the templates or also the example apps and integrations tests. That's why I used two commits (if you only want to merge one of them).

> "This file should not be committed and should be ignored by version control (e.g. inside your .gitignore file)

By the way, the `.gitignore` files in the blitz repo are quite messy and could need a bit of love. Yeah, I know ... there are more important issues ;-)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,6 +53,7 @@ dist
 **/.env.local
 **/.env.*.local
 **/.envrc
+**/next-env.d.ts
 .blitz-*
 .blitz-cli-cache
 .tsbuildinfo
@@ -65,6 +66,4 @@ db.sqlite-journal
 **/db/db.sqlite
 test/integration/**/db.json
 test/**/*/out
-test/**/next-env.d.ts
-examples/**/next-env.d.ts
 .blitz**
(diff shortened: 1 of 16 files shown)


dillonraphael (commented):
All looks good to me. But maybe we shouldn't remove it from the generator? Just have it added to the gitignore

medihack:
Yes, sounds good. I just checked how a freshly created Next.js app does it. It also creates the `next-env.d.ts` initially, but has it added to the `.gitignore`.

There is another thing. The Blitz `next-env.d.ts` has this line in there:

`/// <reference types="next/types/global" />`

This line is deleted right after starting `blitz dev` for the first time. Maybe some legacy stuff of a previous Next.js version. Maybe we could remove this, too.

I will create two new separate PRs for this.

PR #3792, opened 2022-08-24
Remove `-cookie-prefix` appended to the cookiePrefix string

dillonraphael:
Per: https://discord.com/channels/802917734999523368/1012007617700315306/1012085676398886942

Legacy blitz never had the `-cookie-prefix` appended to the cookiePrefix string. This removes it from the generator that is also used for the codemod.

### DIFF
--- a/.changeset/silly-shoes-agree.md
+++ b/.changeset/silly-shoes-agree.md
@@ -2,4 +2,4 @@
 "@blitzjs/generator": patch
 ---
 
-Remove `-cookie-prefix` appended to the cookiePrefix string
+Remove `-cookie-prefix` appended to the cookiePrefix string in the generator template. This will also fix auth and CSRF issues for people upgrading.
--- a/.changeset/silly-shoes-agree.md
+++ b/.changeset/silly-shoes-agree.md
@@ -1,0 +1,5 @@
+---
+"@blitzjs/generator": patch
+---
+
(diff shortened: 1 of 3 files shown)


flybayer (approved):

.changeset/silly-shoes-agree.md
```diff
@@ -0,0 +1,5 @@
+---
+"@blitzjs/generator": patch
+---
+
+Remove `-cookie-prefix` appended to the cookiePrefix string
```
flybayer, line 5: I suggest making it more clear that this also fixes auth and CSRF issues for people upgrading

.changeset/silly-shoes-agree.md
```diff
@@ -2,4 +2,4 @@
 "@blitzjs/generator": patch
 ---
 
-Remove `-cookie-prefix` appended to the cookiePrefix string
+Remove `-cookie-prefix` appended to the cookiePrefix string in the generator template. This will also fix auth and CSRF issues for people upgrading.
```
beerose, line 5: I would also mention what kind of auth and CSRF issues because if we have more than one (or will have), someone reading the changelog may expect ALL of them to be fixed (or at least they won't be sure which ones are solved).
```suggestion
Remove `-cookie-prefix` appended to the `cookiePrefix` config property in the new app template. It will also fix auth and CSRF issues for users upgrading from a legacy framework.
```

Issue #3271, opened 2022-04-07
Add windows testing support to CI


dillonraphael:
Currently when the tests run on windows os the following error happens:

```
Error: spawn cmd.exe ENOENT
 ⎯⎯⎯⎯⎯⎯ Unhandled Error ⎯⎯⎯⎯⎯⎯⎯
WebDriverError: chrome not reachable
```

siddhsuresh:
Hey @dillonraphael @beerose, I saw this issue as one of the points to be completed for the stable release. Can I open a PR to add the windows testing to the workflow?

beerose:
> Hey @dillonraphael @beerose, I saw this issue as one of the issues to be completed for the stable release. Can I open a PR to add the windows testing to the workflow?

That would be amazing!

PR #3799, opened 2022-08-29
Add Windows Testing to CI

siddhsuresh:
Closes: #3271

#### Changes in ```.github/workflows/main.yml```
- [x] Create 4 jobs - Lint, Build, Linux Test, Windows Test
- [x] Enable the remote caching feature in turbo using [felixmosh/turborepo-gh-artifacts](https://github.com/felixmosh/turborepo-gh-artifacts) Github Action

image

image
⬆️ Build and test speeds after first run
https://github.com/siddhsuresh/blitz/actions/runs/2948884996

- [x] **TODO: Add ```TURBO_SERVER_TOKEN``` to GitHub secrets in order to enable the remote caching using GitHub artifacts**

#### Bug Fixes to enable the tests in windows
- [x] Fixed error in generation of route name for **nested routes** in ```@blitzjs/rpc``` in ```packages/blitz-rpc/src/loader-utils.ts```
```js
  return filePath
    .replace(/^.*?[\\/]queries[\\/]/, "/")
    .replace(/^.*?[\\/]mutations[\\/]/, "/")
    .replace(/\\/g, "/") 👈🏼
    .replace(fileExtensionRegex, "")
```
##### Before in windows
image

‼️**The ```"/v2\getCurrentUser"``` gets converted to ```/v2getCurrentUser``` in windows which causes the routePath in the resolverMap in ```blitz-rpc/index-server.ts``` not to match when users are using nested paths.**

##### After in windows
image

- [x] Remove ```NODE_ENV=production``` from the build scripts of the ```apps/*``` as it is not supported in windows [next build will set the NODE_ENV as production by default]

- [x] Update the ```index.test.ts``` file in ```integration-tests/auth``` from explicitly calling the ```seed``` function to now adding the ```prisma reset``` and ```prisma db seed``` commands to the ```package.json```. 
❓The seed function previously called from the test file was not updating the database, causing the test to fail in windows.
- [x] Increasing the hookTimeout time in the ```integration-tests/react-query-utils/vitest.config.ts``` file to pass the tests in windows


dillonraphael:
Ooof this is a juicy one @siddhsuresh 

I'll try and review it this week!

dillonraphael (approved):

Issue #3804, opened 2022-08-30
Show file path on error for the `update project's import` step


dillonraphael:
### What is the problem?

Currently if the `update project's import` step fails with an `Unexpected Error`, we have no clue which file the error is happening on.

![unknown](https://user-images.githubusercontent.com/3496193/187560628-f1efb49a-7c05-42cf-8463-b96f12d2927e.png)

### Paste all relevant code snippets here:

We should put the logic inside a try catch block. Like what @siddhsuresh did here https://github.com/blitz-js/blitz/blob/c213d521c6e48702e5bccd12e0f4ac0d266f1809/packages/codemod/src/upgrade-legacy.ts#L1126-L1146

siddhsuresh:
@dillonraphael can I work on this and add the try...catch block for all the instances where ```getAllFiles``` function is called?

PR #3809, opened 2022-09-01
#3804 Show file path on error

siddhsuresh:
Closes: #3804

- [x] wrap the logic of the steps inside the ```getAllFiles``` function with a ```try...catch``` block to get the file name during errors

### DIFF
--- a/packages/codemod/src/upgrade-legacy.ts
+++ b/packages/codemod/src/upgrade-legacy.ts
@@ -316,7 +316,7 @@ const upgradeLegacy = async () => {
             }
           })
           fs.writeFileSync(path.resolve(appDir, filename), program.toSource())
-        } catch (e: any) {
+        } catch (e) {
           log.error(`Error updating imports in ${file}`)
           throw new Error(e)
         }
@@ -398,7 +398,7 @@ const upgradeLegacy = async () => {
           }
 
           fs.writeFileSync(path.join(path.resolve(file)), program.toSource())
-        } catch (e: any) {
+        } catch (e) {
           log.error(`Error in updating next.js default imports in ${file}`)
           throw new Error(e)
         }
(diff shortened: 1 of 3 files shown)


packages/codemod/src/upgrade-legacy.ts
```diff
@@ -284,37 +284,42 @@ const upgradeLegacy = async () => {
...
-                const index = parsedProgram.value.program.body.indexOf(e)
-                parsedProgram.value.program.body.splice(index, 1)
               }
             }
-          }
-        })
-        fs.writeFileSync(path.resolve(appDir, filename), program.toSource())
+          })
+          fs.writeFileSync(path.resolve(appDir, filename), program.toSource())
+        } catch (e: any) {
```
dillonraphael, line 319: We're being very strict with using any or assertions. You can see how to handle errors here https://github.com/blitz-js/blitz/blob/39fd15d4b5a522740197d1311db16884bd7dfc91/packages/codemod/src/upgrade-legacy.ts#L1301

dillonraphael (changes requested):

packages/codemod/src/upgrade-legacy.ts
siddhsuresh: hey, @dillonraphael I have removed all the manual assertions to ```any``` for errors that can be caught in a file. The inner ```try...catch``` will just pass on the error to the outer catch block like you mentioned that should do the check. I was not sure whether to duplicate the logic in all places. If this is what is expected, I shall do that.

dillonraphael: Yup passing it to the outer catch block should be good @siddhsuresh. But seems the CI is failing.

siddhsuresh: @dillonraphael The CI is now passing.

dillonraphael (approved):

PR #3810, opened 2022-09-01
Fix remote caching in CI

siddhsuresh:
Closes: ?

- [x] Change the server-token given to the ```felixmosh/turborepo-gh-artifacts@v1``` GitHub action from secret value to pre-defined random 36 hex value generated by ```openssl```.

This is requested to be changed as GitHub only allows actions triggered from within the repository the access the secrets defined in the repository, with the exception of the ```GITHUB_TOKEN```

https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow

image

So, whenever a fork triggers the action the windows test fails most of the time due to an exit of a test as the remote caching feature of turbo is not enabled due to a null string being passed as the server-token argument

image

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,6 +9,10 @@ on:
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
+  
+env:
+  TURBO_TOKEN: 05de0230f01174d1f8cb4845a01dc6c895ce28f04ebef2318ab11615791b871c35eabbf8
+  TURBO_TEAM: foo
 
 jobs:
   lint:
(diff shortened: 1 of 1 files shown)


dillonraphael:
Hmm I see what you mean @siddhsuresh. I'd feel better getting an opinion from @flybayer before approving this.

flybayer (approved):

.github/workflows/main.yml
```diff
@@ -37,11 +37,11 @@ jobs:
...
           with:
             repo-token: ${{ secrets.GITHUB_TOKEN }}
-            server-token: ${{ secrets.TURBO_SERVER_TOKEN }}
+            server-token: 05de0230f01174d1f8cb4845a01dc6c895ce28f04ebef2318ab11615791b871c35eabbf8
         - name: Build
-          run: pnpm build -- --api="http://127.0.0.1:9080" --token="${{ secrets.TURBO_SERVER_TOKEN }}" --team="foo"
+          run: pnpm build -- --api="http://127.0.0.1:9080" --token="05de0230f01174d1f8cb4845a01dc6c895ce28f04ebef2318ab11615791b871c35eabbf8" --team="foo"
         - name: Lint
-          run: pnpm lint -- --api="http://127.0.0.1:9080" --token="${{ secrets.TURBO_SERVER_TOKEN }}" --team="foo"
+          run: pnpm lint -- --api="http://127.0.0.1:9080" --token="05de0230f01174d1f8cb4845a01dc6c895ce28f04ebef2318ab11615791b871c35eabbf8" --team="foo"
```
flybayer, line 44: This is great! 

Only thing is that it looks like it would be better to provide the `TURBO_TOKEN` env variable instead of passing `--token` each time.

https://turborepo.org/docs/reference/command-line-reference#--token

siddhsuresh: @flybayer I have updated the workflow.

dillonraphael (approved):

PR #3784, opened 2022-08-23
Migrate recipes

dillonraphael:
Closes: #3253 

### What are the changes and their implications?
Migrate over recipe functionality from legacy framework.

Some recipes still need to be updated to work with blitz 2.0 (https://github.com/blitz-js/blitz/issues/3813)

## Feature Checklist

Exposes paths such as:
- `nextConfig()`
- `blitzClient()`
- `blitzServer()`

Also exports new helper functions `transformNextConfig()` & `wrapAppReturnStatement()`, 

`transformNextConfig()` returns the following:

```ts
program: Program
configObj: []
pushToConfig: (property: j.ObjectProperty) => void
wrapConfig: (func: string | j.CallExpression) => {
   withBlitz: j.Identifier | j.CallExpression 
}
addRequireStatement: (identifier: string, packageName: string) => void
```

_(`wrapConfig().withBlitz` returns whatever is inside `withBlitz()`)_

`wrapAppReturnStatement()` accepts the following:
```ts
function wrapAppReturnStatement(
  program: Program,
  element: string
)
```
_The element is the name of the react component you want to wrap the return statement with_

- [X] Documentation added/updated https://github.com/blitz-js/blitzjs.com/pull/755

PR #3824, opened 2022-09-11
Update main.yml

siddhsuresh:
Closes: ?

> Creating this only due to continuous failures, sometimes due to core dumps or just existing randomly. The tests running will be running in parallel in GitHub actions as separate jobs rather than in the turbo. I shall close this if deemed unnecessary. 

- [x] split test job in the workflow into 1) Unit tests & 2) Integration tests
- [x] Taking inspiration from https://github.com/philips-labs/list-folder-action/blob/main/action.yml and creating another job to search the integration-tests folder to pass on as the argument of the Integration-test matrix
- [x] Running each integration test separately using the above-passed list of subdirectories and running only those tests using the ```--filter``` argument in-built in turbo cli https://turborepo.org/docs/core-concepts/filtering.
- [x] Caching only build and lint jobs while making each test run without a cache.
- [ ] TODO: fix auth test for windows as it is the only test that fails even separately.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,5 +1,3 @@
-# https://github.com/vercel/next.js/commits/canary/.github/workflows/build_test_deploy.yml
-
 name: CI
 
 on:
@@ -74,12 +72,14 @@ jobs:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
           server-token: ${{ env.TURBO_TOKEN }}
       - name: Build
-        run: pnpm build -- --api="http://127.0.0.1:9080"
+        uses: preactjs/compressed-size-action@v2
+        with:
+          pattern: "**/dist/**/*.{js,cjs,mjs}"
       - name: Build Apps
         run: pnpm build:apps -- --api="http://127.0.0.1:9080"
 
-  test:
-    name: "Test: ${{ matrix.os }} (node@${{ matrix.NODE_VERSION }})"
+  Unit-Tests:
+    name: "Unit Test: ${{ matrix.os }} (node@${{ matrix.NODE_VERSION }})"
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


siddhsuresh:
@dillonraphael This updated CI can test the migrate-recipes PR https://github.com/blitz-js/blitz/pull/3784 completely without failing [both linux and windows] 
Example Test Run in fork: https://github.com/siddhsuresh/blitz/pull/44

.github/workflows/main.yml
```diff
@@ -123,6 +175,6 @@ jobs:
         run: pnpm build -- --api="http://127.0.0.1:9080"
         shell: bash
 
-      - name: Test
-        run: pnpm test -- --api="http://127.0.0.1:9080"
-        shell: bash
+      - name: Test Packages
+        run: pnpm test -- --filter=./integration-tests/${{matrix.folder}}
+
```
dillonraphael, line 180: We should keep the `shell: bash` here. This is the reason why we were getting the error `cannot find prisma` with the previous seed setup.

integration-tests/auth/package.json
```diff
@@ -2,14 +2,18 @@
   "name": "test-auth",
   "version": "0.0.0",
   "private": true,
+  "prisma": {
+    "seed": "ts-node --compiler-options {\"module\":\"CommonJS\"} prisma/seed.ts"
+  },
```
dillonraphael, line 7: We shouldn't need this

integration-tests/auth/prisma/seed.ts
```diff
@@ -12,7 +12,11 @@ const seed = async () => {
       hashedPassword,
       role: "user",
     },
-  })
+  }).catch(console.error)
+
+  process.exit(0);
 }
 
+seed()
```
dillonraphael, line 20: We can revert this now too

integration-tests/auth/test/index.test.ts
```diff
@@ -128,7 +128,7 @@ describe("Auth Tests", () => {
       try {
         appPort = await findPort()
         app = await launchApp(appDir, appPort, {cwd: process.cwd()})
-        await seed()
+        // await seed()
```
dillonraphael, line 131: And this

dillonraphael:
Sorry for the back and forth @siddhsuresh 

I did some more research, and it seems you're right. I made a PR here https://github.com/blitz-js/blitz/pull/3832 with the CI passing. 

Some notes I took testing:
- According to the prisma docs, you're correct we should be adding the seed property to the prisma object in package.json
- The test fails with an exit code of 1 when calling the `seed()` function from inside the `beforeAll` in the test file **on windows only.**
- The seed command doesn't seem to work unless calling the seed() function from within the seed file it self. There's a closed issue for this on prisma, but we're using version `4.0.0` which is odd

siddhsuresh:
> Sorry for the back and forth @siddhsuresh
> 
> I did some more research, and it seems you're right. I made a PR here #3832 with the CI passing.
> 
> Some notes I took testing:
> 
> * According to the prisma docs, you're correct we should be adding the seed property to the prisma object in package.json
> * The test fails with an exit code of 1 when calling the `seed()` function from inside the `beforeAll` in the test file **on windows only.**
> * The seed command doesn't seem to work unless calling the seed() function from within the seed file it self. There's a closed issue for this on prisma, but we're using version `4.0.0` which is odd

yeah, I faced this issue in https://github.com/blitz-js/blitz/pull/3799 where I integrated windows testing, this was one of the pain points. The solution I had given was the only way I was able to seed the database.
Okay, @dillonraphael I shall revert to commit before the fix

integration-tests/auth/prisma/seed.ts
```diff
@@ -1,8 +1,8 @@
 import prisma from "./index"
-import {SecurePassword} from "@blitzjs/auth"
+import { SecurePassword } from "@blitzjs/auth"
 
 const seed = async () => {
-  await prisma.$reset()
+  // await prisma.$reset()
```
dillonraphael, line 5: This should still work though. Take a look at my other PR

integration-tests/auth/package.json
```diff
@@ -2,14 +2,18 @@
...
   "scripts": {
     "start:dev": "pnpm run prisma:start && next dev",
     "test": "pnpm run prisma:start && vitest run",
     "test-watch": "vitest",
     "start": "next start",
     "lint": "next lint",
     "clean": "rm -rf .turbo && rm -rf node_modules && rm -rf .next",
-    "prisma:start": "prisma generate && prisma migrate deploy",
+    "prisma:start": "prisma generate && prisma migrate dev && pnpm run prisma:seed",
+    "prisma:seed": "prisma db push --force-reset && prisma db seed",
```
dillonraphael, line 16: If we add back `prisma.$reset()` we only need to run `prisma db seed`

siddhsuresh: okay @dillonraphael, I shall update it

beerose (approved):
Looks good to me!

Thank you for working on this, @siddhsuresh!

integration-tests/auth/package.json
siddhsuresh: @dillonraphael I have reverted the changes. I shall try it again taking you PR as reference

siddhsuresh: I have removed the prisma:seed command as when I tested locally, the same seed command ran twice which caused the failure.

integration-tests/auth/prisma/seed.ts
dillonraphael: @siddhsuresh Let's uncomment the prisma reset

integration-tests/auth/package.json
dillonraphael: @siddhsuresh I don't understand. So where is the seed command being called now?

siddhsuresh: @dillonraphael here https://www.prisma.io/docs/guides/database/seed-database#integrated-seeding-with-prisma-migrate
I was just reading the prisma docs on seeding and found that the prisma migrate dev also runs the seed command defined in the package.json

dillonraphael: @siddhsuresh Ok makes sense. So last question.

Why `prisma db push --force-reset` instead of `prisma migrate reset`?

siddhsuresh: > 

I have removed ```prisma db push --force-reset``` in the latest commit and I am adding the ```prisma migrate reset --force``` in a new commit now to make sure the database does not have any data before the test

siddhsuresh: @dillonraphael any idea why it's failing? It is mentioning some missing packages in the host OS!

siddhsuresh:
@dillonraphael should I include https://playwright.dev/docs/cli#install-system-dependencies in the post install command in package.json?

Issue #3829, opened 2022-09-12
Middleware code does not wrap RPC handler


jan-wilhelm:
### What do you want and why?
In legacy-blitz it was possible to fully wrap queries and mutations in middlewares. This means that you could have middlewares execute code before calling `await next()` and then run code afterwards, and the call to `next()` would internally call the next middleware ... etc until it ultimately calls the handler itself. Therefore, we were able to e.g. measure the execution time of queries and mutations through a middleware.

In Blitz Toolkit, this is no longer possible. It seems that in the [index-server file](https://github.com/blitz-js/blitz/blob/6b788b03bef6a4fad67a82205f1670db06f02aa4/packages/blitz-next/src/index-server.ts#L190) for `blitz-next`, the handler gets called after all the middlewares. Previously, the handler was [added to the middleware stack](https://github.com/blitz-js/legacy-framework/blob/e1fc695d24f6dd85eba8ffce9f88e357edfb562d/nextjs/packages/next/server/api-utils.ts#L131), allowing outer middlewares to completely wrap queries and mutations.

### Possible implementation(s)
Change the implementation back so that the RPC handler gets pushed on the middleware stack (should be a two-line change). I have submitted a PR (#3830) that could close this issue.

PR #3830, opened 2022-09-12
Add RPC handler to middleware stack

jan-wilhelm:
Closes: #3829 

Changed the `api` function factory to treat any Blitz RPC handler the same way as other middlewares, allowing middleware code to run after the handler has run.

## Bug Checklist
- Not sure if this requires an integration test.

### DIFF
--- a/packages/blitz-next/src/index-server.ts
+++ b/packages/blitz-next/src/index-server.ts
@@ -187,8 +187,8 @@ export const setupBlitzServer = ({plugins, onError}: SetupBlitzOptions) => {
     ): NextApiHandler<TResult | void> =>
     async (req, res) => {
       try {
-        await handleRequestWithMiddleware(req, res, middlewares)
-        return handler(req, res, res.blitzCtx)
+        middlewares.push((req, res) => handler(req, res, res.blitzCtx));
+        return await blitz.handleRequestWithMiddleware(req, res, middlewares);
       } catch (error: any) {
         onError?.(error)
         return res.status(400).send(error)


flybayer (approved):

packages/blitz-next/src/index-server.ts
```diff
@@ -187,8 +187,8 @@ export const setupBlitzServer = ({plugins, onError}: SetupBlitzOptions) => {
     ): NextApiHandler<TResult | void> =>
     async (req, res) => {
       try {
-        await handleRequestWithMiddleware(req, res, middlewares)
-        return handler(req, res, res.blitzCtx)
+        middlewares.push((req, res) => handler(req, res, res.blitzCtx));
+        return await handleRequestWithMiddleware(req, res, middlewares);
```
dillonraphael, line 191: @jan-wilhelm The CI is failing because there are some type errors during the build process.

dillonraphael (dismissed):

packages/blitz-next/src/index-server.ts
beerose, line 191: In order to have it typed properly we can do this:
```suggestion
     return await handleRequestWithMiddleware(req, res, [...middlewares, (req, res) => handler(req, res, res.blitzCtx)]);
```
The `handleRequestWithMiddleware` will infer the types properly based on the third argument.

jan-wilhelm: oh yeah, great idea! Thank you!

packages/blitz-next/src/index-server.ts
```diff
@@ -187,8 +187,7 @@ export const setupBlitzServer = ({plugins, onError}: SetupBlitzOptions) => {
     ): NextApiHandler<TResult | void> =>
     async (req, res) => {
       try {
-        await handleRequestWithMiddleware(req, res, middlewares)
-        return handler(req, res, res.blitzCtx)
+     return await handleRequestWithMiddleware(req, res, [...middlewares, (req, res) => handler(req, res, res.blitzCtx)]);
```
beerose, line 190:
```suggestion
         return await handleRequestWithMiddleware(req, res, [...middlewares, (req, res) => handler(req, res, res.blitzCtx)]);
```

beerose (approved):

Issue #3840, opened 2022-09-14
Throw `RedirectError` and `AuthenticationError` on second render 


dillonraphael:
### What is the problem?

We need to throw these errors after mount

- [ ] RedirectError
- [ ] AuthenticationError

PR #3842, opened 2022-09-15
Throw redirect errors when component has mounted in blitz-auth

dillonraphael:
Closes: #3840

Inside `blitz-auth`, make sure the component has mounted before throwing the `RedirectError()`.

### DIFF
--- a/.changeset/heavy-cobras-own.md
+++ b/.changeset/heavy-cobras-own.md
@@ -1,0 +1,7 @@
+---
+"@blitzjs/auth": patch
+"@blitzjs/next": patch
+"blitz": patch
+---
+
+Ensure the component has mounted before throwing a redirect with redirectAuthenticatedTo. This will fix the hydration error from react.
--- a/.changeset/heavy-cobras-own.md
+++ b/.changeset/heavy-cobras-own.md
@@ -4,4 +4,4 @@
 "blitz": patch
 ---
 
-Ensure the component has mounted before throwing a redirect with redirectAuthenticatedTo. This will fix the hydration error from react.
+Fix auth related React hydration errors by not redirecting until after component mount.
--- a/packages/blitz-auth/src/client/index.tsx
+++ b/packages/blitz-auth/src/client/index.tsx
@@ -248,10 +248,12 @@ export function getAuthValues<TProps = any>(
 function withBlitzAuthPlugin<TProps = any>(Page: ComponentType<TProps> | BlitzPage<TProps>) {
   const AuthRoot = (props: ComponentProps<any>) => {
     useSession({suspense: false})
-    const [mounted, setMounted] = useState<boolean>(false)
+    const [mounted, setMounted] = useState(false)
 
     useEffect(() => {
       setMounted(true)
+
+      return () => setMounted(false)
     }, [])
 
     let {authenticate, redirectAuthenticatedTo} = getAuthValues(Page, props)
--- a/packages/blitz-auth/src/client/index.tsx
+++ b/packages/blitz-auth/src/client/index.tsx
@@ -169,8 +169,6 @@ export const useAuthorizeIf = (condition?: boolean) => {
 
   useEffect(() => {
     setMounted(true)
-
-    return () => setMounted(false)
   }, [])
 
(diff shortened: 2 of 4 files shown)


packages/blitz-auth/src/client/index.tsx
```diff
@@ -248,6 +248,11 @@ export function getAuthValues<TProps = any>(
 function withBlitzAuthPlugin<TProps = any>(Page: ComponentType<TProps> | BlitzPage<TProps>) {
   const AuthRoot = (props: ComponentProps<any>) => {
     useSession({suspense: false})
+    const [mounted, setMounted] = useState<boolean>(false)
```
beerose, line 251: You don't need a type here as typescript can infer that one.

```suggestion
    const [mounted, setMounted] = useState(false)
```

packages/blitz-auth/src/client/index.tsx
```diff
@@ -248,6 +248,11 @@ export function getAuthValues<TProps = any>(
...
+
+    useEffect(() => {
+      setMounted(true)
+    }, [])
```
beerose, line 255: Probably also need a clean up function:
```suggestion
    useEffect(() => {
      setMounted(true)
      
      return () => setMounted(false)
    }, [])
```

beerose (changes requested):

packages/blitz-auth/src/client/index.tsx
```diff
@@ -274,9 +281,11 @@ function withBlitzAuthPlugin<TProps = any>(Page: ComponentType<TProps> | BlitzPa
               : formatWithValidation(redirectAuthenticatedTo)
 
           debug("[BlitzAuthInnerRoot] redirecting to", redirectUrl)
```
flybayer, line 283: move this log into the if statement (and might as well add the other url code in there also)

packages/blitz-auth/src/client/index.tsx
```diff
@@ -289,9 +298,11 @@ function withBlitzAuthPlugin<TProps = any>(Page: ComponentType<TProps> | BlitzPa
           const url = new URL(redirectTo, window.location.href)
           url.searchParams.append("next", window.location.pathname)
           debug("[BlitzAuthInnerRoot] redirecting to", url.toString())
```
flybayer, line 300: move this log into the if statement (and might as well add the other url code in there also)

.changeset/heavy-cobras-own.md
```diff
@@ -0,0 +1,7 @@
+---
+"@blitzjs/auth": patch
+"@blitzjs/next": patch
+"blitz": patch
+---
+
+Ensure the component has mounted before throwing a redirect with redirectAuthenticatedTo. This will fix the hydration error from react.
```
flybayer, line 7: It's better to start with the "what" because that's the most important info to users than the "how".
```suggestion
Fix auth related React hydration errors by not redirecting until after component mount.
```

flybayer (commented):
Also this doesn't close the linked issue until the `AuthenticationError`s have been fixed also

packages/blitz-auth/src/client/index.tsx
flybayer: cleanup function not needed. Cleanup function only needed for memory management in cases where you add event handlers or subscriptions

dillonraphael:
> Good work! Few comments. Also this doesn't close the linked issue until the `AuthenticationError`s have been fixed also

Oops. Good catch. Thanks for the review. Just pushed.

packages/blitz-auth/src/client/index.tsx
```diff
@@ -165,7 +165,15 @@ export const useSession = (options: UseSessionOptions = {}): ClientSession => {
 }
 
 export const useAuthorizeIf = (condition?: boolean) => {
-  if (isClient && condition && !getPublicDataStore().getData().userId) {
+  const [mounted, setMounted] = useState(false)
+
+  useEffect(() => {
+    setMounted(true)
+
+    return () => setMounted(false)
```
flybayer, line 173: don't need cleanup functions with useState

flybayer (approved):

Issue #3841, opened 2022-09-15
Make NextJS Feature Custom Page Extensions work with BlitzJS 2.0 RPC


jatidevelopments:
### What do you want and why?

Make NextJS Feature Custom Page Extensions work with BlitzJS 2.0 RPC (https://nextjs.org/docs/api-reference/next.config.js/custom-page-extensions)

Use cases:

We want to have non-pages in the /pages folder then we can declare NextJs Page Extensions in the config file. This enables to write pages with e.g. signup.tsx > signup.page.tsx. 

```
const config = {
  pageExtensions: ['page.tsx', 'ts', 'page.jsx', 'page.js'],
};
```

### Possible implementation(s)

The webpack rule that checks if the pages/api/rpc/[[...blitz]].ts exists need to be changed. Because we need to change the filename to [[...blitz]].page.ts to make the BLitz RPC Layer work with the Custom Page Extension feature. 

Change the @blitzjs/rpc > index-server.cjs > installWebpackConfig function (Line 100)

`test: /[\\/]\[\[\.\.\.blitz]]\.[jt]sx?$/,`

to

`test: /[\\/]\[\[\.\.\.blitz]].+\.[jt]sx?$/,`

beerose:
Thanks for reporting!

File to update: https://github.com/datner/blitz/blob/f6dac093d240a93543b7bd09f55069cb3291d980/packages/blitz-rpc/src/index-server.ts#L91

PR #3843, opened 2022-09-15
Update regex for wildcard blitz route in webpack config

dillonraphael:
Closes: #3841

Allow for custom page extensions for the wildcard blitz route. This PR update's the regex in the webpack config found in `blitz-rpc`.

### DIFF
--- a/.changeset/itchy-spoons-tan.md
+++ b/.changeset/itchy-spoons-tan.md
@@ -1,0 +1,6 @@
+---
+"@blitzjs/rpc": patch
+"blitz": patch
+---
+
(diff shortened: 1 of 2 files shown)


.changeset/itchy-spoons-tan.md
```diff
@@ -0,0 +1,6 @@
+---
+"@blitzjs/rpc": patch
+"blitz": patch
+---
+
+Allow for custom page extensions for the wildcard blitz route. For example [...blitz].api.ts
```
beerose, line 6: Maybe we could add a link to the Next docs: https://nextjs.org/docs/api-reference/next.config.js/custom-page-extensions

beerose (approved):

Issue #3818, opened 2022-09-04
Wrong commit user after creating a new app


noxify:
### What is the problem?

After running the `blitz new` command, the generated code will be commited to git.
The user who did this is not myself, it's https://github.com/hng63

I would expect that my own user is used to make the initial commit.
Or alternative `Blitz.JS CLI` is used as user as defined in the code: https://github.com/blitz-js/blitz/blob/main/packages/generator/src/generators/app-generator.ts#L269-L280

### Paste all your error logs here:

```
...
```

### Paste all relevant code snippets here:

```
...
```

### What are detailed steps to reproduce this?

1. Run `blitz new test-app` and follow the instructions
2. Create a repo at github and push the changes to the remote repo ( or create the remote repo directly via VSCode )
3. check the commit history / contributor list

### Run `blitz -v` and paste the output here:

```
....
```

### Please include below any other applicable logs and screenshots that show your problem:

* https://github.com/noxify/blitz-hackernews/graphs/contributors
* https://github.com/noxify/blitz-international-issue/graphs/contributors

dillonraphael:
I think I've seen someone else with this issue before.

Can you try `rm -rf .git` then running `git init` ?

noxify:
Hi @dillonraphael,

yes this solves the issue.

After running `blitz new blitz-issue-3818`:
![Bildschirmfoto 2022-09-07 um 09 20 05](https://user-images.githubusercontent.com/521777/188814874-79f3dab3-141b-4399-b348-e41c9a2a79ba.png)

After deleting the `.git` directory and running `git init` (+commit):
![Bildschirmfoto 2022-09-07 um 09 20 57](https://user-images.githubusercontent.com/521777/188814876-3f9808c2-7d41-4016-be7a-bb751bd12d0d.png)

Maybe we could just remove the `-c` properties from the command to use the default git config ( = current user name + email )?

What do you think?

PR #3852, opened 2022-09-19
Remove random user git config when generating a new project

dillonraphael:
Closes: #3818

Remove the `-c` flag when running `git commit` during generating a new project.

### DIFF
--- a/.changeset/hip-buttons-dance.md
+++ b/.changeset/hip-buttons-dance.md
@@ -1,0 +1,5 @@
+---
+"@blitzjs/generator": patch
+---
+
+Remove random user git config when generating a new project & commiting.
--- a/.changeset/hip-buttons-dance.md
+++ b/.changeset/hip-buttons-dance.md
@@ -2,4 +2,4 @@
 "@blitzjs/generator": patch
 ---
 
-Remove random user git config when generating a new project & commiting.
+Remove the random user (noop@blitzjs.com) & use user's default git account when commiting a new generated blitz project.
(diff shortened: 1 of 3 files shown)


.changeset/hip-buttons-dance.md
```diff
@@ -0,0 +1,5 @@
+---
+"@blitzjs/generator": patch
+---
+
+Remove random user git config when generating a new project & commiting.
```
beerose, line 5: Hmm, we could be more clear about what changes for the users by saying that now we're going to use users' git account to commit instead of a `Blitz.js CLI` (`noop@blitzjs.com`).

packages/generator/src/generators/app-generator.ts
```diff
@@ -267,17 +267,7 @@ export class AppGenerator extends Generator<AppGeneratorOptions> {
...
-          "user.name='Blitz.js CLI'",
-          "-c",
-          "user.email='noop@blitzjs.com'",
-          "commit",
-          "--no-gpg-sign",
-          "--no-verify",
-          "-m",
-          "Brand new Blitz app!",
-        ],
+        ["commit", "--no-gpg-sign", "--no-verify", "-m", "Brand new Blitz app!"],
```
beerose, line 270: Since we're changing it to use the user's git credentials, I'm not sure we still want `--no-gpg-sign`. Are you sure it's still preferred?

dillonraphael: I believe you're right. In general we should be using the user's local config regardless.

beerose (approved):

Issue #3801, opened 2022-08-29
Only run `prisma generate` if the prisma client hasn't been generated


dillonraphael:
### What is the problem?

The current implementation creates a temporary `blitz-prisma-status.json` file in the temp directory and checks every time the prisma schema file has been changed. 

We can drastically simplify this, by instead just checking if the prisma client files have been generated. We don't need to check every time the schema file has been changed since it's expected to migrate which leads to generating the update prisma client. 

We also need to account for projects that are in monorepos.

Just need to replace this https://github.com/siddhsuresh/blitz/blob/7224bce90a244d739fd5f891f76cffbe58563bf9/packages/blitz/src/cli/utils/codegen-tasks.ts#L42-L53 with potentially a `require.resolve` for the prisma client file.

### Paste all relevant code snippets here:

Something to try, instead of looking for directly in the `node_modules` directory. 

We should be able todo a `require.resolve` but something to keep note is how `pnpm` handles hoisting and directory structure. We may need to traverse outside the `.pnpm` directory found in the root of `node_modules`.

Issue #3776, opened 2022-08-22
Generating prisma client fails when using global `blitz dev` command


dillonraphael:
### What is the problem?

Fails to generate the prisma client when running `blitz dev` with the global cli. It works when using `npx blitz dev`.

### Paste all your error logs here:

```shell
~ > blitz dev
...
Generating Prisma client{
  "code": "ENOENT"
}
...
```

beerose:
The issue seems to result from the wrong path for `schema.prisma`. I believe that it will be fixed by https://github.com/blitz-js/blitz/pull/3855

PR #3855, opened 2022-09-20
Only generate the prisma client if it's not found

dillonraphael:
Closes: #3801 , https://github.com/blitz-js/blitz/issues/3776

Only generate the prisma client if it's not found. It checks for both `@prisma/client` and `.prisma`.

### DIFF
--- a/packages/blitz/src/cli/utils/codegen-tasks.ts
+++ b/packages/blitz/src/cli/utils/codegen-tasks.ts
@@ -40,21 +40,24 @@ export const codegenTasks = async () => {
 
     if (hasPrisma) {
       const foundPrismaClient = resolveFrom.silent(process.cwd(), "@prisma/client")
+      const foundDotPrismaClient = resolveFrom.silent(process.cwd(), ".prisma")
+
       if (
         !foundPrismaClient ||
-        (foundPrismaClient && !fs.existsSync(join(foundPrismaClient, "../../..", ".prisma")))
+        (foundPrismaClient && !fs.existsSync(join(foundPrismaClient, "../../..", ".prisma"))) ||
+        (foundDotPrismaClient && !foundDotPrismaClient)
       ) {
         let prismaSpinner = log.spinner(`Generating Prisma client`).start()
         const result = await runPrisma(["generate"], true)
-        if (result.success) {
-          prismaSpinner.succeed(log.greenText("Generated Prisma client"))
-        } else {
-          prismaSpinner.fail()
-          console.log("\n" + result.stderr)
-          process.exit(1)
+        if (typeof result === "object") {
+          if (result.success) {
+            prismaSpinner.succeed(log.greenText("Generated Prisma client"))
+          } else {
+            prismaSpinner.fail()
+            console.log("\n" + result.stderr)
+            process.exit(1)
+          }
         }
-      } else {
-        log.success("Prisma client was up to date")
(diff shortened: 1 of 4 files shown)


packages/blitz/src/cli/utils/codegen-tasks.ts
```diff
@@ -39,33 +39,23 @@ export const codegenTasks = async () => {
...
+        if (result.success) {
+          prismaSpinner.succeed(log.greenText("Generated Prisma client"))
         } else {
-            log.success("Prisma client was up to date")
+          prismaSpinner.fail()
+          console.log("\n" + result.stderr)
+          process.exit(1)
         }
+      } else {
+        log.success("Prisma client was up to date")
```
beerose, line 57: I'm not sure about this message. According to the code, prisma client **was found**, and not necessarily **up to date**. Saying "up to date" can be misleading.

beerose: If prisma client was found, then we're skipping the `prisma generate` step. IMO we shouldn't print anything in this case.

packages/blitz/src/cli/utils/codegen-tasks.ts
```diff
@@ -39,33 +39,23 @@ export const codegenTasks = async () => {
...
-                console.log("\n" + result.stderr)
-                process.exit(1)
-            }
-            await fs.writeJson(join(process.cwd(), `blitz-${creationDate.birthtimeMs}-prisma-client.json`), {
-                lastModified: lastModified.mtimeMs,
-            })
+      const foundPrismaClient = resolveFrom.silent(process.cwd(), "@prisma/client")
+      if (
+        !foundPrismaClient ||
+        (foundPrismaClient && !fs.existsSync(join(foundPrismaClient, "../../..", ".prisma")))
```
beerose, line 45: Does it work for both pnpm and not-pnpm package managers?

beerose (commented):
Are we sure it works in monorepos? 

From what I saw, it finds a prisma client in the monorepo's root `node_modules`, while I guess it should look for a particular app's `node_modules`.

packages/blitz/src/cli/utils/codegen-tasks.ts
dillonraphael: Just pushed a fix. Tested with flightcontrol as well to get it working.

packages/blitz/src/cli/utils/codegen-tasks.ts
dillonraphael: Ya very valid point

packages/blitz/src/utils/run-prisma.ts
```diff
@@ -1,27 +1,32 @@
...
 
-  return {
-    success: code === 0,
-    stderr: silent ? cp_stderr.join("") : undefined,
+    return {
+      success: code === 0,
+      stderr: silent ? cp_stderr.join("") : undefined,
+    }
+  } else {
+    log.error("Prisma bin not found")
```
beerose, line 30: It should return in each case, as other usages expect this function to return a value, e.g.:
CleanShot 2022-09-23 at 15 44 04@2x

Not sure why CI didn't catch this.

beerose (changes requested):

beerose (approved):

Issue #3862, opened 2022-09-23
Add `blitz console` back to Blitz 2.0


beerose:
We should also switch to esbuild while at it: https://github.com/blitz-js/blitz/issues/1098

siddhsuresh:
@beerose @dillonraphael @flybayer can I work on this?

Issue #3858, opened 2022-09-22
Add `blitz console` option to skip preloading of queries/mutations


johnlk:
### What do you want and why?

I wish there were a way to skip over the pre-loading of queries and mutations when executing `blitz console`. For one, it's very slow especially when you have > 40 queries + mutations as my project does. Second, I almost never use my queries in the console.

When I SSH to my remote server and run `blitz console` it takes > 2 mins to preload queries. Locally, it's about 6 seconds. 

### Possible implementation(s)

Adding an option to skip over the query/mutation preloading. Maybe something like `--skip-preload`

### Additional context

<img width="511" alt="Screen Shot 2022-09-22 at 4 02 44 PM" src="https://user-images.githubusercontent.com/8771036/191769764-6fa28780-8c64-4233-9336-849a97ce2eca.png">

Related to #1098

beerose:
Thanks for the suggestion! Currently blocked on https://github.com/blitz-js/blitz/issues/3862

Issue #1098, opened 2020-09-17
`blitz console` is slow to load - switch to esbuild loader


flybayer:
### What is the problem?

Starting `blitz console` has a long hangup time after "Loading your Code" is finished and before you can start typing.

It gets much worse as you add more queries and mutations.

### Steps to Reproduce

1. `blitz new myapp`
2. `cd myapp`
3. `blitz g resource project`
3. `blitz g resource task`
3. `blitz g resource comment`
3. `blitz db migrate`
3. `blitz console`
3. Try to start typing immediately once the `>` prompt appears. Notice that you can't type for a bit of time.

### Other

The main problem is [`loadBlitz`](https://github.com/blitz-js/blitz/blob/canary/packages/repl/src/utils/load-blitz.ts#L23), which for some reason is super slow on first start. Running `.reload` after initial start (which re-runs loadBlitz) is quite fast.

You can add a log statement [here](https://github.com/blitz-js/blitz/blob/canary/packages/repl/src/utils/load-blitz.ts#L35) and see how slow it is for each module.

We are already using a [lazy require loader](https://github.com/blitz-js/blitz/blob/canary/packages/cli/src/index.ts#L2) which apparently isn't working here.

Probably we need to add a proxy object for each dependency, that only loads the module when it's accessed. But maybe there's a better way.

### Update April 2021

See https://github.com/blitz-js/blitz/issues/1098#issuecomment-823670547

JoseRFelix:
Tried replicating this on Mac OS Catalina 10.15.6, but it worked without any delay.

flybayer:
@JoseRFelix thanks for investigating! This is definitely WAY better in the latest blitz canary. That said, I'm still noticing a 1-3s delay on my project which has a good bit more modules than what you listed here.

I still think we should lazy load these, because as apps get bigger, this will get worse, no matter how optimized it is.

JoseRFelix:
So, I've been thinking about some possible solutions:

1. Have a loading indicator while all the modules are loading, and show input when done. This will make the user have to wait before he can use the console. It is not probably the best solution, but it is the fastest to develop.
2. Have a command that lists the files' names of all available mutations, queries, jobs, integrations, and utils file names. Then, the user will require what he needs just like in javascript, using another command. This would solve the problem but it would increase the learning curve and have the user requiring the files every time.  
[...]

PR #3863, opened 2022-09-25
Migrate console

siddhsuresh:
Closes: #3862
Closes: #3858
Closes: https://github.com/blitz-js/blitz/issues/1098

- [x] migrate required files from legacy framework repo 
- https://github.com/blitz-js/legacy-framework/blob/main/nextjs/packages/next/cli/next-console.ts
- https://github.com/blitz-js/legacy-framework/blob/main/packages/cli/src/commands/console.ts
- [x] use ```esbuild``` instead of ```ts-node``` with reference to https://github.com/blitz-js/blitz/issues/1098#issuecomment-823670547 
- [x] add ```--only-db``` flag allows users to only load the ```db``` module in the initial load of the console but allows usage other modules when changes are made 

> In the following images the "Loaded ... with the time" is only for reference and is not present in the code

### Before esbuild

image

### After esbuild

image

### With ```skip-preload```
image

### If the ```db``` folder does not exist by the user provide the location of ```prisma``` schema
```js
//package.json

"prisma": {
    "schema": "prisma/schema.prisma"
  },
```
image

## Only reloading the modules that are changed and not all when the user makes changes

```ts
const setupFileWatchers = async (repl: REPLServer) => {
  debug("Setting up file watchers...")
  const watchers = [
    watch(await getBlitzModulePaths(), {
      ignoreInitial: true,
    }).on(
      "all",
      async (event:string, path: string) => { 👈🏼
        const modulePath = path
        const modules = await loadBlitz(false, modulePath) 👈🏼
        loadBlitzModules(repl, modules)
      }
    )
  ]
```

For example when a user changes the db/index.ts file only that file is loaded by the code 

image

> pnpm blitz -v
Blitz version: 2.0.0-beta.4 (local)
Windows 11 | win32-x64 | Node: v16.13.0
 Package manager: npm
  System:
    OS: Windows 10 10.0.22621
    CPU: (4) x64 11th Gen Intel(R) Core(TM) i3-1115G4 @ 3.00GHz
    Memory: 570.32 MB / 7.70 GB
  Binaries:
    Node: 16.13.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.17 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 8.1.0 - C:\Program Files\nodejs\npm.CMD


beerose (changes requested):
This is amazing work! Thank you!

I started testing it and it looks like top-level await doesn't work:

CleanShot 2022-09-26 at 13 31 40@2x

siddhsuresh:
> This is amazing work! Thank you!
> 
> I started testing it and it looks like top-level await doesn't work:
> 
> CleanShot 2022-09-26 at 13 31 40@2x

Hey @beerose, I am not getting that error on my end. I just tested in ubuntu through WSL2 and here is the terminal output. (Ignore the speed of my terminal, WSL2 in my laptop is very slow 😅)

[asciicast](https://asciinema.org/a/rrMs9Lb6xl51GB68am0mRtETk)

beerose:
Hmm, that's interesting! I wouldn't expect it to be OS related, but maybe it is? I'm on macos, Node 16.

siddhsuresh:
> Hmm, that's interesting! I wouldn't expect it to be OS related, but maybe it is? I'm on macos, Node 16.

@beerose I am not sure what is going wrong because it works for me on both windows 11 and in ubuntu

siddhsuresh:
> This is amazing work! Thank you!
> 
> I started testing it and it looks like top-level await doesn't work:
> 
> CleanShot 2022-09-26 at 13 31 40@2x

@beerose After researching how the REPL works with top level await, I have introduced this fix.
Now users can write commands without writing the ```await``` as I have overridden the eval code of the NodeJS REPL to resolve all commands as promises

### Before
image

### After 
image

Does this solve the problem on your end?

beerose:
@siddhsuresh we should also change the message:

```
Tips: - Exit by typing .exit or pressing Ctrl-D
      - Use your db like this: await db.project.findMany()
      - ...
```

(remove the await)

beerose:
@siddhsuresh I was also thinking about the `skip-preload` and I'm not sure it's obvious what it means (if someone didn't read the docs first). Maybe we can call thins flag `--skip-resolvers`?

siddhsuresh:
> @siddhsuresh I was also thinking about the `skip-preload` and I'm not sure it's obvious what it means (if someone didn't read the docs first). Maybe we can call thins flag `--skip-resolvers`?

@beerose yeah I agree, I had given the flag name by taking reference from #3858, but also was thinking of changing it to ```only-db``` because it is exactly what the flag would do. What are your thoughts on this? 
If this would not be suitable I agree with naming it ```skip-resolvers```

beerose:
> > @siddhsuresh I was also thinking about the `skip-preload` and I'm not sure it's obvious what it means (if someone didn't read the docs first). Maybe we can call thins flag `--skip-resolvers`?
> 
> @beerose yeah I agree, I had given the flag name by taking reference from #3858, but I was also thinking of changing it to `--only-db` because it is exactly what the flag would do. What are your thoughts on this? If this would not be suitable I agree with naming it `skip-resolvers`

I was also thinking about `only-db` or `db-only`, so totally up to you :)

siddhsuresh:
> > > @siddhsuresh I was also thinking about the `skip-preload` and I'm not sure it's obvious what it means (if someone didn't read the docs first). Maybe we can call thins flag `--skip-resolvers`?
> > 
> > 
> > @beerose yeah I agree, I had given the flag name by taking reference from #3858, but I was also thinking of changing it to `--only-db` because it is exactly what the flag would do. What are your thoughts on this? If this would not be suitable I agree with naming it `skip-resolvers`
> 
> I was also thinking about `only-db` or `db-only`, so totally up to you :)

@beerose I have changed the flag name to ```--only-db```

.changeset/spotty-zoos-film.md
```diff
@@ -0,0 +1,5 @@
+---
+"blitz": patch
+---
+
+Migrate console
```
beerose, line 5: Add `blitz console` CLI command back to toolkit

beerose (approved):
Amazing work!

dillonraphael (approved):

Issue #3748, opened 2022-08-17
Add custom template usage back to toolkit 


madaxen86:
The legacy version supported the use of custom templates.
Docs: 
[Custom templates](https://blitzjs.com/docs/cli-generate#custom-templates)
[blitz.config.ts](https://blitzjs.com/docs/cli-generate#custom-templates)
This does not work at the moment.
`blitz generate` will always use the default templates instead.
@beerose

siddhsuresh:
@beerose can I work on this and add this back to the generate command? If so, should I add the option for the custom template folder in the ```next.config.js``` or in the ```app/blitz-server.ts``` or is some other location preferred?

beerose:
@siddhsuresh we were thinking about using `export default` in `app/blitz-server.ts` for this. Something like:

```
// app/blitz-server.ts

export const {api} = setupBlitzServer(...)

const config = {
  customTemplates: ...
  // .. 
}

export default config
```

PR #3866, opened 2022-09-26
#3748 Add custom template option back to blitz generate command

siddhsuresh:
Closes: #3748

- [x] add type BlitzServerConfig for the types used in the ```blitz-server.ts``` config
- [x] create a new command ```blitz generate custom-templates```  that generates the default templates present in the ```@blitzjs/generator``` to provide an easy starting point for the developer to use the custom templates feature 
   - getting custom template path from the user and using that to save the templates
   - updating the ```app/blitz-sever``` file with the required changes needed to use custom templates
- [x] read the named export ```cliConfig``` exported from ```app/blitz-server.{ts|js}``` and use the ```customTemplates``` property to pass the custom template path 

#### Example Usage

image

```ts
// app/blitz-server.ts
import type { BlitzServerConfig } from "blitz"

...

export const cliConfig: BlitzServerConfig = {
  customTemplates: "app/template",
}
```


packages/blitz/src/cli/commands/generate.ts
```diff
@@ -227,9 +271,20 @@ const generate: CliCommand = async () => {
...
     const generators = generatorMap[selectedType as keyof typeof generatorMap]
 
+    const isTypeScript = await getIsTypeScript()
+    const blitzServerPath = isTypeScript ? "app/blitz-server.ts" : "app/blitz-server.js"
+    const blitzServer = require("path").join(process.cwd(), blitzServerPath)
+    const {register} = require("esbuild-register/dist/node")
+    const {unregister} = register({
+      target: "es6",
+    })
+    const blitzConfig = require(blitzServer).default
```
dillonraphael, line 281: I think we should do this differently. Maybe not export the config object as default and instead add as a property to `setupBlitzServer`.

This may be a design choice though. @flybayer @beerose  what do you think?

dillonraphael (changes requested):

packages/blitz/src/cli/commands/generate.ts
beerose: I suggested this solution as this is what we agreed on during a planning call last week @dillonraphael.

.changeset/slow-impalas-tap.md
```diff
@@ -0,0 +1,6 @@
+---
+"blitz": patch
+"@blitzjs/generator": patch
+---
+
+#3748 Add custom template usage back to toolkit
```
flybayer, line 6: we should flesh this out more, tell the benefit, and show how to use the new command, etc.

packages/blitz/src/cli/commands/generate.ts
```diff
@@ -72,6 +74,7 @@ const args = arg(
     "--parent": String,
     "--dry-run": Boolean,
     "--env": String,
+    "--generate-custom-templates": Boolean,
```
flybayer, line 77: I feel like we should use generator type for this instead of a flag, because flags are things that work with any command but this doesn't.

so instead like this: `blitz generate custom-templates`

packages/blitz/src/cli/commands/generate.ts
flybayer: We agreed to export just the config, not as a property of setupBlitzServer, however I think we should use a named export of `cliConfig` instead of a default export

packages/blitz/src/cli/commands/generate.ts
siddhsuresh: @flybayer yeah I agree, I have now removed the flag and converted it to a blitz generator command

.changeset/slow-impalas-tap.md
```diff
@@ -0,0 +1,6 @@
+---
+"blitz": patch
+"@blitzjs/generator": patch
+---
+
+#3748 Add custom template usage back to toolkit, and a new command `blitz generate custom-templates` to provide an easy starting point for users to customize the default templates.
```
beerose, line 6:
```suggestion
Allow passing custom templates to the `blitz generate` command. Extend the `generate` command with `custom-templates` to provide an easy starting point for users to customize the default templates: `blitz generate custom-templates`.
```

packages/generator/src/generator.ts
```diff
@@ -16,6 +16,88 @@ import {readdirRecursive} from "./utils/readdir-recursive"
...
+) => {
+  const blitzServer = isTypeScript ? "app/blitz-server.ts" : "app/blitz-server.js"
+  const blitzServerPath = require("path").join(process.cwd(), blitzServer)
+  const userConfigModuleSource = fs.readFileSync(blitzServerPath, {encoding: "utf-8"})
+  const userConfigModule = j(userConfigModuleSource, {parser: customTsParser})
+  const program = userConfigModule.get()
+  const cliConfigDeclaration = userConfigModule
+    .find(j.ExportNamedDeclaration, {
+      declaration: {
+        type: "VariableDeclaration",
```
beerose, line 31: Couldn't we use esbuild or something to eval this file and extract default instead?

beerose: Ah right 🤦‍♀️ ok, ignore my question

beerose:
One last thing — we have to update docs for it

dillonraphael (approved):

PR #3867, opened 2022-09-28
update userId to be exclusive union of `number` and `string`

p6l-richard:
Closes: I haven't opened an issue for this. I [documented this](https://twitter.com/richardpoelderl/status/1575166499827744772?s=20&t=sSgCX9pPyrtUTQ0uN1QYsg) in a twitter thread if that's any help. Also, you can see the error in the demo below.

The userId is currently typed as `unknown` which throws a type error when the prisma model is defined as an `Int`
image

Similarly to how prisma provides `XOR` types, I've patched the PublicData["userId"] type to be a mutually exclusive union instead.

I'm not sure where to include the `XOR` & `Without` type utilities, please let me know if they should be moved elsewhere.

Here's a demo of how this makes the types error disappear (& how a regular union isn't enough):
https://user-images.githubusercontent.com/18185649/192862646-ae681846-1db5-4d60-9ff7-4591f9b17a87.mov

## Bug Checklist

## Feature Checklist

### DIFF
--- a/packages/blitz-auth/src/shared/types.ts
+++ b/packages/blitz-auth/src/shared/types.ts
@@ -4,10 +4,20 @@ export interface Session {
   // isAuthorize can be injected here
   // PublicData can be injected here
 }
+/**
+ * XOR is needed to have a real mutually exclusive union type
+ * https://stackoverflow.com/a/53229567/5608461
+ */
+type Without<T, U> = { [P in Exclude<keyof T, keyof U>]?: never };
+type XOR<T, U> =
+ T extends object ?
+ U extends object ?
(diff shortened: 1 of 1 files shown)


flybayer:
Ah, you need to add this `types.ts` file to your repo somewhere: https://github.com/blitz-js/blitz/blob/main/packages/generator/templates/app/types.ts

(except the `views` field in there, that shouldn't be in our template).

That should fix your type error

beerose:
As Brandon said, the error you had is a result of missing `types.ts` file in the root of your application. We are currently working on better docs, so it should be covered in a migration guide. 

Meanwhile, I'm going to close this PR.

Issue #2863, opened 2021-10-20
`blitz generate` doesn't handle string defaults


chrisjkr:
### What is the problem?

It's currently impossible to generate a Prisma model with a string default using `blitz generate`. Not a big deal since it's easy to modify the model manually before migrating, but it's a bad DX. Open to fixing this, if anyone can point me to where to look for and what the intended behaviour is.

### Paste all your error logs here:

No error log, see reproduction steps below.

### Paste all relevant code snippets here:

See below

### What are detailed steps to reproduce this?

```
blitz generate all test name:string:default="test"
blitz generate all test name:string:default=test
blitz generate all test name:string:default='test'
blitz generate all test name:string:default=\"test\"
```

The first 3 commands will produce:

```
> model Test {
>   id        Int      @id @default(autoincrement())
>   createdAt DateTime @default(now())
>   updatedAt DateTime @updatedAt
>   name      String   @default(test)
> }

```

The last one will produce:

```
> model Test {
>   id        Int      @id @default(autoincrement())
>   createdAt DateTime @default(now())
>   updatedAt DateTime @updatedAt
>   name      String
> }

```

Should be:

```
> model Test {
>   id        Int      @id @default(autoincrement())
>   createdAt DateTime @default(now())
>   updatedAt DateTime @updatedAt
>   name      String   @default("test")
> }

```

### Run `blitz -v` and paste the output here:

```
macOS Big Sur | darwin-x64 | Node: v16.9.1

blitz: 0.41.1 (global)
blitz: 0.41.1 (local)

  Package manager: yarn 
  System:
    OS: macOS 11.2.3
    CPU: (8) x64 Apple M1
    Memory: 93.39 MB / 16.00 GB
    Shell: 3.2.0 - /usr/local/bin/fish
  Binaries:
    Node: 16.9.1 - /usr/local/bin/node
    Yarn: 1.22.10 - /opt/homebrew/bin/yarn
    npm: 7.20.5 - /opt/homebrew/bin/npm
    Watchman: 4.9.0 - /opt/homebrew/bin/watchman
  npmPackages:
    @prisma/client: 3.3.0 => 3.3.0 
    blitz: 0.41.1 => 0.41.1 
    prisma: 3.3.0 => 3.3.0 
    react: alpha => 18.0.0-alpha-b81de8633-20211019 
    react-dom: alpha => 18.0.0-alpha-b81de8633-20211019 
    typescript: ~4.3 => 4.3.5 



```

[...]

Issue #2815, opened 2021-10-07
Put generated zod schemas into a separate file


Talor-A:
### What do you want and why?

Per #2754,
The current `generate` script crates queries / mutations with a zod schema in the same file. When users attempt to export this schema to use in e.g. form validation, it will fail as all imports from query / mutation files are replaced. 

Instead, it would be great if they were in a separate file to avoid this unexpected behavior. 

### Possible implementation(s)

For one solution, whenever queries & mutations are generated create a folder such as `validations/` or `schemas/` containing these zod schemas that both the UI and server code can import. 

### Additional context

See #2754

Talor-A:
> Couldn't it be a single file, e.g. `schemas.ts` or `validators.ts` containing all the schemas in a particular queries/mutations directory, for example: `users/queries/validators.ts`? 

I'm not sure, do all files in a ` queries` directory get transformed? Or would this not be an issue?

flybayer:
👆 @beerose's comment as the correct approach here imo.

meehawk:
@beerose is this still relevant? If so I'd love to take it

flybayer:
@meehawk yes!

PR #3869, opened 2022-09-30
feature: Multiple fields forms using templates during generation

siddhsuresh:
Closes: #2038
Closes: #3881
Closes: #2863
Closes: https://github.com/blitz-js/blitz/issues/2815 through https://github.com/blitz-js/blitz/pull/4057

### 🚧Testing
*Please give feedback here if not merged yet or in a discord thread in help or in contributor general if you know to fix the bug*

‼️ **This PR has been published to npm to allow testing**
Continuing PR https://github.com/blitz-js/blitz/pull/2679/

- [x] port from legacy
- [x] read from ```cliConfig``` object in ```app/blitz-server.ts``` to get the user-defined codegen  
- [x]  update unit tests in ```@blitzjs/generator```
- [x] replace hardcoded db resolve in ```packages\generator\src\utils\get-prisma-schema.ts```
- [x] update the logging setup in generator to match with blitz package 
- [x] fix subsequent runs with --parent or belongsTo key word - https://github.com/blitz-js/blitz/pull/2679#issuecomment-1008423700
- [x] make the prisma import by mutations and queries to use the value defined in ```package.json``` 
- [x] https://github.com/blitz-js/legacy-framework/issues/367#issue-1297227082 - Add dropdown option through the new ```LabeledSelectField``` component  to choose parent model
- [x] Add LabeledSelectField to ```formik```
- [x] Add LabeledSelectField to ```react-final-form``` 
- [x] Update/Fix Generator Unit tests
- [x] fix string default value not being wrapped in `"..."` mentioned in #2863  

### TODO
- [x] Declutter the generate.ts and move the relevant code to the respective generators
- [x] Fix the playwright installation problem (again?)
- [x] Add better typing - remove most of the `any` type in the added code

## Feature Checklist

- [x] Documentation added/updated (submit PR to [blitzjs.com repo `main` branch](https://github.com/blitz-js/blitzjs.com))


.changeset/silly-peas-work.md
```diff
@@ -0,0 +1,7 @@
+---
+"toolkit-app": patch
```
siddhsuresh, line 2:
```suggestion

```

siddhsuresh:
@beerose @dillonraphael @flybayer 
I created a new ```LabeledSelectField``` component by directly based it on the existing ```LabeledTextField``` while changing only the input as select than an input.

I did not want to change the existing functionality of the regex, so I kept the API of the component same as the  LabeledTextField, but the functionality of the keywords are a bit different.

> For example
```jsx
<LabeledSelectField name="id" label="User Id" placeholder="User Id" type={users} />
```
```name``` is the field to be shown in the drop-down while the ```type``` holds the array of parent fields.  

Let me know if the API of the new component needs to be changed. I shall then fix those individually

File: apps/toolkit-app/app/blitz-server.ts [hunk 2]
```diff
@@ -19,5 +19,69 @@ const { gSSP, gSP, api } = setupBlitzServer({
...
+        prismaType: "Uuid",
+      },
+      json: {
+        component: "LabeledTextField",
+        inputType: "text",
+        zodType: "any",
+        prismaType: "Json",
+      },
+    },
+  },
```
siddhsuresh, line 85:
```suggestion
```

packages/generator/src/utils/log.ts
```diff
@@ -108,14 +34,14 @@ export const baseLogger = (options?: ISettingsParam): Logger => {
 
   let config
   try {
-    config = loadConfigAtRuntime()
+    config = {} as any // todo: loadConfigAtRuntime()
   } catch {
     config = {}
   }
```
flybayer, line 40: we can remove this now?

packages/generator/templates/app/_forms/formik/LabelSelectField.tsx
```diff
@@ -0,0 +1,68 @@
+import { forwardRef, PropsWithoutRef } from "react";
+import { useFormikContext, ErrorMessage, Field } from "formik";
+
+export interface LabeledSelectFieldProps
+	extends PropsWithoutRef<JSX.IntrinsicElements["select"]> {
+	/** Field name. */
```
flybayer, line 6: this file has tabs instead of spaces

packages/generator/templates/app/src/users/queries/getUsers.ts
```diff
@@ -0,0 +1,31 @@
+import {paginate} from "blitz"
```
flybayer, line 1: file shouldn't be in templates

packages/generator/test/generators/builders/field-values-builder.test.ts
```diff
@@ -0,0 +1,224 @@
+import {FieldValuesBuilder} from "../../../src/generators/template-builders/field-values-builder"
+import {describe, it, expect} from "vitest"
+
+describe("Form Generator", () => {
+  process.env.BLITZ_APP_DIR = process.cwd()
+  const generator = new FieldValuesBuilder()
+
+  it("Should work with simple types", async () => {
```
flybayer, line 8: can you also add a test for two word field names like `orgName` to makes sure they are formatted correctly

packages/generator/templates/form/__ModelName__Form.tsx
```diff
@@ -1,13 +1,22 @@
...
+
 import {z} from "zod"
 export {FORM_ERROR} from "src/core/components/Form"
 
-export function __ModelName__Form<S extends z.ZodType<any, any>>(props: FormProps<S>) {
+function __ModelName__FormSuspense<S extends z.ZodType<any, any>>(props: FormProps<S>) {
   return (
     <Form<S> {...props}>
-      <LabeledTextField name="name" label="Name" placeholder="Name" />
+      {/* template: <__component__ name=__fieldName__ label=__Field_Name__ placeholder=__Field_Name__  type=__inputType__ /> */}
```
flybayer, line 11: I think we should remove the quotes from the generated values and add them in the template if they are needed. Otherwise it's guaranteed we'll run into place where we don't actually want the quotes in the value.

```suggestion
      {/* template: <__component__ name="__fieldName__" label="__Field_Name__" placeholder="__Field_Name__"  type="__inputType__" /> */}
```

packages/generator/templates/form/__ModelName__Form.tsx
```diff
@@ -1,13 +1,22 @@
...
   )
 }
 
+export function ProjectForm<S extends z.ZodType<any, any>>(props: FormProps<S>) {
+  return (
+    <Suspense fallback={<div>Loading...</div>}>
+      <__ModelName__FormSuspense {...props} />
+    </Suspense>
+  )
+}
```
flybayer, line 22: The pattern we've been using is that Suspense is always in the page, not in components.

flybayer (commented):
Awesome!! Great work taking this over and making all this progress.

I did an initial review with some comments. Haven't tested it myself yet.

One question, what happens if prisma is not being used?

packages/generator/package.json
```diff
@@ -44,7 +45,8 @@
     "supports-color": "8.1.1",
     "tslog": "3.3.4",
     "username": "5.1.0",
-    "vinyl": "2.2.1"
+    "vinyl": "2.2.1",
+    "zod": "3.17.3"
```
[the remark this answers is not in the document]
siddhsuresh: @flybayer yes,`zod` is only being used internally to verify the custom schema given by the user. 
Usage in `packages/generator/src/utils/get-codegen.ts` Line 152
```ts
const result = CodegenSchema.safeParse(cliConfig.codegen)
```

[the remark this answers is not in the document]
siddhsuresh: > One question, what happens if prisma is not being used?
Currently, the generator bugs out by creating a `prisma.schema` file with the expected updates. I will be fixing this in some time

packages/generator/src/generator.ts
```diff
@@ -294,12 +309,20 @@ export abstract class Generator<
 
   replaceTemplateValues(input: string, templateValues: any) {
     let result = input
+    // console.log("templateValues", templateValues)
```
Vandivier, line 312: let's remove the commented logs (multiple instances across PR)

packages/generator/src/generator.ts
```diff
@@ -267,7 +272,17 @@ export abstract class Generator<
     if (!this.options.destinationRoot) this.options.destinationRoot = process.cwd()
   }
 
-  abstract getTemplateValues(): Promise<any>
+  // abstract getTemplateValues(): Promise<any>
```
Vandivier, line 275: please remove this commented line or add additional text to clarify why it is here.

packages/generator/src/generator.ts
```diff
@@ -343,46 +381,67 @@ export abstract class Generator<
...
         templatedFile = this.prettier.format(templatedFile, options)
       } catch (error) {
-        console.warn(`Failed trying to run prettier: ` + error)
+        console.error(`Failed trying to run prettier: ` + error)
       }
     }
     return templatedFile
   }
 
+  async preFileWrite(filePath: string): Promise<any> {
```
Vandivier, line 390: can we provide a better type compared to Promise<any>? is it a Promise<AppTemplateValues>?

packages/generator/src/generators/template-builders/field-values-builder.ts
```diff
@@ -0,0 +1,120 @@
...
+      modelName: options.modelName,
+      modelNames: options.modelNames,
+      ModelName: options.ModelName,
+      ModelNames: options.ModelNames,
+      modelNamesPath: this.getModelNamesPath(options.context, options.modelNames),
+    }
+    if (options.extraArgs) {
+      // specialArgs - these are arguments like 'id' or 'belongsTo', which are not meant to
+      // be processed as fields but have their own special logic
+      let specialArgs: {[key in string]: string} = {}
```
Vandivier, line 43:
```suggestion
      const specialArgs: {[key in string]: string} = {}
```

packages/generator/src/utils/log.ts
```diff
@@ -106,16 +32,9 @@ export const newline = () => {
...
 
-  let config
-  try {
-    config = loadConfigAtRuntime()
-  } catch {
-    config = {}
-  }
-
   globalThis._blitz_baseLogger = new Logger({
-    minLevel: config.log?.level || "info",
```
Vandivier, line 117: why remove this?

packages/generator/src/utils/modal-names.ts
```diff
@@ -0,0 +1,14 @@
+import {pluralCamel, pluralPascal, singleCamel, singlePascal} from "../index"
+
+export function modelName(input: string = "") {
```
Vandivier, line 3: plz fix this file name from 'modal-names' to 'model-names' or better yet imo 'model-name-utils'

Vandivier (changes requested):
really great work, some small changes requested.

also do we have unit/ui/component testing on a template using LabeledSelectFieldProps specifically? maybe i missed it, but if we don't have one that would be nice. but i guess it could be a subsequent PR

packages/generator/src/utils/modal-names.ts
siddhsuresh: Yeah, I did not realize the typo, I think I will rename it to `model-names` as it is already present in the utils folder.

packages/generator/src/utils/get-codegen.ts
```diff
@@ -0,0 +1,165 @@
...
+    decimal: {
+      component: "LabeledTextField",
+      inputType: "number",
+      zodType: "number",
+      prismaType: "Decimal",
+    },
+    datetime: {
+      component: "LabeledTextField",
+      inputType: "text",
+      zodType: "string",
```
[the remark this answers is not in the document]
siddhsuresh: thanks!

siddhsuresh:
!preview subtemplate

tordans:
I am looking forward to using this. Three thoughts…

a. ~~Will be part of this PR to generate the fields in the FoomodelForm.tsx ([Code](https://github.com/blitz-js/blitz/pull/3869/files#diff-eae33e603f855f07a8dac2e824b257a6ea9f68429e37f90412ce7d0d4d9a08afR11))? Or, would it be possible to put the field names and types in a code comment there, so its easier to add the form field components?~~ => Already answered, the `template: <Component>` part will be replaced, right? This is great!

b. Could be part of this PR to connect and share the zod validations? The pattern I am using right no (and have to do whenever using the generator …):
   1. Extract and export the zod schema from `createFoomodel` ([Code](https://github.com/blitz-js/blitz/pull/3869/files#diff-3e6ef92aeba5532cb18f185be6d1af0a10652262a81ac5abbe1f52e5378a280eR5-R14)) into `src/foomodel/schema.ts`, name it `FoomodelSchema`
   2. Change the `updateFoomodel` ([Code](https://github.com/blitz-js/blitz/pull/3869/files#diff-83358b15ef5d1adfceef16767a7460f636dbb7204d7706893e916cc79071c1f4R5-R8)) to `FoomodelSchema.merge(z.object({ id: z.number() }))` in order ro reuse it here
   3. Update the `edit.tsx` and `new.tsx` to use the `FoomodelSchema` ([Code](https://github.com/blitz-js/blitz/pull/3869/files#diff-048eed817e219f0cd3e50943ef6e8b0733dfe24d7389b44a9f3733ad2cad6131L45))
  --
  The comment in the `edit.tsx` suggest that this is the way to go and the fact that the server side zod errors are not really usable ATM (see my question in Discord) also makes this an important step.
  This could be a follow up PR, of course…

c. I wonder if it where a good idea to extract the onSubmit inline functions eg. in [edit.tsx](https://github.com/blitz-js/blitz/pull/3869/files#diff-048eed817e219f0cd3e50943ef6e8b0733dfe24d7389b44a9f3733ad2cad6131L47) into a HandleSubmit function. And use the generator to generate the inline types for this function, since my TS complains once the `values` param is not inline anymore.

d. Just a note: A follow up PR could be using a [`<input type="datetime-local" />`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local) for DateTime fields. However, this gets a bit tricky since this input type returns and requires (input) a very specific date time string. I could share our current solution later, if this is interesting.

siddhsuresh:
> I am looking forward to using this. Three thoughts…
> 
> a. ~Will be part of this PR to generate the fields in the FoomodelForm.tsx ([Code](https://github.com/blitz-js/blitz/pull/3869/files#diff-eae33e603f855f07a8dac2e824b257a6ea9f68429e37f90412ce7d0d4d9a08afR11))? Or, would it be possible to put the field names and types in a code comment there, so its easier to add the form field components?~ => Already answered, the `template: <Component>` part will be replaced, right? This is great!
> 
> b. Could be part of this PR to connect and share the zod validations? The pattern I am using right no (and have to do whenever using the generator …):
> 
> 1. Extract and export the zod schema from `createFoomodel` ([Code](https://github.com/blitz-js/blitz/pull/3869/files#diff-3e6ef92aeba5532cb18f185be6d1af0a10652262a81ac5abbe1f52e5378a280eR5-R14)) into `src/foomodel/schema.ts`, name it `FoomodelSchema`
> 2. Change the `updateFoomodel` ([Code](https://github.com/blitz-js/blitz/pull/3869/files#diff-83358b15ef5d1adfceef16767a7460f636dbb7204d7706893e916cc79071c1f4R5-R8)) to `FoomodelSchema.merge(z.object({ id: z.number() }))` in order ro reuse it here
> 3. Update the `edit.tsx` and `new.tsx` to use the `FoomodelSchema` ([Code](https://github.com/blitz-js/blitz/pull/3869/files#diff-048eed817e219f0cd3e50943ef6e8b0733dfe24d7389b44a9f3733ad2cad6131L45))
>    The comment in the `edit.tsx` suggest that this is the way to go and the fact that the server side zod errors are not really usable ATM (see my question in Discord) also makes this an important step.
>    This could be a follow up PR, of course…
> 
> c. I wonder if it where a good idea to extract the onSubmit inline functions eg. in [edit.tsx](https://github.com/blitz-js/blitz/pull/3869/files#diff-048eed817e219f0cd3e50943ef6e8b0733dfe24d7389b44a9f3733ad2cad6131L47) into a HandleSubmit function. And use the generator to generate the inline types for this function, since my TS complains once the `values` param is not inline anymore.
> 
> d. Just a note: A follow up PR could be using a [`<input type="datetime-local" />`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local) for DateTime fields. However, this gets a bit tricky since this input type returns and requires (input) a very specific date time string. I could share our current solution later, if this is interesting.

Hey @tordans, thanks a lot for the insights! Please let me know if you face any issues while testing this.
- a) Yeah it's possible by changing the custom template that you can define like you mentioned
- b) No, currently the way the PR is written, the generator needs the schema to be present in the line. But yes this functionality can be changed in a subsequent PR with similar goals as what you have mentioned and the open issue https://github.com/blitz-js/blitz/issues/2815. 
- c) Hey wouldn't the schema parameter in the form take care of the types to be used which are being updated by the generator? Please correct me if I'm wrong.
- d) Yes please, a PR would be welcome if it will improve the DX.

siddhsuresh:
> b. Could be part of this PR to connect and share the zod validations? The pattern I am using right no (and have to do whenever using the generator …):
> 
> 1. Extract and export the zod schema from `createFoomodel` ([Code](https://github.com/blitz-js/blitz/pull/3869/files#diff-3e6ef92aeba5532cb18f185be6d1af0a10652262a81ac5abbe1f52e5378a280eR5-R14)) into `src/foomodel/schema.ts`, name it `FoomodelSchema`
> 2. Change the `updateFoomodel` ([Code](https://github.com/blitz-js/blitz/pull/3869/files#diff-83358b15ef5d1adfceef16767a7460f636dbb7204d7706893e916cc79071c1f4R5-R8)) to `FoomodelSchema.merge(z.object({ id: z.number() }))` in order ro reuse it here
> 3. Update the `edit.tsx` and `new.tsx` to use the `FoomodelSchema` ([Code](https://github.com/blitz-js/blitz/pull/3869/files#diff-048eed817e219f0cd3e50943ef6e8b0733dfe24d7389b44a9f3733ad2cad6131L45))

@tordans Just an update, I am going to be working on this a new PR, maybe a few design changes which can be debated in that PR.

tordans:
> c) Hey wouldn't the schema parameter in the form take care of the types to be used which are being updated by the generator? Please correct me if I'm wrong.

Once the function is extracted, this automatism is lost, so it needs an explicit type to type the attributes. Did not look into it more, yet.

> d) Yes please, a PR would be welcome if it will improve the DX.

All right, will do on top of this one, once merged.

siddhsuresh:
> Once the function is extracted, this automatism is lost, so it needs an explicit type to type the attributes. Did not look into it more, yet.

From https://github.com/blitz-js/blitz/pull/4057, the automatic generation of zod types will occur in `validations/schema.ts` files [room for change in name]. Would that satisfy this requirement?

siddhsuresh:
!preview multiple-fields

packages/generator/package.json
```diff
@@ -68,7 +69,8 @@
     "react": "18.2.0",
     "typescript": "^4.8.4",
     "unbuild": "0.6.9",
-    "watch": "1.0.2"
+    "watch": "1.0.2",
+    "zod": "3.20.2"
```
flybayer, line 73: having zod as a devDep causes global use of blitz to fail because zod isn't install there.

We need to move this to a regular dep but with version range that supports all of v3, so we don't conflict with user's zod version.


```
/tmp> blitz new blitz-fields
Error: Cannot find module 'zod'
Require stack:
- /Users/b/Library/pnpm/global/5/.pnpm/@blitzjs+generator@0.0.0-multiple-fields-20230126103232/node_modules/@blitzjs/generator/dist/index.cjs
- /Users/b/Library/pnpm/global/5/.pnpm/blitz@0.0.0-multiple-fields-20230126103232/node_modules/blitz/dist/chunks/new.cjs
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:933:15)
    at Function.Module._load (node:internal/modules/cjs/loader:778:27)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at Object.<anonymous> (/Users/b/Library/pnpm/global/5/.pnpm/@blitzjs+generator@0.0.0-multiple-fields-20230126103232/node_modules/@blitzjs/generator/dist/index.cjs:15:11)
    at Module._compile (node:internal/modules/cjs/loader:1103:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1157:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:1005:19) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/Users/b/Library/pnpm/global/5/.pnpm/@blitzjs+generator@0.0.0-multiple-fields-20230126103232/node_modules/@blitzjs/generator/dist/index.cjs',
    '/Users/b/Library/pnpm/global/5/.pnpm/blitz@0.0.0-multiple-fields-20230126103232/node_modules/blitz/dist/chunks/new.cjs'
  ]
}
```

flybayer: let me know when you publish a new version to test

siddhsuresh:
!preview multiple-fields

packages/generator/package.json
siddhsuresh: > let me know when you publish a new version to test

@flybayer I fixed the issue and published

packages/generator/templates/validations/schemas.ts
```diff
@@ -0,0 +1,20 @@
+import {z} from "zod"
+
+if (process.env.parentModel) {
+    export const __ModelName__Schema = z.object({
+      __parentModelId__: z.__parentModelIdZodType__(),
+      // template: __fieldName__: z.__zodType__(),
+    })
+} else {
+    export const __ModelName__Schema = z.object({
```
flybayer, line 9: plus update where this is used
```suggestion
    export const Create__ModelName__Schema = z.object({
      __parentModelId__: z.__parentModelIdZodType__(),
      // template: __fieldName__: z.__zodType__(),
    })
} else {
    export const Create__ModelName__Schema = z.object({
```

flybayer (changes requested):
sweet!! Overall great progress and mostly there, just a few issues

- failing to build brand new app with pnpm
```
> blitz build
Loaded env from /private/tmp/bf/.env.local
Loaded env from /private/tmp/bf/.env
✔ Routes manifest was successfully generated
info  - Linting and checking validity of types ..Failed to compile.

./vitest.config.ts:1:31
Type error: Cannot find module '@next/env' or its corresponding type declarations.

> 1 | import { loadEnvConfig } from "@next/env"
    |                               ^
  2 | import { defineConfig } from "vitest/config"
  3 |
  4 | import react from "@vitejs/plugin-react"
```
- rename `auth/validations.ts` to `auth/schemas.ts` for consistency
- "update" schema needs the `// template` line added to it like "create" schema
- model edit page should use the "update" schema
- generating optional field like `age:number?` does not add `.optional()` to zod schema
- should be able to add new fields by running generate without getting any conflict warning
image
- how to import the `CodegenConfig` type in `blitz-server.ts`?
- generating default is not working at all, it should add default to zod schema and prisma schema `completed:boolean:default=false`
- exporting custom codegen config from blitz-server is not working
- custom codegen config should merge default fieldTypeMap, so user doesn't have to specify everything
- generating a boolean field does not use values from `fieldTypeMap`

go ahead and work on docs

flybayer (approved):

Issue #3860, opened 2022-09-23
Don't rely on hardcoded `db/prisma.schema`  — use whatever is configured in package.json


sharmapukar217:
I want the db folder to be prisma but there's no config available for this.

beerose:
You can configure this in your `package.json` file:

```
"prisma": {
    "schema": "db/schema.prisma"
  },
```

Here you can find relevant docs: https://www.prisma.io/docs/concepts/components/prisma-schema#prisma-schema-file-location

sharmapukar217:
I'm not talking about configuring for prisma, I want blitz js to use `prisma` folder rather than `db`

beerose:
Blitz doesn't really use Prisma internally. Can you describe more about what's causing issues?

sharmapukar217:
when I'm running blitz dev or blitz codegen, or any blitz command running or using prisma, i get "db/schema.prisma" not found 🥲  but its there at `prisma/schema.prisma`

beerose:
Ah okay. It should be fixed with changes introduced in this PR: https://github.com/blitz-js/blitz/pull/3855.

However, I'm going to reopen this issue to make sure we don't rely on hardcoded db directory anywhere else.

beerose:
Places to update:
- [ ] Blitz generator package
- [ ] Codegen tasks

Zeko369:
Codegen tasks - already fixed by @dillonraphael in the last PR
Blitz generator package - in #3874

PR #3874, opened 2022-10-01
Try to load prisma.schema path from pkg json

Zeko369:
Closes: #3860

It first tries to load package.json prisma.schema value (i.e. if you have a `prisma/schema.prisma` folder instead of `db/schema.prisma`) instead of just using `db/schema.prisma`

## Bug Checklist

## Feature Checklist


packages/generator/src/generators/model-generator.ts
```diff
@@ -41,7 +41,10 @@ export class ModelGenerator extends Generator<ModelGeneratorOptions> {
 
   // eslint-disable-next-line require-await
   async write() {
-    const schemaPath = path.resolve("db/schema.prisma")
+    const pkgJson = this.fs.readJSON("package.json", {}) as {[key: string]: any}
```
beerose, line 44: Let's use `cwd`:
```
const packageJsonPath = path.join(process.cwd(), "package.json")
const packageJson = fs.readJSONSync(packageJsonPath)
```

packages/generator/src/generators/model-generator.ts
```diff
@@ -41,7 +41,10 @@ export class ModelGenerator extends Generator<ModelGeneratorOptions> {
...
+    const rawSchemaPath = pkgJson?.["prisma"]?.["schema"] || "db/schema.prisma"
```
beerose, line 45: To simplify:
```suggestion
    const rawSchemaPath = pkgJson?.prisma?.schema || "db/schema.prisma"
```

packages/generator/src/generators/model-generator.ts
Zeko369: The reason for using `this.fs.readJSON` was it loads the root path for us
(so i.e. running this inside of `pages` or `app` will work with `this.fs.readJSON`, but won't with `process.cwd()`)

beerose: I see. it's fine then

beerose:
Could you add a changeset too?

dillonraphael (approved):

Issue #3275, opened 2022-04-09
js project initialization creates invalid glob in lint-staged options


isaacs:
### What is the problem?

invalid glob is `*.{js}`.  Braced groups are only valid if they have more than one member.

### Paste all your error logs here:

```
PASTE_HERE (leave the ``` marks)
```

### Paste all relevant code snippets here:

```
PASTE_HERE (leave the ``` marks)
```

### What are detailed steps to reproduce this?

1. create new js app
2. commit anything to git
3. ``⚠ Detected incorrect braces with only single value: `*.{js}`. Reformatted as: `*.js` ``

### Run `blitz -v` and paste the output here:

```
Loaded env from /Users/isaacs/dev/tierdev/tierweb-blitz/.env.local
Loaded env from /Users/isaacs/dev/tierdev/tierweb-blitz/.env
macOS Monterey | darwin-arm64 | Node: v17.6.0

blitz: 0.45.3 (local)

  Package manager: npm 
  System:
    OS: macOS 12.3.1
    CPU: (10) arm64 Apple M1 Max
    Memory: 3.08 GB / 64.00 GB
    Shell: 5.2.0 - /usr/local/bin/bash
  Binaries:
    Node: 17.6.0 - /usr/local/bin/node
    Yarn: 1.22.18 - /usr/local/bin/yarn
    npm: 8.5.3 - /usr/local/bin/npm
    Watchman: Not Found
  npmPackages:
    @prisma/client: 3.12.0 => 3.12.0 
    blitz: 0.45.3 => 0.45.3 
    prisma: 3.12.0 => 3.12.0 
    react: 18.0.0-beta-149b420f6-20211119 => 18.0.0-beta-149b420f6-20211119 
    react-dom: 18.0.0-alpha-5ca4b0433-20211020 => 18.0.0-alpha-5ca4b0433-20211020 
    typescript: Not Found
```

### Please include below any other applicable logs and screenshots that show your problem:

Invalid glob appears in these files:

```
node_modules/@blitzjs/generator/dist/templates/app/package.js.json:    "*.{js}": ["eslint --fix"]
node_modules/@blitzjs/generator/dist/templates/minimalapp/package.js.json:    "*.{js}": ["eslint --fix"]
```

PR #3877, opened 2022-10-03
refactor(js-templates): remove {} from single values inside package.json

JoaoJesus94:
Closes: #3275 

Remove braces from the `lint-staged` object in `package.js.json` (`full` and `minimal` template)


beerose (approved):
Thank you!

Issue #3706, opened 2022-08-10
missing `loadConfigAtRuntime` for `baseLogger` to change logger config


oltdaniel:
### What is the problem?

Currently there is no way to configure the logger level, as the config is currently a placeholder and marked as ToDo. See https://github.com/blitz-js/blitz/blob/85b15a9cfba1280a0dc146af23b12488d26584d4/packages/blitz/src/logging.ts#L37

> If there is already an issue or card, this can be closed or linked.

### Paste all your error logs here:

```
2022-08-10 09:23:01.256 INFO [...]
```

### Paste all relevant code snippets here:

```js
module.exports = withBlitz({
  blitz: {
    log: { // not accepted via the exposed type
      level: "warn",
    }
  } 
})
```

### What are detailed steps to reproduce this?

1. Try to change the log level of the internal logger of your blitz application without changing code within `node_modules`.

### Run `blitz -v` and paste the output here:

```
Blitz version: 2.0.0-alpha.67 (global)
Blitz version: 2.0.0-alpha.68 (local)
Linux 5.18 | linux-x64 | Node: v16.16.0


 Package manager: npm

  System:
    OS: Linux 5.18 Pop!_OS 22.04 LTS
    CPU: (24) x64 AMD Ryzen 9 5900X 12-Core Processor
    Memory: 50.22 GB / 62.73 GB
    Shell: 5.8.1 - /usr/bin/zsh
  Binaries:
    Node: 16.16.0 - ~/.nvm/versions/node/v16.16.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v16.16.0/bin/yarn
    npm: 8.11.0 - ~/.nvm/versions/node/v16.16.0/bin/npm
  npmPackages:
    @blitzjs/auth: 2.0.0-alpha.68 => 2.0.0-alpha.68 
    @blitzjs/next: 2.0.0-alpha.68 => 2.0.0-alpha.68 
    @blitzjs/rpc: 2.0.0-alpha.68 => 2.0.0-alpha.68 
    @prisma/client: latest => 4.1.1 
    blitz: 2.0.0-alpha.68 => 2.0.0-alpha.68 
    next: 12.2.0 => 12.2.0 
    prisma: 4.0.0 => 4.0.0 
    react: 18.0.0 => 18.0.0 
    react-dom: 18.0.0 => 18.0.0 
    typescript: ^4.5.3 => 4.7.4 
```

dillonraphael:
[...]

PR #3886, opened 2022-10-06
Add BlitzLogger plugin and allow customizing logging

beerose:
Closes: https://github.com/blitz-js/blitz/issues/3706, https://github.com/blitz-js/blitz/discussions/3853

### Problems

We have two logging setups — one in `blitz` package and one in `@blitzjs/generator`. I previously described the problem here: https://github.com/blitz-js/blitz/discussions/3682.

Because of the relations between those two packages, it's impossible to have one logging setup without solving the dependencies between packages as described in https://github.com/blitz-js/blitz/discussions/3682. Therefore, I'm not modifying the generator package's logger in the PR. It's only used for the generator commands, so it's fine as it is anyway. 

Users can provide logging settings in the `blitz-server.ts` file:

```ts 
import {setupBlitzServer} from "@blitzjs/next"
import {BlitzLogger} from "blitz"

const {gSSP, gSP, api} = setupBlitzServer({
  log: {
    logger: BlitzLogger({
      colorizePrettyLogs: true,
      prefix: ["[blitz]"],
    }),
    logLevel: "error",
  },
})
```

By default, `BlitzLogger` will use the same settings as we did so far. It accepts all properties that are accepted by tslog's Logger.

The initial setup in new apps will look like this:

```ts
import { setupBlitzServer } from "@blitzjs/next"
import { AuthServerPlugin, PrismaStorage } from "@blitzjs/auth"
import { simpleRolesIsAuthorized } from "@blitzjs/auth"
import { BlitzLogger } from "blitz"
import db from "db"
import { authConfig } from "./blitz-client"

export const { gSSP, gSP, api } = setupBlitzServer({
  plugins: [
    AuthServerPlugin({
      ...authConfig,
      storage: PrismaStorage(db),
    }),
  ],
  log: {
    logger: BlitzLogger({}),
  },
})
```

We're adding the `log` property along with the `BlitzLogger` for a few reasons:
- so that there's no magic behind logging,
- we want to be explicit about it,
- users can very easily customize logging.

## Feature Checklist


packages/blitz-next/src/index-server.ts
```diff
@@ -50,6 +53,7 @@ export type NextApiHandler<TResult> = (
 type SetupBlitzOptions = {
   plugins: BlitzServerPlugin<RequestMiddleware, Ctx>[]
   onError?: (err: Error) => void
+  log: {logger: ReturnType<typeof BlitzLogger>; logLevel?: BlitzLogLevel}
```
flybayer, line 56: can we simplify this API? Because logLevel will only work with our default logger, so doesn't make sense to have it here I don't think. They can just set `minLevel` in the default logger options

```suggestion
  logger: ReturnType<typeof BlitzLogger>;
```

flybayer (commented):
suggested API change below.

Also, can we automatically add logger to `ctx`, so `ctx.logger` can be used in queries/mutations, etc?

flybayer (approved):
Great! Just need to make sure we update docs if haven't already.

Issue #3806, opened 2022-08-31
Optional chaining from @blitzjs/auth crashes app on old devices


BadCoder1337:
### What is the problem?

After upgrade to Blitz 2.0 I found regression on iOS 12. Regardless of configuration (previously working babel.config.js, next-transpile-modules, etc.) I can't fix it.

### Paste all your error logs here:

```
SyntaxError: Unexpected token '.'
```

### Paste all relevant code snippets here:

https://github.com/blitz-js/blitz/blob/e3523407b9abadfc21f01bc57f8421cd351dc510/packages/blitz-auth/src/client/index.tsx#L72-L84

### What are detailed steps to reproduce this?

1. Init new app with auth
2. Run it
3. Open with Safari 12 (I'm using Browserstack)

### Run `blitz -v` and paste the output here:

```
Blitz version: 2.0.0-beta.3 (global)
Blitz version: 2.0.0-beta.3 (local)
Windows 11 | win32-x64 | Node: v16.15.1


 Package manager: npm

  System:
    OS: Windows 10 10.0.22622
    CPU: (12) x64 AMD Ryzen 5 2600 Six-Core Processor
    Memory: 5.03 GB / 15.93 GB
  Binaries:
    Node: 16.15.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.18 - C:\Program Files\nodejs\yarn.CMD
    npm: 8.11.0 - C:\Program Files\nodejs\npm.CMD
  npmPackages:
    @blitzjs/auth: 2.0.0-beta.3 => 2.0.0-beta.3
    @blitzjs/next: 2.0.0-beta.3 => 2.0.0-beta.3
    @blitzjs/rpc: 2.0.0-beta.3 => 2.0.0-beta.3
    @prisma/client: 4.0.0 => 4.0.0
    blitz: 2.0.0-beta.3 => 2.0.0-beta.3
    next: 12.2.5 => 12.2.5
    prisma: 4.0.0 => 4.0.0
    react: 18.2.0 => 18.2.0
    react-dom: 18.2.0 => 18.2.0
    typescript: ^4.5.3 => 4.8.2

```

### Please include below any other applicable logs and screenshots that show your problem:

![image](https://user-images.githubusercontent.com/16510197/187774101-17d5093b-5515-4960-bf97-14dc88f81d81.png)

dillonraphael:
Thanks for sharing. We'll take a look at this

estevanjantsk:
[...]

PR #3887, opened 2022-10-06
3806 - Optional chaining from @blitzjs/auth crashes app on old devices

estevanjantsk:
Closes: #3806 

This one was a quite interesting issue to solve. Even though it's only one line it gave a lot of work of "investigation".

So basically what happens (I had to check unbuild code to check what they were doing there):

When we build the lib using unbuild, by default they use es2020 (https://github.com/unjs/unbuild/blob/main/src/build.ts#L67)
to build the lib. es2020 does not convert the optional chaining syntax and this was triggering the error. When I changed to es2017 it works because it compiles to an older version, and that's ok. As you can see from the image (I used browserstack) 
Screenshot 2022-10-07 at 01 28 29
the login and signup button appeared. **_So it works!!!_** 💯 

If you want to go deeper xD as you can see from this code:
https://github.com/unjs/unbuild/blob/main/src/builder/rollup.ts#L220 is where they use the version
and the function comes from https://github.com/unjs/unbuild/blob/main/src/builder/rollup.ts#L9 - rollup-plugin-esbuild and this guy does all the hard work.

GG!

OBS: But one thing that I didn't understand quite well is that unbuild is skipping all our ts config...and only taking in consideration the build.config file. So one thing that we could start to think about is: maybe we could build the lib using tsc only? One good thing is that we would not depend on this unbuild lib (tbh I don't know if they will maintain this lib).

OBS2: I really like blitz-js. I really like the project and will try to help as much as I can =) on my spare time. Thanks peeps. All of you rock \,,/. I have been learning a lot from this code.

### DIFF
--- a/packages/blitz-auth/build.config.ts
+++ b/packages/blitz-auth/build.config.ts
@@ -6,6 +6,7 @@ const config: BuildConfig = {
   declaration: true,
   rollup: {
     emitCJS: true,
+    esbuild: {target: "es2017"},
   },
 }
 export default config


dillonraphael:
Hey @estevanjantsk so sorry for the late reply. I'm sure this took a while to investigate and really appreciate that. Our goal is to get a "stable" release for next week and working hard to get through as many fundamental issues as possible. Your contribution doesn't go unnoticed. Give us a little more time and we'll take look. Thanks for the great work.

PS. Can you just run `pnpm install` so the lockfile is up-to-date?

dillonraphael:
Hey @estevanjantsk, was going to get this merged but need the pnpmlock fixed. Are you able to run `pnpm install` and repush the update pnpmlock?

flybayer:
fixed a long time ago in another PR

Issue #3592, opened 2022-07-22
Custom Server TS errors and not using local packages when running blitz dev (using alpha)


lhocke:
### What is the problem?

Running into two issues with blitz alpha
1. attempting to run ```blitz dev``` with a custom server file (either ```server.ts``` or ```server/index.ts```) tries to use global installs of ```@blitzjs/next, @blitzjs/auth, and @blitzjs/rpc```
2. Error 1 can be avoided by invoking the start scripts with either ```yarn dev``` or ```npm dev``` but then throws several typsecript errors

### Paste all your error logs here:

```
$ blitz dev
Loaded env from /Users/lhocke/Code/personal/webdev/comic-blitz/.env.local
Loaded env from /Users/lhocke/Code/personal/webdev/comic-blitz/.env
Using your custom server
✘ [ERROR] Transforming const to the configured target environment ("es5") is not supported yet

    server.ts:6:0:
      6 │ const { PORT = "3000"} = process.env
        ╵ ~~~~~

  The target environment was set to "es5" here:

    tsconfig.json:3:14:
      3 │     "target": "es5",
        ╵               ~~~~~

✘ [ERROR] Transforming destructuring to the configured target environment ("es5") is not supported yet

    server.ts:6:6:
      6 │ const { PORT = "3000"} = process.env
        ╵       ^

  The target environment was set to "es5" here:

    tsconfig.json:3:14:
      3 │     "target": "es5",
        ╵               ~~~~~

✘ [ERROR] Transforming const to the configured target environment ("es5") is not supported yet

    server.ts:7:0:
      7 │ const dev = process.env.NODE_ENV !== "production"
        ╵ ~~~~~

  The target environment was set to "es5" here:

    tsconfig.json:3:14:
      3 │     "target": "es5",
        ╵               ~~~~~

✘ [ERROR] Transforming const to the configured target environment ("es5") is not supported yet

    server.ts:8:0:
      8 │ const app = blitz({ dev })
        ╵ ~~~~~

  The target environment was set to "es5" here:

    tsconfig.json:3:14:
      3 │     "target": "es5",
        ╵               ~~~~~

✘ [ERROR] Transforming const to the configured target environment ("es5") is not supported yet

    server.ts:9:0:
      9 │ const handle = app.getRequestHandler()
        ╵ ~~~~~

  The target environment was set to "es5" here:

    tsconfig.json:3:14:
      3 │     "target": "es5",
        ╵               ~~~~~

[...]

PR #3888, opened 2022-10-07
Custom Server TS error 

siddhsuresh:
Closes: #3592

- The error mentioned in the issue is due to esbuild currently not able to transpile lower than ```es6```.  
- So added the line to the ```esbuild``` options given to build the custom server. 
- This config will override any target config that is configured in ```tsconfig.json``` which ```esbuild``` uses as a fallback

- [x] added ```target:"es6"``` to the esbuild options for custom server

#### Tested using example given in https://nextjs.org/docs/advanced-features/custom-server in typescript import syntax
```ts
// server.ts
import { createServer } from 'http'
import { parse } from 'url'
import next from 'next'

const dev = process.env.NODE_ENV !== 'production'
const hostname = 'localhost'
const port = 3000
// when using middleware `hostname` and `port` must be provided below
const app = next({ dev, hostname, port })
const handle = app.getRequestHandler()

app.prepare().then(() => {
  createServer(async (req, res) => {
    try {
      // Be sure to pass `true` as the second argument to `url.parse`.
      // This tells it to parse the query portion of the URL.
      const parsedUrl = parse(req.url, true)
      const { pathname, query } = parsedUrl

      if (pathname === '/a') {
        await app.render(req, res, '/a', query)
      } else if (pathname === '/b') {
        await app.render(req, res, '/b', query)
      } else {
        await handle(req, res, parsedUrl)
      }
    } catch (err) {
      console.error('Error occurred handling', req.url, err)
      res.statusCode = 500
      res.end('internal server error')
    }
  }).listen(port, (err) => {
    if (err) throw err
    console.log(`> Ready on http://${hostname}:${port}`)
  })
})
```

#### Output when running ```blitz dev```
image


.changeset/moody-spoons-rhyme.md
```diff
@@ -0,0 +1,5 @@
+---
+"blitz": patch
+---
+
+Fix Custom Server TS error - add `es6` taget config to esbuild
```
siddhsuresh, line 5:
```suggestion
Fix Custom Server TS error - add `es6` target config to esbuild 
```

beerose (approved):

PR #3890, opened 2022-10-08
Migrate Routes Command

siddhsuresh:

- [x] migrated ```routes``` cli command from legacy blitz to blitz toolkit
- [x] fix ```getDbFolder``` function in ```packages/blitz/src/cli/utils/next-console.ts``` command  to only check for ```db/schema.prisma``` if schema path defined by ```package.json``` does not exist.
The previous version would return ```db``` regardless if the user stores the schema.prisma file was present or if the user intended another to use the ```db``` folder for any other purpose 
- [x] matched ```blitz routes``` output to the docs https://blitzjs.com/docs/cli-routes

#### Output
image

### DIFF
--- a/.changeset/few-shrimps-leave.md
+++ b/.changeset/few-shrimps-leave.md
@@ -1,0 +1,5 @@
+---
+"blitz": patch
+---
+
+Add `blitz routes` CLI command back to toolkit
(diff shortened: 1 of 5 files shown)


beerose (commented):
You are simply amazing!

Btw. could you also add a changeset?

beerose (approved):

Issue #2162, opened 2021-03-28
[RFC] add GET support to rpc specification


sakulstra:
### What do you want and why?
I want to be able to query(and perhaps even mutate) RPC via `GET` instead `POST` as it unlocks cdn side caching.
This is quite useful for public apis which rarely change, or are processing intensive(swr, rate-limit¹).
Considering cdn a rate-limiting solution(which it isn't), one might even argue that `GET` might be useful for mutations as well.

### Possible implementation(s)
**Always on/Opt in/ Opt out/ per query configuration**
I guess there are multiple ways one could approach this:
1. One could expose queries as `GET/POST` by default and let the consumer choose. `useQuery` would default to  the `POST` way and only the "advanced user", who knows what it is doing set's some property on a per `useQuery` basis to make it use `GET` instead. I'm not very familiar with the blitz source yet - on a first look it seems like we might be able to just add a handler for get requests [alongside the post handler](https://github.com/blitz-js/blitz/blob/canary/packages/core/src/server/rpc-server.ts#L22)?
2. We could enable/disable the `GET` way per default and have a config flag to enable it?
3. One could imagine another of these named export via convention next uses a lot like `export const METHOD = 'POST/GET';` on a query level?
4. ?

### Additional context
Splitting this off of https://github.com/blitz-js/blitz/discussions/1777#discussioncomment-486342 as this discussion is about  different things (the general specification) and I'd like to focus this ticket only on the `GET` support.

**Summing up the points over there**
- Initial idea was to have `POST` for mutations and `GET` for queries[*](https://github.com/blitz-js/blitz/discussions/1777#discussioncomment-303438), but sadly it's not that easy as authenticated queries should probably not be cached on cdn as they might include sensitive info[*](https://github.com/blitz-js/blitz/discussions/1777#discussioncomment-303446).
- GET has more strict size limitations than POST, which has some variance throughout browsers, but is +8k chars on evergreens.[*](https://github.com/blitz-js/blitz/discussions/1777#discussioncomment-303440)

¹I've seen apps use CDN caching headers as soft rate limit solution, but I guess it doesn't properly work like that as it's usually caching per CDN region, not per lambda.

flybayer:
Thanks @sakulstra! I think we should make it opt-in. It's quite nice and simple when all requests use POST and the network requests all look uniform. As well because of the query string limit. Also it helps prevent user data being accidentally cached.

But I'm open to hear other's thoughts on this.

[...]

PR #3891, opened 2022-10-09
Add GET support to RPC specification

siddhsuresh:
Closes: #2162

- [x] Allow opt-in configuration to use GET requests than POST for ```query``` files [ ```mutations``` even with this config will only use POST requests]

#### Example Usage
```ts
export const config: ResolverConfig = {
  httpMethod: "GET",
}
```
- [x] Show warning to the user if URL length exceeds 2083 characters which is the minimum of the maximum URL limit set by browsers (IE and Edge 16) 
  - https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers/417184#417184 
  - https://wiert.me/2010/04/20/maximum-url-lengths/

But, there is no practical limit for URL size for the latest browsers. So, should the warning be removed?
- [x] Update the ```rpc``` tests to accept GET requests 
- [x] Remove updates to toolkit app - currently present to test

#### TODO:
 - [x] Replace `esbuild` usage with `jscodeshift` to get the config from the files

- [x] try shifting to `swc` from `recast` to improve performance 
- [x]  Remove caching layer as it is not needed anymore

#### Output
(Where GET request has been set for ```getProject``` and the default POST request set for ```getProjects```

image

image


apps/toolkit-app/db/migrations/20220427214932_/migration.sql
```diff
@@ -44,4 +44,4 @@ CREATE UNIQUE INDEX "User_email_key" ON "User"("email");
 CREATE UNIQUE INDEX "Session_handle_key" ON "Session"("handle");
 
 -- CreateIndex
-CREATE UNIQUE INDEX "Token_hashedToken_type_key" ON "Token"("hashedToken", "type");
+CREATE UNIQUE INDEX "Token_hashedToken_type_key" ON "Token"("hashedToken", "type");
```
siddhsuresh, line 47:
```suggestion
CREATE UNIQUE INDEX "Token_hashedToken_type_key" ON "Token"("hashedToken", "type");

```

packages/blitz-rpc/src/loader-client.ts
```diff
@@ -39,12 +40,31 @@ export async function transformBlitzRpcResolverClient(
...
+    try {
+      const {register} = require("esbuild-register/dist/node")
+      const {unregister} = register({
+        target: "es6",
+      })
+      const _rpcConfig = require(id).config as ResolverConfig
+      if (_rpcConfig) {
+        resolverConfig.httpMethod = _rpcConfig.httpMethod
+        if (_rpcConfig.httpMethod === "GET") {
+        }
```
siddhsuresh, line 61:
```suggestion

```

packages/blitz-rpc/src/loader-utils.ts
```diff
@@ -1,6 +1,54 @@
...
+}
+
+type _ResolverType = "GET" | "POST"
+
+export function getHttpMethodFromResolverConfig(path: string): _ResolverType {
+  const collection = getCollectionFromSource(path)
+  const config = collection.find(j.ExportNamedDeclaration).filter((p) => {
+    return (
+      p.value.declaration?.type === "VariableDeclaration" &&
+      //@ts-ignore
```
siddhsuresh, line 33: Remove `ts-ignore`

siddhsuresh:
@dillonraphael @beerose what is the recommended way to get the config file from the resolvers?
The ways I have tried till now:
- Use `esbuild` and `require` the resolver file directly and then get the named export `config`. The exists till the [4d70c40](https://github.com/blitz-js/blitz/pull/3891/commits/4d70c40b7110ecae97ab8e84fdca52ee97db7839) commit.
- Use `jscodeshift` and `recast` to use AST to find the named export config by reading the file.
I switched to the second method as it was able to handle re-runs faster when changing the `config.httpMethod` value in the resolvers on `pnpm dev`, which I think is due to the for loop in https://github.com/blitz-js/blitz/blob/de9280473c9fc5c911200372f792b6f3750ce428/packages/blitz-rpc/src/loader-server.ts#L62

and am I missing an easier method to get this config?

packages/blitz-rpc/src/parse-rpc-config.ts
```diff
@@ -0,0 +1,69 @@
...
+}
+
+function parseResolverCacheMiss(content: string): ResolverConfig {
+  // console.time("getResolverConfig ")
+  const resolverConfig = defaultResolverConfig
+  const resolver = parseSync(content, {
+    syntax: "typescript",
+    target: "es2020",
+  })
+  const exportDelaration = resolver.body.find((node) => node.type === "ExportDeclaration")
```
siddhsuresh, line 38: also check if the export is named as config

packages/blitz-rpc/src/data-client/rpc.ts
```diff
@@ -93,18 +95,34 @@ export function __internal_buildRpcClient({
...
+        "?" +
+        new URLSearchParams({
+          params: stringify(serialized.json),
+          meta: stringify(serialized.meta),
+        })
+      if (fullRoutePath.length > 2083) {
+        console.warn(
+          "URL length is greater than 2083 characters. This may cause issues in some browsers. Consider using POST instead of GET.",
+        )
+      }
```
flybayer, line 116: if this is no longer an issue with latest browsers, we can remove it

packages/blitz-rpc/src/loader-client.ts
```diff
@@ -39,12 +41,24 @@ export async function transformBlitzRpcResolverClient(
...
   assertPosixPath(resolverFilePath)
   const routePath = convertPageFilePathToRoutePath(resolverFilePath, options?.resolverPath)
   const resolverName = convertFilePathToResolverName(resolverFilePath)
   const resolverType = convertFilePathToResolverType(resolverFilePath)
+  const resolverConfig: ResolverConfig = {
+    httpMethod: "POST",
+  }
+  if (resolverType === "query") {
+    try {
+      const {httpMethod} = getResolverConfig(id)
```
flybayer, line 54: you already have `src` as the first argument to this `transformBlitzRpcResolverClient` function, so you can directly call `parseResolverCacheMiss(src)` instead of reading the file again from disk.

packages/blitz-rpc/src/index-server.ts
```diff
@@ -37,16 +37,31 @@ const g = getGlobalObject<{blitzRpcResolverFilesLoaded: ResolverFiles | null}>(
...
 export function __internal_addBlitzRpcResolver(
   routePath: string,
+  resolverConfig: ResolverConfig,
   resolver: () => Promise<{default?: Resolver}>,
 ) {
   g.blitzRpcResolverFilesLoaded = g.blitzRpcResolverFilesLoaded || {}
   g.blitzRpcResolverFilesLoaded[routePath] = resolver
+  gConfig.blitzRpcResolverConfig = gConfig.blitzRpcResolverConfig || {}
+  gConfig.blitzRpcResolverConfig[routePath] = resolverConfig
   return resolver
```
flybayer, line 65: You don't need to add a separate resolverConfig object & loading here, because you can get the config directly from `resolver.config` 

```suggestion
export function __internal_addBlitzRpcResolver(
  routePath: string,
  resolver: () => Promise<{default?: Resolver, config: ResolverConfig}>,
) {
  g.blitzRpcResolverFilesLoaded = g.blitzRpcResolverFilesLoaded || {}
  g.blitzRpcResolverFilesLoaded[routePath] = resolver
  return resolver
```

flybayer (commented):
Overall great work! Few comments :)

packages/blitz-rpc/src/index-server.ts
siddhsuresh: thanks, @flybayer. I updated with this logic. I was pretty confused about what the resolver function was returning, which is why I used the same logic I had used for the client for the server also...

packages/blitz-rpc/src/index-server.ts
```diff
@@ -174,14 +178,30 @@ export function rpcHandler(config: RpcConfig) {
       throw new Error("No default export for resolver path: " + routePath)
     }
 
+    let resolverConfig = (await loadableResolver()).config
+    if (!resolverConfig) {
+      resolverConfig = defaultConfig
+    }
```
flybayer, line 184: this doesn't take into account if user exports an empty object as the config. so we should ... spread in the defaults to an object, and let user provided config override the defaults

packages/blitz-rpc/src/index-server.ts
```diff
@@ -174,14 +178,30 @@ export function rpcHandler(config: RpcConfig) {
...
       // We used to initiate database connection here
       res.status(200).end()
       return
-    } else if (req.method === "POST") {
-      // Handle RPC call
-
-      if (typeof req.body.params === "undefined") {
+    } else if (
+      req.method === "POST" ||
+      (req.method === "GET" && resolverConfig?.httpMethod === "GET")
```
flybayer, line 192:
```suggestion
      (req.method === "GET" && resolverConfig.httpMethod === "GET")
```

packages/blitz-rpc/src/index-server.ts
```diff
@@ -192,11 +212,22 @@ export function rpcHandler(config: RpcConfig) {
...
+        if (req.method === "POST") {
+          data = deserialize({
+            json: req.body.params,
+            meta: req.body.meta?.params,
+          })
+        } else {
+          let meta = req.query.meta
+          if (Array.isArray(meta)) {
+            meta = meta[meta.length - 1]
+          }
```
flybayer, line 225: can you add a comment describing why this is necessary?

siddhsuresh: @flybayer this was due to an error when I tested this feature in a slightly bigger project with multiple queries set as `GET`
The error was 
image
Then during debugging I found that it is because of the `?params` words added to the `meta` properties that are not being parsed by `superjson`
image
This was my workaround to use only the first meta value added to avoid this error. 
If this can be fixed in a better way, I shall add those changes.

flybayer: ok that's not good, then we should fix `meta` ever being an array.

packages/blitz-rpc/src/data-client/rpc.ts
```diff
@@ -93,18 +95,29 @@ export function __internal_buildRpcClient({
...
+      },
+    })
+
+    if (httpMethod === "GET") {
+      fullRoutePath +=
+        "?" +
+        new URLSearchParams({
+          params: stringify(serialized.json),
+          meta: stringify(serialized.meta),
+        })
```
flybayer, line 111: Here's your `meta` is an array bug. 

each time this function runs, it keeps appending to fullRoutePath, which is "global" at the module level.

flybayer (changes requested):

packages/blitz-rpc/src/data-client/rpc.ts
```diff
@@ -103,12 +103,10 @@ export function __internal_buildRpcClient({
...
-      fullRoutePath +=
-        "?" +
-        new URLSearchParams({
-          params: stringify(serialized.json),
-          meta: stringify(serialized.meta),
-        })
+      const url = new URL(fullRoutePath, window.location.origin)
+      url.searchParams.set("params", stringify(serialized.json))
+      url.searchParams.set("meta", stringify(serialized.meta))
+      fullRoutePath = url.toString()
```
flybayer, line 109: This is not really better, because it overwrites the original `fullRoutePath`

Probably you should change `fullRoutePath` up above to be `routePath` and then here do like this. So `routePath` is never overwritten (change to const)

```suggestion
      const url = new URL(routePath, window.location.origin)
      url.searchParams.set("params", stringify(serialized.json))
      url.searchParams.set("meta", stringify(serialized.meta))
      fullRoutePath = url.toString()
```

siddhsuresh: @flybayer I made the requested changes

flybayer (commented):
awesome, just one small change left!

packages/blitz-rpc/src/data-client/rpc.ts
```diff
@@ -93,18 +96,25 @@ export function __internal_buildRpcClient({
       serialized = serialize(params)
     }
 
+    const body = JSON.stringify({
+      params: serialized.json,
+      meta: {
+        params: serialized.meta,
+      },
+    })
```
flybayer, line 104: What the heck, I thought I left a comment here but I don't see it anymore.

Let's move this JSON.stringify call down inline with `body: httpMethod === "POST" ? body : undefined,` so that json.stringify only runs for method POST.

dillonraphael (approved):

flybayer (approved):

Issue #2829, opened 2021-10-09
Routes manifest: Extract real/dynamic paths from RouteUrlObject


nickluger:
### What do you want and why?

[Routes manifest](https://blitzjs.com/docs/route-manifest#query-parameters) is a great Single Source of Truth for pages and routes. There are certain places in the app, though, where we would need the filled out URL-path instead of the Next.js dynamic-placeholder path, i.e. we would like to extract

```
/posts/123
```

instead of 
```
/posts/[postId]
```

Currently, the `RouteUrlObject` generated by a `Routes.Post()` call only offers us the properties `pathname` and `query`. Neither can be directly used, for example in:

- Redirects in `getStaticProps` etc.
- Canonical URL link tags like `<link rel="canonical" href={canonicalUrl} />`

### Possible implementation(s)

Add a field to the `RouteUrlObject` like `href` or `as` (following Next.js  Link-component naming) which returns `/posts/123`, instead of `/posts/[postId]`. 

Of course, using `pathname` and `query`one can easily construct the filled out path oneself, but I think it would be more natural to embed it into this than rolling your own helper.

beerose:
Thanks for reporting. I think this is a nice idea! In other issues (#2816, #2619), a suggested solution was to add `toString` method to `RouteUrlObject`, which handles filled-out path scenario. Open for discussion about what would work better!

flybayer:
I vote for `Routes.Post().href`. `href` matches other nextjs/blitz href usage and also `new Url()` usage.

margalit:
Would love if this also exposed a method that prepended the baseUrl. My specific use case is copying a link to the clipboard for sharing.

flybayer:
@margalit should that be a separate field? I'm not personally sure of which places nextjs automatically handles basePath and which they don't.

nickluger:
As things are going to pivot soon, i did not contribute a PR, but here's a workaround. This is type safe on the consumer side.

```ts
import { Routes } from "blitz";
// ** HINT ** You can use any other mapper function ** HINT **
import { mapValues } from "lodash";

type RoutesType = typeof Routes;
type RoutesWithHrefType = {
  [K in keyof RoutesType]: (
    ...params: Parameters<RoutesType[K]>
  ) => ReturnType<RoutesType[K]> & { href: string };
};

export const routes: RoutesWithHrefType = mapValues(
  Routes,
  (Route) =>
    function RouteWithHref(...args: Parameters<typeof Route>) {
[...]

Issue #2816, opened 2021-10-07
make RouteURLObject work directly with res.redirect


okmr-d:
### What do you want and why?
Currently, the expected type of API route' `res.redirect` is string.

It would be nice if we could pass one of our Route Manifest objects into this field.

### Possible implementation(s)

(1) ~add toString() method to RouteURLObject~ (will be covered by blitz-js/blitz#2829).

(2) make RouteURLObject work directly with res.redirect.

### Additional context
similar issue: https://github.com/blitz-js/legacy-framework/issues/79

I think (1) would be convenient, because we can use it anywhere else (for example, canonical tags).

beerose:
Ideally, we could have both implementations. The first one could also cover this issue: https://github.com/blitz-js/blitz/issues/2829 (the suggested solution is slightly different but also works). To keep things tidy, I'll rename this issue to `make RouteURLObject work directly with res.redirect` (solution number 2), and solution number 1 will be covered by https://github.com/blitz-js/blitz/issues/2829. Hope that works.

PR #3892, opened 2022-10-10
feat(cli): add href property to RouteURLObject

a11rew:
Closes: #2829, #2816 

Adds an href property to the Route objects generated in the Route Manifest. 
This href property represents a pathname with the query parameters filled out.

pathname: `projects/[projectID]/users/[userID]/edit`
href: `projects/test-project/users/test-user/edit`

This href could be used in `res.redirect` or canonical URL tags.

## Bug Checklist

## Feature Checklist


a11rew:
@dillonraphael from my testing, in dev linking the cli package to another app does not change where the processes are run so the declarations are not changed. i.e. this call causes the updated declaration and implementations for the route manifest to be output into the blitz packages's node_modules. Not the app it is linked to
https://github.com/blitz-js/blitz/blob/a6e81f156bedf0def9c7b6cd1f856df64e981841/packages/blitz/src/cli/utils/routes-manifest.ts#L518

That might be why you're not seeing the href in the declarations of the app you use to test. I got around that by temporarily hardcoding the path to my apps .blitz folder in node_modules. Screenshot of route with proper type declarations attached
Screenshot from 2022-10-18 09-13-35

 This shouldn't be an issue when this is merged though since like with the other route manifest methods and properties it'll be generated from the app's copy of blitz.

Regarding the manifests not being generated with `dev` but with `codegen`, I think they're actually generated but into the wrong location because of the above.

And as far as I can tell, `blitz dev` and `blitz codegen` call the same manifest generator function so there shouldn't be any difference in what they output (?). I'd appreciate pointers if otherwise 
https://github.com/blitz-js/blitz/blob/a6e81f156bedf0def9c7b6cd1f856df64e981841/packages/blitz/src/cli/utils/routes-manifest.ts#L583-L585

https://github.com/blitz-js/blitz/blob/a6e81f156bedf0def9c7b6cd1f856df64e981841/packages/blitz/src/cli/utils/codegen-tasks.ts#L32

a11rew:
Hi @dillonraphael @beerose, a look here? :sweat_smile:

dillonraphael:
Hey sorry @a11rew! Long overdue response. If you patch the path to your apps .blitz folder in node_modules, did blitz dev work? 

The thing is, when I run blitz codegen, it updates the route manifest correctly (with the href property). But once I run blitz dev, I can see the same file being overwritten which makes me believe it's not so much an issue with finding `.blitz`.

But yes I agree that this odd since both blitz dev & blitz codegen are calling the same routes-manifest function.

a11rew:
Yes @dillonraphael `blitz dev` and `blitz codegen` worked the same - they both add the href property. Weird they don't work the same for you, maybe you don't have this version of the cli (with the href addition) linked? 

Here I have a standalone blitz project, with the cli symlinked with pnpm and the `dotBlitz` path patched like so - 
image


https://user-images.githubusercontent.com/87580113/200022612-f489b589-09b1-4f42-a8e1-af7e2486ffe2.mp4

dillonraphael:
> Yes @dillonraphael `blitz dev` and `blitz codegen` worked the same - they both add the href property. Weird they don't work the same for you, maybe you don't have this version of the cli (with the href addition) linked?
> 
> Here I have a standalone blitz project, with the cli symlinked with pnpm and the `dotBlitz` path patched like so - image
> 
>  2022-11-04.15-58-26.mp4

Interesting. I'll try again.

jeliasson:
Any updates on this? Would love to see the `.href` property.

dillonraphael (approved):

PR #3898, opened 2022-10-12
fix(rpc): allow return undefined in setQueryData updater function

janvennemann:
Closes: ?

Updates the return type for the updater function in `setQueryData` to include undefined. This matches the typings in tanstack query, see [here](https://github.com/TanStack/query/blob/cf0eb6cf70dce2403629a8a57182e0dc9a8180c7/packages/query-core/src/queryClient.ts#L132) and [here](https://github.com/TanStack/query/blob/cf0eb6cf70dce2403629a8a57182e0dc9a8180c7/packages/query-core/src/utils.ts#L61).

## Bug Checklist

## Feature Checklist


beerose (approved):
Thank you!

PR #3906, opened 2022-10-17
Upgrade CI Syntax

siddhsuresh:
- [x] upgrade to the latest Github Action syntax from `set-output` to environment variables to set the integration test folders 
```bash
::set-output name=folders::["auth","get-initial-props","middleware","no-suspense","qm","react-query-utils","rpc","trailing-slash"]
Warning: The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
```
image

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter

- [x] improve the detection of `no-changeset` for internal repository changes and stop changeset workflow to run after the PR is merged.

Previously it checked the latest label event, now it will check for the `no-changeset` label in all the labels that have been assigned to the PR

### DIFF
--- a/.github/workflows/changeset_check.yml
+++ b/.github/workflows/changeset_check.yml
@@ -10,7 +10,7 @@ concurrency:
 
 jobs:
   changeset:
-    if: ${{ github.event.label.name != 'no-changeset' }}
+    if: ${{ !contains(github.event.pull_request.labels.*.name, 'no-changeset') && github.event.pull_request.merged == false }}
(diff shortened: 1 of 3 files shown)


beerose (approved):
Thanks!

Issue #3663, opened 2022-08-03
@blitzjs/rpc does not work without auth plugin


edrickleong:
### What is the problem?

Hi, I am trying to setup the minimal blitzjs project and adding the `@blitz/rpc` without the `@blitz/auth` package. I assumed that blitz rpc would work without the blitz auth plugin but it currently throws an error.

### Paste all your error logs here:

```
Error: Internal Blitz Error: globalThis.__BLITZ_SESSION_COOKIE_PREFIX is not set
```

### Paste all relevant code snippets here:

I created a brand new blitz project and made minimal changes to add @blitz/rpc. The reproduced repo and commit to add @blitz/rpc can be found here https://github.com/edrickleong/blitz-rpc-error/commit/dd56ed6145e3e60a453abe9f5a9398b7f585e7a2

### What are detailed steps to reproduce this?

1. Clone https://github.com/edrickleong/blitz-rpc-error
2. Run `blitz dev`

### Run `blitz -v` and paste the output here:

```
Blitz version: 2.0.0-alpha.65 (local)
macOS Monterey | darwin-x64 | Node: v16.15.1


 Package manager: pnpm

  System:
    OS: macOS 12.5
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 4.25 GB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.15.1 - ~/.volta/tools/image/node/16.15.1/bin/node
    Yarn: 1.22.17 - ~/.volta/tools/image/yarn/1.22.17/bin/yarn
    npm: 8.4.0 - ~/.volta/tools/image/npm/8.4.0/bin/npm
  npmPackages:
    @blitzjs/auth: Not Found
    @blitzjs/next: alpha => 2.0.0-alpha.65 
    @blitzjs/rpc: alpha => 2.0.0-alpha.65 
    @prisma/client: Not Found
    blitz: 2.0.0-alpha.65 => 2.0.0-alpha.65 
    next: 12.2.0 => 12.2.0 
    prisma: Not Found
    react: 18.0.0 => 18.0.0 
    react-dom: 18.0.0 => 18.0.0 
    typescript: ^4.5.3 => 4.7.4 
```

### Please include below any other applicable logs and screenshots that show your problem:

[...]

PR #3943, opened 2022-11-03
Decouple Blitz RPC and Blitz Auth

siddhsuresh:
Closes: #3663 
  
### Implement client-side plugin `middleware` and `events` 

Reference: https://github.com/blitz-js/blitz/pull/3943#issuecomment-1314529266
 
#### **Blitz Auth**

- [x] provide ```beforeHttpRequest``` hook to set `csrf` header on request 
- [x] `beforeHttpResponse` hook to read session info on response and could emit the `blitz:session-created` event to reset the queries
- [x] provide `onRpcError` hook that will run on every `rpc` call to check the type of error returned by the resolver

#### **Blitz RPC**

- [x] consume `beforeHttpRequest` and `beforeHttpResponse` hooks (to set/read headers)
- [x] consume `onRpcError` response hook 
- [x] consume `onSessionCreated` to invalidate queries 

### Dependency changes

- [x] move `@blitzjs/auth` dependency of `@blitzjs/rpc` from `dependencies` to `devDependencies`  

- [x] Integration Tests:
  - [x] Add tests for `beforeHttpRequest` and `beforeHttpResponse` middleware hooks
  - [x] Add tests for `onRpcError` and `onSessionCreated` event hooks
- [x] Unit Tests in `blitz` package:
  - [x] Add tests for `merge`, `pipe` and `reduceBlitzPlugins` functions
- [x] Update `rpc`, `rpc-path-root` and `react-query-utils` integrations tests by removing `@blitzjs/auth` dependency

### Misc Changes to CI
- [x] remove `turborepo` caching action as it is not being used
- [x] fix `playwright` installation for integration tests

## Feature Checklist

- [x] Documentation
<details>
<summary>

## Example Client Plugin

</summary>

```tsx
// src/custom-plugin/plugin.tsx
import { createClientPlugin } from "blitz"
import { ComponentProps, ComponentType } from "react"

type CustomPluginOptions = {
  // ... 
}

function withCustomPlugin<TProps = any>(Page: ComponentType<TProps> | BlitzPage<TProps>) {
  const CustomPluginRoot = (props: ComponentProps<any>) => {    
    // ... custom root component
    return <Page {...props} />
  }
  for (let [key, value] of Object.entries(Page)) {
    CustomPluginRoot[key] = value
  }
  if (process.env.NODE_ENV !== "production") {
    CustomPluginRoot.displayName = `CustomPluginInnerRoot`
  }
  return CustomPluginRoot
}

export const BlitzCustomPlugin = createClientPlugin<CustomPluginOptions, {}>(
  (options?: CustomPluginOptions) => {
    // ... plugin code
    return {
      events: {
        onSessionCreated: async () => {
          // Called when a new session is created - Usually when the user logs in or logs out
        },
        onRpcError: async () => {
          // Called when an RPC call fails
        },
      },
      middleware: {
        beforeHttpRequest: (req) => {
          //make changes to the request options before RPC call
          return req
        },
        beforeHttpResponse: (res) => {
          //make changes to the response before returning to the caller
          return res
        },
      },
      exports: () => ({
        // ... plugin exports
      }),
      withProvider: withCustomPlugin,
    }
  }
)

```
```ts
// src/blitz-client.ts
import {AuthClientPlugin} from "@blitzjs/auth"
import {setupBlitzClient} from "@blitzjs/next"
import {BlitzRpcPlugin} from "@blitzjs/rpc"
import {BlitzCustomPlugin} from "./custom-plugin/plugin"

export const {withBlitz} = setupBlitzClient({
  plugins: [
    AuthClientPlugin({
      cookiePrefix: "web-cookie-prefix",
    }),
    BlitzRpcPlugin({}),
    BlitzCustomPlugin({}),
  ],
})
```
</details>


dillonraphael:
@siddhsuresh I really like how this is looking so far. But the main thing I'm thinking about is what if a user would want to use a different authentication library? For example next-auth?

Maybe a way for users to provider their own CSRFToken without needing blitz-auth?

siddhsuresh:
> @siddhsuresh I really like how this is looking so far. But the main thing I'm thinking about is what if a user would want to use a different authentication library? For example next-auth?
> 
> Maybe a way for users to provider their own CSRFToken without needing blitz-auth?

@dillonraphael I haven't the time yet to test the integration with next-auth, but created the required API for the user to set the csrf token in blitz rpc when blitz auth is not present [If it is, it takes over regardless of the user input here]

Example `blitz-client.ts`
```ts
import { setupBlitzClient } from "@blitzjs/next"
import { BlitzRpcPlugin } from "@blitzjs/rpc"

export const { withBlitz } = setupBlitzClient({
  plugins: [
    BlitzRpcPlugin({
      cookieOptions: {
        prefix: "MY_APP_",
        csrfToken: "AYWj+ihVAByTgVrlPreVdsGo8LbLu4Cnd0CJI1+sWgg=",
        sameSite: "lax",
        domain: "localhost",
        expires: new Date("2021-12-31T23:59:59.999Z"),        
        secure: true,
      },
    }),
  ],
})
```
Using the logic present in https://github.com/blitz-js/blitz/blob/main/packages/blitz-auth/src/server/auth-sessions.ts#L463-L481 to implement in rpc package.

Defaults are set to all cookieOptions except for the csrfToken, also let me know if any of the names should be changed for better DX as I set them currently to get the functionality working.

flybayer:
@siddhsuresh stand by, I need a little time to consider what the ideal implementation is here. I really appreciate your help on it!

apps/toolkit-app/pages/index.tsx
```diff
@@ -7,7 +7,6 @@ import logout from "app/auth/mutations/logout"
 import logo from "public/logo.png"
 import { useMutation } from "@blitzjs/rpc"
 import { Routes, BlitzPage } from "@blitzjs/next"
```
siddhsuresh, line 9:
```suggestion
import { Routes, BlitzPage } from "@blitzjs/next"
import { getSession, useSession } from "@blitzjs/auth"
```

flybayer:
ok @siddhsuresh finally turning my attention to this, this week.

First, I removed the useSession hook from useQuery hooks: https://github.com/blitz-js/blitz/pull/3968

Secondly, we want to untangle these packages via plugins/middleware that other custom plugins can also use. Earlier this year we [stubbed out some ideas](https://github.com/blitz-js/blitz/blob/main/packages/blitz/src/index-browser.tsx#L28-L45) (currently unused).

We have the following needs:

**Blitz Auth**
- provide pre and post request hooks to set csrf header on request and read session info on response
- provide a RPC response hook that will check for AuthenticationError (currently done in rpc.ts)
- emit sessionCreated event

**Blitz RPC**
- consume pre and post request hooks (to set/read headers)
- consume RPC payload response hook (to replace AuthenticationError check in rpc.ts)
- consume sessionCreated event to invalidate queries (currently done in rpc.ts)

We want to use the same approach as we stubbed out, but we may need to adjust names, etc.

Doest this make sense? I'm available to assist as much as you need this week

siddhsuresh:
> ok @siddhsuresh finally turning my attention to this, this week.
> 
> First, I removed the useSession hook from useQuery hooks: https://github.com/blitz-js/blitz/pull/3968
> 
> Secondly, we want to untangle these packages via plugins/middleware that other custom plugins can also use. Earlier this year we [stubbed out some ideas](https://github.com/blitz-js/blitz/blob/main/packages/blitz/src/index-browser.tsx#L28-L45) (currently unused).
> 
> We have the following needs:
> 
> **Blitz Auth**
> - provide pre and post request hooks to set csrf header on request and read session info on response
> - provide a RPC response hook that will check for AuthenticationError (currently done in rpc.ts)
> - emit sessionCreated event
> 
> **Blitz RPC**
> - consume pre and post request hooks (to set/read headers)
> - consume RPC payload response hook (to replace AuthenticationError check in rpc.ts)
> - consume sessionCreated event to invalidate queries (currently done in rpc.ts)
> 
> We want to use the same approach as we stubbed out, but we may need to adjust names, etc.
> 
> Doest this make sense? I'm available to assist as much as you need this week

@flybayer yeah , I think it makes sense to me. I shall give this a try and let you know if I need some help.

flybayer (changes requested):
AHH THIS IS BEAUTIFUL :D

Just needs some small adjustments.

A clarification:

- `event` is something that does not return or modify anything.
- `middleware` is something that takes input and can return a modified version

So we need to split these up accordingly

packages/blitz-next/src/index-browser.tsx
```diff
@@ -136,8 +142,33 @@ const setupBlitzClient = <TPlugins extends readonly ClientPlugin<object>[]>({
...
+      preRequest: plugin.events.preRequest
+        ? pipe(acc.preRequest, plugin.events.preRequest)
+        : acc.preRequest,
+      rpcResponse: plugin.events.rpcResponse
+        ? pipe(acc.rpcResponse, plugin.events.rpcResponse)
+        : acc.rpcResponse,
+      handleError: plugin.events.handleError
+        ? pipe(acc.handleError, plugin.events.handleError)
+        : acc.handleError,
+    }),
```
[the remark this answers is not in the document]
siddhsuresh: yeah @flybayer I agree have replaced this logic for events with the `Promise.all` syntax so that all custom events are also taken into account.

packages/blitz-rpc/src/data-client/rpc.ts
```diff
@@ -101,56 +97,35 @@ export function __internal_buildRpcClient({
...
-          if (response.headers.get(HEADER_CSRF_ERROR)) {
-            const err = new CSRFTokenMismatchError()
-            err.stack = null!
-            throw err
-          }
-        }
-
+        document.addEventListener("blitz-auth:session-created", resetQueryClient)
+        __BLITZ_rpcResponse(response)
+        document.removeEventListener("blitz-auth:session-created", resetQueryClient)
```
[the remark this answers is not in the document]
siddhsuresh: Hey @flybayer, for this I changed this to 
```ts
await Promise.all(__BLITZ_onSessionCreated(resetQueryClient)) // to set the event listener
__BLITZ_beforeHttpResponse(response)
```
Is this what you meant? If not could you please elaborate on what needs to be changed here?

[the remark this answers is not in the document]
siddhsuresh: 🤦‍♂️Oh nevermind I understood what you meant by that. I will make the change when I can next.

packages/blitz-auth/src/client/index.tsx
```diff
@@ -391,8 +397,55 @@ export const AuthClientPlugin = createClientPlugin((options: AuthPluginClientOpt
...
+          getPublicDataStore().clear()
+        }
+      },
+      onSessionCreated: async (resetQueryClient) => {
+        resetQueryClient()
+        debug("Blitz Auth: onSessionCreated")
+      },
+    },
+    middleware: {
+      beforeHttpRequest(options: RequestInit) {
```
flybayer, line 414:
```suggestion
      beforeHttpRequest(req) {
```

packages/blitz-next/src/global.ts
```diff
@@ -1,5 +1,10 @@
 import {QueryClient} from "@tanstack/react-query"
+import {BeforeHttpRequest, BeforeHttpResponse, OnRpcErrorHook, OnSessionCreatedHook} from "blitz"
 
 declare global {
   var queryClient: QueryClient
+  var __BLITZ_beforeHttpRequest: BeforeHttpRequest
+  var __BLITZ_beforeHttpResponse: BeforeHttpResponse
+  var __BLITZ_onRpcError: OnRpcErrorHook
+  var __BLITZ_onSessionCreated: OnSessionCreatedHook
```
flybayer, line 9: these should be in the `blitz` package instead of `blitz-next`

packages/blitz-auth/src/client/index.tsx
```diff
@@ -391,8 +397,55 @@ export const AuthClientPlugin = createClientPlugin((options: AuthPluginClientOpt
...
+        // We don't clear the publicDataStore for anonymous users,
+        // because there is not sensitive data
+        if (error.name === "AuthenticationError" && getPublicDataStore().getData().userId) {
+          getPublicDataStore().clear()
+        }
+      },
+      onSessionCreated: async (resetQueryClient) => {
+        resetQueryClient()
+        debug("Blitz Auth: onSessionCreated")
+      },
```
flybayer, line 411: This should be in blitz-rpc plugin. blitz-auth emits the onSessionCreated event, blitz-rpc responds to it

packages/blitz-rpc/src/data-client/rpc.ts
```diff
@@ -101,56 +97,46 @@ export function __internal_buildRpcClient({
...
-        }
-
+        document.addEventListener(
+          "blitz:session-created",
+          async () => {
+            await Promise.all(__BLITZ_onSessionCreated(resetQueryClient))
+            debug("blitz-rpc:session-created")
+          },
+          {once: true},
+        )
```
flybayer, line 138: this `addEventListener` call must be inside setupBlitzClient instead of here

packages/blitz-rpc/src/data-client/rpc.ts
```diff
@@ -166,18 +152,12 @@ export function __internal_buildRpcClient({
...
               json: payload.error,
               meta: payload.meta?.error,
             }) as any
-            // We don't clear the publicDataStore for anonymous users,
-            // because there is not sensitive data
-            if (error.name === "AuthenticationError" && getPublicDataStore().getData().userId) {
-              getPublicDataStore().clear()
-            }
-
+            await Promise.all(__BLITZ_onRpcError(error))
```
flybayer, line 160: we have mixture here. onSessionCreated is using `document.dispatchEvent` but `onRpcError` does not use document event system and instead just calls the global handler.

We should make this consistent. I'm not sure which is the best way, but I'm leaning towards the document.dispatchEvent way, because it seems more decoupled.

packages/blitz-rpc/src/global.ts
```diff
@@ -1,6 +1,11 @@
 import {QueryClient} from "@tanstack/react-query"
+import {BeforeHttpRequest, BeforeHttpResponse, OnRpcErrorHook, OnSessionCreatedHook} from "blitz"
 
 declare global {
   var queryClient: QueryClient
   var __BLITZ_SUSPENSE_ENABLED: boolean
+  var __BLITZ_beforeHttpRequest: BeforeHttpRequest
+  var __BLITZ_beforeHttpResponse: BeforeHttpResponse
+  var __BLITZ_onRpcError: OnRpcErrorHook
+  var __BLITZ_onSessionCreated: OnSessionCreatedHook
```
flybayer, line 10: I think these shouldn't be needed here if they are defined on `global` in `blitz`

packages/blitz-rpc/src/resolver.ts
```diff
@@ -276,13 +271,19 @@ interface ResolverAuthorize {
...
-    const session: SessionContext = (ctx as any).session
-    session.$authorize(...args)
+    if (!(ctx as any).$session) {
+      log.warn(
+        "You are using the resolver `authorize` function, but `@blitzjs/auth` plugin is not initialised. Consider enabling to use authentication features.",
+      )
+    } else {
+      const session: SessionContext = (ctx as any).session
+      session.$authorize(...args)
+    }
```
flybayer, line 286:
```suggestion
    const session: SessionContext = (ctx as any).session
    if (!session) throw new Error("You are using the resolver.authorize() function, but `@blitzjs/auth` plugin is not initialized. Ensure it is installed and the auth plugin is added to setupBlitzServer()")
    session.$authorize(...args)
```

packages/blitz/src/types.ts
```diff
@@ -2,10 +2,34 @@ import {UrlObject} from "url"
 // Context for plugins to declaration merge stuff into
 export interface Ctx {}
 
+export interface MiddlewareCtx {}
+
```
flybayer, line 6: Because this is same as Ctx, so we need to use Ctx instead
```suggestion
```

packages/blitz/src/types.ts
```diff
@@ -2,10 +2,34 @@ import {UrlObject} from "url"
...
 export interface RouteUrlObject extends Pick<UrlObject, "pathname" | "query"> {
   pathname: string
 }
 
+export type EventHooks = {
+  onSessionCreated?: OnSessionCreated
+  onRpcError?: OnRpcError
+}
+
+export type BeforeHttpRequest = (options: RequestInit) => RequestInit
```
flybayer, line 16:
```suggestion
export type BeforeHttpRequest = (request: RequestInit) => RequestInit
```

packages/blitz/src/types.ts
```diff
@@ -2,10 +2,34 @@ import {UrlObject} from "url"
...
+
+export type BeforeHttpResponse = (response: Response) => Response
+
+export type OnRpcError = (error: Error) => Promise<void>
+
+export type OnRpcErrorHook = (error: Error) => Promise<void>[]
```
flybayer, line 22: These extra xHook types seem redundant — can we remove them?

packages/blitz/src/logging.ts
```diff
@@ -166,6 +166,10 @@ const error = (msg: string) => {
   console.log(`${c.red(msg)}`)
 }
 
+const warn = (msg: string) => {
+  console.log(`${c.hex(brandColor).bold("⚡️ Blitz Warning: ")} ${c.yellow(msg)}`)
+}
```
flybayer, line 171: now unused
```suggestion
```

flybayer (changes requested):
note to self 

- [x] investigate resolver.pipe ctx type and ensure `resolver.authenticate()` correctly modifies ctx type

packages/blitz-rpc/package.json
```diff
@@ -31,6 +30,7 @@
     "supports-color": "8.1.1"
   },
   "devDependencies": {
+    "@blitzjs/auth": "2.0.0-beta.18",
```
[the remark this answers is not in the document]
flybayer: Ah, ok then I think we can keep this here.

packages/blitz/src/types.ts
siddhsuresh: I tried changing the type to Ctx, but it causes type issues in the resolvers
image
So I have reverted the changes in the latest commit

packages/blitz-rpc/src/data-client/rpc.ts
siddhsuresh: I have moved the `addEventListener` to the `useEffect` in `buildWithBlitz` function in the same file. Cannot add in setupBlitzClient because as far as I understand it runs in the server and `document` does not exist there.
Also, I removed the parameter `resetQueryClient` instead of keeping the definition in `blitz-rpc` and running it there. 
Is this fine? 
Or should I move the `resetQueryClient` function to `blitz-next` ?

packages/blitz-next/src/index-browser.tsx
```diff
@@ -37,6 +45,9 @@ const buildWithBlitz = <TPlugins extends readonly ClientPlugin<object>[]>(plugin
         setTimeout(() => {
           document.documentElement.classList.add("blitz-first-render-complete")
         })
+        document.addEventListener("blitz:session-created", async () => {
+          await Promise.all(globalThis.__BLITZ_onSessionCreated())
+        })
```
flybayer, line 50: this needs to go in the `blitz` package because it's a global event that all plugins should use, not just nextjs apps.

It can go in the `setupBlitzClient` function, but just only run it if `isClient === true`

packages/blitz-next/src/index-browser.tsx
```diff
@@ -137,7 +148,53 @@ const setupBlitzClient = <TPlugins extends readonly ClientPlugin<object>[]>({
...
+        onSessionCreated: merge<void, Promise<void>>([]),
+      },
+      exports: {},
+    },
+  )
+
+  globalThis.__BLITZ_beforeHttpRequest = middleware.beforeHttpRequest
+  globalThis.__BLITZ_beforeHttpResponse = middleware.beforeHttpResponse
+  globalThis.__BLITZ_onRpcError = events.onRpcError
+  globalThis.__BLITZ_onSessionCreated = events.onSessionCreated
```
flybayer, line 197: this also needs to go in `blitz` package

flybayer (changes requested):

packages/blitz-rpc/src/data-client/rpc.ts
flybayer: resetQueryClient is now good.

addEventListener needs moved to setupBlitzClient, see my comment below

packages/blitz-rpc/src/data-client/rpc.ts
flybayer: this function is not ran on each render. It's only called when react-query triggers a fetch.

also since this rpcError is an event, we don't have to wait for it to be processed. Which is why dispatchEvent makes sense to me.

packages/blitz/src/types.ts
flybayer: ok I will investigate

packages/blitz-rpc/src/resolver.ts
```diff
@@ -276,12 +271,16 @@ interface ResolverAuthorize {
   <T, C = Ctx>(...args: Parameters<SessionContextBase["$authorize"]>): (
     input: T,
     ctx: C,
-  ) => ResultWithContext<T, AuthenticatedMiddlewareCtx>
+  ) => ResultWithContext<T, MiddlewareCtx>
```
flybayer, line 274:
```suggestion
  ) => ResultWithContext<T, AuthenticatedMiddlewareCtx>
```

packages/blitz-rpc/src/resolver.ts
```diff
@@ -1,9 +1,8 @@
-import {AuthenticatedSessionContext, SessionContext, SessionContextBase} from "@blitzjs/auth"
-import {Await, Ctx, EnsurePromise} from "blitz"
+import type {SessionContext, SessionContextBase} from "@blitzjs/auth"
+import {Await, Ctx, EnsurePromise, MiddlewareCtx} from "blitz"
```
flybayer, line 2:
```suggestion
import {Await, Ctx, EnsurePromise, AuthenticatedMiddlewareCtx} from "blitz"
```

packages/blitz-auth/src/shared/types.ts
```diff
@@ -81,6 +81,9 @@ declare module "blitz" {
   export interface Ctx {
     session: SessionContext
   }
+  export interface MiddlewareCtx extends Omit<Ctx, "session"> {
+    session: AuthenticatedSessionContext
+  }
```
flybayer, line 86:
```suggestion
  export interface AuthenticatedMiddlewareCtx extends Omit<Ctx, "session"> {
    session: AuthenticatedSessionContext
  }
```

flybayer (changes requested):

flybayer:
alright, this is looking really good!! almost there!

packages/blitz-rpc/src/data-client/rpc.ts
siddhsuresh: Ah okay, I have updated the implementation using `CustomEvent`

packages/blitz-next/src/index-browser.tsx
siddhsuresh: Moved this logic to `reduceBlitzPlugins` function in `blitz/index-browser.ts` that will perform the `plugin.reduce`, assign the global middleware functions and create the required event listeners

packages/blitz/src/types.ts
```diff
@@ -2,10 +2,30 @@ import {UrlObject} from "url"
 // Context for plugins to declaration merge stuff into
 export interface Ctx {}
 
+export interface MiddlewareCtx {}
```
siddhsuresh, line 5:
```suggestion
export interface AuthenticatedMiddlewareCtx {}
```

flybayer:
sweet! I see you want to add some integration tests, that's good.

Also, we need to add docs for this

packages/blitz/src/plugin.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export function merge<T, U>([...fns]: Array<(args: T) => U>) {
+  return (args: T) => fns.map((fn) => fn(args))
+}
+
+export function pipe<T>(...fns: Array<(x: T) => T>) {
+  return (x: T) => fns.reduce((v, f) => f(v), x)
+}
+
+export function reduceBlitzPlugins<TPlugins extends readonly ClientPlugin<object>[]>({
```
siddhsuresh, line 12: - [ ] Add `withProvider` to reduce
https://github.com/blitz-js/blitz/blob/8c247e26e7cba72d7d76dd83427a20604afc504f/packages/blitz-next/src/index-browser.tsx#L25-L29

packages/blitz/tests/plugin.test.ts
```diff
@@ -0,0 +1,129 @@
+import {describe, it, expect, spyOn} from "vitest"
+import {ClientPlugin} from "../src/index-server"
+import {reduceBlitzPlugins, merge, pipe} from "../src/plugin"
+
+spyOn(console, "log")
```
siddhsuresh, line 5:
```suggestion
import {describe, it, expect, vi} from "vitest"
import {ClientPlugin} from "../src/index-server"
import {reduceBlitzPlugins, merge, pipe} from "../src/plugin"

vi.spyOn(console, "log")
```

bravo-kernel:
@siddhsuresh this is pretty impressive, don't give up 💃

flybayer (approved):

bravo-kernel:
👏

PR #3953, opened 2022-11-07
Add feature for resolvers to exist outside of Blitz root

paulm17:
Link issue: https://github.com/blitz-js/blitz/discussions/3527

Allow for queries and mutations to exist outside of the blitz root folder, by adding a `includeRPCFolders ` property to the blitz object in `next.config.js`

Example:

```ts
  blitz: {
    includeRPCFolders: ["../no-suspense/app"],
  }
```

## Feature Checklist

- [x] Documentation added/updated (submit PR to [blitzjs.com repo `main` branch](https://github.com/blitz-js/blitzjs.com))

Documentation Page: https://github.com/blitz-js/blitzjs.com/pull/795


packages/blitz-rpc/src/loader-server.ts
```diff
@@ -56,21 +53,26 @@ export async function transformBlitzRpcServer(
...
+  directory: string,
+  rpcFolders: string[],
+  pageExtensions: string[],
+): Promise<string[]> {
+  return recursiveFindResolvers(directory, rpcFolders, buildPageExtensionRegex(pageExtensions))
 }
 
 export async function recursiveFindResolvers(
   dir: string,
+  rpcFolders: string[],
```
flybayer, line 75: move this to the last argument so it can be optional because we only need it at the top level recursion

packages/blitz-rpc/src/loader-server.ts
```diff
@@ -90,18 +93,19 @@ export async function recursiveFindResolvers(
...
       if (!filter.test(part)) {
         return
       }
 
-      const relativeFromRoot = absolutePath.replace(rootDir, "")
-      if (getIsRpcFile(relativeFromRoot)) {
-        arr.push(relativeFromRoot)
-        return
+      if (getIsRpcFile(absolutePath)) {
+        if (!absolutePath.includes("@blitzjs")) {
```
flybayer, line 105: shouldn't need blitzjs ignore here

packages/blitz-rpc/src/loader-utils.ts
```diff
@@ -4,6 +4,7 @@ import {ResolverPathOptions} from "./index-server"
 
 export interface LoaderOptions {
   resolverPath: ResolverPathOptions
+  includeRPCFolders: string[]
```
flybayer, line 7: should be optional

packages/blitz-rpc/src/loader-server.ts
```diff
@@ -17,12 +17,13 @@ export async function loader(this: Loader, input: string): Promise<string> {
   const compiler = this._compiler!
   const id = this.resource
   const root = this._compiler!.context
+  const rpcFolders = this.query.includeRPCFolders
```
flybayer, line 20: needs to default to empty array

flybayer (changes requested):

packages/blitz-rpc/src/loader-server.ts
paulm17: Due to the inclusion of the other "rpc" folders.  If I remove this line then everything breaks.  Other files in the root, do not have this issue.

absolutePath example:

> /Users/paul/development/src/github/bar2/blitz/packages/canvas/node_modules/@blitzjs/auth/node_modules/blitz/node_modules/@blitzjs/generator/templates/app/app/auth/mutations/forgotPassword.ts

Unless there already exists a function to clean this up?

packages/blitz-rpc/src/loader-server.ts
paulm17: As a consequence of moving the param rpcFolders last.  I've had to change the execution of recursiveFindResolvers to the following.

> return recursiveFindResolvers(directory, buildPageExtensionRegex(pageExtensions), undefined, [], directory, rpcFolders)

Is this ok?

packages/blitz-rpc/src/loader-server.ts
flybayer: Ah, then further up in this function we should filter out `node_modules` so we never search inside of it.

paulm17: Ok I have done the following:

```
if (absolutePath.includes("@blitzjs")) {  
    return
}

if (getIsRpcFile(absolutePath)) {                
    arr.push(absolutePath)
    return        
}
```

packages/blitz-rpc/src/loader-server.ts
```diff
@@ -90,17 +100,20 @@ export async function recursiveFindResolvers(
       const pathStat = await promises.stat(absolutePath)
 
       if (pathStat.isDirectory()) {
-        await recursiveFindResolvers(absolutePath, filter, ignore, arr, rootDir)
+        await recursiveFindResolvers(absolutePath, filter, ignore, arr, rootDir, rpcFolders)
```
flybayer, line 103:
```suggestion
        await recursiveFindResolvers(absolutePath, filter, ignore, arr, rootDir)
```

packages/blitz-rpc/src/loader-server.ts
```diff
@@ -90,17 +100,20 @@ export async function recursiveFindResolvers(
       const pathStat = await promises.stat(absolutePath)
 
       if (pathStat.isDirectory()) {
```
flybayer, line 102: here is where you should check for node_modules and abort. because we don't want to call `recursiveFindResolvers` on node_modules

flybayer (changes requested):

dillonraphael:
Really good job @paulm17!

I wonder if it would be better to create 2 new integration test directories instead of modifying the rpc integration test. That way we can cover the test case of not supplying the `includeRPCFolders` property. What do you think @flybayer ?

.changeset/olive-dragons-drum.md
```diff
@@ -0,0 +1,5 @@
+---
+"@blitzjs/rpc": patch
+---
+
+Add feature for resolvers to exist outside of Blitz root
```
flybayer, line 5:
```suggestion
Add ability to put your query and mutation resolvers in a separate monorepo folder, allowing you to use them in multiple apps.
```

flybayer (approved):

dillonraphael (approved):

flybayer:
pnpm lock file needs updated

PR #3958, opened 2022-11-10
Fix builder execute command hangs up

selcukfatihsevinc:
I'm trying to run install my recipe from my another recipe; like this,
```    
builder.addRunCommandStep({
   ...
   command: `blitz install ${myOtherRecipe} --yes`,
});
```
The main install process was hanging up. I don't want to pipe to main process (eg: cp.stderr.pipe(process.stderr))
Adding a stdout data event prevents this situation.

## Bug Checklist


dillonraphael:
Thanks @selcukfatihsevinc for this. Did this effect other recipes?

selcukfatihsevinc:
The recipes in blitz aren't using "addRunCommandStep", they're not affected. I couldn't find a better solution, but putting an empty stdout data event prevented process from hanging. Thanks @dillonraphael

dillonraphael:
@selcukfatihsevinc Ok cool. This should be good then. Although I'm really curious about a situation where you would want to install a blitz recipe from a blitz recipe. It's like inception 😂

PR #3961, opened 2022-11-10
Fix form paths when running blitz generate all

dillonraphael:
Closes: ?

Update the generated form path from `app` to `src` directory


flybayer (approved):

.changeset/giant-students-carry.md
```diff
@@ -0,0 +1,5 @@
+---
+"@blitzjs/generator": patch
+---
+
+Update form paths when running blitz generate all
```
flybayer, line 5: let's be more specific in changelogs and PR titles. This should be **fix** because it's broken :)

Issue #3964, opened 2022-11-14
Runtime error after upgrade to beta 15 (and later) because of enhancePrisma


medihack:
### What is the problem?

Upgrading from 2.0.0-beta.14 to 2.0.0-beta.15 (or later) gives me the below runtime error. I see that there was a commit named `Export enhancePrisma from the server entry point`. Do I have to change anything on my site? Also tried to delete `.next` and `node_modules` folder without success.

When I console log `enhancePrisma` in the `db/index.ts` file, it is a function on the server, but `undefined` on the client in beta 15. In beta 14 both were defined. So it seems there is an export issue.

**Update**:
I investigated a bit further. `index-browser.tsx` misses now an export of `enhancePrisma` after this [commit](https://github.com/blitz-js/blitz/commit/1b798d9a03242713517a56793b8e7c51d0fb1130). I see you [had problems](https://github.com/blitz-js/blitz/pull/3928) when importing it in the `index-browser.ts` (I never had those issues with `useSession`), but simply deleting that export isn't the solution.

### Paste all your error logs here:

```
Unhandled Runtime Error
TypeError: (0 , blitz__WEBPACK_IMPORTED_MODULE_0__.enhancePrisma) is not a function

Source
db/index.ts (4:37) @ eval

  2 | import { enhancePrisma } from "blitz"
  3 | 
> 4 | const EnhancedPrisma = enhancePrisma(PrismaClient)
    |                                     ^
  5 | 
  6 | export * from "@prisma/client"
  7 | const db = new EnhancedPrisma()
```

### Paste all relevant code snippets here:

This is how my db/index.ts file looks (nothing special there).
```
import { PrismaClient } from "@prisma/client"
import { enhancePrisma } from "blitz"

const EnhancedPrisma = enhancePrisma(PrismaClient)

export * from "@prisma/client"
const db = new EnhancedPrisma()
export default db
```

### What are detailed steps to reproduce this?

I just upgraded from beta 14 to beta 15. Nothing more.

### Run `blitz -v` and paste the output here:

```
Blitz version: 2.0.0-beta.17 (global)
Blitz version: 2.0.0-beta.15 (local)
Linux 5.15 | linux-x64 | Node: v16.16.0


 Package manager: npm

  System:
    OS: Linux 5.15 Ubuntu 20.04.4 LTS (Focal Fossa)
    CPU: (16) x64 AMD EPYC 7B13
    Memory: 27.31 GB / 62.80 GB
[...]

PR #3965, opened 2022-11-14
Export enhancePrisma for client again

medihack:
Closes: #3964

## Bug Checklist

## Feature Checklist


medihack:
`enahncePrisma` was removed from the client side by this [merged PR ](https://github.com/blitz-js/blitz/pull/3928). Unfortunately, this leads to a runtime error on the client described in #3964. I am not sure what @beerose issue exactly was with `Exporting it from the browser entry point caused issues when using useSession — Module not found: Can't resolve 'fs'` (never had that one and also using `useSession`), but deleting the export creates a new issue.

flybayer:
Yeah this should be exported from both client and server.

@beerose can you elaborate on what the initial problem was behind https://github.com/blitz-js/blitz/pull/3928 ?

beerose:
> Yeah this should be exported from both client and server.
> 
> @beerose can you elaborate on what the initial problem was behind #3928 ?

I tried to use blitz-auth in a new next app, and the Blitz setup didn't work unless I removed the enhancePrisma from the browser entry point (the error was as mentioned in the PR). Maybe the issue is somewhere else, but it's worth checking if adding it back won't break pure next + blitz-auth apps.

flybayer:
@beerose this does require adding withBlitz() to next.config.js — did you perhaps not have that when you got the error?

beerose:
> @beerose this does require adding withBlitz() to next.config.js — did you perhaps not have that when you got the error?

Yeah, that seems to solve the issue.

flybayer (approved):

PR #3969, opened 2022-11-15
remove Fallback port(3000)

nkilm:
Remove fallback PORT - 3000, which is used in blitz and forwared as it is to nextjs.

Before
https://github.com/blitz-js/blitz/blob/936211a657202aab4ba68791eb2f2aefa02db590/packages/blitz/src/cli/utils/next-utils.ts#L158

After modification
```js
 availablePort = await detect({port: config.port})
```


packages/blitz/src/cli/utils/next-utils.ts
```diff
@@ -155,7 +155,7 @@ async function createCommandAndPort(config: ServerConfig, command: string) {
   let spawnCommand: string[] = [command]
   let availablePort: number
 
-  availablePort = await detect({port: config.port ? config.port : 3000})
+  availablePort = await detect({port: config.port})
   spawnCommand = spawnCommand.concat(["-p", `${availablePort}`])
```
flybayer, line 159: all of this should be removed :)

flybayer (changes requested):
thank you!!

packages/blitz/src/cli/utils/next-utils.ts
nkilm: @flybayer `AvailablePort` is to be removed completely? i.e, line number 156 to 159
> Sorry, I'm still learning all these😅

flybayer: yes :)

nkilm: Okay

nkilm: Please check now

flybayer (approved):

flybayer:
@nkilm we were having CI issues that I think are resolved now.

There are a number of tests failing that I suspect are because of this change, but I'm not sure. Can you investigate? You'll want to run the failing tests locally and see if they pass or fail there. You can run the tests for a single integration test by going into the `integration-tests/X` folder and running `pnpm test`

nkilm:
@flybayer Okay I will test locally.

nkilm:
#### @flybayer Test results for `pnpm test` in root directory

It was showing the following error
image

### FInal result
image

packages/blitz/src/cli/utils/next-utils.ts
```diff
@@ -153,18 +153,14 @@ function getSpawnEnv(config: ServerConfig) {
...
 async function createCommandAndPort(config: ServerConfig, command: string) {
   let spawnCommand: string[] = [command]
-  let availablePort: number
-
-  availablePort = await detect({port: config.port ? config.port : 3000})
-  spawnCommand = spawnCommand.concat(["-p", `${availablePort}`])
 
   if (config.hostname) {
     spawnCommand = spawnCommand.concat(["-H", `${config.hostname}`])
   }
```
flybayer, line 159: we should remove this too

Issue #465, opened 2020-05-07
Allow installer transforms to run on multiple files


aem:
### What do you want and why?

Currently the transform executor API theoretically could support the user selecting multiple files to transform, or using a detector function to dynamically select all files that pass a custom test.

### Possible implementation(s)

There's two options here. First, we could continue to use the `selectTargetFiles` field which already exists on the API, and just require the author to do whatever they'd like to detect the total list of files to modify.

Alternatively, we could change that prop to something like `detectFile` which takes in a file's AST and the `recast` API (similar to the transform API) and returns a boolean, telling us whether or not to execute the transform on the file. In that case, we'd have to target the entire codebase recursively with the transform which makes it a bit slower, but has the benefit that it will prevent authors from having to write too much boilerplate.

There is of course the third option where we support both 😬. That's just a stretch goal though

PR #3974, opened 2022-11-16
Allow recipes to transform multiple files

tmcw:
Closes: #465

This adds a `multi` property to recipes that allows them to run codemods on multiple files: for example:

```ts
builder.addTransformFilesStep({
  stepId: "migrateTests",
  stepName: "Migrate tests",
  explanation: `Migrate Jest syntax to Vite`,
  singleFileSearch: "app/**/*.test.ts",
  multi: true,
  transform(program: j.Collection<j.Program>) {
    // …
    return program;
  },
});
```

This is unfortunately not fully finished, because I can't get `globby` to play nice with my environment: whether I use dynamic import or static import, I always get an error telling me to use dynamic import instead. I'll keep trying but I don't know how to get past that. So it currently includes the `glob` module which is CJS. This should be switched back when the ESM issues get resolved.

As far as I can tell, this is PR is necessary for supporting vitest with a recipe, because that recipe would have to modify all test files.

## Bug Checklist

## Feature Checklist


dillonraphael:
I've been having the same issue with globby, so you're not alone. Let me know if you figure anything out. But this PR idea is great!

dillonraphael:
Hey @tmcw any luck with this?

tmcw:
I haven't had much luck getting `globby` to work correctly, so kind of gave up on this PR, unfortunately 😢

PR #3991, opened 2022-11-24
fix resolverPath:root to make it work with monorepo resolvers outside of root directory

paulm17:
Does not close any particular issue.

This is the 2nd fix after the initial PR for https://github.com/blitz-js/blitz/pull/3953.  Hopefully all use cases work now.  🙏 

This change does not affect the documentation.

## Bug Checklist

## Feature Checklist


packages/blitz-rpc/src/loader-client.ts
```diff
@@ -41,7 +41,8 @@ export async function transformBlitzRpcResolverClient(
 ) {
   assertPosixPath(id)
   assertPosixPath(root)
-  const resolverFilePath = "/" + posix.relative(root, id)
+  let resolverFilePath = "/" + posix.relative(root, id)
+  resolverFilePath = normalize(join(normalize(root), resolverFilePath.replace("/", sep)))
```
siddhsuresh, line 45:
```suggestion
  const resolverFilePath = normalize(join(normalize(root), "/" + posix.relative(root, id).replace("/", sep))).replace(/\\/g, "/")
```

packages/blitz-rpc/src/loader-server-resolvers.ts
```diff
@@ -41,7 +41,8 @@ export async function transformBlitzRpcResolverServer(
   assertPosixPath(id)
   assertPosixPath(root)
 
-  const resolverFilePath = "/" + posix.relative(root, id)
+  let resolverFilePath = "/" + posix.relative(root, id)
+  resolverFilePath = normalize(join(normalize(root), resolverFilePath.replace("/", sep)))
```
siddhsuresh, line 45:
```suggestion
  const resolverFilePath = normalize(join(normalize(root), "/" + posix.relative(root, id).replace("/", sep))).replace(/\\/g, "/")
```

siddhsuresh (commented):
Hey @paulm17, adding the `.replace(/\\/g, "/")` makes it windows compatible

packages/blitz-rpc/src/loader-utils.ts
```diff
@@ -69,6 +69,15 @@ export function convertPageFilePathToRoutePath({
...
       path = path.replace(join(normalize(appRoot), extraPath.replace("/", sep)), "")
+
+      // change paths included in includeRPCfolders setting only
+      if (
+        normalize(absoluteFilePath).includes(
+          normalize(extraPath.replace("../", "").replace("../", sep)),
+        )
+      ) {
+        path = join(normalize(extraPath.replace("../", "").replace("../", sep)), path)
+      }
```
flybayer, line 80: can you explain how this fixes it? 

We should try to find another way to fix this that doesn't rely on hard coding `../` values.

lastly, can you update the integration test with a reproduction of the bug you have in your app? I need that to understand what the problem is and so we can make sure it doesn't break in the future.

paulm17: > We should try to find another way to fix this that doesn't rely on hard coding ../ values.

Yes, this is what I mentioned in discord about needing some help.  I tried some functions like normalise and I couldn't find a better way.  I absolutely agree with your comment.

> can you explain how this fixes it?

Primarily, I had to update resolverFilePath in other files to reflect the original change from relative to absolute path in the original PR.  Doing this change, ensured that the output was always uniform.

Secondly, if our next config, blitz param is like this:

```
resolverPath: "root",
includeRPCFolders: ["../../modules"],
```

The request urls become:

> http://localhost:3001/api/rpc/modules/dam/queries/getDam

Otherwise if I remove that code.  It becomes:

> http://localhost:3001/api/rpc/dam/queries/getDam

I think this approach is correct, as if there was another workspace with duplicate names for the components and or queries/mutations.  This may lead to issues.  I think we should cover all use cases.

paulm17: Would this be a better change?

```
// change paths included in includeRPCfolders setting only
const baseRPCFolder = sep + extraPath.replace(/^(?:\.\.\/)+/, "")

if (normalize(absoluteFilePath).includes(baseRPCFolder)) {
    path = join(baseRPCFolder, path)
}
```

flybayer: @paulm17 get the integration test added, and then I look at how to refactor this

paulm17:
Hope it's ok.  I had asked for confirmation on discord on what to do and just went ahead with doing a PR.  Let me know if there's anything else I need to do to get this resolved.  Thanks!

integration-tests/rpc-path-root/test/index.test.js
```diff
@@ -49,11 +49,16 @@ function runTests(dev = false) {
...
+        const data = await fetchViaHTTP(
+          appPort,
+          "/api/rpc/no-suspense/app/queries/getNoSuspenseBasic",
+          null,
+          {
+            method: "POST",
+            headers: {"Content-Type": "application/json; charset=utf-8"},
+            body: JSON.stringify({params: {}}),
+          },
+        ).then((res) => res.ok && res.json())
```
flybayer, line 61: this test should not change. For this app with it's current config, the rpc path should be `/api/rpc/queries/getNoSuspenseBasic`.

so we need to app a **new** integration test to reproduce and test for the issue that @paulm17 is having

packages/blitz-rpc/src/loader-utils.ts
```diff
@@ -67,8 +67,19 @@ export function convertPageFilePathToRoutePath({
     path = resolverBasePath(path)
   } else if (resolverBasePath === "root") {
     path = path.replace(normalize(appRoot), "")
+
     for (const extraPath of extraRpcBasePaths) {
       path = path.replace(join(normalize(appRoot), extraPath.replace("/", sep)), "")
+
+      let baseRPCFolder = sep + extraPath.replace(/^(?:\.\.\/)+/, "")
```
flybayer, line 74: we still need to fix this to not hard code `../`

flybayer (changes requested):

flybayer:
closed in favor of https://github.com/blitz-js/blitz/pull/4045

Issue #3900, opened 2022-10-14
Blitz install stuck on "Generating file diff" in new project


mikeyXmomo:
### What is the problem?

`blitz install tailwind` stuck on "Generating file diff". I waited for an hour and still not done yet

### Paste all your error logs here:

```
Loaded env from /Users/macbookair/Documents/Projects/recipe-test/.env.local
Loaded env from /Users/macbookair/Documents/Projects/recipe-test/.env
✅ Installed 3 dependencies
✅ Installed 3 dependencies
✅ Successfully created postcss.config.js, tailwind.config.js
✅ Successfully created app/core/styles/index.css

+––––––––––––––––––––––––+
⎪   Import stylesheets   ⎪
+––––––––––––––––––––––––+
Imports the stylesheet we just added into your app

⠏Generating file diff...
```

### Paste all relevant code snippets here:

```
blitz install tailwind
```

### What are detailed steps to reproduce this?

1.Updated Blitz to version: 2.0.0-beta.12 with `sudo npm install -g blitz`
2. Created a new Blitz app with `blitz new`
3. `cd `into folder
4. Attempted to install tailwind with `blitz install tailwind` and stuck on "Generating file diff"

### Run `blitz -v` and paste the output here:

```
blitz -v
Blitz version: 2.0.0-beta.12 (global)
Blitz version: 2.0.0-beta.12 (local)
macOS Monterey | darwin-arm64 | Node: v18.7.0


 Package manager: npm

  System:
    OS: macOS 12.6
    CPU: (8) arm64 Apple M1
    Memory: 79.05 MB / 8.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.7.0 - ~/.nvm/versions/node/v18.7.0/bin/node
    Yarn: 3.2.2 - ~/.nvm/versions/node/v18.7.0/bin/yarn
    npm: 8.19.2 - ~/.nvm/versions/node/v18.7.0/bin/npm
  npmPackages:
    @blitzjs/auth: 2.0.0-beta.12 => 2.0.0-beta.12
    @blitzjs/next: 2.0.0-beta.12 => 2.0.0-beta.12
    @blitzjs/rpc: 2.0.0-beta.12 => 2.0.0-beta.12
    @prisma/client: 4.0.0 => 4.0.0
    blitz: 2.0.0-beta.12 => 2.0.0-beta.12
    next: 12.2.5 => 12.2.5
[...]

PR #3995, opened 2022-11-28
Fix blitz install for recipes that use the path helper

dillonraphael:
Closes: #3900

Check if `src/pages` directory exists in the installer paths helper, else use the root `pages` directory.

## Bug Checklist


packages/blitz/src/installer/utils/paths.ts
```diff
@@ -18,12 +18,23 @@ function getBlitzPath(type: string) {
...
 }
 
+function findPageDir() {
+  const srcPagePath = `src/pages`
+  const srcPage = fs.existsSync(path.resolve(srcPagePath))
+
+  if (srcPage) {
+    return srcPagePath
+  } else {
+    return `pages`
```
maotora, line 28: I think previous versions of blitz were using `app/pages` instead of just `pages`.

maotora:
@dillonraphael Hey nice catch, I was facing the same error trying to install the tailwind recipe. 
I just left a comment in one of your files. I think previous versions of blitz were using `app/pages` or something similar instead of just `pages`.

maotora:
Thanks for your response, @dillonraphael!  

I just tested this and I found there is another edge case to cover for Tailwindcss to get our coding files which no longer reside at pages or app directories. 
We need to update `tailwind.config.js` content array to reflect the directory change as well.

I came up with something like this which fixed it;

```js
// tailwind.config.js
module.exports = {
  content: ["src/**/*.{js,ts,jsx,tsx}"],
  theme: {
    extend: {},
  },
  plugins: [],
}
```

Which worked!

dillonraphael:
> 

Good find @maotora. Would you like to open another PR for updating the tailwind recipe (since this is tailwind recipe specific)?

maotora:
@dillonraphael sure I can do that. Will make a PR and tag you then.

maotora:
@dillonraphael done I've made a PR: [here](https://github.com/blitz-js/blitz/pull/3997)

cassilup:
/Still experiencing this in 2.0.2 (released last week).

Doc0x1:
Still experiencing this in 2.0.3 as well. I haven't tested to make sure but I think this might be an issue for projects that are using the App Router instead of the older Pages system.

siddhsuresh:
yeah @Doc0x1, that seems likely, we would need to add additional handling to also check for the existence of the `app` directory in the same area as the change in this PR.

PR #3997, opened 2022-12-04
Fix blitz recipe paths

maotora:
Closes: https://github.com/blitz-js/blitz/issues/4118

I have added support for `src/{components,pages}` directories for the new blitz version which creates pages and components there and made it backward compatible so if `app/pages` exists it uses that instead of the `src` directory.

## Bug Checklist


dillonraphael (commented):
Good work @maotora

We should probably add this to the `blitz/src/installer/utils/path.ts` file so we can keep things DRY (https://github.com/blitz-js/blitz/blob/fe2e4eb1e99ab5161300add7f1be24c89c855775/packages/blitz/src/installer/utils/paths.ts#L8).

dillonraphael:
> Hey @dillonraphael sorry for the late response, I got quite occupied lately.
> 
> I think in that function there is already support for `app` and `src` directories.
> 
> Screenshot 2022-12-13 at 21 57 04
> 
> Or did you mean I should edit the path for `_app` and `_document` files?

Correct. Let's use the `getBlitzPath` helper

recipes/chakra-ui/index.ts
```diff
@@ -104,6 +106,8 @@ function removeDefaultStyleElement(program: Program) {
   return program
 }
 
+const appDirectory = fs.existsSync(path.resolve("src/core")) ? "src" : "app"
```
dillonraphael, line 109: Let's use the `getBlitzPath` recipe helper here instead.

recipes/next-ui/index.ts
```diff
@@ -75,6 +77,8 @@ function removeDefaultStyleElement(program: Program) {
   return program
 }
 
+const appDirectory = fs.existsSync(path.resolve("src/core")) ? "src" : "app"
```
dillonraphael, line 80: Plus here

recipes/tailwind/templates/config/tailwind.config.js
```diff
@@ -1,6 +1,11 @@
 // tailwind.config.js
+const fs = require("fs")
+const path = require("path")
+
+const appDirectory = fs.existsSync(path.resolve("src/pages")) ? "src" : "{pages,app}"
```
dillonraphael, line 5: And here

dillonraphael (changes requested):

dillonraphael:
@maotora Looks good! Just need to fix some conflicts and we can get this merged

recipes/next-ui/index.ts
```diff
@@ -1,6 +1,8 @@
 import {addImport, paths, Program, RecipeBuilder} from "blitz/installer"
 import type {NodePath} from "ast-types/lib/node-path"
 import j, {JSXIdentifier} from "jscodeshift"
+import fs from "fs"
+import path from "path"
```
tordans, line 5: Are those imports needed?

recipes/tailwind/templates/config/tailwind.config.js
```diff
@@ -1,6 +1,8 @@
 // tailwind.config.js
+const {paths} = require("blitz/installer")
+
 module.exports = {
-  content: ["{pages,app}/**/*.{js,ts,jsx,tsx}"],
+  content: [`${paths.appSrcDirectory({config: true})}/**/*.{js,ts,jsx,tsx}`],
```
tordans, line 5: I wonder if this is a performant solution. It does mean, that on every run TW needs to first to the `fs.existsSync`, doesn't it?

I suggest adding the two possible places and let do the check for files. 

I don't know how the Next 13 app+pages-structure will work but assume that the Next 13 app directory needs to go into the blitz-src-folder as well, right? So there is no clash to be expected there.

```diff
-   content: [`${paths.appSrcDirectory({config: true})}/**/*.{js,ts,jsx,tsx}`],
+   content: ["{src,app}/**/*.{js,ts,jsx,tsx}"],
```

This would also simplify the lookup by removing the config-option.

---

About that, what does this `return "{pages,app}"` – the `pages, app` – actually stand for? The TW config is in root, so there is no "pages" folder on root, right? I never understood why the default config puts it like that. I changed it to just "`./src/*…`" for my setup and that works fine. Hence my change to have this part `{src,app}` above.

tordans (commented):
Looks like a good fix to me. It will close https://github.com/blitz-js/blitz/issues/4020. 

I left a few notes that came to mind when reading through the PR.

recipes/next-ui/index.ts
maotora: Nope, I forgot to remove them, thanks!

recipes/tailwind/templates/config/tailwind.config.js
maotora: @tordans great advice!

So to answer your question, I believe the earlier versions of blitz had a pages directory in the root together with app directory hence why we used `{pages,app}`.

@dillonraphael  I'm not sure if we still need to support `{app,pages,src}` or if we can go with `{app,src}` this time as adding something like `{pages,app,src}` may result in a "No such file or directory" error for recent projects.

dillonraphael: There is still a chance the user may have the `pages` directory in their root. For example if they are using a default nextjs setup, most likely they will have their `pages` directory in the root and not in `src`.

maotora: Okay, so from the above explanations I believe it won't error if I put all the directories in to consideration ie. `{app,pages,src}` and we can get rid of the `isConfig` param.

tordans: > Okay, so from the above explanations I believe it won't error if I put all the directories in to consideration ie. {app,pages,src} and we can get rid of the isConfig param.

I tested and confirmed it in my blitz app just now, having `{app,pages,src}` but only a `src` directory is no issue.

If we where to follow the https://tailwindcss.com/docs/guides/nextjs recommendations to the letter, this would become:

```
  content: [
    "./src/**/*.{js,ts,jsx,tsx}",
    "./app/**/*.{js,ts,jsx,tsx}",
    "./pages/**/*.{js,ts,jsx,tsx}",
  ],
```

(They spread it, but they also have a `./` at the start.)

maotora: Thanks, @tordans I have updated the code to have `./{pages,src,app}/**/` I've tested and this seems to work.

.changeset/many-fans-fetch.md
```diff
@@ -0,0 +1,7 @@
+---
+"@blitzjs/recipe-chakra-ui": patch
+"@blitzjs/recipe-next-ui": patch
+"@blitzjs/recipe-tailwind": patch
+---
+
+This change is to reflect the directory change from app/pages to src/pages. It checks for which directories exists first so it is compatible with either versions.
```
tordans, line 7:
```suggestion
This change reflects the directory change from app/pages to src/pages for the receipts chakra-ui, next-ui, tailwind. It checks for which directories exists and is compatible with either versions.
```

packages/blitz/src/installer/utils/paths.ts
```diff
@@ -18,14 +18,28 @@ function getBlitzPath(type: string) {
...
-    return srcPagePath
-  } else {
-    return `pages`
+  switch (srcPage) {
+    case "src": {
+      return srcPagePath
+    }
+    default: {
+      return `pages`
+    }
```
tordans, line 42: TBH I am not sure if this change here helps the clarity. We cannot get rid of the const `srcPagePath`, one has to read another function to understand what is true/false; and the switch is harder to read for me than the if-else.

recipes/tailwind/templates/config/tailwind.config.js
```diff
@@ -1,6 +1,7 @@
 // tailwind.config.js
+
```
tordans, line 2:
```suggestion
```

Files like the vitest.config dont have a new line either.

tordans (commented):
Looks good to me! Minor subjective suggestions.

packages/blitz/src/installer/utils/paths.ts
maotora: I thought factoring out some logic would be helpful that's why I made a call to `getAppSourceDir` in `findPageDir` function.

flybayer (approved):

PR #3998, opened 2022-12-05
Prisma helper function for nested array fields

dillonraphael:

It helps convert array fields into prisma Api format. It will convert an array like [1,2,3] to something like {create: [1,2,3]} and will use the correct prisma CRUD method based on initial data.

For example:
```
 value = [ {name: 1}, {id: 2, name: 3} ]
 initial = [ {id: 2, name: 2}, {id: 3} ]
 prepareArrayField(value, initial)
 
// returns
 {
   create: [ {name: 1} ],
   update: [ { data: {name: 3}, where: {id: 2} } ],
   remove: [ {id: 3} ]
 }
```

## Feature Checklist

- [ ] Unit test added (see [test docs](https://blitzjs.com/docs/contributing#running-tests) if needed)


flybayer (commented):
this looks good. Can you share a realistic example of how this looks in practice?

Issue #4001, opened 2022-12-08
BlitzServerMiddleware has `any` types


flybayer:
### What is the problem?

BlitzServerMiddleware has `any` as the type for `req`, `res, and `next`

```ts
    BlitzServerMiddleware(async (req, res, next) => {
    }),
```

### To Fix

[This code](https://github.com/blitz-js/blitz/blob/main/packages/blitz/src/index-server.ts#L61-L67) needs fixed. The `middleware` argument should be `RequestMiddleware` instead of using a generic

ETLopes:
Does it makes sense to be this way?

```
export const BlitzServerMiddleware = <
  TRequest extends IncomingMessage = IncomingMessage,
  TResponse extends ServerResponse = ServerResponse,
  TMiddleware extends RequestMiddleware<TRequest, TResponse> = RequestMiddleware<TRequest, TResponse>,
>(
  middleware: TMiddleware,
): BlitzServerPlugin => ({
  requestMiddlewares: [middleware],
})
```

PR #4002, opened 2022-12-09
fix: The middleware argument should be RequestMiddleware instead of using a generic

ETLopes:
Closes: [#4001](https://github.com/blitz-js/blitz/issues/4001)

BlitzServerMiddleware has any as the type for req, res, and next`
The middleware argument was changed instead of using a generic.

## Bug Checklist

## Feature Checklist


dillonraphael:
Thanks @ETLopes this is close to what @flybayer wanted, but the issue with removing the generic is that older blitz middleware packages like https://github.com/ntgussoni/blitz-guard won't work.

I tested blitz guard with this PR, and I get type errors. This isn't your fault at all, and really we should get blitz guard updated.

PR #4018, opened 2022-12-23
Update README.md|s to use `src/` instead of `app/`

tordans:
I assume that after the Next 13 release the default folder for Blitz.js was change from `app` to `src`, right?
A fresh blitz app does use `src` and the docs say `src` https://blitzjs.com/docs/file-structure.
However, those README|s did not … – now they do.


dillonraphael:
Sweet! Thanks @tordans

dillonraphael (approved):

PR #4019, opened 2022-12-23
Reference `vitest.config.ts` instead of jest for README|s

tordans:
- The readmes reference the file that is present, now
- `apps/toolkit-app-passportjs/README.md` still uses jest, so that stayed as is

# Question

- [ ] Is it by design, that some folders have a `vitest.config.ts` _and_ `.js` file? I picket the `.ts` file for the readme…


apps/toolkit-app/README.md
```diff
@@ -122,7 +122,7 @@ __name__
 ├── .eslintrc.js
 ├── babel.config.js
 ├── blitz.config.ts
-├── jest.config.ts
+- `vitest.config.ts` contains config for Vitest tests. You can [customize it if needed](https://vitejs.dev/config/).
```
dillonraphael, line 125:
```suggestion
├── vitest.config.ts
```

dillonraphael (changes requested):

tordans:
@dillonraphael thanks for the review an fix.

- Could you give a hint at the question from the PR description: 
  >  Is it by design, that some folders have a vitest.config.ts and .js file? I picket the .ts file for the readme…
- This PR requires a changeset, yes? Something like "Update README|s to reference vitest instead of jest"

dillonraphael:
> @dillonraphael thanks for the review an fix.
> 
> * Could you give a hint at the question from the PR description:
>   > Is it by design, that some folders have a vitest.config.ts and .js file? I picket the .ts file for the readme…
> * This PR requires a changeset, yes? Something like "Update README|s to reference vitest instead of jest"

Not by design. Ideally we should be using .ts

Also I wouldn't say this PR needs a changeset. I tagged it as such.

dillonraphael (approved):

PR #4023, opened 2022-12-30
Remove redundant `vitest.config.js` files (the `.ts` files remain)

tordans:
Follow up to https://github.com/blitz-js/blitz/pull/4019#issuecomment-1367518718. Those `.js` files where identical to `.ts` files in the same folder, so we can remove them. The READMEs reference the `.ts` files.


dillonraphael:
@tordans This is for the generator incase someone selects to use javascript instead of typescript

tordans:
> This is for the generator incase someone selects to use javascript instead of typescript

Makes sense, closing…

Issue #4021, opened 2022-12-24
`blitz g` with brackets can break the cli


tordans:
### What is the problem?

**This is what I get after breaking thinks:**

```
% blitz help
blitz:22: command not found: endAt:dateTime
% blitz g all post title:string author:string content:string
blitz:23: command not found: endAt:dateTime
% blitz help
blitz:24: command not found: endAt:dateTime
```

**This is likely how I broke thinks…**

I suspect the `now()` caused the issue.

```
blitz generate all date startAt:dateTime:default=now() endAt:dateTime title:string locationName:string locationUrl:string description:string id:uuid --dry-run 
```

**How _not_ to fix this…**

- `npm uninstall -g blitz`, then reinstall does not help
- force-cleaning the npm cache does not help
- removing `node_modues`, then reinstalling all does not help

**How to fix it:**

I had to remove files in a npm temp folder, which "cached" broken code.

```
// 
$ echo TMPDIR
/var/folders/_2/w9vhc5hx1fg2vfmnypv3wvmm0000gp/T/
    
% cd $TMPDIR
% cd node-jiti
% ls
  blitz-auth-build.config.ts.42bb7c8e.js
  blitz-build.config.ts.531e0b43.js
  blitz-next-build.config.ts.b5eaaddf.js
  blitz-rpc-build.config.ts.f41b01c7.js
  codemod-build.config.ts.8d2cf2eb.js
  generator-build.config.ts.515f44ba.js
  pkg-template-build.config.ts.40d86f9a.js
    
$ cat pkg-template-build.config.ts.40d86f9a.js

"use strict"; Object.defineProperty(exports, "__esModule", { value: true }); exports.default = void 0;

const config = {
  entries: ["./src/index-browser", "./src/index-server"],
  externals: ["index-browser.cjs", "index-browser.mjs", "react"],
  declaration: true,
  rollup: {
    emitCJS: true,
    esbuild: {
      target: ["es2015"] } } };var _default =



config; exports.default = _default; /* v7-eb5970095fe3c5ea */

$ cat /private/var/folders/_2/w9vhc5hx1fg2vfmnypv3wvmm0000gp/T/node-jiti/generator-build.config.ts.515f44ba.js

"use strict";Object.defineProperty(exports, "__esModule", { value: true });exports.default = void 0;

const config = {
  entries: ["./src/index"],
  externals: ["react"],
  declaration: true,
  rollup: {
    emitCJS: true,
    esbuild: {
[...]

PR #4024, opened 2022-12-31
Guard `blitz generate` input against unwanted characters

tordans:
Closes: https://github.com/blitz-js/blitz/issues/4021

The input strings passed to `blitz g` are checked against an allow list of characters via regex. The helper `throw`s if an unwanted character is found.

I think it would be enough to check in `prisma/fields.ts`. The main reason for added it there was so the tests in `prisma/fields.test.ts` can be used to test the regex. However, `generators/model-generator.ts` has the raw string (right?), so this would be the place it first.
To keep the code simpler, we could remove it from `generators/model-generator.ts` again – let me know what you thing.

Also, please give the regex and everything a good review. I am new new to this :-).

## Bug Checklist

## Feature Checklist

- [ ] ~Documentation added/updated (submit PR to [blitzjs.com repo `main` branch](https://github.com/blitz-js/blitzjs.com))~


packages/generator/src/utils/checkInputOrRaise.ts
```diff
@@ -0,0 +1,14 @@
+export const checkInputsOrRaise = (inputs: string[] | string) => {
+  if (typeof inputs === "string") {
+    checkInputOrRaise(inputs)
+  } else {
+    inputs.forEach((input) => checkInputOrRaise(input))
+  }
+}
+
+const checkInputOrRaise = (input: string) => {
+  const regex = /^[a-zA-Z0-9-_:=\?[\]\s]+$/
```
Vandivier, line 10: 🎉

packages/generator/src/utils/checkInputOrRaise.ts
```diff
@@ -0,0 +1,14 @@
...
+  if (!regex.test(input)) {
+    throw new Error("Input can only a-z, A-Z, 0-9, _, -, [, ], ?")
+  }
```
siddhsuresh, line 13: Hey, @tordans thanks for the PR! 
A little design nitpick for the error to be displayed, I feel this could be better DX than the generic error.
```suggestion
 if (!regex.test(input)) {
    const firstInvalidCharacter = input.match(/[^a-zA-Z0-9-_:=\?[\]\s]/)
    if (firstInvalidCharacter) {
      log.branded(
        "Blitz Generator Parser Error: " +
          chalk.red(
            `Input contains invalid character: "${firstInvalidCharacter[0]}" in ${firstInvalidCharacter.input} at position ${firstInvalidCharacter.index}`,
          ),
      )
      throw new Error(
        "Input should only contain alphanumeric characters, spaces, and the following characters: - _ : = ? [ ]",
      )
    }
  }
```

siddhsuresh:
Hey @tordans, will you be able to commit the change requested?

tordans:
> are you able to commit the change requested? We can merge it then.

@siddhsuresh I looked into it and create https://github.com/blitz-js/blitz/pull/4053 as a follow up. I don't know enough about regex to add tests without spending too much time on it.

I hope this can be merged as a first step and https://github.com/blitz-js/blitz/pull/4053 continued as an improvement. 

My primary goal was, to prevent the "catastrophic failure" (as in, "nothings works anymore"). Since it looks like I was the first to have this issue, it also looks like it's not too common, so having a rougher error message might be enough…

siddhsuresh:
> > are you able to commit the change requested? We can merge it then.
> 
> @siddhsuresh I looked into it and create #4053 as a follow up. I don't know enough about regex to add tests without spending too much time on it.
> 
> I hope this can be merged as a first step and #4053 continued as an improvement.
> 
> My primary goal was, to prevent the "catastrophic failure" (as in, "nothings works anymore"). Since it looks like I was the first to have this issue, it also looks like it's not too common, so having a rougher error message might be enough…

Hey, @tordans, if it's fine with you I could commit the change and the changes to the tests to make it work.
Also, I think we should be able to finish these changes in this PR only, so I don't this the second PR would be necessary

tordans:
@siddhsuresh 
> Hey, …if it's fine with you I could commit the change and the changes to the tests to make it work.

Absolutely, always! Thanks! :-)

flybayer (approved):

PR #4032, opened 2023-01-03
prisma/field: Improve TS guard

tordans:
The previous solution `_fieldName as string` did solve TS warning inside the if-statement but not the line below, which still had this TS error. (This might be specific to my TS config, though…)

```javascript
      const relationType = Relation[_fieldName]
      // ^- const _fieldName: string | undefined
      //    Type 'undefined' cannot be used as an index type.ts(2538)
```


dillonraphael:
Thanks @tordans. It's always better to check `typeof` instead of doing type assertion.

dillonraphael (approved):

flybayer (approved):

PR #4039, opened 2023-01-05
Update package Zod to 3.20.2

tordans:
Zod 3.20 introduce `.string().datetime()` which looks like a very handy addition (more at https://github.com/colinhacks/zod/releases/tag/v3.20).

Right now, it is [3.19.1](https://github.com/blitz-js/blitz/blob/main/apps/toolkit-app/package.json#L41).

---

PS: Just in case, the Zod's [Changelog](https://github.com/colinhacks/zod/blob/master/CHANGELOG.md) is out of date.

---


tordans:
> fyi, you can update zod version in your app without this change.

Yes, we are. However, the changes to `.datetime()` look interesting for things like https://github.com/blitz-js/blitz/pull/3869/files#diff-d53f9302d286d7148a0cd2b137533349a17d66de9356256bbcfa6b4458061481R99.

> this change just updates internal zod version used for dev and test, not production. however, to get the build to pass, you must update all zod versions in this repo, not just the toolkit-app package :)

I just changed all zods and force-pushed on this branch.

flybayer (approved):
thank you!

PR #4050, opened 2023-01-11
internal: Publish PRs to npm

siddhsuresh:
# 🚀 Features

based on https://github.com/withastro/astro/blob/main/.github/workflows/snapshot-release.yml

✔️ Action controlled using the `!preview {tag_name}` command
✔️ Allow only members with admin access to publish a PR to npm
✔️ Using https://github.com/changesets/changesets/blob/main/docs/snapshot-releases.md feature of changesets to create a new version of blitz with the format of `0.0.0-next-{tag_name}-{Time}`
✔️ End users can use the in-progress PRs through the `@next--{tag_name}` tag 

PS. Cannot test this outside fork since the workflow must exist in `main` for the `issue_comment` event to be triggered


.github/workflows/pr-release.yml
```diff
@@ -0,0 +1,44 @@
...
+        env:
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+      - name: Comment on PR
+        uses: peter-evans/create-or-update-comment@v2
+        with:
+          issue-number: ${{ github.event.pull_request.number }}
+          body: |
+            The new version of this PR is available as next in npm:
+            ```
+            npm install <package>@next${{ github.event.pull_request.number }}
```
flybayer, line 43:
```suggestion
            npm install <package>@next/${{ github.event.pull_request.number }}
```

.github/workflows/pr-release.yml
```diff
@@ -0,0 +1,44 @@
...
+  pull_request:
+    types: [issue_comment]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  changeset:
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'release/next') && github.event.pull_request.merged == false && contains(github.event.comment.body,'/publish') && github.event.comment.author_association == 'MEMBER' }}
```
flybayer, line 13: what was the thought behind requiring label **and** a comment?

Seems like we could maybe remove the label requirement

siddhsuresh: the label I thought would help in organizing and keeping track of the PRs currently published. 
But think you are right I will remove that check and let the maintainer decide on the label

siddhsuresh:
I will be testing this in https://github.com/blitz-js/blitz/pull/3869 now

siddhsuresh:
test

siddhsuresh:
test

siddhsuresh:
test

flybayer (commented):
instead of requiring the tag name in the `!preview` command, could we use short commit sha instead?

otherwise good to go

siddhsuresh:
> instead of requiring the tag name in the `!preview` command, could we use short commit sha instead?
> 
> otherwise good to go

done @flybayer, I have added a fallback to the commit short SHA if we don't mention the tag to be published

PR #4051, opened 2023-01-12
Remove useEffect from reset password templates

tordans:
Closes: -

This signals, that this is by design and not a mistake because `router.isReady` is dependent on `router.query.token`.

## Change Checklist

- ~Changeset added (run `pnpm changeset` in the root directory)~
- ~Integration test added (see [test docs](https://blitzjs.com/docs/contributing#running-tests) if needed)~
- ~Documentation added/updated (submit PR to [blitzjs.com repo `main` branch](https://github.com/blitz-js/blitzjs.com))~


apps/toolkit-app/src/pages/auth/reset-password.tsx
```diff
@@ -16,6 +16,7 @@ const ResetPasswordPage: BlitzPage = () => {
 
   useEffect(() => {
     setToken(router.query.token as string)
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [router.isReady])
```
flybayer, line 20: Thanks @tordans! I was looking at the source of this change (https://github.com/tordans/blitz/commit/60de057477d0cc08743da3fc9230f074932849cb) and I think it's not the best implementation.

We should remove this useEffect entirely and then pass `token: router.query.token` directly in the `await resetPasswordMutation({ ...values, token: router.query.token })` call

flybayer (changes requested):

apps/toolkit-app/src/pages/auth/reset-password.tsx
tordans: Thanks @flybayer, I updated the code in a way that hopefully fits what you had in mind. 

In our own app, I extracted the handleSubmit which makes TS aware of the possible `token===undefined`. To catch this case properly, I added a pageState and extra message for when the token is missing. See https://github.com/FixMyBerlin/trassenscout/commit/3f84eb657c159437d352d0a5c852c623b50ae679 — Let me know if this is something I should add here or to a follow up PR.

.changeset/perfect-baboons-relate.md
```diff
@@ -0,0 +1,6 @@
+---
+"toolkit-app": patch
```
siddhsuresh, line 2:
```suggestion
```

flybayer (approved):

PR #4053, opened 2023-01-12
Guard `blitz generate` input against unwanted characters – Part 2

tordans:
This is a follow up to https://github.com/blitz-js/blitz/pull/4024 adding @siddhsuresh suggestion from https://github.com/blitz-js/blitz/pull/4024#discussion_r1062460708

However, this needs working tests ATM. If someone can finish this, that would be great. I don't know enough about regex to finish it in a timely manner.


tordans:
Closing in favor of https://github.com/blitz-js/blitz/pull/4024#issuecomment-1380183377.

PR #4054, opened 2023-01-13
Add missing Layout.tsx for generated mimimalapp

bravo-kernel:

Current `minimalapp` is generated without `src/core/Layout.tsx` causing errors and build to fail.


flybayer (approved):
thank you!!

PR #4061, opened 2023-01-18
Update to `"typescript": "4.9.4"`

tordans:
Closes: -

See https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html; Version 4.9 introduces the new `satisfies` operator.

## Feature Checklist

- [-] Integration test added (see [test docs](https://blitzjs.com/docs/contributing#running-tests) if needed)
- [-] Documentation added/updated (submit PR to [blitzjs.com repo `main` branch](https://github.com/blitz-js/blitzjs.com))

### DIFF
--- a/.changeset/angry-actors-greet.md
+++ b/.changeset/angry-actors-greet.md
@@ -1,0 +1,12 @@
+---
+"toolkit-app": patch
+"test-full-auth-with-rpc": patch
+"blitz": patch
+"@blitzjs/auth": patch
+"@blitzjs/next": patch
+"@blitzjs/rpc": patch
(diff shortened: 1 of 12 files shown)


tordans:
It looks like we are on [4.9.5 already](https://www.npmjs.com/package/typescript?activeTab=versions), now. Let me know which version is feasable for Blitz.

siddhsuresh:
> It looks like we are on [4.9.5 already](https://www.npmjs.com/package/typescript?activeTab=versions), now. Let me know which version is the right fit for Blitz.

@tordans, I will test with the versions in some time and update here

PR #4067, opened 2023-01-26
Add secure-password as an Optional peerDependency

siddhsuresh:
Closes: https://discordapp.com/channels/802917734999523368/1067888303606599750

Move `secure-password` from a mandatory dependency of `@blitz/auth` to an optional peerDependency

## Bug Checklist


dbrxnds:
Great improvement! Hoping this will be part of a beta soon

.changeset/thick-peas-jog.md
```diff
@@ -0,0 +1,7 @@
+---
+"@blitzjs/auth": major
+---
+
+Make secure-password an `optional peerDependency`. 
+
+Now, to use the **Blitz wrapper** for password hashing, install `secure-password` in your projects as a depenendency.
```
flybayer, line 7:
```suggestion
BREAKING CHANGE: secure-password is now an `optional peerDependency`, if you are using `SecurePassword` api, you need to now install `secure-password` in your application.

This helps users who do not use SecurePassword from having native package build issues.
```

dbrxnds:
This should be reopened again. Blitz exports some code that in turn imports stuff from secure-password so it will break as it cannot resolve that module. 

The secure-password stuff should probably be exported seperately or something.

PR #4068, opened 2023-01-26
Fix max-width of body in new app template

siddhsuresh:

image

### DIFF
--- a/packages/generator/templates/minimalapp/src/styles/Home.module.css
+++ b/packages/generator/templates/minimalapp/src/styles/Home.module.css
@@ -51,7 +51,7 @@
   width: 100%;
   flex-direction: row;
   align-self: center;
-  max-width: var(--screen-width);
+  max-width: calc(min(var(--screen-width),700px));
 }
 
 .globe {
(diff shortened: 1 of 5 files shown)


packages/generator/templates/app/src/styles/Home.module.css
```diff
@@ -51,7 +51,7 @@
   width: 100%;
   flex-direction: row;
   align-self: center;
-  max-width: var(--screen-width);
+  max-width: calc(min(var(--screen-width),700px));
```
flybayer, line 54:
```suggestion
  max-width: calc(min(var(--screen-width),800px));
```

packages/generator/templates/minimalapp/src/styles/Home.module.css
```diff
@@ -51,7 +51,7 @@
   width: 100%;
   flex-direction: row;
   align-self: center;
-  max-width: var(--screen-width);
+  max-width: calc(min(var(--screen-width),700px));
```
flybayer, line 54:
```suggestion
  max-width: calc(min(var(--screen-width),800px));
```

flybayer (approved):
otherwise good!

Issue #4065, opened 2023-01-20
Client tests appear to be broken on new apps (beta 22)


jhonnymichel:
### What is the problem?

In beta 22, the vitest client tests are broken in a new app (`blitz new`).

The only frontend test that exists is initially skipped. If I remove the "skip" flag on `test/index.test.tsx` and run the test, it fail with this error:

```
Error: NextRouter was not mounted. https://nextjs.org/docs/messages/next-router-not-mounted
 ❯ Object.useRouter node_modules/src/client/router.ts:135:11
 ❯ BlitzProvider node_modules/@blitzjs/next/dist/chunks/index-browser.cjs:462:27
 ❯ renderWithHooks node_modules/react-dom/cjs/react-dom.development.js:16305:18
 ❯ mountIndeterminateComponent node_modules/react-dom/cjs/react-dom.development.js:20074:13
 ❯ beginWork node_modules/react-dom/cjs/react-dom.development.js:21587:16
 ❯ beginWork$1 node_modules/react-dom/cjs/react-dom.development.js:27426:14
 ❯ performUnitOfWork node_modules/react-dom/cjs/react-dom.development.js:26560:12
 ❯ workLoopSync node_modules/react-dom/cjs/react-dom.development.js:26466:5
 ❯ renderRootSync node_modules/react-dom/cjs/react-dom.development.js:26434:7
 ❯ recoverFromConcurrentError node_modules/react-dom/cjs/react-dom.development.js:25850:20
```

Looks like there's something wrong in `BlitzProvider`. The error happens inside of it, even when trying to render it inside `<RouterContext>`.

It's a clean install so I believe the scaffold for frontend tests could be broken.

is it possible that the `test/utils` files generated with blitz new is outdated? The file is different from this one I found in Blitz sourcecode:  https://github.com/blitz-js/blitz/blob/main/integration-tests/utils/blitz-test-utils.tsx

### Paste all your error logs here:

```
Error: NextRouter was not mounted. https://nextjs.org/docs/messages/next-router-not-mounted
 ❯ Object.useRouter node_modules/src/client/router.ts:135:11
 ❯ BlitzProvider node_modules/@blitzjs/next/dist/chunks/index-browser.cjs:462:27
 ❯ renderWithHooks node_modules/react-dom/cjs/react-dom.development.js:16305:18
 ❯ mountIndeterminateComponent node_modules/react-dom/cjs/react-dom.development.js:20074:13
[...]

PR #4072, opened 2023-01-30
Fixing tests for blitz apps that use vitest

jhonnymichel:
Closes: #4065 

Some Blitz internals have internal logic to work properly on userland tests. In those places, to check if we're running inside a user test environment we checked for `JEST_WORKER_ID` being present in the env. 

Nowadays, new blitz projects are generated with vitest as the test runner, and it looks like some places in the codebase were not properly updated to support vitest in userland tests.

In addition to that, blitz internal tests seemed to rely on the fact userland tests ran on jest (and blitz internal tests run with vitest). This test runner difference between the blitz repo tests and userland tests is not necessarily true anymore.

This PR fixes all that:

- A new environment variable was added to the env of some integration tests, so the tests know the current test environment is a BLITZ (internal, own repo) TEST ENVIRONMENT.

> It looks like this BLITZ_TEST_ENVIRONMENT strategy already existed at some point and then was dropped. as I see a mention to it in the codebase, but the env var is nowhere to be found.

- updated some places to stop checking for just "JEST_WORKER_ID", adding support for vitest through the `isNotInUserTestEnvironment` function.
- updated the generated example test that was broken because of the issues and because it was outdated in relation to the generated index page.

## Bug Checklist


packages/generator/templates/app/test/index.test.tsx
```diff
@@ -28,10 +28,9 @@
       })
     }
   ))
- 
+
    const { getByText } = render(<Home />)
-   const linkElement = getByText(/Documentation/i)
+   const linkElement = getByText(/Blitz Docs/i)
```
jhonnymichel, line 33: Is this the right place to change this? Using my local dev version of blitz, I couldn't manage to create a new app with the updated template.

Also, should this file be updated in `apps/toolkit-app` too?

jhonnymichel:
brand new generated blitz app using blitz/main, running `index.test.ts` test (that uses useMutation):
```
 FAIL  test/index.test.tsx > renders blitz documentation link
Error: Either the file path to your resolver is incorrect (must be in a "queries" or "mutations" folder that isn't nested inside "pages" or "api") or you are trying to use Blitz's useQuery to fetch from third-party APIs (to do that, import useQuery directly from "@tanstack/react-query").
 ❯ validateQueryFn ../../packages/blitz-rpc/dist/chunks/rpc.cjs:46:11
 ❯ Object.sanitizeMutation ../../packages/blitz-rpc/dist/chunks/rpc.cjs:54:3
 ❯ Proxy.useMutation ../../packages/blitz-rpc/dist/index-browser.cjs:134:41
 ❯ UserInfo src/pages/index.tsx:17:28
     15| const UserInfo = () => {
     16|   const currentUser = useCurrentUser()
     17|   const [logoutMutation] = useMutation(logout)
       |                            ^
     18|
     19|   if (currentUser) {
 ❯ renderWithHooks ../../node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom.development.js:16305:18
 ❯ mountIndeterminateComponent ../../node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom.development.js:20074:13
 ❯ beginWork ../../node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom.development.js:21587:16
 ❯ beginWork$1 ../../node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom.development.js:27426:14
 ❯ performUnitOfWork ../../node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom.development.js:26560:12
 ❯ workLoopSync ../../node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom.development.js:26466:5
 ```
 
 Brand new generated blitz app using my branch, running all tests:
 
 ```
 ✓ src/auth/mutations/resetPassword.test.ts (1) 899ms
 ✓ src/auth/mutations/forgotPassword.test.ts (2) 2215ms
 ✓ test/index.test.tsx (1)

 Test Files  3 passed (3)
      Tests  4 passed (4)
   Start at  21:55:28
   Duration  3.57s (transform 548ms, setup 208ms, collect 1.03s, tests 3.16s)
   ```

File: .changeset/plenty-kiwis-greet.md [hunk 2]
```diff
@@ -0,0 +1,9 @@
+---
+"test-middleware": patch
+"test-qm": patch
+"test-rpc": patch
```
siddhsuresh, line 4:
```suggestion
```

packages/generator/templates/app/test/index.test.tsx
siddhsuresh: yes @jhonnymichel, it's the correct location for the change.

Yeah you could update the files in `apps/toolkit` with the new design and test

packages/blitz/src/utils/index.ts
```diff
@@ -143,3 +143,11 @@ export function interopDefault(mod: any) {
...
   return str.length > maxLength ? str.substring(0, maxLength - 3) + "..." : str
 }
+
+export function isNotInUserTestEnvironment() {
+  if (process.env.VITE_BLITZ_TEST_ENVIRONMENT) {
+    return true
+  }
+
+  return process.env.JEST_WORKER_ID === undefined && process.env.VITEST_WORKER_ID === undefined
+}
```
siddhsuresh, line 153: Hey @jhonnymichel, I was just taking a look at this, and haven't tested it yet. 

Logically for a function named `isNotInUserTestEnvironment` it should return `true` in normal execution and `false` in tests. 

So shouldn't after checking the `env` variable `VITE_BLITZ_TEST_ENVIRONMENT` that you are setting for the tests, shouldn't the function return `false` because it is in the test env. 

If it's a misconception on my part, or if I missing something, do let me know.

```suggestion
export function isNotInUserTestEnvironment() {
  if (process.env.VITE_BLITZ_TEST_ENVIRONMENT) {
    return false
  }

  return process.env.JEST_WORKER_ID === undefined && process.env.VITEST_WORKER_ID === undefined
}
```

jhonnymichel: The idea is to return `false` if we're in a 'blitz user' (a next app using the blitz toolkit) test environment.  it will return `true` while running in the blitz repo (this repo) tests.

That's why the function is called `isNotInUserTestEnvironment` instead of just `isNotInTestEnvironment`. the name is also not mine, I just moved the function around, but it already existed with the same strategy -- although it had a bug, according to my judgement.

Previously this check (wanting to know if we're running inside a blitz app, not inside the blitz repo) existed, but it was simply `if (process.env.JEST_WORKER_ID === undefined)`, because the expectation was that the blitz repo tests ran in vitest, and user (a next app using the blitz toolkit) tests ran in jest.

This not being the case anymore, we need a flag to differentiate the two cases.

Also, this flag strategy already existed in the codebase at some point, probably when the blitz repo tests also ran with Jest. and now I'm kinda recovering it and using it when needed.

siddhsuresh: Hey @jhonnymichel, thanks a lot for the clarification!

siddhsuresh (approved):
LGTM

zenhob:
I think my upgrade to Blitz 2.0 is blocked by this! :+1:

flybayer (approved):

Issue #4070, opened 2023-01-28
Next 13 App Directory / Server Components


tordans:
This ticket is taken over by @siddhsuresh. (@tordans opened it but all the text and work is by @siddhsuresh.)

---

This is a master issue to keep track of the development for blitz to work seamlessly with react server components and the new Next 13 `app` directory.

**Links:**

- Next.js App directory docs: https://beta.nextjs.org/
- React 18 Sever Components RFC: [rfcs/0188-server-components.md at main · reactjs/rfcs · GitHub](https://github.com/reactjs/rfcs/blob/main/text/0188-server-components.md)

# 🔧 Auth
## New Blitz Auth Hook getServerSession
- [ ] Create a new auth hook getServerSession that has the following API:  https://github.com/blitz-js/blitz/pull/4079
```ts
getServerSession(RequestCookies | ReadonlyRequestCookies, Headers) => SessionContext
```
Example Usage in React Server Component in app directory in Next 13
```ts
import {cookies, headers} from "next/headers"
import {getServerSession} from "src/blitz-server"
import getCurrentUser from "src/users/queries/getCurrentUser"

export default async function Home() {
  const session = await getServerSession(cookies(), headers())
  const user = await getCurrentUser(null, {session})
  return(
    <>
        ...
    </>
  )
}
```

# 🔧 RPC
- [ ] Add "use client" directive to the required files containing `react-query` to work with server components. https://github.com/blitz-js/blitz/pull/4079
- [ ] Add logic to use the next router in `react-query` for state only if in pages directory https://github.com/blitz-js/blitz/pull/4079
- [ ] New Blitz React Server Component Wrapper https://github.com/blitz-js/blitz/pull/4079
Create a new wrapper functionBlitzProvider to be imported from setupBlitzClient in src/blitz-client.ts
Example Usage
```ts
// layout.ts
import BlitzProvider from "src/blitz-client"

export default function RootLayout({children}: {children: React.ReactNode}) {
  return (
    <html lang="en">
      <head>
        ...
      </head>
      <body>
        <BlitzProvider>
          ...
        </BlitzProvider>
      </body>
    </html>
  )
}
```

# 🔧 `@blitzjs/next`
- [ ] Add "use client" directive to the required files containing `react-query` or react `hooks` to work with server components. https://github.com/blitz-js/blitz/pull/4079

# 🔧 Generator
[...]

PR #4079, opened 2023-02-06
feature: Next 13 App directory

siddhsuresh:
Solves Major Changes mentioned here: https://github.com/blitz-js/blitz/issues/4070

### 🚧Testing
*Please give feedback here if not merged yet or in a discord thread in help or in contributor general if you know to fix the bug*

**‼️ This PR has been published to npm to allow testing**

Install Using tag `@next--app-dir`

Example: https://github.com/siddhsuresh/Blitz-Next-App-Dir-Example

### 🔧 New Blitz Auth Function `getBlitzContext`

This function will use the [`cookies`](https://beta.nextjs.org/docs/api-reference/cookies) and [`headers`](https://beta.nextjs.org/docs/api-reference/headers) provided by the server component and returns the current session. 

#### API
```ts
getBlitzContext() => Ctx
```

#### Usage

**Example Usage in React Server Component in `app` directory in Next 13**

```ts
import {getBlitzContext} from "src/blitz-server"
import getCurrentUser from "src/users/queries/getCurrentUser"

export default async function Home() {
  const ctx = await getBlitzContext()
  const user = await getCurrentUser(null, ctx)
  return(
    <>
        ...
    </>
  )
}
```

  
### 🔧 New Blitz React Server Component Wrapper

- [x] Create a new wrapper function`BlitzProvider` to be imported from `setupBlitzClient` in `src/blitz-client.ts`

##### Example Usage

This provider should wrap the app and should be ideally placed at the root layout.ts file. 

```tsx
// layout.ts
import BlitzProvider from "src/blitz-client"

export default function RootLayout({children}: {children: React.ReactNode}) {
  return (
    <html lang="en">
      <head>
        ...
      </head>
      <body>
        <BlitzProvider>
          ...
        </BlitzProvider>
      </body>
    </html>
  )
}
```

### Blitz Next

- [x] Remove `<RouterContext.Provider value={router}>` from the `BlitzProvider`. Instead wrap the ErrorBoundary with the `withRouter` hook from `next/router`

### Dependencies

- [x] Bump Nextjs Version in packages and tests to 13.1.6

## Feature Checklist


packages/blitz-next/src/error-boundary.test.tsx
```diff
@@ -209,50 +211,52 @@ test("withErrorBoundary HOC", () => {
...
+//             errorBoundaryRef.current?.resetErrorBoundary()
+//           }}
+//         >
+//           recover
+//         </button>
+//         <ErrorBoundary ref={errorBoundaryRef} FallbackComponent={ErrorFallback}>
+//           {explode ? <Bomb /> : children}
+//         </ErrorBoundary>
+//       </>
+//     )
```
siddhsuresh, line 246: Fix This:

packages/blitz-auth/package.json
```diff
@@ -44,9 +44,13 @@
   },
   "peerDependencies": {
     "blitz": "2.0.0-beta.23",
+    "next": "13.1.6",
```
flybayer, line 47: can this be `*` so it's very loose?

packages/blitz-rpc/package.json
```diff
@@ -35,18 +35,22 @@
     "supports-color": "8.1.1"
   },
   "peerDependencies": {
+    "@tanstack/query-core": "4.24.4",
+    "@tanstack/react-query": "4.24.4",
```
flybayer, line 39: do we have to move this to peerDeps? it's a breaking change

flybayer (changes requested):

packages/blitz-rpc/package.json
siddhsuresh: No I missed to revert this. `@tanstack/query-core` should stay in peerDeps but `@tanstack/react-query` should go back to a regular dependency

packages/blitz-auth/package.json
siddhsuresh: yeah I will make the change now

siddhsuresh:
!preview app-dir

siddhsuresh:
!preview app-dir

packages/blitz-rpc/package.json
```diff
@@ -35,18 +36,20 @@
     "supports-color": "8.1.1"
   },
   "peerDependencies": {
+    "@tanstack/query-core": "4.24.4",
```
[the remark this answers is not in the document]
siddhsuresh: @flybayer I don't think this is necessary since we have `react-query` as the dependency which in turn has `@tanstack/query-core` as its dependency. So I feel this being in peer-deps would be better mainly for pnpm. 
Do you think its better to move to deps anyway?

siddhsuresh:
!preview app-dir

flybayer (approved):

bravo-kernel:
👏

PR #4081, opened 2023-02-10
fix: type error in reset-password.tsx

exKAZUu:
Follow up: #4051

Changes in #4051 cause an type error since `token` may be null.
This PR also removes unused an import statement.

## Bug Checklist

- [ ] ~~Integration test added (see [test docs](https://blitzjs.com/docs/contributing#running-tests) if needed)~~

### DIFF
--- a/packages/generator/templates/app/src/pages/auth/reset-password.tsx
+++ b/packages/generator/templates/app/src/pages/auth/reset-password.tsx
@@ -7,6 +7,7 @@ import { BlitzPage, Routes } from "@blitzjs/next"
 import { useRouter } from "next/router"
 import { useMutation } from "@blitzjs/rpc"
 import Link from "next/link"
+import { assert } from "blitz"
 
 const ResetPasswordPage: BlitzPage = () => {
   const router = useRouter()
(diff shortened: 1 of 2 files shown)


tordans:
Sorry about that.

However with this change the UI will fail silently, right?

In https://github.com/blitz-js/blitz/pull/4051#discussion_r1092191342 i suggested to add states for this that levied an explicit message. 

Another option would be to set the token to "not provided" so it fails the mutation and shows an error.

packages/generator/templates/app/src/pages/auth/reset-password.tsx
```diff
@@ -36,7 +35,7 @@ const ResetPasswordPage: BlitzPage = () => {
           }}
           onSubmit={async (values) => {
             try {
-              await resetPasswordMutation({ ...values, token })
+              token && await resetPasswordMutation({ ...values, token })
```
flybayer, line 38: thanks for the PR! 

instead of this, we should add an `assert` statement just above this line :)

flybayer (changes requested):

exKAZUu:
Thank you for your comments! I've added `assert` statement instead of failing silently.

flybayer (approved):

PR #4124, opened 2023-04-09
Add formatError property to Blitz server- & rpchandler setup

dbrxnds:
Closes: https://discord.com/channels/802917734999523368/1094607326108983400

Add a `formatError` callback property to `setupBlitzServer` and `rpcHandler`. Allowing users to add custom logic before an error gets returned to the client.

Example: Showing less descriptive errors in production such as "Internal server error occured" rather than a complete Prisma error.

### Questions
1. Would it be possible to pass the Blitz session public data to this callback function? This could allow for some logic based on roles / userid. I tried to make this work but I wasn't too sure about how to go about it, so I left it out for now.
2. Would it be okay to extract the statusCode from the formattedError (if there is one) and use that for the response?

## Bug Checklist

## Feature Checklist


.changeset/chatty-gifts-whisper.md
```diff
@@ -0,0 +1,7 @@
+---
+"@blitzjs/next": minor
+"@blitzjs/rpc": minor
+"toolkit-app": patch
```
siddhsuresh, line 4:
```suggestion
```

apps/toolkit-app/src/pages/api/rpc/[[...blitz]].ts
```diff
@@ -1,4 +1,20 @@
 import { rpcHandler } from "@blitzjs/rpc"
 import { api } from "src/blitz-server"
 
-export default api(rpcHandler({ onError: console.log }))
+class ErrorWithStatusCode extends Error {
```
Comment (siddhsuresh) [inline apps/toolkit-app/src/pages/api/rpc/[[...blitz]].ts:4]: Hey @dbrxnds , This looks good, I feel it would be better to export this error maybe something like `BlitzErrorWithStatusCode` in the main blitz package.
This would make it easier to maintain and reuse as the code is now hardcoding `status Code` attribute to be present in the custom error for it to work.

Comment (dbrxnds) [reply apps/toolkit-app/src/pages/api/rpc/[[...blitz]].ts:4]: This is only inside the tookit-app for testing now. Would you still prefer it to be exported from blitz package? Or should I just remove it from this code so people can make up their own logic for it?

siddhsuresh (approved):

flybayer (approved):

PR #4148, opened 2023-05-22
fix: enormous memory consumption of the dev server

jayu:
There was not issue for that, I've decided to fix the problem right away

### Context

We have a blitz v2 app with ~290 queries and mutations. 

Each query/mutation indirectly imports up to 900 source code files. 
We use dependency injection and as a result each endpoint imports the whole backend code.

We use next@12 and have SWC enabled.

Our dev server used to take up to 18GB or RAM, which is enormous amount of memory for parsing and merging together 900 source code files + compiling the font-end.

The memory was allocated for short period of time, and then reduced to 6-8GB.

It was killing 2020 MacBook Pro with 16BG of Ram.

image

I dig into the problem for quite a while and I found out that the issue was basically using dynamic imports for importing routes code.

Dynamic imports in Webpack automatically creates a separate chunk. 

Separate chunk means a separate isolated compilation of code. Chunking backend code does not make any sense unless it's run as lambda function.

With dynamic imports and the size of our app, Webpack was bundling the same code ~290 times instead of doing it once. 

That created huge memory peaks in quite random moments, eg. a 30 seconds after `pages/api/rpc/[[...blitz]].ts` finished to compile.

To fix that I simply changed the dynamic import into `require` call, which do not create separate chunk in Webpack.

We are using the patched version  for more than a week now and everything works fine. Compilation is faster, there are no more memory peaks and most importantly we reduced the team frustration. 

Preparing a reproducible demo that showcase the memory usage might not be easy without real production code, but at least I can prove that the chunks are created for each query/mutation file when using dynamic imports. Let me know if that's needed.

## Bug Checklist

- [ ] ~~Integration test added (see [test docs](https://blitzjs.com/docs/contributing#running-tests) if needed)~~
  - I believe existing tests should cover this change, cos it only changes how to code is bundled without affecting runtime

### DIFF
--- a/.changeset/good-oranges-pretend.md
+++ b/.changeset/good-oranges-pretend.md
@@ -1,0 +1,5 @@
+---
+"@blitzjs/rpc": patch
+---
+
+Fixes enormous memory consumption of the dev server
--- a/.changeset/good-oranges-pretend.md
+++ b/.changeset/good-oranges-pretend.md
@@ -19,7 +19,7 @@ const config = {
 }
 ```
 
-When `resolversDynamicImport` is set to `true`, the default import strategy will be "import" instead of "require".
+When `resolversDynamicImport` is set to `true`, the import strategy will be "import" instead of "require".
 
 ### On Vercel
 
--- a/.changeset/good-oranges-pretend.md
+++ b/.changeset/good-oranges-pretend.md
@@ -2,4 +2,25 @@
 "@blitzjs/rpc": patch
 ---
 
-Fixes enormous memory consumption of the dev server
+Fixes enormous memory consumption of the dev server by changing the default import strategy to "require" instead of "import" which in webpack causes multiple chunks to be created for each import.
+
+## Blitz Configuration
+
+To configure this behaviour, you can add the following to your next.config.js:
+
+```js
+/**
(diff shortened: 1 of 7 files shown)


siddhsuresh (approved):
LGTM! Thanks for the PR!

.changeset/good-oranges-pretend.md
```diff
@@ -0,0 +1,26 @@
...
+ * @type {import('@blitzjs/next').BlitzConfig}
+ **/
+const config = {
+  blitz: {
+    resolversDynamicImport: true,
+  },
+}
+```
+
+When `resolversDynamicImport` is set to `true`, the default import strategy will be "import" instead of "require".
```
siddhsuresh, line 22:
```suggestion
When `resolversDynamicImport` is set to `true`, the import strategy will be "import" instead of "require".
```

flybayer:
I just pushed a commit with a fix and refactor.

- `resolversDynamicImport` was not actually being passed to webpack
- We had duplicate loader option types, so I removed the one, now we have proper type safety to catch the first problem
- clean up the logic in the loader by defaulting the dynamic import config at the webpack config level

flybayer (approved):

jayu:
Thanks for help with shipping this 🙏

PR #4162, opened 2023-06-26
Configure the Verbosity of Blitz RPC Logging

siddhsuresh:
Closes: ?

This PR allows the user to control the verbosity of Blitz RPC Logging.

Previously we printed the `input`, `output` of the resolver by default and understandably with the increasing size of resolver outputs then generates a lot of clutter while debugging or during production.

### Now we can configure Blitz RPC in the following way,

In your `[[...blitz]].ts` api file you can see the following settings
```ts
logging?: {
  /**
   * allowList Represents the list of routes for which logging should be enabled
   * If whiteList is defined then only those routes will be logged
   */
  allowList?: string[]
  /**
   * blockList Represents the list of routes for which logging should be disabled
   * If blockList is defined then all routes except those will be logged
   */
  blockList?: string[]
  /**
   * verbose Represents the flag to enable/disable logging
   * If verbose is true then Blitz RPC will log the input and output of each resolver
   */
  verbose?: boolean
  /**
   * disablelevel Represents the flag to enable/disable logging for a particular level
   */
  disablelevel?: "debug" | "info"
}
```
```ts
import { rpcHandler } from "@blitzjs/rpc"

export default api(
  rpcHandler({
    onError: console.log,
    formatError: (error) => {
      error.message = `FormatError handler: ${error.message}`
      return error
    },
   logging: {
...
}
  })
)
```

Example:
```ts
export default api(
  rpcHandler({
    onError: console.log,
    formatError: (error) => {
      error.message = `FormatError handler: ${error.message}`
      return error
    },
    logging: {
      verbose: true,
      blockList: ["getCurrentUser", ...], //just write the resolver name [which is the resolver file name]
    },
  })
)
```

This is enable verbose blitz rpc logging for all resolvers except the resolvers `getCurrentUser` and others mentioned in the `blockList`

## Feature Checklist

### DIFF
--- a/packages/blitz-rpc/src/index-server.ts
+++ b/packages/blitz-rpc/src/index-server.ts
@@ -170,6 +170,9 @@ interface RpcConfig {
 }
 
 function isBlitzRPCVerbose(routePath: string, config: RpcConfig, level: string) {
+  if (!config.logging) {
+    return true
+  }
   const isLevelDisabled = config.logging?.disablelevel === level
   if (config.logging?.verbose) {
     // If whiteList array is defined then allow only those routes in whiteList
@@ -193,14 +196,7 @@ function isBlitzRPCVerbose(routePath: string, config: RpcConfig, level: string) 
   return false
 }
 
-export function rpcHandler(
-  config: RpcConfig = {
-    // keeping this for backwards compatibility
-    logging: {
-      verbose: true,
-    },
-  },
-) {
+export function rpcHandler(config: RpcConfig) {
   return async function handleRpcRequest(req: NextApiRequest, res: NextApiResponse, ctx: Ctx) {
     const resolverMap = await getResolverMap()
     assert(resolverMap, "No query or mutation resolvers found")
--- a/packages/blitz-rpc/src/index-server.ts
+++ b/packages/blitz-rpc/src/index-server.ts
@@ -148,15 +148,15 @@ interface RpcConfig {
   formatError?: (error: Error) => Error
   logging?: {
     /**
-     * WhiteList Represents the list of routes for which logging should be enabled
-     * If whiteList is defined then only those routes will be logged
+     * allowList Represents the list of routes for which logging should be enabled
+     * If allowList is defined then only those routes will be logged
      */
-    whiteList?: string[]
+    allowList?: string[]
     /**
-     * BlackList Represents the list of routes for which logging should be disabled
-     * If blackList is defined then all routes except those will be logged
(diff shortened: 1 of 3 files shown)


siddhsuresh:
!preview blitz-rpc-verbose

flybayer (commented):
this will be good to have.

- rename black/white list to allowList and blockList (they are more inclusive language)
- verbose should be default true, even if `config.logging` is an object without verbose set in it
- we need ability to separately enable/disable input vs output logging. The setup we had with old blitz, i think, logged output as a debug log so that you could set log level `info` to only get input and not output logged.

packages/blitz-rpc/src/index-server.ts
```diff
@@ -213,11 +267,16 @@ export function rpcHandler(config: RpcConfig) {
...
-        log.info(customChalk.dim("Starting with input:"), data ? data : JSON.stringify(data))
+        if (isBlitzRPCVerbose(resolverName, config, "info")) {
+          log.info(customChalk.dim("Starting with input:"), data ? data : JSON.stringify(data))
+        }
         const startTime = Date.now()
         const result = await resolver(data, (res as any).blitzCtx)
         const resolverDuration = Date.now() - startTime
-        log.info(customChalk.dim("Result:"), result ? result : JSON.stringify(result))
+
+        if (isBlitzRPCVerbose(resolverName, config, "info")) {
```
flybayer, line 277:
```suggestion
        if (isBlitzRPCVerbose(resolverName, config, "debug")) {
```

packages/blitz-rpc/src/index-server.ts
```diff
@@ -231,21 +290,26 @@ export function rpcHandler(config: RpcConfig) {
...
-        )
+        if (isBlitzRPCVerbose(resolverName, config, "debug")) {
+          log.debug(
+            customChalk.dim(
+              `Finished: resolver:${prettyMs(resolverDuration)} serializer:${prettyMs(
+                serializerDuration,
+              )} total:${prettyMs(duration)}`,
+            ),
+          )
+        }
```
flybayer, line 312:
```suggestion
        if (isBlitzRPCVerbose(resolverName, config, "info")) {
          log.info(
            customChalk.dim(
              `Finished: resolver:${prettyMs(resolverDuration)} serializer:${prettyMs(
                serializerDuration,
              )} total:${prettyMs(duration)}`,
            ),
          )
        }
```

flybayer (approved):

PR #4168, opened 2023-07-04
Update build pipeline

datner:
- chore: add nix
- build(pkg-template)!: change build system to tsup
- build(generator)!: change build system to tsup
- build(codemod)!: change build system to tsup
- build(blitz-rpc)!: change build system to tsup
- build(blitz-next)!: change build system to tsup
- build(blitz-auth)!: change build system to tsup
- build(blitz)!: change build system to tsup
- build: use turborepo pipelines for coordination
- chore: update dev script
- chore: update lockfile

The PR is segmented meaningfully such as to allow an easier PR experience.
Please use the commits tab instead of the files tab or i'll cry

Nothing is user-facing, so I am not sure if a changeset is required?

Closes: Nothing. I did this because I wanted to :)

## Bug Checklist

## Feature Checklist


datner:
I made fixes for the things that broke. Now I'm searching for an elegant way to handle the import swapping. I have some ideas

flybayer:
Hey @datner! Thanks for your contribution. There's some real good things here like the turborepo updates which didn't exist when we set this up. However, we're less keen on changing the build system, since there's so many things that could cause bugs or unexpected behavior. Can you elaborate more on why you propose changing it?

Aside from the build system, we'd like you to pull the non-build system related changes into a separate, smaller PR that we can merge regardless of whether we move forward with the build system change

datner:
@flybayer 
> Can you elaborate more on why you propose changing it?

Sure, the initial reason was quite simple: unbuild isn't enough (as we use it):
1. It can't _actually_ handle multiple entries, which requires us to always be careful what we import or export in bundles else we will duplicate things over and over and over needlessly bloating and introducing subtle bugs
2. it doesn't transpile correctly. While it follows esm best practices, it fails to correctly translate esm to cjs. Which requires doing exotic things like declaring `module.exports = loader` in an esm file (where it should be `export default loader`)
3. it's clunky, with setup and clean required outside of its own execution
4. It does magic transformations. given entry `src/a/b/c/d/e.ts` you get out `e.mjs`. I didn't test what happens when theres duplication, but that behavior is unintuitive. (Which can be solved by `mkdist`)
5. It is not "monorepo ready", not in the sense that it can't resolve dependencies, it uses esbuild it works great. But in the sense that it is not made for _development in monorepos_, only for publishing. So it lacks `watch`. Something that is stated to not be changed as the promoted solution is `jiti`. Not saying that `jiti` isn't a solution, but incremental build were picked over runtime stubbing.
6. It's not very popular with 10x less downloads than tsup.

Now, there is a very important caveat. We are using a year-old version, _and_ we are using unbuild very naively. 
The _real_ motivator for change were `watch` and `cpx`. `wait-on` is fine, I don't have issues with it but I don't think it's actually needed if the project is configured correctly.
I could go savvy and use `unbuild` "the right way" with all their best practices, which will probably fix many of the issues outlined above but I am just more familiar with `tsup` so I picked it, as they are quite similar in the end 🤷🏻 
I'll make another PR for this purpose

> Aside from the build system, we'd like you to pull the non-build system related changes into a separate, smaller PR that we can merge regardless of whether we move forward with the build system change

Easy, good thing I separated the commits like I did :)

datner:
@flybayer @siddhsuresh 
I've tried to create a version that uses `unbuild`, but it's becoming hacky so it's not gonna happen :)
`jiti` can't stub any flavor of `JSX` so most bundles are blocked from it, meaning that hacking a `watch` is once again required.
`unbuild`, unlike `tsup` don't have an opt-in or opt-out of rollup, which as you knew removes directives for ideological reasons that are not actually coherent to me. (Yes it will mark the entire bundled thing as "use client" and it'll be imported to the client. But that was the status quo before the directive anyways. You can't _use_ code from the server in the client if it does not have the "use client" directive. wtf?) Which is "solveable" with hacks, file-to-file transpiling (kinda), or some plugin. None of these options make sense to me. It is also significantly slower and louder despite using esbuild aswell. I can't explain why.
`unbuild`, again unlike `tsup`, lacks a notion like public dir (`templates` for engine)
`unbuild` does not allow you to control transformation of files through hooks. So while there are many hooks, they can't actually help us with things like adding `"use client"` or renaming imports. Something `tsup` does support.
There are likely other walls I've smashed into that I now forget.
The one good thing (very good thing!!!) that `unbuild` does, is that it encourages you to configure your repo correctly. Since it can infer from your deps, exports, bin, and so on.

PR #4173, opened 2023-07-12
blitz-rpc: Cleanup Method for Event Listeners

siddhsuresh:
Closes: ?

Add cleanup method for potential additional event listeners.

## Bug Checklist


siddhsuresh:
converting this PR to draft for now, will investigate further, but there is no issue with the current state since the plugin.ts is being called only once after continuous testing

flybayer (approved):

Issue #4172, opened 2023-07-11
Update to Prisma 5.0 


cfragkos:
### What is the problem?

New Prisma 5.0 version is not working with @blitz/auth PrismaStorage

### Paste all your error logs here:

```
Argument of type 'PrismaClient<PrismaClientOptions, unknown, Args> & EnhancedPrismaClientAddedMethods' is not assignable to parameter of type 'PrismaClientWithSession'.
  The types of 'session.update' are incompatible between these types.
    Type '<T extends SessionUpdateArgs<Args>>(args: SelectSubset<T, SessionUpdateArgs<Args>>) => Prisma__SessionClient<GetFindResult<SessionPayload<Args>, T>, never, Args>' is not assignable to type '(args: { data: Partial<SessionModel>; where: { handle?: string | undefined; }; }) => Promise<SessionModel>'.
      Types of parameters 'args' and 'args' are incompatible.
        Type '{ data: Partial<SessionModel>; where: { handle?: string | undefined; }; }' is not assignable to type '{ select?: SessionSelect<Args> | null | undefined; include?: SessionInclude<Args> | null | undefined; data: (Without<SessionUpdateInput, SessionUncheckedUpdateInput> & SessionUncheckedUpdateInput) | (Without<...> & SessionUpdateInput); where: SessionWhereUniqueInput; }'.
          Types of property 'where' are incompatible.
            Type '{ handle?: string | undefined; }' is not assignable to type 'SessionWhereUniqueInput'.
              Type '{ handle?: string | undefined; }' is not assignable to type '{ id: string | number; handle: string | number; } & { id?: number | undefined; handle?: string | undefined; AND?: SessionWhereInput | SessionWhereInput[] | undefined; ... 10 more ...; user?: (Without<...> & UserWhereInput) | ... 2 more ... | undefined; }'.
                Property 'id' is missing in type '{ handle?: string | undefined; }' but required in type '{ id: string | number; handle: string | number; }'.
```

### Paste all relevant code snippets here:

```
import { setupBlitzServer } from "@blitzjs/next"
import { AuthServerPlugin, PrismaStorage } from "@blitzjs/auth"
import { simpleRolesIsAuthorized } from "@blitzjs/auth"
import { BlitzLogger } from "blitz"
import db from "db"
import { authConfig } from "./blitz-client"

export const { gSSP, gSP, api } = setupBlitzServer({
  plugins: [
    AuthServerPlugin({
      ...authConfig,
[...]

PR #4174, opened 2023-07-12
blitz-auth: Update Prisma Adapter for v5

siddhsuresh:
Closes: https://github.com/blitz-js/blitz/issues/4172

- [x] make `handle` a required parameter which removed the need for `id` causing the type error

## Feature Checklist

### DIFF
--- a/.changeset/fifty-drinks-cry.md
+++ b/.changeset/fifty-drinks-cry.md
@@ -1,0 +1,8 @@
+---
+"@blitzjs/auth": patch
+"blitz": patch
+---
+
+blitz-auth: Support for Prisma v5
+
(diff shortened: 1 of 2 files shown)


pgvr:
Hi @siddhsuresh, I don't know whether this is relevant but in my case the Session model has a String ID, so I'm not sure the typings would work.

image-yYql3Qgs

siddhsuresh:
@pgvr yes you are correct. I had stopped working while fixing this exact issue at the time. I will be pushing a better fix soon.

cfragkos:
Come on guys what is taking so long

flybayer (approved):

Issue #4175, opened 2023-07-20
Warning: You're using `next/head` inside the `app` directory


wmadden:
### What is the problem?

This warning is triggered when using the Next.js App Router. Blitz includes `@next/head` in a few places, notably: [‎packages/blitz-next/src/index-browser.tsx‎](https://github.com/blitz-js/blitz/blob/0ba2f4e77aeb1a2af1afe5b0121cff56bb9d8bfc/packages/blitz-next/src/index-browser.tsx#L4) and [‎packages/blitz-next/src/error-component.tsx‎](https://github.com/blitz-js/blitz/blob/0ba2f4e77aeb1a2af1afe5b0121cff56bb9d8bfc/packages/blitz-next/src/error-component.tsx#L4)

### Paste all your error logs here:

```
Warning: You're using `next/head` inside the `app` directory, please migrate to the Metadata API. See https://nextjs.org/docs/app/building-your-application/upgrading/app-router-migration#step-3-migrating-nexthead for more details.
```

### Paste all relevant code snippets here:

N/A

### What are detailed steps to reproduce this?

Run a Next.js / Blitz dev server

### Run `blitz -v` and paste the output here:

```
Blitz version: 2.0.0-beta.30 (local)
macOS Ventura | darwin-arm64 | Node: v16.14.2


 Package manager: pnpm

  System:
    OS: macOS 13.4
    CPU: (8) arm64 Apple M1
    Memory: 77.27 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 16.14.2 - ~/.nodenv/versions/16.14.2/bin/node
    Yarn: Not Found
    npm: 8.5.0 - ~/.nodenv/versions/16.14.2/bin/npm
  npmPackages:
    @blitzjs/auth: Not Found
    @blitzjs/next: Not Found
    @blitzjs/rpc: Not Found
    @prisma/client: Not Found
    blitz: Not Found
    next: 13.4.8 => 13.4.8
    prisma: Not Found
    react: Not Found
    react-dom: Not Found
    typescript: 4.9.4 => 4.9.4
```

PR #4176, opened 2023-07-20
blitz-next: Fix `next/head` warning in app directory

siddhsuresh:
Closes: https://github.com/blitz-js/blitz/issues/4175

## Bug Checklist


.changeset/chatty-scissors-jump.md
```diff
@@ -0,0 +1,7 @@
+---
+"next-blitz-auth": patch
```
siddhsuresh, line 2:
```suggestion
```

packages/blitz-next/src/error-component.tsx
```diff
@@ -31,7 +31,7 @@ export class ErrorComponent<P = {}> extends React.Component<P & ErrorProps> {
   render() {
     const {statusCode} = this.props
     const title = this.props.title || statusCodes[statusCode] || "An unexpected error has occurred"
-
+    const Head = dynamic(() => import("next/head"))
```
flybayer, line 34: - this should not be in a render function, should be at top level in the file.
- what affect does this have on legacy apps? I think this will trigger suspense?

siddhsuresh: @flybayer just verified there is no impact on legacy apps

Issue #4191, opened 2023-08-06
Getting Blitzjs CodeSandbox.Io Ready - Inifnite Loops Everywhere! RPC to Blame? 


PerfectProductions:
### What is the problem?

Hey loving this amazing platform, unfortunatley unable to introduce to peers and collaborate as codesandbox instance throws an infinite loop by default straight off a fresh install!

Would love to help get this rectified asap. 

Vid: https://www.awesomescreenshot.com/video/19721584?key=898e7cee95511db0313152a07fda91f8

### Paste all your error logs here:

```
PASTE_HERE (leave the ``` marks)
```

### Paste all relevant code snippets here:

```
PASTE_HERE (leave the ``` marks)
```

### What are detailed steps to reproduce this?

1. Create local repo with 'blitz new blitzApp' and push to github.
2. Create Account and Login to codesandbox.io
3. Import Repository from step one and start the application.
4. Receive Inifnite loop on homepage basic fresh default brand new installation.

Video: https://www.awesomescreenshot.com/video/19721584

Frequency: Happens every time. 

### Run `blitz -v` and paste the output here:

```
yarn run v1.22.19
$ /project/home/perfectproductions/workspace/node_modules/.bin/blitz -v
Blitz version: 2.0.0-beta.31 (local)
Linux 5.15 | linux-x64 | Node: v16.17.0


 Package manager: npm

  System:
    OS: Linux 5.15 Debian GNU/Linux 11 (bullseye) 11 (bullseye)
    CPU: (2) x64 AMD EPYC
    Memory: 864.48 MB / 2.01 GB
    Shell: 5.1.4 - /bin/bash
  Binaries:
    Node: 16.17.0 - /tmp/yarn--1691333664060-0.268803763317665/node
    Yarn: 1.22.19 - /tmp/yarn--1691333664060-0.268803763317665/yarn
    npm: 8.15.0 - /usr/local/bin/npm
  npmPackages:
    @blitzjs/auth: 2.0.0-beta.31 => 2.0.0-beta.31 
    @blitzjs/next: 2.0.0-beta.31 => 2.0.0-beta.31 
    @blitzjs/rpc: 2.0.0-beta.31 => 2.0.0-beta.31 
    @prisma/client: 4.6.1 => 4.6.1 
    blitz: 2.0.0-beta.31 => 2.0.0-beta.31 
    next: 13.4.5 => 13.4.5 
    prisma: 4.6.1 => 4.6.1 
[...]

PR #4199, opened 2023-08-16
blitz-auth & blitz-rpc: Fix unexpected infinite loops in codesandbox

siddhsuresh:
Closes: https://github.com/blitz-js/blitz/issues/4191

Clears `HEADER_SESSION_CREATED` header so make sure it is set only when intended. In local testing (happens randomly only in dev env and codesandbox and its kind) it seems to fix the issue. Going to publish a temporary release to check in codesandbox.

Okay found the actual issue:

Screenshot 2023-08-17 at 13 49 07

## Bug Checklist


packages/blitz-auth/src/server/auth-sessions.ts
```diff
@@ -151,9 +151,11 @@ export async function getSession(
   res: ServerResponse,
   appDir = false,
 ): Promise<SessionContext> {
+  // clear any existing session-create headers
```
siddhsuresh, line 154:
```suggestion
```

siddhsuresh:
!preview fix-infinite-loop

WaysToGo:
@siddhsuresh I am getting the same infinite loop issue in the docker container deployed in AWS. getCurrentUser is called continuously.

siddhsuresh:
@WaysToGo does this fix work for you?

can you minimally reproduce the issue ? 

Or maybe a video or screen shot with the network tab of the request or repo? Could also help in understanding the issue

WaysToGo:
https://github.com/user-attachments/assets/607064db-2c97-4b6c-9346-c507250cc185

WaysToGo:
https://github.com/user-attachments/assets/a70f70a5-2a05-4ef7-af52-2208cd72e728

it seems like everytime new cookie is been set and triggers reload.

WaysToGo:
@siddhsuresh any update on this?

PR #4261, opened 2023-12-13
Add helpful error message when RPC resolvers have the same path

siddhsuresh:

- [x] display the following error when two resolvers resolve to the same path, previously the second or following resolvers with the same just override the previous
- [x] In development, replace the resolver with this error for better visibility. 

image

## Bug Checklist

### DIFF
--- a/packages/blitz-rpc/src/index-server.ts
+++ b/packages/blitz-rpc/src/index-server.ts
@@ -32,7 +32,13 @@ function getGlobalObject<T extends Record<string, unknown>>(key: string, default
 }
 
 type Resolver = (...args: unknown[]) => Promise<unknown>
-type ResolverFiles = Record<string, () => Promise<{default?: Resolver; config?: ResolverConfig}>>
+type ResolverFiles = Record<
+  string,
+  {
+    absoluteResolverPath: string
+    resolver: () => Promise<{default?: Resolver; config?: ResolverConfig}>
+  }
+>
 export type ResolverPathOptions = "queries|mutations" | "root" | ((path: string) => string)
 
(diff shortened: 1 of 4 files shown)


flybayer (approved):

packages/blitz-rpc/src/index-server.ts
```diff
@@ -50,10 +56,47 @@ export function loadBlitzRpcResolverFilesWithInternalMechanism() {
...
 export function __internal_addBlitzRpcResolver(
   routePath: string,
+  absoluteResolverPath: string,
   resolver: () => Promise<{default?: Resolver; config?: ResolverConfig}>,
 ) {
   g.blitzRpcResolverFilesLoaded = g.blitzRpcResolverFilesLoaded || {}
-  g.blitzRpcResolverFilesLoaded[routePath] = resolver
+  const existingResolver = g.blitzRpcResolverFilesLoaded[routePath]
+  if (existingResolver && existingResolver.absoluteResolverPath !== absoluteResolverPath) {
+    console.log("existingResolver", existingResolver)
```
siddhsuresh, line 65:
```suggestion
```

packages/blitz-rpc/src/rpc-logger.ts
```diff
@@ -165,8 +165,11 @@ export class RpcLogger {
   public error(e: any) {
     if (typeof e === "string") {
       this.#logger.error(e)
+    } else if (e instanceof Error) {
+      this.#logger.error(e)
+    } else {
+      this.#logger.error(new Error(e))
```
siddhsuresh, line 171:
```suggestion
    if (typeof e === "string" || e instanceof Error) {
      this.#logger.error(e)
    } else {
      this.#logger.error(new Error(e))
```

PR #4275, opened 2024-01-13
Page role authentication does not work when using roles instead of role

cmoileo:

I have changed the User model to make it able to have several roles instead of a single role.


packages/blitz-auth/src/client/index.tsx
```diff
@@ -347,8 +340,12 @@ function withBlitzAuthPlugin<TProps = any>(Page: ComponentType<TProps> | BlitzPa
...
           typeof authenticate === "object" &&
           authenticate.redirectTo &&
-          authenticate.role &&
-          !authorizeRole(authenticate.role, publicData.role as string)
+          authenticate.roles &&
+          (!Array.isArray(authenticate.roles)
+            ? authorizeRole(authenticate.roles, publicData.roles as string[])
+            : authenticate.roles.some((role: string) =>
+                authorizeRole(role, publicData.roles as string[]),
+              ))
```
siddhsuresh, line 348: Here a possible implementation, since `authorizeRole` can handle an array of roles
```suggestion
          authenticate.role &&
          !authorizeRole(authenticate.role, [
          publicData.role,
          ...(Array.isArray(publicData.roles) ? publicData.roles : []),
        ] as string[])
```

packages/blitz-auth/src/client/index.tsx
```diff
@@ -196,25 +197,21 @@ export const useAuthorizeIf = (condition?: boolean, role?: string | Array<string
...
   }
 }
 
-const authorizeRole = (role?: string | Array<string>, currentRole?: string) => {
-  if (role && currentRole) {
-    if (Array.isArray(role)) {
-      if (role.includes(currentRole)) {
-        return true
-      }
+const authorizeRole = (roles?: string | Array<string>, currentRoles?: string[]) => {
```
siddhsuresh, line 209: 🔨 I have not tested it myself, but the api would need to look like this, to handle a case were a single `currentRole` exists or an array
```ts
const authorizeRole = (role?: string | Array<string>, currentRole?: string | Array<string>) => {
  if (role && currentRole) {
    if (Array.isArray(role) && role.some((r) => currentRole.includes(r))) {
      return true
    } else {
      if (currentRole === role) {
        return true
      }
    }
  }
  return false
}
```

packages/blitz-auth/src/client/index.tsx
```diff
@@ -259,7 +252,7 @@ export type RedirectAuthenticatedToFn = (
 
 export type BlitzPage<P = {}> = React.ComponentType<P> & {
   getLayout?: (component: JSX.Element) => JSX.Element
-  authenticate?: boolean | {redirectTo?: string | RouteUrlObject; role?: string | Array<string>}
+  authenticate?: boolean | {redirectTo?: string | RouteUrlObject; roles?: string | Array<string>}
```
siddhsuresh, line 255: 🔨must not change this api
```suggestion
  authenticate?: boolean | {redirectTo?: string | RouteUrlObject; role?: string | Array<string>}
```

packages/blitz-auth/src/client/index.tsx
```diff
@@ -184,7 +185,7 @@ export const useSession = (options: UseSessionOptions = {}): ClientSession => {
   return session
 }
 
-export const useAuthorizeIf = (condition?: boolean, role?: string | Array<string>) => {
+export const useAuthorizeIf = (condition?: boolean, roles?: string | Array<string>) => {
```
siddhsuresh, line 188: 🔨 no changes needed for this function, please revert

packages/blitz-auth/src/client/index.tsx
```diff
@@ -54,7 +54,7 @@ export const parsePublicDataToken = (token: string) => {
   }
 }
 
-const emptyPublicData: EmptyPublicData = {userId: null, role: null}
+const emptyPublicData: EmptyPublicData = {userId: null, roles: ["user"]}
```
siddhsuresh, line 57: 🔨
```suggestion
const emptyPublicData: EmptyPublicData = {userId: null, role: null, roles: null}
```

packages/blitz-next/src/error-boundary.test.tsx
```diff
@@ -426,7 +427,7 @@ test("supports reset via resetKeys right after error is triggered on component m
 test("should support not only function as FallbackComponent", () => {
   const consoleError = console.error as MockedFunction<(args: unknown[]) => void>
 
-  const FancyFallback = forwardRef(({error}: ErrorFallbackProps) => (
+  const FancyFallback = React.forwardRef(({error}: ErrorFallbackProps) => (
```
siddhsuresh, line 430: why these changes?

packages/blitz-auth/src/client/auth-client.test.ts
```diff
@@ -98,7 +98,7 @@ describe("publicDataStore", () => {
         ret = data
       })
       getPublicDataStore().clear()
-      expect(ret).toEqual({userId: null, role: null})
+      expect(ret).toEqual({userId: null, roles: ["user"]})
```
siddhsuresh, line 101: 🔨 do not change existing tests.

.changeset/great-emus-confess.md
```diff
@@ -0,0 +1,50 @@
...
+"@blitzjs/recipe-passenger": major
+"@blitzjs/recipe-quirrel": major
+"@blitzjs/recipe-reflexjs": major
+"@blitzjs/recipe-render": major
+"@blitzjs/recipe-secureheaders": major
+"@blitzjs/recipe-stitches": major
+"@blitzjs/recipe-styled-components": major
+"@blitzjs/recipe-tailwind": major
+"@blitzjs/recipe-theme-ui": major
+"@blitzjs/recipe-vanilla-extract": major
```
siddhsuresh, line 47:
```suggestion
"@blitzjs/auth": patch
```

.changeset/great-emus-confess.md
```diff
@@ -0,0 +1,50 @@
...
+---
+
+I have changed User model in all of Prisma's schemas. I've changed the role key, changing it from a single string to an array of string called "roles". Then, I've changed all of functions where user is created to grant him a default roles array ["user"] instead of a single string "user". I've also changed all of types that where related to User's role and I've changed line 351 of file "blitz/packages/blitz-auth/src/client/index.tsx", to make the condition accept an array of string for roles.
```
siddhsuresh, line 50: 🔨 this is a user facing changeset that will be present in the github release, so the text here should be on the user facing changes implemented here

siddhsuresh (changes requested):
Hey @cmoileo thanks for the PR and interest to contribute to fix this issue!

There are a few issues with the current state of this implementation, I see that you are updating all the usages of `publicData.role` into `publicData.roles` which would break all apps that depend on this.

For the completion of this particular issue, we need to implement new handling to authorise correctly either when the user creates the session using `role` or `roles`

The only file that would need an implementation change would be `packages/blitz-auth/src/client/index.tsx`, please clean the PR from the other changes for the next re-review.

I have provided my suggestions. Let me know what you think.

packages/blitz-next/src/error-boundary.test.tsx
cmoileo: I had problem passing these tests, forwardRef was not defined. Adding the React indicator fixed that problème.

cmoileo:
Hello @siddhsuresh, thank's for your review ! :)

I've removed every changes that are outside the packages/blitz-auth/src/client/index.tsx file, and I've pushed the changes.  Don't we need to changes the User model aswell ?I juste reset as it where before.

siddhsuresh:
> Don't we need to changes the User model aswell ?I juste reset as it where before.

@cmoileo no we do not need to change the User model, the `role` and `roles` are used for a given session's `publicData` which is user defined and configurable, with a few preset attributes like `role` and `roles`.

These are configurable in the `types.ts` file and the `blitz-(server|client).ts` files.

On a basic relook, the build will fail with the current changes, please address it when you can. Thanks!

flybayer:
Hey, going to close this for now. Feel free to open a new PR once everything is fixed.

Issue #4278, opened 2024-01-15
BlitzCliConfig does not work when blitz-server.ts is in /src/app directory


ReadB:
### What is the problem?

`blitz-server.ts` was moved from `src/blitz-server.ts`to `src/app/blitz-server.ts` some time between 2.0.0-beta.35 and v2.0.2

This causes the BlitzCliConfig to not function, tested using `blitz generate` command and `codegen: { fieldTypeMap: {} }` config.

### Paste all your error logs here:

n/a

### Paste all relevant code snippets here:

blitz-server.ts
```ts
import type { BlitzCliConfig } from "blitz"
///...
export const cliConfig: BlitzCliConfig = {
  codegen: {
    fieldTypeMap: {
      string: {
        component: "LabeledTextField",
        inputType: "text",
        zodType: "string",
        prismaType: "String",
      },
      boolean: {
        component: "LabeledBooleanField",
        inputType: "checkbox",
        zodType: "boolean",
        prismaType: "Boolean",
      },
    },
  },
}
```

### What are detailed steps to reproduce this?

1. Run `blitz new app`
2. Run `cd app`
3. Modify the BlitzCliConfig in `src/app/blitz-server.ts` as above
4. Run `blitz generate all test somestring:string someboolean:boolean`
5. Find `src/app/tests/components/TestForm.tsx` and observe the field someboolean is using the default LabeledTextField component instead of LabeledBooleanField

Moving `src/app/blitz-server.ts` back to the original `src/blitz-server.ts` location resolves the issue.

### Run `blitz -v` and paste the output here:

```
Blitz version: 2.0.2 (global)
Blitz version: 2.0.2 (local)
Linux 5.19 | linux-x64 | Node: v18.18.2


 Package manager: npm

  System:
    OS: Linux 5.19 Ubuntu 22.04.3 LTS 22.04.3 LTS (Jammy Jellyfish)
    CPU: (8) x64 Intel(R) Core(TM) i7-6700 CPU @ 3.40GHz
    Memory: 8.99 GB / 15.50 GB
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 18.18.2 - ~/.nvm/versions/node/v18.18.2/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v18.18.2/bin/yarn
    npm: 9.8.1 - ~/.nvm/versions/node/v18.18.2/bin/npm
  npmPackages:
[...]

PR #4279, opened 2024-01-15
Fix: BlitzCliConfig does not work when blitz-server.ts is in /src/app directory

siddhsuresh:
Closes: https://github.com/blitz-js/blitz/issues/4278

- [x] search for any subsdirectory inside the `app|src` directories to find `blitz-sever.ts` since we don't need to have a fixed location for it.

## Bug Checklist

### DIFF
--- a/packages/generator/src/generator.ts
+++ b/packages/generator/src/generator.ts
@@ -35,7 +35,7 @@ export const customTemplatesBlitzConfig = async (
   codemod = false,
 ) => {
   const {globby} = await import("globby")
-  const blitzServer = await globby(["{app,src}/**/blitz-server.{ts,js}"], {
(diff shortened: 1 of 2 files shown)


packages/generator/src/generator.ts
```diff
@@ -35,7 +35,7 @@ export const customTemplatesBlitzConfig = async (
   codemod = false,
 ) => {
   const {globby} = await import("globby")
-  const blitzServer = await globby(["{app,src}/**/blitz-server.{ts,js}"], {
+  const blitzServer = await globby(["{app,src}/**/**/blitz-server.(ts,js)"], {
```
siddhsuresh, line 38:
```suggestion
  const blitzServer = await globby(["{app,src}/**/**/blitz-server.{ts,js}"], {
```

flybayer (approved):

PR #4288, opened 2024-01-23
perf: add filter to query only non expired sessions

siddhsuresh:

- [x] add check to make sure we only query the sessions whose `expiredAt` dates are after the current date

## Bug Checklist


.changeset/chatty-ants-bake.md
```diff
@@ -0,0 +1,6 @@
+---
+"@blitzjs/auth": patch
+"blitz": patch
+---
+
+(internal) perf: add filter to select only non expired sessions
```
flybayer, line 6: remove "internal"

flybayer (approved):

PR #4290, opened 2024-01-24
Added codemod support for legacy projects with a /src folder

Doc0x1:
Changes the codemod upgrade-legacy tool to add support for legacy projects which have folders nested inside a /src directory
## Bug Checklist

## Feature Checklist

### DIFF
--- a/.changeset/serious-impalas-guess.md
+++ b/.changeset/serious-impalas-guess.md
@@ -1,0 +1,5 @@
+---
+"@blitzjs/codemod": patch
+---
+
+Added support to codemod upgrade-legacy for projects that have their pages folder nested in a src/ folder
--- a/packages/codemod/src/upgrade-legacy.ts
+++ b/packages/codemod/src/upgrade-legacy.ts
@@ -32,6 +32,12 @@ class ExpectedError extends Error {
   }
 }
 
+const findPagesDirectory = () => {
+  const srcPagesDir = path.join("src", "pages")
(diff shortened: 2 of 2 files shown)


siddhsuresh (commented):
thanks for the PR @Doc0x1! this is a good start. 

1. https://github.com/blitz-js/blitz/blob/a09685a63a01a01b697047b3200427f12489b214/packages/codemod/src/upgrade-legacy.ts#L49

I feel the above line we should make the following change 
```diff
- const appDir = path.resolve("app")
+ const appDir = fs.existsSync(path.resolve("app")) ? fs.existsSync(path.resolve("src"))
```
since in this case the user has the files only in src directory and not in the app

2. https://github.com/blitz-js/blitz/blob/a09685a63a01a01b697047b3200427f12489b214/packages/codemod/src/upgrade-legacy.ts#L1317

we should use the same utility here.

other than these, looks good!

Doc0x1:
Okay @siddhsuresh sounds good, I'll implement those changes once I'm back home

Doc0x1:
@siddhsuresh Okay, should be good to go for review. Let me know if there's anything else. Also I decided to throw an Error around line 49 inside a try catch statement in the event that appDir is just an empty string to add a bit of error handling towards the start.

siddhsuresh (commented):
LGTM!

siddhsuresh (approved):

Issue #4307, opened 2024-03-05
Windows Compatibility Issue: process.kill("SIGABRT") not supported


lmisea:
### What is the problem?

When running the `yarn blitz dev` command in our project on a Windows machine, an error message is thrown:

```
Error: kill ENOSYS
    at ChildProcess.kill (node:internal/child_process:509:13)
    at Object.onRebuild (<project folder>\node_modules\blitz\dist\chunks\next-commands.cjs:187:20)
    ...
```

This error originates from the **next-commands.cjs** file, where a process is attempting to be terminated with the `process.kill("SIGABRT")` signal. This signal is not supported on Windows, causing the ENOSYS error.

A potential fix for this issue is to replace `process.kill("SIGABRT") `with `process.kill("SIGINT")` in the next-commands.cjs file for Windows only users. This will likely require checking if the user OS is Windows first and if so, making the change. SIGINT is a process termination signal that is supported by Windows.

After making this change, running `yarn blitz dev` does not result in any errors.

It would be great if this could be addressed in the library itself to improve the developer experience for Windows users.

### Paste all your error logs here:

```
Error: kill ENOSYS
    at ChildProcess.kill (node:internal/child_process:509:13)
    at Object.onRebuild (C:\Users\iseal\Projects\seconds\node_modules\blitz\dist\chunks\next-commands.cjs:187:20)
    at C:\Users\iseal\Projects\seconds\node_modules\esbuild\lib\main.js:1317:27
    at runOnEndCallbacks (C:\Users\iseal\Projects\seconds\node_modules\esbuild\lib\main.js:1162:65)
    at C:\Users\iseal\Projects\seconds\node_modules\esbuild\lib\main.js:1307:17
    at handleRequest (C:\Users\iseal\Projects\seconds\node_modules\esbuild\lib\main.js:730:15)
    at handleIncomingPacket (C:\Users\iseal\Projects\seconds\node_modules\esbuild\lib\main.js:756:7)
    at Socket.readFromStdout (C:\Users\iseal\Projects\seconds\node_modules\esbuild\lib\main.js:632:7)
    at Socket.emit (node:events:518:28)
    at Socket.emit (node:domain:488:12) {
  errno: -4054,
  code: 'ENOSYS',
  syscall: 'kill'
}
```

### Paste all relevant code snippets here:

[...]

PR #4308, opened 2024-03-07
fix: process.kill("SIGABRT") not supported on windows

lmisea:
Closes: #4307

- Use `process.kill("SIGINT")` instead of `process.kill("SIGABRT")` on `startCustomServer()` at **next-utils.ts** on blitz package.
- This solves a minor Windows compatibility issue because `process.kill("SIGABRT")` is not supported on Windows.

## Bug Checklist

- [x] Changeset added (run `pnpm changeset` in the root directory) (patch bump)

## Feature Checklist

### DIFF
--- a/.changeset/clever-insects-shave.md
+++ b/.changeset/clever-insects-shave.md
@@ -1,0 +1,5 @@
+---
+"blitz": patch
+---
+
+Use process.kill("SIGINT") instead of process.kill("SIGABRT") on startCustomServer() for better compatibility with operative systems
--- a/.changeset/clever-insects-shave.md
+++ b/.changeset/clever-insects-shave.md
@@ -2,4 +2,4 @@
 "blitz": patch
 ---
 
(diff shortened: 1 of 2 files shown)


siddhsuresh (approved):
Hey @lmisea thanks for the PR! LGTM!

Can you run `pnpm changeset` and set the type to `patch` for `blitz` in the root directory to create the changeset explaining this change (it will be included in the release)

Another way would be clicking here,
image

.changeset/clever-insects-shave.md
```diff
@@ -0,0 +1,5 @@
+---
+"blitz": patch
+---
+
+Use process.kill("SIGINT") instead of process.kill("SIGABRT") on startCustomServer() for better compatibility with operative systems
```
siddhsuresh, line 5:
```suggestion
Use `SIGINT` signal instead of `SIGABRT` to stop the process, when using custom server for better compatibility with operative systems
```

PR #4313, opened 2024-03-15
Use this.rootContext instead of webpack internals

timneutkens:
Ensures the root context is read from the public API that webpack exposes. This is the first step for Turbopack support as Turbopack includes `this.rootContext` as well.

What is unclear to me is why it's checking for `server` only and not for edge, does Blitz support edge runtime? If so you can leverage the `isServer` condition in webpack() to remove the internals usage completely 👍

Closes: ?

## Bug Checklist

- [ ] Changeset added (run `pnpm changeset` in the root directory)
- [ ] Integration test added (see [test docs](https://blitzjs.com/docs/contributing#running-tests) if needed)

## Feature Checklist

- [ ] Changeset added (run `pnpm changeset` in the root directory)
- [ ] Integration test added (see [test docs](https://blitzjs.com/docs/contributing#running-tests) if needed)
- [ ] Documentation added/updated (submit PR to [blitzjs.com repo `main` branch](https://github.com/blitz-js/blitzjs.com))

PR #4314, opened 2024-03-16
Turbopack support for Blitz

timneutkens:
- Includes the changes of #4313

This PR includes the changes required to make the Blitz loaders work with Turbopack. The biggest challenge is that Blitz currently uses private fields, namely `_compiler` (anything `_` prefixed in webpack is private) and `_compiler` is not available in Turbopack. Even if we were to mock the `_compiler` being there Turbopack does not have the concept of multiple compilers running so the `name` can't be relied upon to decide when to compile or not.

Potentially a better solution for the webpack implementation is to use the `this.target` property (also not available in Turbopack) for the webpack implementation, that way you don't have to rely on internals that could change in the future.
Even better would be passing an option for when to apply, e.g. using `isServer` in the custom webpack config that is applied to Next.js.

Now for the Turbopack side of things:
- We've landed support for conditionally running loaders on e.g. `browser` or `node`, PR: https://github.com/vercel/next.js/pull/63333
	- This is available on the latest canary: 14.2.0-canary.25, which I've published for you earlier this evening to try the changes in this PR.
	- There's a small configuration checking warning that we still have to resolve so it will show warnings about config being incorrect but the config I've shared below is the correct one.
- I've refactored the Blitz loader to just always run the transform when there is no name, instead we leverage the loader configuration to correctly apply at which layer we want the loader to run
- I've changed the require()/import() path to be relative instead of absolute since Turbopack does not support reading absolute paths since they cause uncacheable output as the path is specific to your machine.

Here's the `next.config.js` edits I've made to run the Blitz rpc loaders:

Notably Turbopack's glob matcher doesn't support `[` and `]` yet and you can't reliably escape them, we'll work on fixing that, but for now I just used `*` before and after the `...blitz` path to wildcard match `[[` and `]]`.

```tsx
const { withBlitz } = require("@blitzjs/next")

const loaderClient = require.resolve("@blitzjs/rpc/dist/loader-client.cjs")
const loaderServer = require.resolve("@blitzjs/rpc/dist/loader-server.cjs")
const loaderServerResolvers = require.resolve("@blitzjs/rpc/dist/loader-server-resolvers.cjs")

console.log("loaderClient", loaderClient)
console.log("loaderServer", loaderServer)
console.log("loaderServerResolvers", loaderServerResolvers)

/** @type {import('next').NextConfig} */
const nextConfig = {
  experimental: {
    turbo: {
      resolveAlias: {
        "cross-spawn": "./turbopack/empty.js",
        "npm-which": "./turbopack/empty.js",
        fs: { browser: "./turbopack/empty.js" },
      },
      rules: {
        "**/*...blitz*.{jsx,tsx,js,ts}": {
          default: {
            loaders: [{ loader: loaderServer, options: {} }],
            as: "*.ts",
          },
        },
        "**/{queries,mutations}/**": {
          browser: {
            loaders: [
              {
                loader: loaderClient,
                options: {},
              },
            ],
            as: "*.ts",
          },
          default: {
            loaders: [
              {
                loader: loaderServerResolvers,
                options: {},
              },
            ],
            as: "*.ts",
          },
        },
      },
    },
  },
}

module.exports = withBlitz(nextConfig)
```

Please give it a try and let me know if it's working as intended, as far as I can tell it's applying the loaders correctly now but I'm not familiar enough with Blitz to know for certain that e.g. edge cases are working correctly too 🙏 

Closes: ?

## Bug Checklist

## Feature Checklist


packages/blitz-rpc/src/server/loader/server/loader-server.ts
```diff
@@ -68,7 +89,7 @@ export async function transformBlitzRpcServer(
 
     code += `__internal_addBlitzRpcResolver('${routePath}','${slash(
       resolverFilePath,
-    )}',() => ${importStrategy}('${slash(resolverFilePath)}'));`
+    )}',() => ${importStrategy}('${slash(relative(context, resolverFilePath))}'));`
```
sokra, line 92:
```suggestion
    )}',() => ${importStrategy}('${slash(relative(context, resolverFilePath).replace(/\\/g, "/")}'));`
```

for windows

packages/blitz/src/index-browser.tsx
```diff
@@ -84,6 +84,5 @@ if (typeof window !== "undefined" && process.env.NODE_ENV === "development") {
 
 export * from "./utils"
 export * from "./types"
-export * from "./utils/enhance-prisma"
```
flybayer, line 87: this is a breaking change

siddhsuresh: I introduced this in my latest commit. will revert

flybayer: actually is fine

flybayer:
Remaining

- [ ] auto add turbo config in `withBlitz`
- [ ] add test for turbo mode

gengjiawen:
@timneutkens currently a blitzjs project build consume too much memory, can't even setup build on a 2gb machine (about 3.5g from my bizzare observation), will turbo improve this case ?

cfragkos:
> @timneutkens currently a blitzjs project build consume too much memory, can't even setup build on a 2gb machine (about 3.5g from my bizzare observation), will turbo improve this case ?

No, turbopack, at the moment is for the dev server. Building will remain the same.

gengjiawen:
> > @timneutkens currently a blitzjs project build consume too much memory, can't even setup build on a 2gb machine (about 3.5g from my bizzare observation), will turbo improve this case ?
> 
> No, turbopack, at the moment is for the dev server. Building will remain the same.

But I guess it's very near, since turbo is almost 100% tests passed.

flybayer:
!preview

flybayer:
!preview turbopack

packages/generator/templates/pages/package.ts.json
```diff
@@ -27,7 +27,7 @@
     "@blitzjs/rpc": "latest",
     "@prisma/client": "4.6.1",
     "blitz": "latest",
-    "next": "13.4.5",
```
gengjiawen, line 30: This better stays on next 13. since next14 broken on vercel deploy.

siddhsuresh:
!preview turbopack

flybayer (approved):

PR #4340, opened 2024-05-21
Remove `.node-version` file

tordans:
I was looking for the recommended node version to use.

I am assuming this file is of date and it is likely not used anymore so might as well be deleted?
It was created 2 years ago when migrating to this Monorepo.


siddhsuresh (approved):
Thanks! yeah this is way too outdated right now and not really being used

Issue #4344, opened 2024-06-01
enhanced security by default


Vandivier:
### What do you want and why?

for streamers, blitz logs expose sensitive data by default on the terminal

### Possible implementation(s)

```
// Note: This stays in the /pages folder for the time being

import { rpcHandler } from "@blitzjs/rpc"
import { api } from "src/app/blitz-server"

const getBlitzLogLevel = (): "info" | "debug" | undefined => {
  const requestedLevel = process.env.BLITZ_LOG_DISABLE_LEVEL
  if (requestedLevel === "info" || requestedLevel === "debug") {
    return requestedLevel
  }
  return undefined
}

export default api(
  rpcHandler({
    onError: (error, ctx) => console.log(error),
    logging: {
      disablelevel: getBlitzLogLevel(),
    },
  })
)

```

Vandivier:
related https://github.com/Vandivier/ladderly-3/issues/182

siddhsuresh:
@Vandivier this sounds like a good default, will you be willing to send a PR?

Vandivier:
@siddhsuresh yes, thanks! i'll get a PR up in the next 1-3 days

PR #4347, opened 2024-06-05
feat: hide raw db query out from terminal

Vandivier:
closes https://github.com/blitz-js/blitz/issues/4344


Vandivier:
fyi @siddhsuresh

PR #4348, opened 2024-06-11
fix: use the correct SHA for `pr-release.yml`

RedYetiDev:
This PR fixes the `pr-release.yml` workflow to use the correct SHA, rather than trying to get a non-existant `${{ github.event.pull_request.head.ref }}`.

This PR has the workflow get the SHA associated with the issue, verify it is the correct SHA, and then use it.


siddhsuresh (approved):
Thanks a lot for the PR! LGTM

PR #4356, opened 2024-07-03
add error register in rpc handler

siddhsuresh:
add blitz custom error register in rpc handler in case it has not been registered

### DIFF
--- a/packages/blitz/src/errors.ts
+++ b/packages/blitz/src/errors.ts
@@ -129,3 +129,5 @@ export function registerBlitzErrorClasses() {
     globalThis._BLITZ_ERROR_CLASS_REGISTERED = true
   }
 }
+
+registerBlitzErrorClasses()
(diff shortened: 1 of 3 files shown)


flybayer (approved):

packages/blitz/src/errors.ts
```diff
@@ -89,7 +89,10 @@ export class PaginationArgumentError extends Error {
   }
 }
 
-if (isNotInUserTestEnvironment() && !globalThis._BLITZ_ERROR_CLASS_REGISTERED) {
+let _blitzErrorClassRegistered = false
+
+export function registerBlitzErrorClasses() {
+  if (_blitzErrorClassRegistered || isNotInUserTestEnvironment()) return
```
siddhsuresh, line 95:
```suggestion
  if (_blitzErrorClassRegistered || !isNotInUserTestEnvironment()) return
```

Issue #4358, opened 2024-07-14
when generating the Route manifest, a route using a HOC wrapper will not be generated correctly


bezalel6:
### What is the problem?

when generating the Route manifest, a route using a HOC wrapper will be named after the HOC. which would obv be a problem, especially since a HOC is meant to be used across an application.

### Paste all your error logs here:

```
```

### Paste all relevant code snippets here:

for example:
```
const Products: BlitzPage = () => {
    return <ProductsList source={{source: 'all-products'}}></ProductsList>
}

export default withSuspense(Products)
```
the generated route will be withSuspence, and not Products. 

### What are detailed steps to reproduce this?

exporting a default component and wrapping it with any HOC. the generated route will be named after the HOC and not the component itself

### Run `blitz -v` and paste the output here:

```
Blitz version: 2.0.9 (global)
Blitz version: 2.0.10 (local)
Windows 11 | win32-x64 | Node: v20.15.0


 Package manager: npm

  System:
    OS: Windows 10 10.0.22631
    CPU: (16) x64 AMD Ryzen 7 3700X 8-Core Processor
    Memory: 15.60 GB / 31.91 GB
  Binaries:
    Node: 20.15.0 - C:\Program Files\nodejs\node.EXE
    Yarn: Not Found
    npm: 10.7.0 - C:\Program Files\nodejs\npm.CMD
  npmPackages:
    @blitzjs/auth: ^2.0.10 => 2.0.10
    @blitzjs/next: ^2.0.10 => 2.0.10
    @blitzjs/rpc: ^2.0.10 => 2.0.10
    @prisma/client: ^5.16.2 => 5.16.2
    blitz: ^2.0.10 => 2.0.10
    next: ^14.2.5 => 14.2.5
    prisma: ^5.16.2 => 5.16.2
    react: ^18.3.1 => 18.3.1
    react-dom: ^18.3.1 => 18.3.1
    typescript: ^5.5.3 => 5.5.3
```

PR #4359, opened 2024-07-14
modify parseDefaultExportName so it does not grab an HOC, but the fir…

bezalel6:
…st Argument it takes. presumably the actual component

Closes: #4358 

- Updated `parseDefaultExportName` to improve parsing of default exports.
- The function now handles cases where the default export is a higher-order component (HOC).
- If a declaration like `export default function ...` exists, it returns the function name as before.
- If no declaration is found but a valid function name inside parentheses exists (e.g., `export default functionName1(functionName2)`), it returns `functionName2` instead, treating `functionName1` as a HOC.

## Bug Checklist

## Feature Checklist

### DIFF
--- a/.changeset/little-boats-rest.md
+++ b/.changeset/little-boats-rest.md
@@ -1,0 +1,5 @@
+---
+"blitz": patch
+---
+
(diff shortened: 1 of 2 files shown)


siddhsuresh (approved):
Thanks for the PR! LGTM

